PR #1, opened 2019-08-06
Make generic, removing dependency on node_runtime

ascjones:
Enables use with any custom substrate runtime.


ascjones:
Just testing that this works, test currently failing with:

https://github.com/paritytech/substrate/blob/741dc1562d39dd255445f52124bb508cb1dec8d6/core/sr-api-macros/src/decl_runtime_apis.rs#L212

PR #3, opened 2019-08-07
Update to parity-scale-codec + misc

ascjones:
Future should now return `Error` in case of `Decode` error.


dvc94ch:
Rebased #4 on this branch

PR #7, opened 2019-08-10
Refactor

dvc94ch:
* nonce handling
* uses default values from metadata
* uses functions from metadata
* removes dependency on srml_system::Trait and srml_balances::Trait.
* adds doc comments
* adds system, balances, contracts modules

depends on paritytech/substrate#3370


src/metadata.rs
```diff
@@ -61,26 +106,38 @@ pub struct StorageMetadata {
...
+                Some(StorageMap {
+                    _marker: PhantomData,
+                    prefix,
+                    hasher,
+                    default,
+                })
             }
             _ => None,
         }
+        .ok_or(MetadataError::StorageTypeError)
```
ascjones: Simpler just to return `Ok` and `Err` in the branches above?

src/srml/balances.rs
```diff
@@ -0,0 +1,129 @@
...
+    ) -> Box<dyn Future<Item = <Self::Balances as Balances>::Balance, Error = Error> + Send>
+    {
+        let free_balance_map = || {
+            Ok(self
+                .metadata()
+                .module("Balances")?
+                .storage("FreeBalance")?
+                .get_map::<
+                <Self::Balances as System>::AccountId,
+                <Self::Balances as Balances>::Balance>()?)
```
ascjones: This formatting done by rustfmt? Looks odd

src/lib.rs
```diff
@@ -91,266 +108,256 @@ impl<T: srml_system::Trait, SE: SignedExtension> ClientBuilder<T, SE> {
...
-    fn connect(&self) -> impl Future<Item = rpc::Rpc<T, SE>, Error = error::Error> {
+impl<T: System> Clone for Client<T> {
+    fn clone(&self) -> Self {
+        Self {
+            url: self.url.clone(),
+            genesis_hash: self.genesis_hash.clone(),
+            metadata: self.metadata.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
dvc94ch: It requires System: Clone, which is annoying from a usability perspective

src/srml/balances.rs
dvc94ch: that is rustfmt... it's not perfect...

PR #16, opened 2019-08-28
An attempt to remove ModuleCalls in srml/mod.rs and relevant parts

daogangtang:
In original implementation, there needs ModuleCalls to delegate the call of those srml module, I think this is owed to the problem of the project structure.

This design leads to a severe problem: if user of this library wants to introduce a new runtime module, he needs to impl ModuleCalls directly in his module file, and this is prevented by Rust for orphan rule.

So, I just do some changes for it.


daogangtang:
@dvc94ch Hi, the main purpose is to solve "if user of this library wants to introduce a new runtime module, he needs to impl ModuleCalls directly in his module file, and this is prevented by Rust for orphan rule.", so user has to clone this repository to his disk, and adds code in this repository, that is not what a library's purpose.

Code is just to show idea, and not clean now, with your opinion, I can clean them.

ascjones:
@daogangtang yes we absolutely want to support custom runtime modules, it's my attempt at namespacing the modules api that has caused this. I will look at this as soon as I have finished my current task.

dvc94ch:
@daogangtang renaming files causes large diffs which make seeing the change harder. You can run the ignored tests with `cargo test -- --ignored`. The `System` and `Balances` traits are specific to the runtime used, I don't think it's a good idea to have `BasicSystem` assume that the types are equal to whatever the `node-runtime` uses. The node runtime is a runtime for testing purposes, the polkadot, the node-template and custom runtimes may use other types.

ascjones:
Superseded by #24

PR #20, opened 2019-09-18
Opaque events

ascjones:
Removes the requirement for a top level `Event` enum to match the substrate runtime. Instead all the events for a given named module can be retrieved and decoded directly to the module `Event` enum.

This decouples the use of events from a specific substrate chain/runtime definition.


ascjones:
This doesn't work, so alternative work in progress using dynamic decoding.

PR #29, opened 2019-10-23
Fix for substrate master

thiolliere:
change introduced on master by https://github.com/paritytech/substrate/pull/3821/files

basically Debug is now required in wasm as well but deriving debug in no-std must be made with https://crates.parity.io/substrate_debug_derive/index.html to avoid bloating the wasm blob.

### DIFF
--- a/src/srml/system.rs
+++ b/src/srml/system.rs
@@ -21,10 +21,9 @@ use runtime_primitives::traits::{
     CheckEqual,
     Hash,
     Header,
-    MaybeDebug,
     MaybeDisplay,
-    MaybeSerializeDebug,
-    MaybeSerializeDebugButNotDeserialize,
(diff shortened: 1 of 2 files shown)


src/srml/system.rs
```diff
@@ -49,7 +50,8 @@ pub trait System: 'static + Eq + Clone + std::fmt::Debug {
     /// The block number type used by the runtime.
     type BlockNumber: Parameter
         + Member
-        + MaybeSerializeDebug
+        + MaybeSerializeDeserialize
```
thiolliere: Deserialize has been bound now because some requirement has changed on blocknumber as well, it should be a problem.

Stefie:
Thanks for solving that so quickly!

PR #31, opened 2019-11-05
update deps for jsonrpc

insipx:
In relation to this: https://github.com/paritytech/substrate/pull/4012

allows subxt to  compile against latest substrate master


koushiro:
@ascjones Please update **v0.1 branch**, otherwise compiling `ink!` will fail.

PR #46, opened 2019-11-26
Fix the module index needed to form a Call

en:
Same as events, modules with no calls have no definition in the top
level enum. Now use two fields, _index_for_calls_ and _index_for_events_ to
track the module index needed for Call and Event respectively.


en:
Closing, I will re-submit a new PR.

Issue #35, opened 2019-11-18
Contracts::put_code: InvalidTransaction::BadProof


ascjones:
Since upgrading to latest signing #33 and Extrinsic version 4, the `put_code` test fails with `Failed to submit extrinsic: Transaction pool error: Invalid transaction validity: InvalidTransaction::BadProof`. 

Oddly the `transfer_balance` test succeeds, so not sure why the signature is good for one and not the other.

ascjones:
Fixed by #47

PR #47, opened 2019-11-27
Fix the module index used to create a encoded func

en:
Same as Events, modules with no calls should be ignored when calculating
the module index.


en:
I think this PR also fixed #35

PR #57, opened 2019-12-19
Add a method to fetch current finalized head hash

maciejhirsz:


src/rpc.rs
```diff
@@ -286,7 +293,7 @@ impl<T: System + Balances + 'static> Rpc<T> {
                                 TransactionStatus::Future
                                 | TransactionStatus::Ready
                                 | TransactionStatus::Broadcast(_) => None,
-                                TransactionStatus::Finalized(block_hash) => {
+                                TransactionStatus::InBlock(block_hash) => {
```
ascjones: Was this variant renamed in substrate?

PR #78, opened 2020-03-20
Fix dep version conflict

Satoshi-Kusumoto:
Current Subxt works well with either a running alpha3 Substrate node or a alpha3 node-template node (if given a proper Runtime).

However, when I tried to use node-template and subxt as dependency, and do a lazy impl for the Runtime like the following
```rust
impl System for MyNodeRuntime {
    type Index = <node_template::Runtime as system::Trait>::Index;
```
I will get a compiler error:
```
error[E0271]: type mismatch resolving `<sp_runtime::MultiSigner as sp_runtime::traits::IdentifyAccount>::AccountId == sp_core::crypto::AccountId32`
  --> src/main.rs:57:26
   |
57 |             let xt = cli.xt(signer, None).await?;
   |                          ^^ expected struct `sp_core::crypto::AccountId32`, found a different struct `sp_core::crypto::AccountId32`
   |
   = note: perhaps two different versions of crate `sp_core` are being used?
```
So there is definitely a version conflict.
This fix, following the toml file of Substrate and node-template, solves this problem. I think before a stable release of all the related things on Crates.io, it would be better to keep the same dependency format as Substrate repo to avoid version conflict.
(All the tests against a Substrate node still pass)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,11 @@ serde_json = "1.0"
 url = "1.7"
 codec = { package = "parity-scale-codec", version = "1.2", default-features = false, features = ["derive", "full"] }
 
-frame-metadata = { version = "11.0.0-alpha.3", package = "frame-metadata" }
+[dependencies.frame-metadata]
+git = 'https://github.com/paritytech/substrate.git'
+package = 'frame-metadata'
+rev = '013c1ee167354a08283fb69915fda56a62fee943'
+version = '11.0.0-alpha.3'
 
 [dependencies.frame-support]
 git = 'https://github.com/paritytech/substrate.git'
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ keywords = ["parity", "substrate", "blockchain"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
+hex = "0.4.0"
 log = "0.4"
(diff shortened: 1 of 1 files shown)


ascjones:
Yes I also had a similar issue because the `node-template` is using github dependencies. 

However I am inclined to leave it as it is because I aim to release to `crates.io` very soon (just need to update to use the `jsonrpsee` from crates.io)

You can probably work around this by using [patch](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html). Alternatively you can build your `subxt` runtime from the constituent types rather than depending on the whole runtime itself, it was designed in this way.

Satoshi-Kusumoto:
I see. Being able to depend on the whole runtime may offer some convenience. Surely this is a trade-off. The current design philosophy also has its reason.
By the way, any rough ideas about the release time? Say couple of days?

ascjones:
> Being able to depend on the whole runtime may offer some convenience. Surely this is a trade-off.

It does an it is.

However once stable versions of the `substrate` crates are available on `crates.io`, the runtime can be constructed using those, but of course they will have to be the same version as those used in `subxt`.

> By the way, any rough ideas about the release time? 

I believe substrate `alpha.4` dependencies will be published early next week, so I will publish this as soon as those are avaialable.

Satoshi-Kusumoto:
Cool!

Issue #82, opened 2020-03-30
Create raw payload from rpc client


whalelephant:
Hi,
I would like to get feedback on a proposed method for `Client` such that 
`client.create_raw_payload(AccountId, Call)` returns the payload to be signed.

The reason is so that the payload may be signed by other signers (maybe not in the `XtBuilder` and use this or other tools to create `uncheckedExtrinsic`.

ascjones:
Sounds like a good idea, though perhaps it would be on the `XtBuilder` API?

whalelephant:
From my understanding `Xtbuilder` currently accept a signer type with trait `Pair` implemented for signing. The proposed method is so that the only requirement for creating the raw payload is the pubkey / account ID

ascjones:
Yeah you're right. Potentially could make signer `Option<P>` then provide an `XtBuilder` impl with `P = ()`. But feel free to make a PR with your idea, and we can see how it looks.

whalelephant:
ok cool, I don't have permission to push my working branch for the PR at the moment. Please help. thanks!

ascjones:
If you have a fork you should be able to push to that and create a PR from there

ascjones:
Closed in #83

PR #83, opened 2020-04-01
create raw payload in Client

whalelephant:
ref to  #82  - explore the method on `Client`


whalelephant:
Do let me know if you want it structured differently and remove dups. Thanks!!

PR #91, opened 2020-04-18
Support embedded full/light node clients.

### LINKED ISSUE #49: Setup CI
- build
- test
- fmt

**Integration Tests**

If possible run integration tests against a recent substrate node instance (docker).

dvc94ch:
This let's the `ClientBuilder` specify a `jsonrpsee::Client` and implements a `jsonrpsee::ClientTransport` that wraps a substrate light client.

TODO:
- [x] use node template from official repo, depends on [0]
- [ ] in a follow up remove support for the subxt_test macro. Due to usability improvements, it's no longer useful, and migrate the remaining tests to using the node-template where possible.

- [0] https://github.com/paritytech/substrate/pull/6313

Closes #49

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = [".", "proc-macro"]
+members = [".", "client", "proc-macro"]
 
 [package]
 name = "substrate-subxt"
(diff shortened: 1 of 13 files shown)


dvc94ch:
any ideas how to write tests for this? `node-cli` isn't published to crates.io, maybe we need a test-node here. might also be useful for testing in the future?

dvc94ch:
@ascjones can you please review? While the light client still doesn't work, it now also supports spawning full nodes, which means we can start working on removing the `#[ignored]` from our tests.

dvc94ch:
Enabled all non subxt tests to use the node-template from git. They are no longer `#[ignore]`

ascjones:
I'll have a look tomorrow

client/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    // Spawn background task.
+    let session = RpcSession::new(to_front.clone());
+    let mut from_front = from_front.compat();
+    task::spawn(poll_fn(move |cx| {
+        loop {
+            match Pin::new(&mut from_front).poll_next(cx) {
+                Poll::Ready(Some(message)) => {
+                    let mut to_front = to_front.clone().sink_compat();
+                    let message = message.unwrap();
```
ascjones, line 246: What does an `Err` here mean?

dvc94ch:
@ascjones added a README section

Cargo.toml
```diff
@@ -16,33 +16,44 @@ description = "Submit extrinsics (transactions) to a substrate node via RPC"
...
+frame-system = { version = "2.0.0-rc3", package = "frame-system" }
+node-template = { git = "https://github.com/paritytech/substrate" }
+pallet-balances = { version = "2.0.0-rc3", package = "pallet-balances" }
+sp-keyring = { version = "2.0.0-rc3", package = "sp-keyring" }
+substrate-subxt-client = { path = "client" }
+tempdir = "0.3.7"
+
+[patch.crates-io]
+sc-network = { git = "https://github.com/paritytech/substrate" }
+sc-service = { git = "https://github.com/paritytech/substrate" }
```
ascjones, line 59: What are these patches for? Something not yet included in `rc3`? It might prevent publishing the main crate

dvc94ch: The problem is that node template isn't published, but it's required for the tests. So this makes sure that when it's checked out locally it uses the same versions as the template. I was hoping this trick would allow publish to crates.io

dvc94ch:
The library still is light weight, only compiling the tests takes longer, if you add it as a dep, without the light client enabled it shouldn't increase compile times

dvc94ch:
I'd love this to be the defacto standard way of building substrate based applications. We're using it heavily and once your metadata work is finished, I think it'll be on track to be that. We're building tools like dart bindgen that will allow us to rapidly create flutter ui's using subxt.

PR #93, opened 2020-04-21
Double map and plain storage support, introduce macros

dvc94ch:
Simplifies trait bounds.
Allows setting a custom client.
Adds support for plain and double-map storage items.


dvc94ch:
I added a proc-macro to generate client code for substrate modules. I'll look into generating test cases too which should fix your comment.

dvc94ch:
Sorry, this PR has gotten a little out of hand. It still needs more field testing, and the `substrate-subxt` crate should probably be split into `substrate-subxt-core` and `substrate-subxt`, so the proc macros can be used for the default modules.

```rust
#[module]
pub trait Balances: System {
    type Balance: Parameter
        + Member
        + AtLeast32Bit
        + codec::Codec
        + Default
        + Copy
        + MaybeSerialize
        + Debug
        + From<<Self as System>::BlockNumber>;
}

#[derive(Clone, Decode, Default)]
pub struct AccountData<Balance> {
    pub free: Balance,
    pub reserved: Balance,
    pub misc_frozen: Balance,
    pub fee_frozen: Balance,
}

#[derive(Encode, Store)]
pub struct AccountStore<'a, T: Balances> {
    #[store(returns = AccountData<T::Balance>)]
    pub account_id: &'a <T as System>::AccountId,
}

#[derive(Call, Encode)]
pub struct TransferCall<'a, T: Balances> {
    pub to: &'a <T as System>::Address,
    #[codec(compact)]
    pub amount: T::Balance,
}

#[derive(Debug, Decode, Eq, Event, PartialEq)]
pub struct TransferEvent<T: Balances> {
    pub from: <T as System>::AccountId,
    pub to: <T as System>::AccountId,
    pub amount: T::Balance,
}

impl Balances for KusamaRuntime {
    type Balance = u128;
}

subxt_test!({
    name: transfer_test_case,
    runtime: KusamaRuntime,
    account: Alice,
    step: {
        state: {
            alice: AccountStore { account_id: &alice },
            bob: AccountStore { account_id: &bob },
        },
        call: TransferCall {
            to: &bob,
            amount: 10_000,
        },
        event: TransferEvent {
            from: alice.clone(),
            to: bob.clone(),
            amount: 10_000,
        },
        assert: {
            assert_eq!(pre.alice.free, post.alice.free - 10_000);
            assert_eq!(pre.bob.free, post.bob.free + 10_000);
        },
    },
});
```

dvc94ch:
Mmh one problem is that the nonces are wrong when running multiple tests in parallel. Is it possible to disable nonce checking on dev chains?

4meta5:
https://stackoverflow.com/questions/44947914/how-to-limit-the-number-of-test-threads-in-cargo-toml

dvc94ch:
Due to possible races it looks like it's the only way to do it. If we shared the client and set the account based on the thread I'd, arbitrary interleaving of tests will cause failures.

proc-macro/src/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    let mut bindings = vec![];
+    for variant in s.variants() {
+        for binding in variant.bindings() {
+            bindings.push(binding);
+        }
+    }
+    bindings
+}
+
+pub fn module_name<'a>(generics: &'a syn::Generics) -> &'a syn::Path {
```
ascjones, line 20: Can these lifetime be elided?

dvc94ch:
Is it ok if I just add a test for plain? The system, balances and contracts modules don't contain any `double_maps`. I can follow up with splitting subxt into a core module and update system, balances and contracts to use the proc-macros. the generic-asset module contains some `double_map`'s, I could add support for it.

PR #95, opened 2020-04-27
adding support for StorageValue and StorageDoubleMap

brenzi:
* plain insert without unit tests. might want to add some.


ascjones:
Thanks @brenzi but this has now been implemented via #93

PR #99, opened 2020-04-30
Staking support

Demi-Marie:
This is incomplete, but I would still like a review.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,4 @@ substrate-subxt-client = { version = "0.4.0", path = "client" }
 tempdir = "0.3.7"
 test-node = { path = "test-node" }
 wabt = "0.10.0"
+assert_matches = "1.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "client", "proc-macro", "test-node"]
 
 [package]
 name = "substrate-subxt"
-version = "0.9.0"
+version = "0.10.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
@@ -42,7 +42,7 @@ sp-rpc = { version = "2.0.0-rc4", package = "sp-rpc" }
 sp-core = { version = "2.0.0-rc4", package = "sp-core" }
 sc-rpc-api = { version = "0.8.0-rc4", package = "sc-rpc-api" }
 sp-transaction-pool = { version = "2.0.0-rc4", package = "sp-transaction-pool" }
-substrate-subxt-client = { version = "0.1.0", path = "client", optional = true }
+substrate-subxt-client = { version = "0.2.0", path = "client", optional = true }
 substrate-subxt-proc-macro = { version = "0.9.0", path = "proc-macro" }
 sp-std = "2.0.0-rc4"
 application-crypto = { version = "2.0.0-rc4", package = "sp-application-crypto", default-features = false }
@@ -59,6 +59,6 @@ wabt = "0.9.2"
 frame-system = { version = "2.0.0-rc4", package = "frame-system" }
 pallet-balances = { version = "2.0.0-rc4", package = "pallet-balances" }
 sp-keyring = { version = "2.0.0-rc4", package = "sp-keyring" }
-substrate-subxt-client = { version = "0.1.0", path = "client" }
+substrate-subxt-client = { version = "0.2.0", path = "client" }
 tempdir = "0.3.7"
 test-node = { path = "test-node" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,12 +53,35 @@ sp-authority-discovery = "2.0.0-rc4"
 pallet-staking = "2.0.0-rc4"
 
 [dev-dependencies]
-async-std = { version = "=1.5.0", features = ["attributes"] }
+async-std = { version = "1.5.0", features = ["attributes"] }
 env_logger = "0.7.1"
 wabt = "0.9.2"
+wabt-sys = "=0.7.1" # pinned because 0.7.2 fails to compile
 frame-system = { version = "2.0.0-rc4", package = "frame-system" }
 pallet-balances = { version = "2.0.0-rc4", package = "pallet-balances" }
 sp-keyring = { version = "2.0.0-rc4", package = "sp-keyring" }
 substrate-subxt-client = { version = "0.2.0", path = "client" }
 tempdir = "0.3.7"
 test-node = { path = "test-node" }
+
+[patch.crates-io]
+frame-metadata = { git = "https://github.com/paritytech/substrate", package = "frame-metadata" }
+frame-support = { git = "https://github.com/paritytech/substrate", package = "frame-support" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", package = "sp-runtime" }
+sp-version = { git = "https://github.com/paritytech/substrate", package = "sp-version" }
+pallet-indices = { git = "https://github.com/paritytech/substrate", package = "pallet-indices" }
+sp-rpc = { git = "https://github.com/paritytech/substrate", package = "sp-rpc" }
+sp-core = { git = "https://github.com/paritytech/substrate", package = "sp-core" }
+sc-rpc-api = { git = "https://github.com/paritytech/substrate", package = "sc-rpc-api" }
+sp-transaction-pool = { git = "https://github.com/paritytech/substrate", package = "sp-transaction-pool" }
+sp-std = { git = "https://github.com/paritytech/substrate" }
+application-crypto = { git = "https://github.com/paritytech/substrate", package = "sp-application-crypto", default-features = false }
+sp-finality-grandpa = { git = "https://github.com/paritytech/substrate" }
+sp-consensus-babe = { git = "https://github.com/paritytech/substrate" }
+pallet-im-online = { git = "https://github.com/paritytech/substrate" }
+sp-authority-discovery = { git = "https://github.com/paritytech/substrate" }
+pallet-staking = { git = "https://github.com/paritytech/substrate" }
+sc-service = { git = "https://github.com/paritytech/substrate" }
+sc-cli = { git = "https://github.com/paritytech/substrate" }
+sc-transaction-pool = { git = "https://github.com/paritytech/substrate" }
+jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,8 @@ keywords = ["parity", "substrate", "blockchain"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [features]
+kusama = []
 default = ["kusama"]
-kusama = ["polkadot"]
 
 [dependencies]
 log = "0.4"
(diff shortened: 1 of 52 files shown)


dvc94ch:
I think it might be better to split the crate in to core and use the proc macros for deriving the boilerplate, as it's going to be a lot for a module like staking...

DemiMarie-parity:
@dvc94ch I was not aware that there were macros available, thanks!  Some of the struct and type definitions were copied verbatim from the staking module in the runtime, but I can switch to using the runtime crate from `crates.io` instead.

ascjones:
proc macro split is merged now so can try it out here

DemiMarie-parity:
The proc macros seem broken.  At the very least, I could not use them to obtain the validator set.

dvc94ch:
Can you elaborate? What does cargo expand say? What exactly did you try?

ascjones:
@DemiMarie-parity presumably #111 is what you mean

DemiMarie-parity:
@ascjones indeed it is.

DemiMarie-parity:
The proc macros are now fixed.

src/frame/staking.rs
```diff
@@ -111,6 +114,7 @@ impl Default for ValidatorPrefs {
 /// The subset of the `frame::Trait` that a client must implement.
 #[module]
 pub trait Staking: System {}
+impl<T: System> Staking for T {}
```
dvc94ch, line 117: Why this? A runtime that implements system may not impl staking

DemiMarie-parity: I did this because `Staking` has no new items beyond those in `System`, but you are correct that this impl is too broad.

DemiMarie-parity:
@dvc94ch can you re-review?

DemiMarie-parity:
@dvc94ch I fixed everything except adding tests.  The difficulty with testing is that I am not sure how to set up validators.  Many of these cannot change without breaking [Ledgeracio](/paritytech/ledgeracio)’s on-device component, which is implemented as C code running on a Ledger device, so breaking changes should be extremely rare.

dvc94ch:
Ultimately it's up to @ascjones, looks good from my end

ascjones:
#119 merged now, so branch needs updating

DemiMarie-parity:
@ascjones can you review?

README.md
```diff
@@ -6,6 +6,9 @@ A library to **sub**mit e**xt**rinsics to a [substrate](https://github.com/parit
 
 See [examples](./examples).
 
+If you use `#[derive(Call)]` without `#[module]` in the same module, you will get errors
+complaining about an undefined method with a name starting with `with_`.
```
ascjones, line 10: Please open an issue for this, or provide some context here

src/frame/staking.rs
```diff
@@ -0,0 +1,313 @@
...
+    pub index: EraIndex,
+    /// Moment of start expresed as millisecond from `$UNIX_EPOCH`.
+    ///
+    /// Start can be none if start hasn't been set for the era yet,
+    /// Start is set on the first on_finalize of the era to guarantee usage of `Time`.
+    pub start: Option<u64>,
+}
+
+/// Data type used to index nominators in the compact type
+pub type NominatorIndex = u32;
```
ascjones, line 60: Why is this and ValidatorIndex concrete types, whereas Session::SessionIndex is an associated type?

src/signer.rs
```diff
@@ -152,3 +147,35 @@ where
         )))
     }
 }
+
+impl<T, S, E> Signer<T, S, E> for Box<dyn Signer<T, S, E> + Send + Sync>
```
ascjones, line 151: What's this for?

DemiMarie-parity: It allows a boxed `Signer` to be used where a `Signer` is required.

README.md
DemiMarie-parity: It took me a long time to figure this out, so I added this for the benefit of others.

src/frame/staking.rs
DemiMarie-parity: Fixed

src/signer.rs
dvc94ch: I think what @ascjones means is that `Box<T>` implements `Deref` so anything taking a `&dyn Signer` should work.

Cargo.toml
```diff
@@ -27,22 +31,23 @@ serde_json = "1.0"
...
-sp-finality-grandpa = "2.0.0-rc2"
-sp-consensus-babe = "0.8.0-rc2"
-pallet-im-online = "2.0.0-rc2"
-sp-authority-discovery = "2.0.0-rc2"
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-finality-grandpa = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-consensus-babe = { git = "https://github.com/paritytech/substrate", branch = "master" }
+pallet-im-online = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-authority-discovery = { git = "https://github.com/paritytech/substrate", branch = "master" }
+polkadot = { git = "https://github.com/paritytech/polkadot", branch = "master", optional = true, default-features = false, features = ["std"], package = "polkadot-primitives" }
```
ascjones, line 50: We publish to crates.io so we can't have github dependencies, need to find out whether there are plans to publish `polkadot-primitives`

ascjones:
Any progress on some tests for this?

DemiMarie-parity:
@ascjones I have some manual tests, but automated tests are going to need to wait until Ledgeracio is finished.  There are no tests for any of the other FRAME modules either, so this is not a regression.

ascjones:
>  There are no tests for any of the other FRAME modules either,

The other two existing modules `balances` and `contracts` do have tests.

DemiMarie-parity:
@ascjones would you be willing to help with the tests?  I am less worried about the types getting out of sync, as that would require offline signers (such as the Ledgeracio device) to be updated as well.

dvc94ch:
So how's this PR coming along? I'm working on deploying a testnet but there seems no simple way to get the keys configured in the genesis config loaded in the keystore. I was thinking about building a cli tool for that, but it seems it's what ledgeracio intends to do. Is it usable enough to call author_setKey and check that the keys are loaded?

DemiMarie-parity:
@dvc94ch That is out of scope for ledgeracio.  Ledgeracio is a special-purpose tool intended for staking operations, not a general purpose CLI for interacting with Substrate-based blockchains.  I would love to see such a tool, but it would be a much larger amount of work than ledgeracio itself.

dvc94ch:
How is setKeys not a staking operation? You can only set the public keys with the staking module, you need to add the key to the keystore too...

DemiMarie-parity:
The staking module has a *lot* of dependencies, which makes testing hard.  Should I move that to a separate crate?

DemiMarie-parity:
@ascjones how would you prefer me to implement testing?  The two options I can think of are:

- Add new modules to the test node, and try to come up with test mocks where possible.  This will be a lot of effort as `substrate-subxt` becomes more and more comprehensive.  There is also a risk that the test node could diverge form the production node.
- Change the test suite to use an actual Substrate node on a development network.  This has the advantage of “flying what you test and testing what you fly” ― since this is how `substrate-subxt` will be used in production, it is a better fit for real-world usage.  It also represents a potentially substantial one-time effort.

ascjones:
Hi @DemiMarie-parity, apologies for the late response.

For now I would accept tests which simply against the substrate `--dev` chain (using the default node runtime). This should have an `#[ignored]` attribute (or we can add a feature for these type of tests). This can then be run manually (the contracts tests are currently run this way). In the future we could run these tests on CI, but for now it is not too much work for me to run them locally before merging PRs.

> Add new modules to the test node, and try to come up with test mocks where possible. This will be a lot of effort as substrate-subxt becomes more and more comprehensive. There is also a risk that the test node could diverge form the production node.

I agree, we should not be adding every single module to the test node since it would become brittle, however we might consider it for core modules that are reasonably stable. Staking may well be one of those but I think it's something we can consider in the future.

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+/// Clipped Exposure of validator at era.
+///
+/// This is similar to [`ErasStakers`] but number of nominators exposed is reduced to the
+/// `T::MaxNominatorRewardedPerValidator` biggest stakers.
+/// (Note: the field `total` and `own` of the exposure remains unchanged).
+/// This is used to limit the i/o cost for the nominator payout.
+///
+/// This is keyed fist by the era index to allow bulk deletion and then the stash account.
+///
+/// Is it removed after `HISTORY_DEPTH` eras.
```
dvdplm, line 238:
```suggestion
/// It is removed after `HISTORY_DEPTH` eras.
```

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+#[derive(Encode, Copy, Clone, Debug, Store)]
+pub struct ErasStakersClippedStore<T: Staking> {
+    #[store(returns = Exposure<T::AccountId, T::Balance>)]
+    /// Era index
+    pub era: EraIndex,
+    /// Stash account of the validator
+    pub validator_stash: T::AccountId,
+}
+
+/// The active era information, it holds index and start.
```
dvdplm, line 249:
```suggestion
/// The active era information, holds index and start.
```

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+#[derive(Encode, Copy, Clone, Debug, Hash, PartialEq, Eq, Ord, PartialOrd, Store)]
+pub struct ActiveEraStore<T: Staking> {
+    #[store(returns = Option<ActiveEraInfo>)]
+    /// Marker for the runtime
+    pub _runtime: PhantomData<T>,
+}
+
+/// Declare the desire to validate for the origin controller.
+///
+/// Effects will be felt at the beginning of the next era.
```
dvdplm, line 262:
```suggestion
/// Effective from the beginning of the next era.
```
or perhaps "If successful, effective from the beginning of the next era."?

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+///
+/// The dispatch origin for this call must be _Signed_ by the controller, not the stash.
+/// And, it can be only called when [`EraElectionStatus`] is `Closed`.
```
dvdplm, line 265:
```suggestion
/// Can only be called when [`EraElectionStatus`] is `Closed`.
```

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+///
+/// # <weight>
+/// - Independent of the arguments. Insignificant complexity.
+/// - Contains a limited number of reads.
```
dvdplm, line 269: Does "limited" here mean "exact number varies but is in all cases bounded"?

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+pub struct ValidateCall<T: Staking> {
+    /// Runtime marker.
+    pub _runtime: PhantomData<T>,
+    /// Validation preferences.
+    pub prefs: ValidatorPrefs,
+}
+
+/// Declare the desire to nominate `targets` for the origin controller.
+///
+/// Effects will be felt at the beginning of the next era. This can only be called when
```
dvdplm, line 287: same as above

src/frame/staking.rs
```diff
@@ -0,0 +1,315 @@
...
+/// [`EraElectionStatus`] is `Closed`.
+///
+/// The dispatch origin for this call must be _Signed_ by the controller, not the stash.
+/// And, it can be only called when [`EraElectionStatus`] is `Closed`.
```
dvdplm, line 291: same as above

src/rpc.rs
```diff
@@ -96,6 +96,18 @@ impl From<u32> for BlockNumber {
     }
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+/// System properties for a Substrate-based runtime
+pub struct Properties {
```
dvdplm, line 102: I'd prefer a less generic name, e.g. `SystemProperties`.

dvdplm (commented):
Documentation nitpicks and questions.

Demi-Marie:
@dvdplm I don’t have write access to this repository anymore, so I will close this PR and open another one.

dvc94ch:
You don't work for parity anymore?

Demi-Marie:
@dvc94ch I still work for Parity, but am no longer part of the @paritytech GitHub org.

DemiMarie:
@dvc94ch can we go ahead and merge this as-is?  This is probably *the* last remaining blocker for Ledgeracio.

DemiMarie:
Also @ascjones @dvdplm.

src/frame/staking.rs
```diff
@@ -0,0 +1,252 @@
...
+/// The dispatch origin for this call must be _Signed_ by the controller, not the stash.
+/// Can only be called when [`EraElectionStatus`] is `Closed`.
+#[derive(Call, Encode, Debug)]
+pub struct NominateCall<T: Staking> {
+    /// The targets that are being nominated
+    pub targets: Vec<T::Address>,
+}
+
+#[cfg(all(test, feature = "integration-tests"))]
+mod tests {
```
dvdplm, line 200: Please clean up the tests before asking for a review. This is PoC-level stuff here. :)

DemiMarie: The `integration-tests` feature is necessary because the tests don’t work without a running Polkadot and/or Kusama node.  I could have written

```rust
#[cfg(test)]
#[cfg(feature = "integration-tests")]
```
or added `#[cfg(feature = "integration-tests")]` on each test, but this is equivalent and shorter.

dvc94ch: I think he is referring to looking how it is handled already by other tests. Even better would be to add the staking module to the test runtime we use for tests

DemiMarie: I tried adding the staking module, but it would up pulling in a huge portion of Kusama.  I think we would be better off using an actual Kusama node instead.

dvc94ch: Is there a reason to use a feature gate instead of #[ignore]?

DemiMarie: No, fixed.

DemiMarie: I also fixed the test suite to retry on temporary errors.

dvdplm: No I was referring to the test code itself: it contains a bunch of my lame attempts at figuring out how to test this. It's not something we want to actually review. :/

I think @ascjones added the `integration-tests` feature as a way to (eventually) run these tests in CI. There's a problem though: the contracts tests he added in #163 assumes the "remote" node has the contracts pallet, which is true for a substrate node, but isn't for polkadot/kusama. This means `cargo test --feature integration-tests` will pass or fail tests depending on which flavour of node you're running. :/

@demi the retry thing doesn't actually work for me. If you use `_and_watch()` and check on what the outcome actually is you'll see that it's not ok.

dvdplm (approved):
I think this is good enough. The tests are imperfect but I do not think this is the PR where we fix the tricky problem of solid integration testing and state setup.

src/frame/staking.rs
ascjones: > EDIT: I think the problem is that if you run a single node the chain isn't progressing so when we submit several extrinsics they end up failing in random order with "Extrinsic Usurped" or "Priority is too low: (1155024648500 vs 1155024648500)"

Yeah usually that is probably because the nonce needs incrementing. I came up with a rudimentary solution here which appears to work consistently: https://github.com/paritytech/substrate-subxt/blob/aj-contract-call/src/frame/contracts.rs#L175

> I tried adding the staking module, but it would up pulling in a huge portion of Kusama. I think we would be better off using an actual Kusama node instead.

I agree. I assume these tests would work against `node-runtime` too? If so then https://github.com/paritytech/substrate-subxt/issues/169 might be a solution.

That said for this PR I am happy for the tests to be run manually against a local node of whatever flavour.

README.md
ascjones: I've raised an issue for it to improve the error message: https://github.com/paritytech/substrate-subxt/issues/170. But probably we should remove from the readme since it doesn't make sense without any context.

src/frame/system.rs
```diff
@@ -101,6 +101,8 @@ pub trait System {
         + MaybeSerialize
         + Debug
         + MaybeDisplay
+        + Encode
+        + Decode
```
ascjones, line 105: I think these are redundant, they come with `Paramater`

ascjones (commented):
Looks good, just the `kusama` feature needs removing I think

src/frame/system.rs
dvdplm: q: `Parameter` also comes with `fmt::Debug`, so should I drop that too?

PR #102, opened 2020-05-04
Split subxt

dvc94ch:
@ascjones all tests pass


ascjones:
@dvc94ch thanks I'll take a look later today. Can you merge master again?

src/events.rs
```diff
@@ -194,6 +201,10 @@ impl<T: System> EventsDecoder<T> {
                         // PhantomData is size 0
                         return Ok(())
                     }
+                    if name.contains("DispatchError") {
+                        let error = DispatchError::decode(input)?;
+                        return Err(error.into())
+                    }
```
dvc94ch, line 207: Not sure if this makes any sense, not had a `DispatchError::Other` since you updated to alpha.7

proc-macro/src/store.rs
```diff
@@ -89,21 +85,27 @@ pub fn store(s: Structure) -> TokenStream {
         .filter_map(|bi| bi.ast().attrs.iter().filter_map(parse_returns_attr).next())
         .next()
         .expect("#[store(returns = ..)] needs to be specified.");
+    let filtered_fields: Vec<_> = fields
+        .iter()
+        .filter(|(field, _)| field.to_string() != "_runtime")
```
ascjones, line 90: Could we ignore marker fields by `Phantom<T>` rather than by the field name which is brittle?

proc-macro/src/test.rs
```diff
@@ -305,27 +317,45 @@ impl Step {
...
+                state_param,
+                } = state;
+                let state_struct = quote! {
+                    struct State<#(#state_param),*> {
+                        #(#state_name: #state_param,)*
+                    }
+                };
+                let build_struct = quote! {
+                    #(
+                        let #state_name = client.fetch(#state, None).await.unwrap().unwrap();
```
ascjones, line 336: Might be better to have `expect` instead of `unwrap` so at least it's easier to debug if the fetch fails.

proc-macro/src/test.rs
```diff
@@ -335,20 +365,13 @@ impl Step {
                 .unwrap();
 
             #(
-                assert_eq!(
-                    result.find_event::<#event_name<_>>().unwrap(),
-                    Some(#event)
-                );
+                let event = result.find_event::<#event_name<_>>().unwrap().unwrap();
```
ascjones, line 368: Same here prefer `expect` for easier debugging tests

src/events.rs
ascjones: Probably because I increased the gas limit of instantiate which no longer causes the `DispatchError` to be thrown. Still we need to handle contract call errors - could you remove this from this PR we can fix it separately via #103

src/frame/contracts.rs
```diff
@@ -103,133 +99,63 @@ pub struct CallCall<'a, T: Contracts> {
...
+            ).unwrap();
+            let code_hash = buf.into();
+        },
+        step: {
+            call: PutCodeCall {
+                _runtime: PhantomData,
+                code: &wasm,
+            },
+            event: CodeStoredEvent {
+                code_hash,
```
ascjones, line 145: This test fails for me
```
thread 'frame::contracts::tests::test_put_code_and_instantiate' panicked at 'assertion failed: `(left == right)`
  left: `Some(CodeStoredEvent { code_hash: 0x4d8b39b728b35fee98001095400a61f5c2ef91e0c807b1f1754b17f0397360ba })`,
 right: `Some(CodeStoredEvent { code_hash: 0x0000000000000000000000000000000000000000000000000000000000000000 })`', src/lib.rs:1:1
```

dvc94ch: Weird, how can the code hash possibly be all zeroes? Anyway, I updated the test to not care, can you try if it works for you now?

src/lib.rs
```diff
@@ -334,11 +340,7 @@ where
         account_id: &<T as System>::AccountId,
         call: C,
     ) -> Result<SignedPayload<Encoded, <E as SignedExtra<T>>::Extra>, Error> {
-        let account_nonce = self
-            .fetch(AccountStore(account_id), None)
-            .await?
-            .unwrap()
-            .nonce;
+        let account_nonce = self.account(account_id).await?.unwrap().nonce;
```
ascjones, line 343: I know this was already there but I'd like the `unwrap()` replacing with `ok_or("Account not found")?`

src/lib.rs
```diff
@@ -365,12 +367,7 @@ where
         let account_id = S::Signer::from(signer.public()).into_account();
         let nonce = match nonce {
             Some(nonce) => nonce,
-            None => {
-                self.fetch(AccountStore(&account_id), None)
-                    .await?
-                    .unwrap()
-                    .nonce
-            }
+            None => self.account(&account_id).await?.unwrap().nonce,
```
ascjones, line 370: Same here replace `unwrap` with `ok_or`

src/frame/contracts.rs
ascjones: Yes passes now

src/lib.rs
dvc94ch: I don't know if that can actually ever happen. We need a test case for `Option<T>` and `T`. I think if the returns type is declared as `Option<T>` it should work out. Fetch always returns a default value. I'll remove this for now.

PR #115, opened 2020-05-26
Add support for sudo

### LINKED ISSUE #112: module_with_calls is not public caused I failed to compose two calls
Hi there,

This issue might be quite the same with [How could I send a sudo call](https://github.com/paritytech/substrate-subxt/issues/84). 

I'm using subxt-0.6, which works fast and as expected. But I find out that this interface [module_with_calls](https://github.com/paritytech/substrate-subxt/blob/master/src/metadata.rs#L92) become private while I try to upgrade to 0.7. Here is code sample with subxt-0.6
```rust
let proposal = client.metadata().module_with_calls("MyModule")
		.and_then(|module| module.call("my_call", my_call_args))
		.map_err(|_| crate::Error::SubxtError("failed to compose a sudo call"))?;
let call = Call::new("Sudo", "sudo", proposal);
```

So now I cannot upgrade subxt to 0.7 or higher, can that interface change to be public? Or maybe you can give me a better solution instead of changing that interface.

Thanks!

dvc94ch:
@Dengjianping Do you want to finish it? Test is still failing because `ensure_signed` doesn't accept transactions with root origin. Closes #112

### DIFF
--- a/src/frame/mod.rs
+++ b/src/frame/mod.rs
@@ -34,6 +34,7 @@ use sp_core::storage::StorageKey;
 
 pub mod balances;
 pub mod contracts;
+pub mod sudo;
 pub mod system;
 
 /// Store trait.
--- a/src/frame/sudo.rs
+++ b/src/frame/sudo.rs
@@ -1,0 +1,61 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+// This file is part of substrate-subxt.
+//
+// subxt is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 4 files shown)


Dengjianping:
Sorry for trying this so late.

I ran your sudo example as well as my own, I got error like this:
```
Rpc(Request(Error { code: ServerError(1010), message: "Invalid Transaction", data: Some(String("BadProof")) }))
```
It looks there's problem on the signer.

dvc94ch:
Have you tried `purge-chain --dev`?

Dengjianping:
Yes, I purged the whole chain. By the way, current rev I'm using is kind of old, does older rev cause this problem? I see that subsxt is using rc2.

dvc94ch:
@Dengjianping yes, that could definitely cause weird issues. I'm closing this PR for now.

PR #120, opened 2020-05-31
Fix optional store items.

dvc94ch:


proc-macro/src/store.rs
```diff
@@ -127,7 +137,7 @@ pub fn store(s: Structure) -> TokenStream {
                 #args
             ) -> core::pin::Pin<Box<dyn core::future::Future<Output = Result<#ret, #subxt::Error>> + Send + 'a>> {
                 let #marker = core::marker::PhantomData::<T>;
-                Box::pin(self.fetch(#build_struct, None))
+                Box::pin(self.#fetch(#build_struct, None))
```
en, line 140: Could the second parameter be exposed? It is convenient for the user to query the store at a specified block height.

dvc94ch: is this really used often? one down side is that no storage parameter can be called hash now, maybe it should have a weird name like __hash or something, but that will show up in the docs

en: Thanks, a project I am working on needs this feature. It happens during the handshake of interchain communication, a chain needs to ensure that a state is committed to the opposite chain at a specified height.

PR #121, opened 2020-06-04
Zero sized types

dvc94ch:


ascjones:
Can you describe briefly what this fix was about?

dvc94ch:
Using a zero sized type like `()` is ok, it's not an error. For our [identity](https://github.com/sunshine-protocol/substrate-identity) module, we set the System AccountData type to `()`.

src/events.rs
```diff
@@ -87,51 +86,24 @@ pub struct EventsDecoder<T> {
...
-        decoder.register_type_size::<u32>("SessionIndex")?;
-        decoder.register_type_size::<u32>("PropIndex")?;
-        decoder.register_type_size::<u32>("ProposalIndex")?;
-        decoder.register_type_size::<u32>("AuthorityIndex")?;
-        decoder.register_type_size::<u64>("AuthorityWeight")?;
-        decoder.register_type_size::<u32>("MemberCount")?;
-        decoder.register_type_size::<T::AccountId>("AccountId")?;
-        decoder.register_type_size::<T::BlockNumber>("BlockNumber")?;
-        decoder.register_type_size::<T::Hash>("Hash")?;
-        decoder.register_type_size::<u8>("VoteThreshold")?;
```
ascjones, line 116: While this is an extremely imperfect solution, IIRC these type size registrations are required in case we receive events from modules not defined in `subxt` here: https://github.com/paritytech/substrate-subxt/blob/master/src/rpc.rs#L479 so we can skip over them. Of course with improvements to the metadata, potentially integrating `https://github.com/paritytech/scale-info` this would no longer be required. Also `desub` has an alternative approach which utilizes mappings from `polkadot-js` https://github.com/paritytech/desub. See https://github.com/paritytech/substrate-subxt/issues/72

dvc94ch: I see, I'll revert for now. scale-info sounds cool

dvc94ch: Done. Is there any eta on scale-info? Seems like that would solve a lot of problems

ascjones: I'm currently working on generics for a `v0.1` release, and once I've implemented associated type representation I will experiment with substrate integration.

src/lib.rs
```diff
@@ -364,8 +362,8 @@ where
     /// Returns an events decoder for a call.
     pub fn events_decoder<C: Call<T>>(&self) -> Result<EventsDecoder<T>, Error> {
```
ascjones, line 363: This no longer needs to return a `Result`

PR #126, opened 2020-06-22
Events sub

dvc94ch:
When subscribing to events there are two common use cases:

- Finding events emitted after submitting an extrinsic
- Subscribing to a specific event

This makes it easier to subscribe to a specific event.

@ascjones would it be possible to create a new release after this? Looks like we have all we need for sunshine-identity for now.

### DIFF
--- a/src/frame/balances.rs
+++ b/src/frame/balances.rs
@@ -114,10 +114,12 @@ mod tests {
             Error,
             RuntimeError,
         },
+        events::EventsDecoder,
         signer::{
             PairSigner,
             Signer,
         },
+        subscription::EventSubscription,
         system::AccountStoreExt,
         tests::{
             test_client,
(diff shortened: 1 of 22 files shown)


dvc94ch:
Test failure is spurious

src/rpc.rs
```diff
@@ -425,58 +434,13 @@ impl<T: System> ExtrinsicSuccess<T> {
...
-                extrinsic: ext_hash,
-                events,
-            })
-        } else {
-            Err(format!("No events found for block {}", block_hash).into())
-        }
-    }
-    unreachable!()
+/// Returns the index of an extrinsic in a block.
+pub fn find_extrinsic<T: Runtime>(
```
ascjones, line 438: This `pub fn` could be made inline since it's only used above in `submit_and_watch_extrinsic`

src/frame/balances.rs
```diff
@@ -201,4 +203,28 @@ mod tests {
...
+    async fn test_transfer_subscription() {
+        env_logger::try_init().ok();
+        let alice = PairSigner::new(AccountKeyring::Alice.pair());
+        let bob = AccountKeyring::Bob.to_account_id();
+        let (client, _) = test_client().await;
+        let sub = client.subscribe_events().await.unwrap();
+        let mut decoder = EventsDecoder::<TestRuntime>::new(client.metadata().clone());
+        decoder.with_balances();
+        let mut sub = EventSubscription::<TestRuntime>::new(sub, decoder);
+        sub.filter_event::<TransferEvent<_>>();
```
ascjones, line 217: Might be nice to have a higher level api for subscribing to a particular event type and doing the decoding, but no need for this PR

ascjones:
Build is failing

PR #135, opened 2020-07-06
Document the #[module] macro

dvdplm:


proc-macro/src/lib.rs
```diff
@@ -30,6 +30,54 @@ use synstructure::{
     Structure,
 };
 
+/// Register type sizes for [EventsDecoder](struct.EventsDecoder.html)
```
ascjones, line 33: I think we should also mention the module name constant, which will be the same as the trait ident. This in turn must match the name of the substrate pallet. Indeed the codegen for events and calls depends on this constant, so we might mention that this attribute is required

proc-macro/src/lib.rs
```diff
@@ -30,6 +30,54 @@ use synstructure::{
...
+///
+/// Example:
+///
+/// ```ignore
+/// #[module]
+/// pub trait Herd: Husbandry {
+/// 	type Hooves: HoofCounter;
+/// 	type Wool: WoollyAnimal;
+/// 	#[module(ignore)]
+/// 	type Digestion: EnergyProducer + std::fmt::Debug;
```
ascjones, line 47: Indentation

proc-macro/src/lib.rs
```diff
@@ -30,6 +30,54 @@ use synstructure::{
...
+///     // Registers this modules types.
+///     fn with_herd(&mut self);
+/// }
+///
+/// impl<T: Herd> HerdEventsDecoder for
+///     substrate_subxt::EventsDecoder<T>
+/// {
+///     fn with_herd(&mut self) {
+///         self.with_husbandry();
+///         self.register_type_size::<T::Hooves>("Hoves");
```
ascjones, line 74:
```suggestion
///         self.register_type_size::<T::Hooves>("Hooves");
```

proc-macro/src/lib.rs
```diff
@@ -30,6 +30,54 @@ use synstructure::{
...
+///         self.register_type_size::<T::Wool>("Wool");
+///     }
+/// }
+/// ```
+///
+/// The size of the following types is registered by default: `bool, u8, u32, AccountId, AccountIndex, AuthorityId, AuthorityIndex, AuthorityWeight, BlockNumber, DispatchInfo, Hash, Kind, MemberCount, PhantomData, PropIndex, ProposalIndex, ReferendumIndex, SessionIndex, VoteThreshold`
```
ascjones, line 80:
```suggestion
/// The following type sizes are registered by default: `bool, u8, u32, AccountId, AccountIndex, AuthorityId, AuthorityIndex, AuthorityWeight, BlockNumber, DispatchInfo, Hash, Kind, MemberCount, PhantomData, PropIndex, ProposalIndex, ReferendumIndex, SessionIndex, VoteThreshold`
```

proc-macro/src/module.rs
```diff
@@ -69,7 +69,7 @@ fn events_decoder_trait_name(module: &syn::Ident) -> syn::Ident {
 fn with_module_ident(module: &syn::Ident) -> syn::Ident {
     format_ident!("with_{}", module.to_string().to_snake_case())
 }
-
+/// Attribute macro that registers the type sizes used by the module.
```
ascjones, line 72: Might want to mention about the module name const which is required for the other derive macros to work

proc-macro/src/lib.rs
dvdplm: Sorry, I don't see it, what is not right here?

ascjones: Looks like 1 space instead of 4
image

dvdplm: Oh, we use spaces here, good catch. Odd that my github shows me the right white space (as did my editor, but you're right, I had one space and one tab here).

PR #136, opened 2020-07-06
Add macOS's metadata files to gitignore

dvdplm:


dvdplm:
This is not something we need or want. Closing.

PR #137, opened 2020-07-06
Document the `Call` derive macro

dvdplm:


proc-macro/src/lib.rs
```diff
@@ -36,7 +36,43 @@ pub fn module(args: TokenStream, input: TokenStream) -> TokenStream {
...
+    /// #[derive(Call)]
+    /// pub struct FunStuffCall<T: MyTrait> {
+    ///     /// Runtime marker.
+    ///     pub _runtime: PhantomData<T>,
+    ///     /// The argument passed to the call..
+    ///     pub something: Vec<u8>,
+    /// }
+    /// ```
+    ///
+    /// When building a [Client](../substrate_subxt/struct.Client.html) parametrised to `MyRuntime`, you have access to
```
ascjones, line 68:
```suggestion
    /// When building a [Client](../substrate_subxt/struct.Client.html) parameterised to `MyRuntime`, you have access to
```

proc-macro/src/lib.rs
```diff
@@ -36,7 +36,43 @@ pub fn module(args: TokenStream, input: TokenStream) -> TokenStream {
...
+    ///  two new methods: `fun_stuff()` and `fun_stuff_and_watch()` by way of the derived `FunStuffExt` trait. The fields
```
ascjones, line 69: Might be good to quickly mention the difference between the two, i.e. `_and_watch()` allows waiting for events that are triggered as a result of this call.

proc-macro/src/lib.rs
```diff
@@ -36,7 +36,43 @@ pub fn module(args: TokenStream, input: TokenStream) -> TokenStream {
...
+    /// #[module]
+    /// pub trait MyTrait: System + Balances {}
+    ///
+    /// #[derive(Call)]
+    /// pub struct FunStuffCall<T: MyTrait> {
+    ///     /// Runtime marker.
+    ///     pub _runtime: PhantomData<T>,
+    ///     /// The argument passed to the call..
+    ///     pub something: Vec<u8>,
+    /// }
```
ascjones, line 65: I think it would be useful to mention that this struct here maps to corresponding substrate runtime call e.g.

```
decl_module! {
  ...
  pub fn fun_stuff(origin, something: Vec<u8>) -> DispatchResult { .. }
  ...
}

PR #138, opened 2020-07-06
Refactor: extract functions and types for creating extrinsics

ascjones:
Extracts the types and routines for creating extrinsics from the client so they can be used independently of the `Client` to construct extrinsics.

## todo

- [x] Sense check public API.


ascjones:
@niklasad1 I think remember you asking a while ago in Riot about constructing extrinsics offline - maybe this PR would be useful. /cc @dvdplm

Issue #141, opened 2020-07-19
improve events subscriptions


4meta5:
Might be nice to have a higher level api for subscribing to a particular event type and doing the decoding, but no need for this PR

_Originally posted by @ascjones in https://github.com/paritytech/substrate-subxt/pull/126#discussion_r444773385

We need an iterator-like API for subscribing to events for @sunshine-protocol CC @dvc94ch

dvc94ch:
@ascjones any progress on dynamic event decoding using scale-info?

ascjones:
No. It depends on integrating `scale-info` with substrate which I haven't even started yet. I'm busy with `ink! 3.0` work atm but will see if I can block out some time to look into it.

dvc94ch:
Any progress? We need more sophisticated event subscriptions, and without this, they won't work very reliably.

ascjones:
@dvc94ch I've started experimenting with this: I'm part way through implementing SCALE encoding for `scale-info` here: https://github.com/paritytech/scale-info/pull/19.

I'll do some more on it next week and let you know how I progress.

jsdw:
We have this now, and in the not distant future we'll start relyong on `scale_value` and so may provide back Value's for unknown events rather than skipping over the bytes, for more flexibility (but also for more robust/deduplicated decode logic)

PR #142, opened 2020-07-20
event subscriptions

4meta5:
The goal is to add a method on `Client<T>` that returns `EventSubscription<T>` for a specific Event. Related #141 

Still need to test.

### DIFF
--- a/examples/transfer_subscribe.rs
+++ b/examples/transfer_subscribe.rs
@@ -37,11 +37,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let dest = AccountKeyring::Bob.to_account_id().into();
 
     let client = ClientBuilder::<DefaultNodeRuntime>::new().build().await?;
-    let sub = client.subscribe_events().await?;
-    let mut decoder = EventsDecoder::<DefaultNodeRuntime>::new(client.metadata().clone());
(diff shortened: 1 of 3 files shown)


examples/transfer_subscribe.rs
```diff
@@ -37,11 +37,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let dest = AccountKeyring::Bob.to_account_id().into();
 
     let client = ClientBuilder::<DefaultNodeRuntime>::new().build().await?;
-    let sub = client.subscribe_events().await?;
-    let mut decoder = EventsDecoder::<DefaultNodeRuntime>::new(client.metadata().clone());
-    decoder.with_balances();
```
4meta5, line 42: @dvc94ch is there a way to form this call within `event_subscription` method I added below from the `E: Event<T>` associated type? 

Like this, but I can't use the proc_macro2 or quote....
```rust
let decoder_with_module: proc_macro2::TokenStream = quote!(quote::format_ident!(
        "decoder.with_{}()",
        E::MODULE.to_string().to_snake_case()
    )).into();
    decoder_with_module;
```

4meta5:
can't register the module types with the event decoder so this is a dead end

requires dynamic event decoding

PR #147, opened 2020-08-03
Fetch all from storage map

ascjones:
*WIP* Closes #144.

Mistakenly pushed the first part of this in https://github.com/paritytech/substrate-subxt/commit/663934ca374983308fa3cceccbbb58f59b6785b3.

## todo

- [ ] Create an `Iterator` to perform paging requests
- [ ] Add client config to override default page size
- [ ] Add method for creating the iterator in `proc_macro` e.g. `fn accounts_iter()`

PR #148, opened 2020-08-04
Storage map iter

dvc94ch:
Completes #147

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,7 @@ use sc_rpc_api::state::ReadProof;
 use sp_core::{
     storage::{
         StorageChangeSet,
+        StorageData,
         StorageKey,
     },
     Bytes,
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -185,6 +195,42 @@ impl<T: Runtime> Clone for Client<T> {
...
+    start_key: Option<StorageKey>,
+    keys: Vec<StorageKey>,
+}
+
+impl<T: Runtime, F: Store<T>> KeyIter<T, F> {
+    /// Returns the next key value pair from a map.
+    pub async fn next(&mut self) -> Result<Option<(StorageKey, F::Returns)>, Error> {
+        loop {
+            if let Some(key) = self.keys.pop() {
+                if let Some(value) = self.client.rpc.storage(&key, self.hash).await? {
```
ascjones, line 218: What about `self.client.rpc.query_storage` with the whole batch of keys for a single roundtrip?

src/lib.rs
```diff
@@ -185,6 +195,42 @@ impl<T: Runtime> Clone for Client<T> {
...
+    _marker: PhantomData<F>,
+    count: u32,
+    hash: Option<T::Hash>,
+    start_key: Option<StorageKey>,
+    keys: Vec<StorageKey>,
+}
+
+impl<T: Runtime, F: Store<T>> KeyIter<T, F> {
+    /// Returns the next key value pair from a map.
+    pub async fn next(&mut self) -> Result<Option<(StorageKey, F::Returns)>, Error> {
```
ascjones, line 215: I'm wondering whether it's worth making this `impl Stream` too? Then you could do e.g. `client.account_iter().collect::<Vec<_>>().await`

src/lib.rs
```diff
@@ -200,45 +246,59 @@ impl<T: Runtime> Client<T> {
         &self.metadata
     }
 
-    /// Fetch a StorageKey with default value.
-    pub async fn fetch_or_default<F: Store<T>>(
+    /// Fetch a StorageKey an optional storage key.
```
ascjones, line 249: I realise this comment was already there but it doesn't make any sense

src/lib.rs
dvc94ch: Would be nice, but it seems terribly annoying to do. Maybe I'm missing a simple solution, but basically once you go from manually implemented futures and streams to async syntax there isn't an easy way back.

ascjones: Yeah. I can't think of a nice and elegant way to do it - maybe a future PR

src/lib.rs
```diff
@@ -185,6 +196,53 @@ impl<T: Runtime> Clone for Client<T> {
...
+    start_key: Option<StorageKey>,
+    buffer: Vec<(StorageKey, StorageData)>,
+}
+
+impl<T: Runtime, F: Store<T>> KeyIter<T, F> {
+    /// Returns the next key value pair from a map.
+    pub async fn next(&mut self) -> Result<Option<(StorageKey, F::Returns)>, Error> {
+        loop {
+            if let Some((k, v)) = self.buffer.pop() {
+                return Ok(Some((k, Decode::decode(&mut &v.0[..])?)))
```
4meta5, line 219: @dvc94ch why don't we decode the key here? It is a raw `sp_core::sp_store::StorageKey`, but I'm pushing onto a vector `Vec<(T::BountyId, BountyState<T>)>`. I guess I can manually decode it...

PR #152, opened 2020-08-13
Add a method to fetch an unhashed key, close #100

### LINKED ISSUE #100: Support query unhashed storage
In some scenarios, we need to query [unhashed storage](https://github.com/paritytech/substrate/blob/32ae0fd2221456d275471a2304fb30eba9736b14/frame/grandpa/src/lib.rs#L261) directly. Can the new [fetch](https://github.com/paritytech/substrate-subxt/blob/194c7bcf2f368571266c4f7a7cfdfda4caa73205/src/lib.rs#L209) add this support? Or should we simply expose rpc.storage? cc @dvc94ch

en:


src/lib.rs
```diff
@@ -258,6 +258,15 @@ impl<T: Runtime> Client<T> {
         &self.metadata
     }
 
+    /// Fetch the value under an unhashed storage key
+    pub async fn fetch_unhashed(
+        &self,
+        key: &StorageKey,
+        hash: Option<T::Hash>,
+    ) -> Result<Option<StorageData>, Error> {
```
ascjones, line 266: It might make sense to also perform the decoding rather than return just the raw data?

src/lib.rs
```diff
@@ -258,6 +258,19 @@ impl<T: Runtime> Client<T> {
...
+        &self,
+        key: StorageKey,
+        hash: Option<T::Hash>,
+    ) -> Result<Option<V>, Error> {
+        if let Some(data) = self.rpc.storage(&key, hash).await? {
+            Ok(Some(Decode::decode(&mut &data.0[..])?))
+        } else {
+            Ok(None)
+        }
+    }
```
ascjones, line 272: One more thing: could you modify `fetch` below :point_down: to call this method just to reduce the duplication.

PR #156, opened 2020-08-30
Only return an error if the extrinsic failed.

dvc94ch:
If an extrinsic in a block fails, all extrinsics will report a failure. We check the phase when watching an extrinsic and ignore errors of other extriniscs.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -62,13 +62,7 @@ pub enum Error {
     TypeSizeUnavailable(String),
     /// Runtime error.
     #[error("Runtime error: {0}")]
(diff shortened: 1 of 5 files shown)


dvc94ch:
All tests pass, it's a transient error. Not sure why we keep getting virual method called after the test runner completes.

ascjones (approved):
LGTM

ascjones:
> All tests pass, it's a transient error. Not sure why we keep getting virual method called after the test runner completes.

Yeah that keeps happening, perhaps something to do with the test client?

PR #162, opened 2020-09-11
Remove visibility limiter `pub (crate)` from Metadata

enfipy:
I faced that I could not use Metadata's main functions outside of the crate and don't see any reason why these functions should be invisible to outside.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -93,10 +93,7 @@ impl Metadata {
             .ok_or(MetadataError::ModuleNotFound(name))
     }
 
-    pub fn module_with_calls<S>(
-        &self,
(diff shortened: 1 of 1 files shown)


ascjones (approved):
LGTM. 

Please run `cargo +nightly fmt` to satisfy the CI before I merge

ascjones:
Fails to build now because of missing docs on public methods

enfipy:
@ascjones Some odd error in tests, how can I fix it?

ascjones:
> @ascjones Some odd error in tests, how can I fix it?

The tests passed, just that weird error on the end happens occasionally.

PR #168, opened 2020-09-21
Display RawEvent data in hex

liuchengxu:
So that printing the `RawEvent` won't take more space than the whole screen.


ascjones (approved):
LGTM

PR #171, opened 2020-09-22
Address review grumbles from #99

dvdplm:


ascjones (approved):
LGTM. CI is failing on fmt check though

PR #173, opened 2020-09-23
Upgrade to substrate 2.0.0

dvdplm:
Update dependencies and sort out incompatibilities.


src/frame/contracts.rs
```diff
@@ -116,6 +116,7 @@ pub struct InstantiatedEvent<T: Contracts> {
 }
 
 #[cfg(test)]
+#[cfg(feature = "integration-tests")]
```
dvdplm, line 119: All tests in this module are integration tests so `cargo check --tests` will warn about unused items without this.

src/frame/system.rs
```diff
@@ -118,7 +118,7 @@ pub trait System {
 }
 
 /// Type used to encode the number of references an account has.
-pub type RefCount = u8;
+pub type RefCount = u32;
```
dvdplm, line 121: @dvc94ch ty! Was scratching my head on this one. :)

src/frame/contracts.rs
ascjones: Means the individual test ones are redundant now, but anyway I'll fix that in #165

ascjones (approved):
LGTM (test failure just the usual `pure virtual method called`)

PR #194, opened 2020-12-02
Remove some substrate dependencies to improve compile time

ascjones:
Removes some big substrate dependencies which were only being used for one or two types. I have just copied these over verbatim, which is acceptable since we are targeting a specific version of substrate at the moment.

This shaves about a minute off the compile time, down from 3:42 to 2:46 on my machine.

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -47,7 +47,7 @@ use sp_core::{
     twox_128,
     Bytes,
 };
-use csp_rpc::{
+use sp_rpc::{
     list::ListOrValue,
     number::NumberOrHex,
 };
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -97,9 +97,9 @@ impl From<u32> for BlockNumber {
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/rpc.rs
```diff
@@ -386,7 +424,7 @@ impl<T: Runtime> Rpc<T> {
         let mut xt_sub = self.watch_extrinsic(extrinsic).await?;
 
         while let status = xt_sub.next().await {
-            log::info!("received status {:?}", status);
+            // log::info!("received status {:?}", status);
```
dvdplm, line 427: remove or uncomment

src/rpc.rs
```diff
@@ -108,6 +109,43 @@ pub struct SystemProperties {
     pub token_symbol: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
dvdplm, line 112: Would be good to add a big note to all the types that must align with substrate's.

dvdplm (changes requested):
LGTM, and the speedup is real! Left a few nitpicks.

dvdplm (approved):

PR #198, opened 2020-12-07
Make RawEvent Cloneable && Add a wrapper rpc call && export request for custom rpc call.

alannotnerd:
* RawEvent needs to implement Clone for multithread usage.
* Export request method, for non-standard rpc.
* Wrapped `chain_getHead` rpc call.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,7 @@ extern crate substrate_subxt_proc_macro;
 #[cfg(feature = "client")]
 pub use substrate_subxt_client as client;
 
+pub use jsonrpsee::common::Params;
 pub use sp_core;
 pub use sp_runtime;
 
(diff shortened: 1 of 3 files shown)


ascjones:
Thanks! Need to run `cargo +nightly fmt` to make CI happy.

alannotnerd:
> Thanks! Need to run `cargo +nightly fmt` to make CI happy.

Thanks for suggestion! I have formatted code.

src/lib.rs
```diff
@@ -378,6 +395,12 @@ impl<T: Runtime> Client<T> {
         Ok(head)
     }
 
+    /// Get a block hash of the latest block
```
dvdplm, line 398:
```suggestion
    /// Get the block hash of the latest block
```

src/lib.rs
```diff
@@ -259,6 +259,23 @@ impl<T: Runtime> Client<T> {
...
+    pub async fn request<Ret>(
+        &self,
+        method: impl Into<String>,
+        params: impl Into<jsonrpsee::common::Params>,
+    ) -> Result<Ret, jsonrpsee::client::RequestError>
+    where
+        Ret: jsonrpsee::common::DeserializeOwned,
+    {
+        self.rpc().client().request(method, params).await
+    }
```
dvdplm, line 277: This seems more like a convenience method and could perhaps live in your own code as well? Or why do you think this is important to add?
My concern here is that this leaks types from `jsonrpsee`, which makes it harder for us to change the implementation under the hood (although we likely already leak plenty in other places but the less we do it the better).

alannotnerd: Many rpc methods are not covered by subxt, and there is no obvious way except changing library code to add customized call. For example, I have to change Client implementation if I want to use `chain_getHead`. Further more, substrate allows user to add their own runtime module which may contain extra private rpc methods. 

In conclusion, It needs to expose an api for user to make their own private rpc call.

alannotnerd:
Trying to convince you again!

Issue #187, opened 2020-10-29
Prioritize a transaction by tip


Dengjianping:
Now, it looks subxt  doesn't support tip to prioritize a transaction.

dvdplm:
Related to #331.

jsdw:
As of ¢490 it's possible to configure the extra params of a transaction, including tip and mortality.

PR #204, opened 2020-12-14
feat: Configurable mortal extrinsic construction

### LINKED ISSUE #202: Extrinsic construction should default to `Era::mortal`
Currently, it looks like extrinsic construction with `subxt` [defaults to `Era::Immortal`](https://github.com/paritytech/substrate-subxt/blob/80aff7eb6e66b4e7cdd504b385fe73ef2878894d/src/extrinsic/extra.rs#L288). Further, it looks like the `subxt` public api does not expose a way for a user to construct a mortal transaction. 

Immortal transactions are dangerous as they expose a potential replay attack vector; thus I believe defaulting to (or even only allowing) mortal transactions is best practice for an extrinsic construction lib. [Here is a section on the polkadot wiki that elaborates on the aforementioned	.](https://wiki.polkadot.network/docs/en/build-protocol-info#transaction-mortality)

If my understanding is correct, implementation wise one way to achieve this is by allowing [`Extrinsic::new()`](https://github.com/paritytech/substrate-subxt/blob/80aff7eb6e66b4e7cdd504b385fe73ef2878894d/src/extrinsic/extra.rs#L236) to accept an argument for `Era`, which could [be constructed with this](https://github.com/paritytech/substrate/blob/dbf2163250833e6ac898e7f6c3c8f89f08a7c19d/primitives/runtime/src/generic/era.rs#L66) - passing in `current` block number, and `period` blocks the transaction should live for, with a reasonable default like 64. `current` block number could be fetched under the hood by RPC and `period` could be filled by a default value, which I think would result in no breaking changes for current users (not counting the change that the ext is now mortal).

cc @dvdplm, @joepetrowski

emostov:
Closes #202 

## public api updates

~~This PR allows for users to configure both mortal and immortal extrinsics through a new `SignedOptions` struct, where they can pass in the desired `era_period` for a mortal ext or `None` for immortal ext. I chose this design because I believe it will make it easier to append more options, such as tip (#187), with minimal maintenance.~~

This PR changes default extrinsic construction to `Era::mortal`, where the `mortal_period` defaults to 64 blocks. The `mortal_period` can be set via the client method `set_mortal_period`, where `None` indicates an immortal transaction.

This PR also exposes a `DEFAULT_MORTAL_PERIOD` to give users a reasonable starting point.

## review notes

~~If there is no immediate major grumbles I go ahead and update tests~~ No need to update tests


src/lib.rs
```diff
@@ -455,12 +473,23 @@ impl<T: Runtime> Client<T> {
             self.account(signer.account_id(), None).await?.nonce
         };
         let call = self.encode(call)?;
+        let era_opts = if opts.era_period.is_some() {
```
niklasad1, line 476:
```suggestion
        if let Some(era_opts) = if opts.era_period.is_some() {
```

src/lib.rs
```diff
@@ -455,12 +473,23 @@ impl<T: Runtime> Client<T> {
...
+            let era_period = opts.era_period.unwrap();
```
niklasad1, line 477: Remove unwrap

src/lib.rs
```diff
@@ -455,12 +473,23 @@ impl<T: Runtime> Client<T> {
...
+            let current_block = self.block(None::<T::Hash>).await?.unwrap().block;
```
niklasad1, line 478: Return err instead of unwrap?

src/lib.rs
```diff
@@ -498,12 +527,13 @@ impl<T: Runtime> Client<T> {
         &self,
         call: C,
         signer: &(dyn Signer<T> + Send + Sync),
+        opts: SignedOptions,
```
niklasad1, line 530: NOTE: breaking change.

src/lib.rs
```diff
@@ -193,6 +198,18 @@ pub struct Client<T: Runtime> {
     page_size: u32,
 }
 
+/// Construction options for a signed extrinsic
+#[derive(Clone)]
```
niklasad1, line 202:
```suggestion
#[derive(Copy, Clone)]
```

niklasad1 (commented):
Looks overall good but I don't know the rationale why `mortal transactions` wasn't implemented before

dvdplm (commented):
Good stuff here. My main question here is whether we should do more with the proc macro and ask users to annotate their `structs` with an attribute somehow, e.g.:

```rust
#[derive(Call)]
pub struct MyVeryOwnCall<'a, T: Balances> {
  pub to: &'a <T as System>::Address,
  pub amount: T::Balance,
  [#Call(signed_options)]
  pub era_period: u64, 
}
```
In other words, piggy back on the args parsing machinery we already have and add some annotation to guide its usage? Or maybe I'm completely wrong here and all extrinsics need the era info anyway?

ascjones (commented):
I suggest moving the era into the `Signer`, in the same way that it handles the `nonce`. Instead of having this extra options argument to pass around.

ascjones:
> Good stuff here. My main question here is whether we should do more with the proc macro and ask users to annotate their `structs` with an attribute somehow, e.g.:
> 
> ```rust
> #[derive(Call)]
> pub struct MyVeryOwnCall<'a, T: Balances> {
>   pub to: &'a <T as System>::Address,
>   pub amount: T::Balance,
>   [#Call(signed_options)]
>   pub era_period: u64, 
> }
> ```
> 
> In other words, piggy back on the args parsing machinery we already have and add some annotation to guide its usage? Or maybe I'm completely wrong here and all extrinsics need the era info anyway?

I think it's good enough to have the global default and leave it up to the callers at runtime to configure the options per call.

src/lib.rs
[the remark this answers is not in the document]
niklasad1: use  `if let Some(era_period) = opt.era_period {  ... }` instead here then the unwrap is not needed

src/lib.rs
```diff
@@ -193,6 +198,18 @@ pub struct Client<T: Runtime> {
...
+pub struct SignedOptions {
+    /// The period, measured in blocks, that transaction will live for, starting from a checkpoint
+    /// block. A good default is 64 (64 * 6secs = 6min 40sec).
+    ///
+    /// `era_period == None`: immortal transaction.
+    /// `0 <= era_period <= 65536`: rounded up to the closest power of 2, starting at 4.
+    /// `65536 < era_period`: 65536.
```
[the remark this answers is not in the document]
niklasad1: last line means: min(era_period, 65536)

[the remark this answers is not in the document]
niklasad1: Also, can you explain why `era period` is capped at `65536`?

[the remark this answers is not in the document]
emostov: That number came from my interpretation of these comments: https://github.com/paritytech/substrate/blob/dbf2163250833e6ac898e7f6c3c8f89f08a7c19d/primitives/runtime/src/generic/era.rs#L50-L65

src/lib.rs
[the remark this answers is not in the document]
emostov: I'm not totally familiar with the rationale, but since this is public and all of those seem potentially helpful to a user I don't see why not?

emostov:
> I suggest moving the era into the Signer, in the same way that it handles the nonce. Instead of having this extra options argument to pass around.

Alternatively, or in addition to the above, this option and other similar could be provided as part of the Client, so would get applied to all calls from that client instance (if such an option is likely to be the same across all transactions).

> Alternatively, or in addition to the above, this option and other similar could be provided as part of the `Client`, so would get applied to all calls from that client instance (if such an option is likely to be the same across all transactions).

I agree this should be moved to the `Signer` or `Client` as it would likely be the same across transactions for users. I think of mortality preferences being something non-specific to account, so in that sense I think it makes sense to put on the `Client` - but I think both setups would work.

src/extrinsic/mod.rs
```diff
@@ -59,6 +65,7 @@ pub async fn create_signed<T>(
     nonce: T::Index,
     call: Encoded,
     signer: &(dyn Signer<T> + Send + Sync),
+    era_opts: Option<(u64, u64, T::Hash)>,
```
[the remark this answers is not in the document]
emostov: Yes, great point. Will change

src/lib.rs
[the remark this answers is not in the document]
emostov: Just learned from @joepetrowski and wanted to clarify - this number is the maximum that can be encoded, but `BlockHashCount` is the maximum amount of blocks that the transaction is actually valid for (on Polkadot and Kusama that is 2400)

src/lib.rs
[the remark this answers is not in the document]
emostov: This is not longer public, so just deriving Clone and Copy.

src/extrinsic/extra.rs
```diff
@@ -271,12 +273,14 @@ impl<T: System + Balances + Clone + Debug + Eq + Send + Sync> SignedExtra<T>
         tx_version: u32,
         nonce: T::Index,
         genesis_hash: T::Hash,
+        era_info: (Era, T::Hash)
```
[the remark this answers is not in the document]
emostov: I changed this so it is `Option<T::Hash>` and we pass in none when it is immortal - this results in an extra conditional to determine options variant and if we should use the genesis hash.

src/extrinsic/mod.rs
```diff
@@ -41,6 +44,9 @@ use crate::{
     Error,
 };
 
+/// A reasonable default for `era_period`
```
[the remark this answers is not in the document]
emostov: Bringing offline discussion here - I am not using any science to determine this is reasonable, I saw it in `txwrapper`, it is power of 2, and subjectively it seems like an amount of time where the Tx could sit in a pool for a little if needs to but it would be hard to create a replay attack within that time period. Certainly open to ideas for how to come up with a better number (or just better rationale for this number).

There also is no default in the metadata.

emostov:
Updated: mortal period now lives on the client. I think ready for another round of reviews

src/lib.rs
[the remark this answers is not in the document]
dvdplm: > I'm not totally familiar with the rationale

https://rust-lang.github.io/api-guidelines/debuggability.html#c-debug (the whole guide is worth a read or two)

> This is no longer public, so just deriving Clone and Copy.

👍

src/lib.rs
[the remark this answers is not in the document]
dvdplm: TIL!
Seems like there's a lot more complexity here than I first thought. Would be good to capture this in the comment (and perhaps upstream as well?).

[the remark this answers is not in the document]
joepetrowski: It's a funny nuance. It's not that the transaction is invalid per se, it's that the runtime stores `BlockHashCount` blocks in state as reference. If the runtime doesn't have the transaction's checkpoint block to reference against, it's _unable_ to say whether the block is valid or not, so just rejects it on those grounds.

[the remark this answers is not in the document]
niklasad1: Alright, that's annoying for this usecase. 

If I understand this correctly `BlockHashCount` is FRAME-related and configured by each runtime. Thus, if I want to use my custom runtime with `BlockHashCount == 3` and `era_period > 3` then it will be rejected and would hard to figure out as a user of `substrate-subxt`.

Next question is: can we fetch `BlockHashCount` in the `Client constructor/builder` via RPC [similary as other runtime stuff](https://github.com/paritytech/substrate-subxt/blob/master/src/lib.rs#L165-#L171) then do `min(block_hash_count, mortality_period)`?

src/lib.rs
```diff
@@ -178,10 +187,26 @@ impl<T: Runtime> ClientBuilder<T> {
...
+/// Construction options for a signed extrinsic
+#[derive(Copy, Clone)]
+struct ClientSignedOptions {
+    /// The period, measured in blocks, that an extrinsic will live for, starting from a checkpoint
+    /// block.
+    ///
+    /// Rules for how the period will be encoded in an extrinsic:
+    /// `mortal_period == None`: immortal transaction
+    /// `0 <= mortal_period <= 65536`: rounded up to the closest power of 2, starting at 4
+    /// `65536 < mortal_period`: Min(65536, mortal_period)
```
niklasad1, line 206: I think it is better to refer to https://docs.rs/sp-runtime/2.0.0/sp_runtime/generic/enum.Era.html#variant.Mortal

It's not obivious with the edge case when `BlockHashCount` is less than `mortality_period`

src/extrinsic/mod.rs
```diff
@@ -41,6 +44,9 @@ use crate::{
     Error,
 };
 
+/// A reasonable default for `mortal_period`
+pub const DEFAULT_MORTAL_PERIOD: u64 = 64;
```
niklasad1, line 48: If I understand this correctly, this would works as long `BlockHashCount` >= `64` but we can't know for use because it's defined by the runtime. Thus, I prefer to fetch `block_hash_count` from the runtime and use it as `default_mortal_period` (could be something within the range 1..=block_hash_count)

src/lib.rs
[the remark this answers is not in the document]
joepetrowski: Yes, it's available over RPC: https://github.com/paritytech/substrate/blob/31499cd29ed30df932fb71b7459796f7160d0272/frame/system/src/lib.rs#L508

src/extrinsic/mod.rs
emostov: I think your proposed approach is ideal, but after looking into it a little bit it seems like this is not possible without implementing constants for metadata. I think we would need support for this: https://docs.rs/frame-metadata/11.0.0-alpha.3/frame_metadata/struct.ModuleMetadata.html#structfield.constants. You can even see here we currently have a place holder: https://github.com/paritytech/substrate-subxt/blob/8a76bebcff6534421b498ac874e7e6ffe09d256a/src/metadata.rs#L169 ... I could be wrong though, maybe there is a way to just key into storage?

emostov: https://github.com/paritytech/substrate-subxt/issues/206

Also realizing though right after making that issue, it is ok to have a `mortal_period` that is too high. In that case the extrinsic mortality will simply end at `BlockHashCount`.

niklasad1: all right, I see

emostov:
Main update from the last review is linking directly to the sp-runtime generic `Era` substrate in the doc comments for `mortal_period`:
https://github.com/paritytech/substrate-subxt/blob/722c38d01ee41502debbe3ee566bc6abe0f5072a/src/lib.rs#L199-L206

I would think this is the preferred approach as `substrate`  is the conceptual layer and docs dealing with higher level concepts should come from there.

emostov:
Updated to now close #206 

See: 
[src/extrinsic/mortal_period.rs](src/extrinsic/mortal_period.rs)
& 
https://github.com/paritytech/substrate-subxt/blob/f09adc18755e04baadc2dbd5f75b463abf8090c4/src/lib.rs#L184

niklasad1 (commented):
Nice, much cleaner now.

However, I'm not sure of these `default fallbacks` isn't better just to fail and propagate the error to the user and why is it not user-configurable to set the `mortality period`?

src/metadata.rs
```diff
@@ -164,7 +173,33 @@ impl Metadata {
             }
         }
         string
-    }
+	}
+
+	/// Derive a default mortal period
+	pub(crate) fn derive_mortal_period(&self) -> Result<u64, MetadataError> {
+		let block_hash_count = if let Ok(system_meta) = self.module("System") {
```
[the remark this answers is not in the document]
emostov: BlockHashCount is type BlockNumber:
https://github.com/paritytech/polkadot/blob/116df38725078fd7a4219851d186ab734ece7668/runtime/common/src/lib.rs#L58

In polkadot master BlockNumber is use u32:
https://github.com/paritytech/polkadot/blob/2c6920c40fb5ffb607cd2d3607dd182a76cee5d3/core-primitives/src/lib.rs#L28

src/metadata.rs
```diff
@@ -165,6 +174,32 @@ impl Metadata {
...
+            if let Ok(milliseconds) = babe_meta.constant("ExpectedBlockTime") {
+                milliseconds.value::<u64>()?
+            } else {
+                FALLBACK_EXPECTED_BLOCK_TIME
+            }
+        } else {
+            FALLBACK_EXPECTED_BLOCK_TIME
+        };
+
+        Ok((BASELINE_MORTAL_PERIOD / block_time)
```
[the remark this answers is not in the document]
emostov: We now pass an error up in this scenario now

emostov:
@niklasad1 See my responses to feedback for specifics, but I updated `derive_mortal_period` to not use any fallback values and instead bubble up an error. I think this setup removes some ambiguity about what is happening under the hood. One thing I am on the fence about is logging a message in addition to returning the error:
https://github.com/paritytech/substrate-subxt/blob/a80c64461f4a3c4e23960c5b42b30100b418f57c/src/lib.rs#L194-L197

Is this overkill?

src/metadata.rs
```diff
@@ -165,6 +171,28 @@ impl Metadata {
         }
         string
     }
+
+    /// Derive a mortal period
+    pub(crate) fn derive_mortal_period(&self) -> Result<u64, MetadataError> {
```
niklasad1, line 176: I was thinking a little bit about having concrete types `u32 and u64` here which might not work for all runtimes ideally we would read the types from the runtime but it seems quite complicated to get it to work and `pallet-babe` is not implemented for `substrate-subxt`.

Is it fair to assume that `BABE` is used for all runtimes?

However, for example the `runtime type parameter` was passed in here we could do:
```rust
    /// Derive a mortal period
    pub(crate) fn derive_mortal_period<T: Runtime>(&self) -> Result<impl AtLeast32BitUnsigned, MetadataError> {
        let block_hash_count = self
            .module("System")
            .and_then(|sys| sys.constant("BlockHashCount"))
            .and_then(|count| count.value::<<T as System>::BlockNumber>())?;

        let expected_block_time = self
            .module("Babe")
            .and_then(|babe| babe.constant("ExpectedBlockTime"))
            // babe frame trait doesn't exist yet....
            .and_then(|e| e.value::<<T as Babe>::ExpectedBlockTime>())?;

        // maybe convert by try_from or something.
        match expected_block_time {
            0 => Err(MetadataError::ZeroValue("Babe::ExpectedBlockTime")),
            expected_block_time => {
                Ok((BASELINE_MORTAL_PERIOD / expected_block_time)
                    .next_power_of_two()
                    .min(block_hash_count.next_power_of_two()))
            }
        }
    }
```

//cc @ascjones

src/metadata.rs
```diff
@@ -165,6 +171,28 @@ impl Metadata {
...
+        let block_hash_count: u64 = self
+            .module("System")
+            .and_then(|sys| sys.constant("BlockHashCount"))
+            .and_then(|count| count.value::<u32>())
+            .map(Into::into)?;
+        let expected_block_time = self
+            .module("Babe")
+            .and_then(|babe| babe.constant("ExpectedBlockTime"))
```
niklasad1, line 184: Is it fair to assume that `BABE` is used for all runtimes?

src/lib.rs
```diff
@@ -119,6 +128,7 @@ pub struct ClientBuilder<T: Runtime> {
     url: Option<String>,
     client: Option<jsonrpsee::Client>,
     page_size: Option<u32>,
+    mortal_period: Option<Option<u64>>,
```
niklasad1, line 131: What's up with `Option<Option<_>>` rather than `Option<_>`?

src/lib.rs
```diff
@@ -203,6 +246,7 @@ impl<T: Runtime> Clone for Client<T> {
             runtime_version: self.runtime_version.clone(),
             _marker: PhantomData,
             page_size: self.page_size,
+            signed_options: self.signed_options.clone(),
```
niklasad1, line 249: The type implements `Copy`

```suggestion
            signed_options: self.signed_options,
```

src/lib.rs
```diff
@@ -170,18 +187,43 @@ impl<T: Runtime> ClientBuilder<T> {
             rpc.system_properties(),
         )
         .await;
+        let metadata = metadata?;
+        let mortal_period = if let Some(period) = self.mortal_period {
+            period
+        } else {
+            match metadata.derive_mortal_period() {
```
niklasad1, line 194: I had in mind to pass `mortal_period` into `derive_mortal_period` in order to check if it was valid but I have no strong opinions on it.

niklasad1:
> Is this overkill?

It is fine from my side but you might want to print the actual error because it might fail for other reasons than `mortal_period` :)

src/lib.rs
emostov: `None` => User has not set anything
`Some(64)` => User set a mortal period of 64
`Some(None)` => User has set it to immortal

src/metadata.rs
emostov: No, I think Aura is popular as well.

emostov: I think we can use TimeStamp::MinimumPeriod and double it

https://github.com/paritytech/substrate/blob/332399d16668a6c769f1a7db154bb9ea3b50e61c/frame/timestamp/src/lib.rs#L129-L133

src/lib.rs
emostov: Honestly it is not even clear to me what would be an invalid mortal period a user could set. I think we could check if `mortal_period <= BlockHashCount.next_power_of_two()`, but the tx will still be processed by the node with larger values.  Knowing that, it seems artificially limiting, especially if for some reason a runtime follows a different rule.

src/lib.rs
```diff
@@ -170,18 +187,44 @@ impl<T: Runtime> ClientBuilder<T> {
...
+                Err(e) => {
+                    log::error!("{}", e);
+                    log::error!("`Metadata::derive_mortal_period` failed. You may need to set `mortal_period` prior to invoking `Client::build`");
```
niklasad1, line 197:
```suggestion
                    log::error!("`Metadata::derive_mortal_period` failed: {}. You may need to set `mortal_period` prior to invoking `Client::build`", e);
```

src/lib.rs
niklasad1: Fair enough, I was thinking of the initial annoyance, if the `mortal_period > BlockHashCount`

src/metadata.rs
```diff
@@ -165,6 +171,36 @@ impl Metadata {
...
     }
+
+    /// Derive a mortal period
+    pub(crate) fn derive_mortal_period(&self) -> Result<u64, MetadataError> {
+        let block_hash_count: u64 = self
+            .module("System")
+            .and_then(|sys| sys.constant("BlockHashCount"))
+            .and_then(|count| count.value::<u32>())
+            .map(Into::into)?;
+        let expected_block_time = self
```
niklasad1, line 182: :+1:

src/lib.rs
niklasad1: Ok, it would be more readable with an enum such as `MortalityPeriod`.

src/metadata.rs
```diff
@@ -40,6 +40,9 @@ use sp_core::storage::StorageKey;
 
 use crate::Encoded;
 
+/// 5 min `mortal_period` in milliseconds, to be adjusted based on expected block time
+pub const BASELINE_MORTAL_PERIOD: u64 = 5 * 60 * 1000;
```
dvdplm, line 44: Does this need to be `pub`?

src/lib.rs
```diff
@@ -455,12 +519,31 @@ impl<T: Runtime> Client<T> {
...
+                Some(signed_block) => signed_block.block,
+                None => return Err("RPC chain_getBlock returned None when Some(signed_block) was expected".into()),
+            };
+            let current_number =
+                (*current_block.header().number()).saturated_into::<u64>();
+            let current_hash = current_block.hash();
+
+            (
+                Era::mortal(mortal_period, current_number),
+                Some(current_hash),
```
dvdplm, line 535: dq: why do we need to return both the block number and the block hash? Isn't that redundant info? Could we get away with only the hash and return only `Era::Mortal(period, current_hash)`/`Era::Immortal`?

src/lib.rs
```diff
@@ -170,18 +196,43 @@ impl<T: Runtime> ClientBuilder<T> {
...
             _marker: PhantomData,
             page_size: self.page_size.unwrap_or(10),
+            signed_options: ClientSignedOptions { mortal_period },
         })
     }
 }
 
+/// Construction options for a signed extrinsic
+#[derive(Copy, Clone)]
+struct ClientSignedOptions {
```
dvdplm, line 227: Would `SigningOptions` work here?

src/lib.rs
```diff
@@ -170,18 +196,43 @@ impl<T: Runtime> ClientBuilder<T> {
...
+    /// The period, measured in blocks, that an extrinsic will live for, starting from a checkpoint
```
dvdplm, line 228:
```suggestion
    /// The period of validity, measured in blocks, that an extrinsic will live for, starting from a checkpoint
```

src/lib.rs
```diff
@@ -112,13 +121,23 @@ use crate::{
     },
 };
 
+/// The mortality of a transaction
```
dvdplm, line 124:
```suggestion
/// The period of validity of a transaction
```

src/extrinsic/extra.rs
```diff
@@ -271,21 +272,28 @@ impl<T: System + Balances + Clone + Debug + Eq + Send + Sync> SignedExtra<T>
...
+            hash
+        } else {
+            self.genesis_hash
+        };
         (
             CheckSpecVersion(PhantomData, self.spec_version),
             CheckTxVersion(PhantomData, self.tx_version),
             CheckGenesis(PhantomData, self.genesis_hash),
-            CheckEra((Era::Immortal, PhantomData), self.genesis_hash),
+            CheckEra((self.era_info.0, PhantomData), era_hash),
```
dvdplm, line 296: This is a bit clunky. Ideally we'd have `CheckEra(PhantomData, self.era_info)` but that seems tricky to make work?

dvdplm (commented):
I'm a bit out of my depth here, but left a few comments and questions.

src/extrinsic/extra.rs
emostov: The inner tuple is what actually gets encoded, while the `era_hash` is not part of the `CheckEra` struct in Substrate and is just used here for implementing `additional_signed`. We could make a function like `create_check_era(PhantomData, self.era_info)`, but it seems like that would just create unnecessary code.

Substrate FRAME nodes do not need the hash because they have a map of canon block number => block hash, which we do not have

src/lib.rs
emostov: `Era::mortal` is a function that returns the enum variant `Era::Mortal`. The function takes in a mortal period and block number, does some math and spits out the variant `Mortal`, which holds the mortal period and a derived `phase` - so the `Mortal` variant is not aware of the block number or block hash. See: https://docs.rs/sp-runtime/2.0.0/src/sp_runtime/generic/era.rs.html#66-76

We need the `current_hash` in addition to the `Mortal` variant because the `additional_signed` for `CheckEra` needs to return the hash of the block. For `Era::Immortal` we just pass `None` for the block hash because we end up just using the genesis hash, which we already know.

src/metadata.rs
emostov: Nope, now removed.

src/lib.rs
emostov: Yup, updated.

niklasad1 (commented):
LGTM

src/metadata.rs
ascjones: As we speak I am working on upgrading substrate metadata to use `scale-info` for full type information and once we have that we should be able to do runtime specific builds for `subxt`

ascjones:
What's the status of this? Waiting for a review from @dvdplm?

dvdplm (approved):
Better late than never?

ascjones:
Just needs merge conflicts fixing now

ascjones:
Not sure as to the mergability of this since the recent refactoring. Certainly we need to functionality still.

ascjones:
I believe this has now been superseded by https://github.com/paritytech/subxt/pull/490 /cc @jsdw

PR #207, opened 2020-12-17
feat: Enable retrieval of constants from runtime metadata

### LINKED ISSUE #205: Add constants to metadata
While storage, events, and errors are present in `subxt` `Metadata`, constants are missing. 

While `subxt` users may likely find numerous use cases for constants, a couple use case within `subxt` itself come to mind:

- `System::Constants::TransactionByteFee` and `TransactionPayment::Constants::ExtrinsicBaseWeight` could be used to do offchain fee calculations
- `System::Constants::BlockHashCount` could be used for creating a [better default mortality period for an extrinsic](https://github.com/paritytech/substrate-subxt/pull/204#discussion_r543404026_)

It looks like [`constants` are defined as a field of `ModuleMetadata`](https://docs.rs/frame-metadata/12.0.0/frame_metadata/struct.ModuleMetadata.html#structfield.constants)

emostov:
Closes #205 

Example of new public api usage and outputs

```Rust
    let client = ClientBuilder::<KusamaRuntime>::new().build().await?;
    let block_hash_count_meta = client.metadata()
        .module("System")?
        .constant("BlockHashCount")?;

    println!("name: {:#?}\n value: {:#?}\n ty: {:#?}\n documentation: {:#?}\n",
        block_hash_count_meta.name(),
        block_hash_count_meta.value::<u32>()?,
        block_hash_count_meta.ty(),
        block_hash_count_meta.documentation()
    );

  // name: "BlockHashCount"
  // value: 2400
  // ty: "T::BlockNumber"
  // documentation: [
  //    " The maximum number of blocks to allow in mortal eras.",
  // ]

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -70,6 +70,12 @@ pub enum MetadataError {
     /// Default error.
     #[error("Failed to decode default: {0}")]
     DefaultError(CodecError),
+    /// Failure to decode constant value.
+    #[error("Failed to decode constant value: {0}")]
+    ConstantValueError(CodecError),
+    /// Constant is not in metadata.
+    #[error("Constant {0} not found")]
+    ConstantNotFound(&'static str),
 }
 
 /// Runtime metadata.
@@ -166,7 +172,7 @@ pub struct ModuleMetadata {
     index: u8,
     name: String,
     storage: HashMap<String, StorageMetadata>,
-    // constants
+    constants: HashMap<String, ModuleConstantMetadata>,
 }
 
 impl ModuleMetadata {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by niklasad1 — APPROVED** [review]
nice, LGTM

File: src/metadata.rs [hunk 1]
```diff
@@ -175,6 +181,16 @@ impl ModuleMetadata {
             .get(key)
             .ok_or(MetadataError::StorageNotFound(key))
     }
+
+    /// Get a constant's metadata by name
+    pub fn constant(
+        &self,
+        key: &'static str,
```
ascjones, line 188:
```suggestion
        key: &str,
```
No need to restrict it to `'static`

ascjones: I see not some of the other similar methods are also `'static`, but I don't think they need to be if we make the respective `Error` variants have a `String` instead.

ascjones (approved):
LGTM

We can figure out whether to allow non `'static` keys in another PR.

PR #210, opened 2020-12-30
Fix 'Option' return of storage

wuminzhe:
Suppose a substrate storage's return type is `Option<Vec<_>>`. So its metadata modifier is `Optional` and value type is `Vec<_>`.

If the storage returns `Some([])`, the hex data is not `0x0100` but `0x00`, which will be decoded to `None`. This is wrong.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@ pub use sp_core;
 pub use sp_runtime;
 
 use codec::Decode;
+use frame_metadata::StorageEntryModifier;
 use futures::future;
 use jsonrpsee::client::Subscription;
 use sp_core::{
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -274,10 +275,19 @@ impl<T: Runtime> Client<T> {
         &self,
         key: StorageKey,
         hash: Option<T::Hash>,
+        modifier: StorageEntryModifier,
     ) -> Result<Option<V>, Error> {
-        if let Some(data) = self.rpc.storage(&key, hash).await? {
-            Ok(Some(Decode::decode(&mut &data.0[..])?))
+        if let Some(mut data) = self.rpc.storage(&key, hash).await? {
+            let bytes = if modifier == StorageEntryModifier::Optional {
+                let mut bytes = vec![1u8];
```
ascjones, line 282: If I'm reading this correctly then this will always return `Some` for `StorageEntryModifier::Optional` values - is that correct?

ascjones (commented):
Thanks for this - sorry for the delayed review. I'd really like to see some tests for this if possible.

wuminzhe:
sorry for the late.

For a storage whose return type is `Option<Vec<u8>>`, the rpc will returns
- `0x00` for `Some([])`
- `0x080001` for `Some([0, 1])`

So, if subxt want to use the `Option<Vec<u8>>` to decode the returns, the hex shoud be change to:
- `0x00` to `0x0100`
- `0x080001` to `0x01080001`

Or, find a way to use `Vec<u8>` to decode.

subxt has upgrade to 0.14, I will update the PR.

PR #217, opened 2021-01-11
Make type explicit

c410-f3r:
Newer compiler releases require this statement


h4x3rotab:
LGTM

niklasad1 (approved):
LGTM, because [Index implements AtLeast32Bit](https://docs.rs/sp-arithmetic/2.0.1/sp_arithmetic/traits/trait.AtLeast32Bit.html) so this is fine. Personally, I prefer `One::one()` over `1_u32` but.... doesn't matter.

ascjones (approved):

PR #222, opened 2021-01-19
[ci]: use latest nightly

niklasad1:


niklasad1:
Error looks legit, maybe better to use `nightly-2020-10-01` for now?!

.github/workflows/rust.yml
```diff
@@ -15,15 +15,18 @@ jobs:
...
     - name: setup
-      run: |
-          rustup install nightly-2020-10-01 --profile default
-          rustup target add wasm32-unknown-unknown --toolchain nightly-2020-10-01
+      uses: actions-rs/toolchain@v1
+      with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+          components: wasm32-unknown-unknown, rustfmt
```
ascjones, line 23: `wasm32-unknown-unknown` is a target so should be added to a `target` field

ascjones (approved):
LGTM

PR #223, opened 2021-01-26
Fix the wrong rpc call in query_storage_at()

h4x3rotab:
`query_storage_at()` is supposed to call `state_queryStorageAt(keys, hash)` instead of `state_queryStorage(keys, from, to)`. 

Before this fix, a call to the function is equal to `state_queryStorage(keys, hash, null)`. So the responded `StorageChangeSet` includes everything from the given hash to the latest block. I found this problem because the assertion in the iterator fails:

https://github.com/paritytech/substrate-subxt/blob/4725a33ea25c066bb7773a2e9d301498ea156568/src/lib.rs#L250


shawntabrizi (approved):

ascjones (approved):
Thanks!

PR #252, opened 2021-03-18
Spawn test node process for integration tests

ascjones:
Attempt to fix ongoing issues with the integration tests on CI.

- Introduces `TestNodeProcess` helpers to help launch local substrate binary for testing.
- Replaces the use of the embedded node in the tests with a spawned substrate node.
- Moves tests out of `lib.rs` into their own file.

Rel #178 

# todo

- [x] Document installation of substrate node on the `path`
- [x] Download substrate exe in CI script

PR #253, opened 2021-03-19
Convert integration tests to use test node process

ascjones:
Previously the tests were hidden behind the `integration-tests` feature which was only run manually. 

This PR removes the `integration-test` feature and now runs the tests on CI using the test node process introduced in #252.


niklasad1 (approved):
great

ascjones:
Actually running those tests has caught a few regressions, will fix later

PR #254, opened 2021-03-19
Remove test-node crate

ascjones:
Integration tests now use a spawned test node introduced in #252, so this crate is no longer required. 

Should lower maintenance burden and speed up the build a bit.


ascjones:
Build time down from 21 to 7 minutes :tada:

niklasad1 (approved):

niklasad1:
It would be even faster if we remove `subxt-client` to another repo ^^

PR #256, opened 2021-03-23
Fix 'Option' return of storage

wuminzhe:
Renew old PR #210 

For example

For a storage that has a return type `Option<Vec<u8>>`, the storage data through RPC will returns:
- `0x00` for `Some([])`
- `0x080001` for `Some([0, 1])`

So, if subxt want to use the `Option<Vec<u8>>` to decode the returns, the hex should be changed to:
- `0x00` to `0x0100`
- `0x080001` to `0x01080001`  

which is prefixed with a `01` for scale codec `Some`

Or, find a way to use `Vec<u8>` to decode.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@ use codec::{
     Codec,
     Decode,
 };
+use frame_metadata::StorageEntryModifier;
 use futures::future;
 use jsonrpsee_http_client::{
     HttpClient,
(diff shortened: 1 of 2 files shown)


HackFisher:
Related? https://github.com/paritytech/substrate-subxt/pull/120

wuminzhe:
After reading the macro code, I feel this PR is not correct.

PR #259, opened 2021-04-01
Workaround for pallet missing declare error type

jasl:
Ref Substrate issue https://github.com/paritytech/substrate/issues/8501

I'd like fix that issue, but I think subxt still need to do a workaround because it would crash inside subxt.

In short,
when pallet missing declare Error type, the errors field in metadata will be empty,
then, if Subxt parsing block meets error which not list in metadata, it will trigger https://github.com/paritytech/substrate-subxt/blob/master/src/metadata.rs#L256
then, it will crash inside Subxt

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -99,6 +99,9 @@ pub enum RuntimeError {
     /// Module error.
     #[error("Runtime module error: {0}")]
     Module(ModuleError),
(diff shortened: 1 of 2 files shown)


jasl:
Ops... seems CI failure not my mistake

h4x3rotab:
We can close it since #294 was merged

PR #260, opened 2021-04-10
Reexport support crates

mattsse:
This reexports the internally used `frame_support` and `codec` crates which gives easy access to `codec::Encode`, which is required for deriving your own `Call`s and `frame_support::{Parameter}` that is helpful for module type bounds.


ascjones:
Partially resolved via #294

PR #261, opened 2021-04-10
Document the Store proc macro

mattsse:
This supersedes #260 and documents the `Store` derive macro.


ascjones:
Store macro is removed in https://github.com/paritytech/substrate-subxt/pull/294

PR #263, opened 2021-04-12
[jsonrpsee http client]: support tokio1 & tokio02.

### LINKED ISSUE #262: Support build on tokio0.2
**Description**

According to https://github.com/paritytech/jsonrpsee/issues/254, it could be great to support builds on tokio0.2.

niklasad1:
Closes #262 

As `substrate` and some of the Rust ecosystem hasn't updated to `tokio 1` yet, let's make it configurable to support either `tokio0.2` or `tokio 1`.

//cc @akru can you try this to ensure that it actually fixes your issue ^^

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ thiserror = "1.0.24"
 futures = "0.3.13"
 jsonrpsee-types = "=0.2.0-alpha.3"
 jsonrpsee-ws-client = "=0.2.0-alpha.3"
-jsonrpsee-http-client = { version = "=0.2.0-alpha.3", default-features = false, optional = true }
+jsonrpsee-http-client = { version = "=0.2.0-alpha.3", default-features = false }
 num-traits = { version = "0.2.14", default-features = false }
 serde = { version = "1.0.124", features = ["derive"] }
 serde_json = "1.0.64"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -17,16 +17,19 @@ keywords = ["parity", "substrate", "blockchain"]
...
+tokio1 = ["jsonrpsee-http-client/tokio1"]
 
 [dependencies]
 log = "0.4.14"
 thiserror = "1.0.24"
 futures = "0.3.13"
 jsonrpsee-types = "=0.2.0-alpha.3"
 jsonrpsee-ws-client = "=0.2.0-alpha.3"
-jsonrpsee-http-client = "=0.2.0-alpha.3"
+jsonrpsee-http-client = { version = "=0.2.0-alpha.3", default-features = false, optional = true }
```
ascjones, line 32: Why is it `optional`, in case we use `default-features  = false` when depending on subxt. Wouldn't that cause a compilation error?

niklasad1: right good catch, it doesn't need to be optional.

However, if tokio1 or tokio02 is not enabled on jsonrpsee it will be a compilation error anyway. So doesn't really make a difference in practice.

ascjones (approved):
LGTM

PR #267, opened 2021-04-21
Expose the rpc client in Client

liuchengxu:
This makes the life of the downstream user easier.

With `Client`, we can only access the public methods exposed in it
but that does not cover all the RPC interfaces provided by the chain.
Therefore, the user of this library has to build another `RpcClient` for
the uncovered RPC interfaces, which can be unnecessary.


liuchengxu:
Ping @ascjones

ascjones (approved):

PR #284, opened 2021-06-01
Reorg the order of deps

liuchengxu:
This patch is simply for easier editor operation when replacing the
substrate deps. No code changes.


ascjones (approved):
LGTM

PR #285, opened 2021-06-08
deps: update jsonrpsee 0.2.0

niklasad1:
The motivation is to avoid pinning certain alpha versions and to avoid
breaking users builds without having to some `Cargo.lock` updating.

I had to add the `tokio1 compat` flag to `async-std` in order to because we don't use `async-std` anymore in `jsonrpsee`

### DIFF
--- a/src/subscription.rs
+++ b/src/subscription.rs
@@ -180,7 +180,8 @@ impl<T: Runtime> FinalizedEventStorageSubscription<T> {
                 return Some(storage_change)
             }
             let header: T::Header =
-                read_subscription_response("HeaderSubscription", &mut self.subscription).await?;
+                read_subscription_response("HeaderSubscription", &mut self.subscription)
+                    .await?;
             self.storage_changes.extend(
                 self.rpc
                     .query_storage_at(&[self.storage_key.clone()], Some(header.hash()))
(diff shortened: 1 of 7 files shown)


src/subscription.rs
```diff
@@ -199,8 +203,22 @@ impl<T: Runtime> EventStorageSubscription<T> {
...
             Self::Finalized(event_sub) => event_sub.next().await,
         }
     }
 }
+
+async fn read_subscription_response<T>(sub: &mut Subscription<T>) -> Option<T>
+where
+    T: DeserializeOwned,
+{
+    match sub.next().await {
```
niklasad1, line 216: Note: this is a breaking change in `jsonrpsee` before all failures were ignored and this would only return on valid responses on the `subscription stream`.

We could change this API in `subxt` to send back the error here but I didn't want to make any breaking changes.

This error should only occur if the server doesn't send back a response on the stream that is the `type T` on the subscription.

src/subscription.rs
```diff
@@ -199,8 +203,22 @@ impl<T: Runtime> EventStorageSubscription<T> {
...
+        Ok(Some(next)) => Some(next),
+        Ok(None) => return None,
```
ascjones, line 218: Isn't `return` redundant here and below?

niklasad1: nice catch, fixed

ascjones (approved):
LGTM

PR #286, opened 2021-06-10
Update to substrate master

ascjones:
There have been no releases of substrate dependencies on crates.io for a while, so this PR changes the substrate dependencies to `master` for now so we can sync up with the latest version.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,7 +25,7 @@ jobs:
 
     - name: download-substrate
       run: |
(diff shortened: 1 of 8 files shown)


gilescope:
I suggest we point to one of the tagged releases instead: https://github.com/paritytech/substrate-subxt/pull/295
These are released on a regular basis (and are easier to upgrade as there's one thing to change across the board).

gilescope:
(It's a branch but effectively a tagged release - only bugfixes will get backported to these kind of branches)

ascjones:
Closing because of #294

PR #295, opened 2021-09-17
Updated to later version of substrate so that it compiles.

gilescope:
(Referencing substrate this way you only have to search and replace one string to upgrade)


ascjones:
Needs `fmt`.

gilescope:
Not sure what's up with that CI...

ascjones (approved):

PR #307, opened 2021-11-02
fix extrinsics retracted

paulormart:
Since `Retracted` extrinsics may fit in a different block, i think there is no need to raise an error, just wait until it gets a `Finalized`, `Dropped` or `FinalityTimeout`.

In this way the caller most likely will get the finalised block for the specific extrinsic and do not lose it's sight.


paulormart:
reference to pull request #294

ascjones:
Please target `master` now

h4x3rotab:
A rebase to master would be nice

ascjones (approved):
LGTM.

Though this highlights again the inadequacy of this API, I'm thinking we should give more control to the user by accessing the `TransasctionStatus` as part of the events subscription, then they can react as they choose. I will open an issue for this soon.

niklasad1 (approved):

PR #308, opened 2021-11-02
Codegen: enum variant indices

sander2:
This fix is required for calls that embed other calls, such as `sudo` and `batch`. Without it, these calls fail if the explicit index provided in the metadata does not match the implicit index of the enum variant


ascjones:
#294 now merged, so please update this PR to `master`

sander2:
> #294 now merged, so please update this PR to `master`

@ascjones done

ascjones (approved):
Good find :+1:

ascjones:
Codegen unit tests need fixing

sander2:
> Codegen unit tests need fixing

Should be fixed now

PR #310, opened 2021-11-04
fix: keep processing a block's events after encountering a dispatch error

sander2:
When processing a block, the code immediately returned any encountered errors. Any events produced in the block after the error would be silently dropped. This pr changes the behavior: an error is now handled in the same way as events. That is, they are added to the `VecDeque` and are returned only when they get popped from there. This ensures that all events from the block are correctly processed

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -42,6 +42,7 @@ use sp_core::Bytes;
 
 /// Raw bytes for an Event
 #[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq, Clone))]
 pub struct RawEvent {
     /// The name of the pallet from whence the Event originated.
     pub pallet: String,
--- a/src/subscription.rs
+++ b/src/subscription.rs
@@ -49,7 +49,7 @@ pub struct EventSubscription<'a, T: Config> {
     block: Option<T::Hash>,
     extrinsic: Option<usize>,
     event: Option<(&'static str, &'static str)>,
-    events: VecDeque<RawEvent>,
+    events: VecDeque<Raw>,
     finished: bool,
 }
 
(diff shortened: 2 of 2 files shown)


h4x3rotab:
I have the impression we have made some similar fix locally. In the original code, it recognizes an error produced by any extrinsic including the one we don't care and fails too early.

ascjones (approved):
LGTM.

This seems to be another hangover from when this block of code was designed only to return the events triggered by a specific extrinsic, in which case it wouldn't have been a bug since it would have skipped over any events/errors which were not for that extrinsic.

We should consider adding some unit tests for this block of code since it is so critical and has had a few bugs in it. One way to do it would be to add a `Mock` variant to `EventStorageSubscription`.

sander2:
We have an integration test for this in our client code, but I wasn't able to replicate the test in the subxt repo - I didn't manage to create a single block with 1 failing and 1 succeeding tx. Fwiw, we've been using this patch locally for a long time, so it's been somewhat battle-tested already. I'll look into adding a unit test though

sander2:
@ascjones I added some unit tests. I had to refactor a little bit, because not only did I need to mock `EventStorageSubscription`, but also the event decoder.

I also added a modified version of #283 based on [your comment](https://github.com/paritytech/subxt/pull/283#issuecomment-961876200), so that I could integrate the test for it right away. If you prefer separate prs, I'm happy to remove the last commit here.

ascjones:
> If you prefer separate prs, I'm happy to remove the last commit here.

Yes separate PR please

sander2:
> Yes separate PR please

Ok, updated. I'll update the other pr after this one is merged then.

insipx (approved):
LGTM

PR #317, opened 2021-11-10
Update jsonrpsee

ascjones:


niklasad1 (approved):
LGTM,

we should check that the URLs contains the port number because default port numbers are not assigned by the `WS client` anymore FWIW.

ascjones:
> we should check that the URLs contains the port number because default port numbers are not assigned by the WS client anymore FWIW.

Our default URL contains the port: https://github.com/paritytech/subxt/blob/793c945fbd2de022f523c39a84ee02609ba423a9/src/client.rs#L93.

I assume if the user supplies a URL without the port then `RpcClient::try_from_url(url)` will fail?

niklasad1:
> I assume if the user supplies a URL without the port then RpcClient::try_from_url(url) will fail?

Exactly

h4x3rotab (approved):

PR #319, opened 2021-11-12
Add some basic usage docs to README.

ascjones:


README.md
```diff
@@ -8,7 +8,41 @@ A library to **sub**mit e**xt**rinsics to a [substrate](https://github.com/parit
 
 ## Usage
 
-See [examples](./examples).
+### Downloading metadata from a Substrate node
+
+Use the `subxt-cli` CLI tool to download the metadata for your target runtime from a Substrate node.
```
dvdplm, line 13: I guess it could be any type of node yes? Polkadot, kusama, <your parachain here>?

```suggestion
Use the `subxt-cli` CLI tool to download the metadata for your target runtime from a node.
```

README.md
```diff
@@ -8,7 +8,41 @@ A library to **sub**mit e**xt**rinsics to a [substrate](https://github.com/parit
...
+
+```bash
+cargo run -p subxt-cli -- metadata -f bytes > metadata.scale
```
dvdplm, line 16: Sure would be good to have a release on crates so we could write `cargo install subxt-cli` here. :/

README.md
```diff
@@ -8,7 +8,41 @@ A library to **sub**mit e**xt**rinsics to a [substrate](https://github.com/parit
...
+```
+
+This defaults to querying the metadata of a locally running node on the default `http://localhost:9933/`. If querying
+a different node then the `metadata` command accepts a `--url` argument.
+
+### Generating the runtime API from the downloaded metadata
+
+Simply declare a module and decorate it with the `subxt` attribute which points at the downloaded metadata for the 
+target runtime:
```
dvdplm, line 25:
```suggestion
Declare a module and decorate it with the `subxt` attribute which points at the downloaded metadata for the 
target runtime:
```

README.md
```diff
@@ -8,7 +8,41 @@ A library to **sub**mit e**xt**rinsics to a [substrate](https://github.com/parit
...
+
+```rust
+#[subxt::subxt(runtime_metadata_path = "metadata.scale")]
+pub mod node_runtime { }
+```
+
+**Important:** `runtime_metadata_path` resolves to a path relative to the directory where your project's `Cargo.toml`
```
dvdplm, line 32: dq: relative the *workspace* Cargo.toml or the current crate's Cargo.toml?

README.md
ascjones: Indeed, I suppose I meant to say any node built *using* substrate

README.md
ascjones: Absolutely!

ascjones: I suppose there's nothing stopping us releasing a `subxt-cli` `0.1`

README.md
ascjones: Ah whatever `std::env::var("CARGO_MANIFEST_DIR")` resolves to, which is would be the workspace I believe.

README.md
ascjones: https://github.com/paritytech/subxt/pull/320

README.md
jsdw: I was curious so I had a play; it looks like, however you try compiling a crate in a workspace, CARGO_MANIFEST_DIR is always relative to the crate itself and not the workspace.

Which makes sense I think :)

jsdw (approved):
Assuming this'll be tweaked when the binary is published to crates.io, lgtm!

h4x3rotab (approved):

README.md
ascjones: Ah yes the clue is in the name!

PR #322, opened 2021-11-12
Additional parameters for SignedExtra

kvinwang:
We need to control the transaction longevity in our application. subxt does's have an API to do so.
We can use a custom SignedExtra implementation to fill the CheckMortality parameters. However, a custom SignedExtra has no way to be passed in extra infomation to construct the CheckMortality. This PR add an associated type to allow SignedExtra to carry some custom data to construct the signed extra.  Check [here](https://github.com/Phala-Network/phala-blockchain/blob/db56f7a6e6f490aa68d6ba2af6a498e597e8ea04/crates/phaxt/src/extra.rs#L66-L87) for more detail.

### DIFF
--- a/src/extrinsic/extra.rs
+++ b/src/extrinsic/extra.rs
@@ -230,6 +230,8 @@ impl SignedExtension for ChargeTransactionPayment {
 pub trait SignedExtra<T: Config>: SignedExtension {
     /// The type the extras.
     type Extra: SignedExtension + Send + Sync;
+    /// The additional config parameters.
+    type Parameters: Default + Send + Sync;
 
     /// Creates a new `SignedExtra`.
     fn new(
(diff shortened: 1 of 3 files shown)


h4x3rotab:
This should be an up-to-date verison of #202

src/extrinsic/extra.rs
```diff
@@ -230,13 +230,16 @@ impl SignedExtension for ChargeTransactionPayment {
 pub trait SignedExtra<T: Config>: SignedExtension {
     /// The type the extras.
     type Extra: SignedExtension + Send + Sync;
+    /// The extra config parameters.
+    type Config: Default + Send + Sync;
```
jsdw, line 234: Nit: I was initially a bit confused as there is also a `Config` trait above. Perhaps this should be called osmething more unique (and vague) like `Other`, since presumably its goal is simply to allow other arbitrary data to be provided?

h4x3rotab: How about "Parameters"?

kvinwang: Parameters sounds better. Renamed.

insipx (approved):

h4x3rotab (approved):
LGTM

jsdw (approved):

PR #333, opened 2021-11-26
Ensure metadata is in sync with running node during tests

### LINKED ISSUE #329: Keep Metadata and Substrate version aligned in CI
Currently, we use the latest substrate binary (or whatever is available on your machine for local runs) for tests, but we use a hardcoded metadata blob. These two things can get out of sync.

Let's ask the node for metadata when spawning it, rather than relying on a hardcoded blob.

jsdw:
Adds a `test-runtime` crate with the sole purpose of acquiring metadata from the current available substrate binary, and generating/exposing the subxt node runtime types from it, to use during tests.

This ensures that the metadata we're using is always in sync with the substrate runtime the integration tests are running against.

A next step would be to pick a pinned version of substrate to run during CI that we know we're happy with, and to run a nightly pipeline using latest substrate to spot compatibility breakages.

Closes #329


.github/workflows/rust.yml
```diff
@@ -70,6 +77,13 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v2
 
+      - name: Download Substrate
+        run: |
+            curl "https://releases.parity.io/substrate/x86_64-debian:stretch/latest/substrate/substrate" --output substrate --location
```
jsdw, line 82: I don't much like the repetition of this curl thing, but the alternate way of sharing artifacts between jobs works out to be about as verbose...

test-runtime/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition = "2021"
+
+[dependencies]
+subxt = { path = ".." }
+sp-runtime = { package = "sp-runtime", git = "https://github.com/paritytech/substrate/", branch = "master" }
+codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive", "full", "bit-vec"] }
+
+[build-dependencies]
+hex = "0.4.3"
+reqwest = { version = "0.11.6", features = ["blocking", "json"] }
```
ascjones, line 13: Since we already depend on `subxt` above we could avoid this extra (another!) http client dependency. Should be able to use `subxt::RpcClient`

test-runtime/build.rs
```diff
@@ -0,0 +1,152 @@
...
+                panic!("Cannot connect to substrate node after {} retries", retries);
+            }
+            let res = reqwest::blocking::Client::new()
+                .post(format!("http://localhost:{}", port))
+                .json(&json!({
+                   "id": 1,
+                   "jsonrpc": "2.0",
+                   "method": "state_getMetadata",
+                }))
+                .send();
```
ascjones, line 70: If we use `subxt::RpcClient` as mentioned above, you could just do e.g. https://github.com/paritytech/subxt/blob/7e26f9559c26a9644eb4cd189e06ab1ad8bc16dd/src/rpc.rs#L413

test-runtime/build.rs
```diff
@@ -0,0 +1,152 @@
...
+        "
+        #[subxt::subxt(
+            runtime_metadata_path = \"{}\",
+            generated_type_derives = \"Debug, Eq, PartialEq\"
+        )]
+        pub mod node_runtime {{
+            #[subxt(substitute_type = \"sp_arithmetic::per_things::Perbill\")]
+            use sp_runtime::Perbill;
+        }}
+    ",
```
ascjones, line 109: Could use a raw string literal here (`r#".."#`) to avoid the escapes.

test-runtime/build.rs
```diff
@@ -0,0 +1,152 @@
...
+        .expect("Couldn't write runtime rust output");
+
+    // Re-build if we point to a different substrate binary:
+    println!("cargo:rerun-if-env-changed={}", SUBSTRATE_BIN_ENV_VAR);
+    // Re-build if this file changes:
+    println!("cargo:rerun-if-changed=build.rs");
+}
+
+/// Returns the next open port, or None if no port found in range.
+fn next_open_port() -> Option<u16> {
```
ascjones, line 125: I wonder whether we could extract `TestNodeProcess` and other test utils to their own crate. Not only could we reuse the node spawning logic, but we could publish that crate for use by other projects for testing their nodes. I would certainly use it (this code is originally copied from `cargo-contract`).

Just an idea, don't need to do it here.

tests/integration/main.rs
```diff
@@ -15,13 +15,15 @@
...
 
 #[cfg(test)]
 mod client;
 #[cfg(test)]
 mod frame;
 
-pub use runtime::node_runtime;
+pub use test_runtime::{
+    node_runtime,
+    METADATA,
```
ascjones, line 27: Where's that used?

ascjones (commented):
Main thing is, it would be good to remove `reqwest` since we already have `ureq` and `jsonrpsee` http client dependencies in the workspace.

Also, I remember someone in my presenetation (@athei or @dvdplm?) suggesting we could even enable the option for the macro to fetch the metadata from the running substrate node, which may be something we can do in the future for this use case.

test-runtime/Cargo.toml
jsdw: Good point! I'll make use of that or one of the other http deps instead!

test-runtime/build.rs
jsdw: I need to format in the string path though :)

ascjones: I thought it did...one way to find out...

test-runtime/build.rs
jsdw: I had a very similar thought doing this! I didn't really fancy the amount of churn in this PR, but I can imagine it would be useful!

tests/integration/main.rs
jsdw: Ah yes, I thought it would be useful to make the metadata available because I saw that metadata was used, but then I realised that it was used via the subxt Client, and so currently this export isn't helping with anything.

test-runtime/Cargo.toml
athei: > Also, I remember someone in my presentation (@athei or @dvdplm?) suggesting we could even enable the option for the macro to fetch the metadata from the running substrate node, which may be something we can do in the future for this use case.

Yeah this was my suggestion. To download the metadata inside the proc macro.

test-runtime/build.rs
ascjones: Will be worth doing, separate PR probably better - must remember remove the duplication here.

test-runtime/build.rs
jsdw: Ah, indeed the raw string does work! I'm not sure why I thoguht it didn't now..

test-runtime/build.rs
niklasad1: why aren't we just binding to `0.0.0.0:0` instead and letting the OS find the next free port?

ascjones: That is copied from code I wrote in haste, so most likely not the best way to do it. If that works then let's do it.

test-runtime/Cargo.toml
jsdw: I think that's best left for a separate PR if we want it. It does strike me that adding the feature to the macro would only really be beneficial for this sort of testing though, so I actually quite like the idea that it's separate and that the macro doesn't become a little more complex to handle it :)

test-runtime/build.rs
jsdw: I do like the `0.0.0.0:0` approach normally; the complexity is that you then need to read the actual port used back to connect to the service, so I think I'd need to monitor the stdout (unless you have a better idea?).

In the other tests, we need to grab 3 ports by the looks of things, so it is a little trickier!

jsdw: But yes, in this build script, it's probably a little more robust to scan for a port than rely on any specific stdout from substrate on the matter (but if there's a better way to see which port was used, that'd be great!).

test-runtime/build.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut retries = 0;
+        let mut wait_secs = 1;
+        loop {
+            if retries >= MAX_RETRIES {
+                panic!("Cannot connect to substrate node after {} retries", retries);
+            }
+            let res =
+                subxt::RpcClient::try_from_url(&format!("http://localhost:{}", port))
+                    .await
+                    .unwrap()
```
ascjones, line 66: This might panic if it can't connect

ascjones (approved):
Other than the `unwrap()` LGTM

test-runtime/build.rs
jsdw: Ah yes, I think I took it to be if the url was malformed, and was surprised to then have to move it after the await..

jsdw: Actually, I did have a perusal through the code, and it's all OK as is (if the url is WS, it tries to establish a connection (hence the await), but if it's HTTP it won't try to connect and will only error if not an http(s) URL). 

I'll change to an expect to document this!

niklasad1: yeah, I think we should kill `http` anyway because it won't work for subscriptions.

niklasad1 (approved):

test-runtime/README.md
```diff
@@ -0,0 +1,10 @@
+# test-runtime
+
+The logic for this crate exists mainly in the `build.rs` file.
+
+At compile time, this crate will:
+- Spin up a local `substrate` binary (set the `SUBSTRATE_NODE_PATH` env var to point to a custom binary, otehrwise it'll look for `substrate` on your PATH).
+- Obtain metadata from this node.
```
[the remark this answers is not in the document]
ascjones: That's a good idea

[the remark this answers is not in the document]
jsdw: I assume that the `metadata dump` command would itself start the substrate node and such. wrapping this logic into a separate command that's useful on its own does sound like a good idea to me too!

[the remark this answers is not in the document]
dvdplm: https://github.com/paritytech/substrate/issues/10418

PR #335, opened 2021-11-29
Add nightly CI check against latest substrate.

### LINKED ISSUE #334: Pin PR CI tests against specific substrate; daily compat tests against latest master.
Following #329, we could also pin the integration tests carried out in CI to a specific version of substrate, so that breaking changes in substrate don't start breaking out PRs here.

We still want to know about such breaking changes, and for this we can add a new CI file that does scheduled daily runs like:

```
name: Daily compatibility check against substrate master
on:
  schedule:    
    - cron: "0 9 * * *"
```

There would be a lot of code duplication if we copied all of the CI steps over (you can import jobs I think from other files), but it is probably sufficient just to run a daily `cargo test` against latest substrate.

I'm not yet sure how a cron job failure here would be reported to us; I assume as a red tick on master. We'd need to make sure it was fairly obvious and easy to receive notifications for.

jsdw:
Rel #334


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,41 @@
...
+          override: true
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v1.3.0
+
+      - name: Cargo test
+        uses: actions-rs/cargo@v1.0.3
+        with:
+          command: test
+          args: --all-targets --workspace
```
niklasad1, line 41: maybe do something like: https://github.community/t/create-an-issue-case-the-workflow-fails/17247

jsdw: Ooh, great idea! I'll look at that :)

jsdw:
I realised that the reason that the doc step was failing was because the `tokio1` features was missing from `async-std` and causing a panic that seemed to block the build step from progressing (I'm not yet sure why it didn't just bail!). That's fixed now.

jsdw:
Just FYI: this issue was opened to address the question of "which substrate binary to pin to for PR runs": https://github.com/paritytech/ci_cd/issues/272

ascjones (approved):
LGTM

PR #349, opened 2021-12-02
Make system properties an arbitrary JSON object, plus CI fixes

### LINKED ISSUE #316: System properties should be optional
[There][] are optional properties and empty by default in almost all templates what I have seen (see [example][]).

```rust
/// System properties for a Substrate-based runtime
#[derive(serde::Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
#[serde(rename_all = "camelCase")]
pub struct SystemProperties {
    /// The address format
    pub ss58_format: u8,
    /// The number of digits after the decimal point in the native token
    pub token_decimals: u8,
    /// The symbol of the native token
    pub token_symbol: String,
}
```

[This]: https://github.com/paritytech/subxt/blob/master/src/rpc.rs#L147-L151
[example]: https://github.com/substrate-developer-hub/substrate-node-template/blob/master/node/src/chain_spec.rs#L118-L119
[but there]: https://github.com/substrate-developer-hub/substrate-parachain-template/blob/master/node/src/chain_spec.rs#L118-L120

ascjones:
Closes #316.

System properties are just an arbitrary JSON object defined in the chainspec, so this should be reflected in `subxt`. See https://github.com/paritytech/substrate/blob/b7409a2257c7c231b485ee827daa73d05d303047/client/chain-spec/src/lib.rs#L148

Additionally to satisfy CI:

- Makes timestamp test not flaky
- Fixes clippy errors and warnings

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -157,7 +157,14 @@ impl<T: Config> Client<T> {
         &self.metadata
     }
 
-    /// Returns the system properties
+    /// Returns the properties defined in the chain spec as a JSON object.
+    ///
+    /// # Note
+    ///
+    /// Some chains use this to define common properties such as `token_decimals` and `token_symbol`
+    /// required for UIs, but this is merely a convention. So it is up to the library user to
+    /// deserialize the JSON into the appropriate type or otherwise extract the properties defined
+    /// in the target chain's spec.
     pub fn properties(&self) -> &SystemProperties {
         &self.properties
     }
--- a/src/client.rs
+++ b/src/client.rs
@@ -162,7 +162,7 @@ impl<T: Config> Client<T> {
     /// # Note
     ///
     /// Many chains use this to define common properties such as `token_decimals` and `token_symbol`
-    /// required for UIs, but this is merely a convention. So it is up to the library user to
+    /// required for UIs, but this is merely a convention. It is up to the library user to
     /// deserialize the JSON into the appropriate type or otherwise extract the properties defined
     /// in the target chain's spec.
     pub fn properties(&self) -> &SystemProperties {
--- a/src/client.rs
+++ b/src/client.rs
@@ -161,7 +161,7 @@ impl<T: Config> Client<T> {
     ///
     /// # Note
     ///
-    /// Some chains use this to define common properties such as `token_decimals` and `token_symbol`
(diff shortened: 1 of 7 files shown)


src/rpc.rs
```diff
@@ -141,17 +141,8 @@ impl From<u32> for BlockNumber {
...
-pub struct SystemProperties {
-    /// The address format
-    pub ss58_format: u8,
-    /// The number of digits after the decimal point in the native token
-    pub token_decimals: u8,
-    /// The symbol of the native token
-    pub token_symbol: String,
-}
+/// Arbitrary properties defined in the chain spec as a JSON object.
+pub type SystemProperties = serde_json::Map<String, serde_json::Value>;
```
niklasad1, line 145: could this even be `serde_json::Value` instead to be more permissive?

niklasad1 (approved):

src/rpc.rs
ascjones: I could, but the type is specified to be `Map` on the substrate side so I think it should mirror that. I think it would need properties, it doesn't really make sense to have discrete values for the properties.

niklasad1:
> I could, but the type is specified to be Map on the substrate side so I think it should mirror that. I think it would need properties, it doesn't really make sense to have discrete values for the properties.

I don't think it matters anyway, let's ignore my comment :D

src/client.rs
```diff
@@ -157,7 +157,14 @@ impl<T: Config> Client<T> {
         &self.metadata
     }
 
-    /// Returns the system properties
+    /// Returns the properties defined in the chain spec as a JSON object.
+    ///
+    /// # Note
+    ///
+    /// Some chains use this to define common properties such as `token_decimals` and `token_symbol`
```
dvdplm, line 164:
```suggestion
    /// Many chains use this to define common properties such as `token_decimals` and `token_symbol`
```

src/client.rs
```diff
@@ -157,7 +157,14 @@ impl<T: Config> Client<T> {
...
+    /// required for UIs, but this is merely a convention. So it is up to the library user to
```
dvdplm, line 165:
```suggestion
    /// required for UIs, but this is merely a convention. It is up to the library user to
```

dvdplm (approved):

PR #352, opened 2021-12-03
Refactor type generation, remove code duplication

### LINKED ISSUE #328: Refactor duplication for generation of composite fields
Extracted from #309.

There is some code duplication between `type_def` and `struct_def` for generating composite fields for structs and enum variants. This can be consolidated and would ahve avoided bugs such as #327.

ascjones:
Closes #328. Pure refactoring of the codegen to unify code generating structs and fields, the duplication of which has caused bugs such as: https://github.com/paritytech/subxt/pull/327.

The main goal of this refactoring was to consolidate the generation of code which produces a set of fields, give some `scale_info::Field`s. This is required in several different places:

1. Generation of standalone structs from the type registry
2. Generation of enum variants, which themselves are represented as a sequence of `scale_info::Field`s.
3. Generation of [`call`](https://github.com/paritytech/subxt/pull/352/files#diff-150e3adf812594d63375d8b942ddc19adf74456c59d8805a560866ae3ccf9bc7R336) and `event` structs, extracted from the variants of the top level `Call` and `Event` types. 

This refactoring moves most of the core code generation for types which consist of a set of fields to the [`CompositeDef`](https://github.com/paritytech/subxt/pull/352/files#diff-e13ca4a58deba66024618e9ebdca7de031d6245c380de8e4f7d06c718978e426R42) type. This type is an "intermediate representation" of some set of fields for generation of a `struct` or an `enum` variant.  A good place to start is to look at it's [`ToTokens` implementation](https://github.com/paritytech/subxt/blob/66db15f89cf3d5808c2156d004ed0e5511481a5a/codegen/src/types/composite_def.rs#L112) and the [`field_tokens` method](https://github.com/paritytech/subxt/blob/66db15f89cf3d5808c2156d004ed0e5511481a5a/codegen/src/types/composite_def.rs#L214) which performs the actual codegen of the fields. You will see that all logic to do with adding compact attributes and phantom data fields is consolidated here, where previously it was spread around in various places. See code removed in `struct_def.rs` and `type_def.rs`.

I hope that it also improves the modularity of the types codegen to make it easier to understand and maintain. I think there could be more improvements here but this is a step in the right direction.


codegen/src/types/composite_def.rs
```diff
@@ -0,0 +1,320 @@
...
+    EnumVariant,
+}
+
+/// Encapsulates the composite fields, keeping the invariant that all fields are either named or
+/// unnamed.
+#[derive(Debug)]
+pub struct CompositeDefFields {
+    named: bool,
+    fields: Vec<CompositeDefField>,
+}
```
jsdw, line 166: I wonder whether a slightly better fitting type here would be something like:

```
enum CompositeDefFields {
   Named(Vec<(syn::Ident, CompositeDefField)>),
   Unnamed(Vec<CompositeDefField>)
}
```

And then remove the `name` from `CompositeDefField`. You wouldn't have to check a boolean or handle the option<names> in that case :)

jsdw (approved):
Barring a small suggestion this looks good to me!

codegen/src/types/composite_def.rs
ascjones: Indeed, that is precisely the type as it currently is: https://github.com/paritytech/subxt/blob/415e1222d23ca5b00a68128f63c74e4554780a4e/codegen/src/struct_def.rs#L40

It's been a while since I did this but I remember getting into difficulties with this type across it's different usage contexts, hence the relaxing of the strict type.

I will look again at this to see whether it can be improved.

codegen/src/types/composite_def.rs
```diff
@@ -0,0 +1,320 @@
...
+                    true,
+                );
+
+                quote! {
+                    #derives
+                    pub struct #name #type_params #fields
+                }
+            }
+            CompositeDefKind::EnumVariant => {
+                let fields = self.fields.field_tokens(None, None, false);
```
dvdplm, line 135: I wonder if it would make sense to have separate `field_tokens` methods for Struct vs Enums, e.g. `struct_tokens()`/`enum_tokens()`, to make the signature a bit easier on the eyes here.

codegen/src/types/type_def_params.rs
```diff
@@ -0,0 +1,76 @@
...
+// You should have received a copy of the GNU General Public License
+// along with subxt.  If not, see <http://www.gnu.org/licenses/>.
+
+use super::TypeParameter;
+use crate::types::CompositeDefField;
+use quote::quote;
+use std::collections::HashSet;
+
+#[derive(Clone, Debug, Default)]
+pub struct TypeDefParameters {
```
dvdplm, line 23: SOme docs would be good here. Maybe on on the module level as well.

codegen/src/types/type_def_params.rs
```diff
@@ -0,0 +1,76 @@
...
+        let mut used_type_params = HashSet::new();
+        for field in fields {
+            field.type_path.parent_type_params(&mut used_type_params)
+        }
+        for used_type_param in &used_type_params {
+            self.unused.remove(used_type_param);
+        }
+    }
+
+    /// Construct a [`core::marker::PhantomData`] type for the type unused type params.
```
dvdplm, line 46: s/for the type unused type params./for the unused type params./

codegen/src/types/type_def_params.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn unused_params_phantom_data(&self) -> Option<syn::TypePath> {
+        if self.unused.is_empty() {
+            return None
+        }
+        let mut params = self.unused.iter().collect::<Vec<_>>();
```
dvdplm, line 51: Maybe `let mut params = Vec::from_iter(self.unused);`?

dvdplm (approved):
LGTM, left a few nitpicks.

codegen/src/types/type_def_params.rs
ascjones: Hm, actually I have just changed this from `HashSet` to a `BTreeSet` for consistent ordering of params, so the conversion to `Vec` is no longer required.

That said I'm never 100% sure when to use the explicit form e.g. `Vec::from_iter`. Although probably better than the turbofish in this case, perhaps `let mut params: Vec<_> = self.unused.iter().collect();` implicit form is most idiomatic?

codegen/src/types/composite_def.rs
ascjones: This change has now been applied, let me know if it looks good

codegen/src/types/composite_def.rs
ascjones: Done, let me know if that works now.

codegen/src/types/type_def_params.rs
ascjones: Done

PR #354, opened 2021-12-03
Implement new API for `sign_and_submit_then_watch`

### LINKED ISSUE #312: Finer grained event subscription API
Currently the `Event` subscription API attempts to simplify subscribing to events, at the cost of not giving the user very much control over which events are received from the subscription, and making some assumptions which may be false. See https://github.com/paritytech/subxt/pull/310, https://github.com/paritytech/subxt/pull/307, https://github.com/paritytech/subxt/pull/283, https://github.com/paritytech/subxt/pull/306 and maybe more.

We should consider an API similar to [`polkadot.js`](https://polkadot.js.org/docs/api/start/api.tx.subs), which will give the user full access to the transaction status etc.

```typescript
// Make a transfer from Alice to BOB, waiting for inclusion
const unsub = await api.tx.balances
  .transfer(BOB, 12345)
  .signAndSend(alice, (result) => {
    console.log(`Current status is ${result.status}`);

    if (result.status.isInBlock) {
      console.log(`Transaction included at blockHash ${result.status.asInBlock}`);
    } else if (result.status.isFinalized) {
      console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
      unsub();
    }
  });
```

In this way the user receives all events, and is responsible for filtering and acting upon them.

One obvious thing we need to consider is how to handle subscription cancellation ergonomically in Rust with async/await. 

See also https://github.com/paritytech/subxt/issues/309#issuecomment-960532382

jsdw:
This PR is an attempt to implement the API for subscribing to transaction state and obtaining events as described in #312. 

Closes #312


jsdw:
So, I'm fairly happy with this API now overall.

Some things that come out of this PR that could be addressed separately:
- More docs, tests and examples of the new API. (especially examples, to show how you can dig into each stage of a transaction in more detail if needed).
- Making `TransactionProgress` impl `Stream` (probably an upstream change in `jsonrpsee` is easiest).
- Tidy up the event subscription stuff. Now that the new API doesn't need to subscribe to events, we can probably simplify and have a second look at the use cases for the separate event subscription API. (I'd like to think that we could strip it right down, and rely on a `Stream` impl to provide combinators for filtering and working with events).

insipx (approved):
Looks pretty clean to me, nj

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+pub struct TransactionProgress<'client, T: Config> {
+    sub: Option<RpcSubscription<TransactionStatus<T::Hash, T::Hash>>>,
+    ext_hash: T::Hash,
+    client: &'client Client<T>,
+}
+
+impl<'client, T: Config> std::fmt::Debug for TransactionProgress<'client, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TransactionProgress")
+            .field("sub", &self.sub)
```
dvdplm, line 285: Does this actually print something useful?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+        let sub = match &mut self.sub {
+            Some(sub) => sub,
+            None => return Ok(None),
+        };
+
+        // Return the next item otherwise:
+        let res = sub.next().await?;
+        Ok(res.map(|status| {
+            match status {
+                TransactionStatus::Future => TransactionProgressStatus::Future,
```
dvdplm, line 319: Would `TransactionStatus::Pending` be more idiomatic?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+
+    /// Wait for the transaction to be finalized, and return a [`TransactionInBlock`]
+    /// instance when it is, or an error if there was a problem waiting for finalization.
+    ///
+    /// **Note:** consumes `self`. If you'd like to perform multiple actions as the state of the
+    /// transaction progresses, use [`TransactionProgress::next()`] instead.
+    pub async fn wait_for_finalized(
+        mut self,
+    ) -> Result<TransactionInBlock<'client, T>, Error> {
+        while let Some(status) = self.next().await? {
```
dvdplm, line 407: Are transport-level errors handled somewhere else or do we ignore them with the blanket `RpcError::Custom` below?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    /// use [`TransactionProgress::next()`] instead.
+    pub async fn wait_for_finalized_success(self) -> Result<TransactionEvents<T>, Error> {
+        let evs = self.wait_for_finalized().await?.wait_for_success().await?;
+        Ok(evs)
+    }
+}
+
+/// Possible transaction progess states returned from our [`TransactionProgress::next()`] call.
+#[derive(Debug)]
+pub enum TransactionProgressStatus<'client, T: Config> {
```
dvdplm, line 445: Could possibly be shortened into `TransactionStatus`?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    /// Transaction is part of the future queue.
+    Future,
+    /// Transaction is part of the ready queue.
+    Ready,
+    /// The transaction has been broadcast to the given peers.
+    Broadcast(Vec<String>),
+    /// Transaction has been included in block with given hash.
+    InBlock(TransactionInBlock<'client, T>),
+    /// The block this transaction was included in has been retracted.
+    Retracted(T::Hash),
```
dvdplm, line 455: Is it safe to retry the extrinsic after this? Or do users need to rebuild it from scratch? Either way it'd be good to document.

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    /// Maximum number of finality watchers has been reached,
+    /// old watchers are being removed.
```
dvdplm, line 457: Not sure I understand what this means. What is a "finality watcher"? What's the max? And isn't this a timeout?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    FinalityTimeout(T::Hash),
+    /// Transaction has been finalized by a finality-gadget, e.g GRANDPA
+    Finalized(TransactionInBlock<'client, T>),
+    /// Transaction has been replaced in the pool, by another transaction
+    /// that provides the same tags. (e.g. same (sender, nonce)).
+    Usurped(T::Hash),
```
dvdplm, line 463: Is this safe to retry (with a new nonce)?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    /// Transaction has been dropped from the pool because of the limit.
+    Dropped,
+    /// Transaction is no longer valid in the current state.
```
dvdplm, line 466: When can this happen? Worth mentioning in the docs?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    pub fn as_in_block(&self) -> Option<&TransactionInBlock<'client, T>> {
+        match self {
+            Self::InBlock(val) => Some(val),
+            _ => None,
+        }
+    }
+}
+
+/// This struct represents a transaction that has made it into a block.
+pub struct TransactionInBlock<'client, T: Config> {
```
dvdplm, line 491: Would it make sense to provide an easy way for users to query the finalization status of one of these? Imagine a scenario where the users' app takes an optimistic stance, awaits "InBlock" status and then prepares things on its end anticipating finalization and then, before enacting whatever the app does when the extrinsic was successful, it checks if the block is actually final. Maybe `is_final()`?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    ext_hash: T::Hash,
+    client: &'client Client<T>,
+}
+
+impl<'client, T: Config> std::fmt::Debug for TransactionInBlock<'client, T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TransactionInBlock")
+            .field("block_hash", &self.block_hash)
+            .field("ext_hash", &self.ext_hash)
+            .field("client", &"<client>")
```
dvdplm, line 502: Is this what is stopping us from `#[derive(Debug)]` instead of the manual impl? Maybe we can have a manual `Debug` impl on `Client` instead?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+                )?;
+                return Err(runtime_error.into())
+            }
+        }
+
+        Ok(events)
+    }
+
+    /// Fetch all of the events associated with this transaction. This succeeds whether
+    /// the transaction was a success or not; it's up to you to handle the error and
```
dvdplm, line 548:
```suggestion
    /// the transaction was a success or not; it's up to the user to handle the error and
```

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+    /// success events however you prefer.
```
dvdplm, line 549:
```suggestion
    /// success events however preferred.
```

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+                    dispatch_error,
+                )?;
+                return Err(runtime_error.into())
+            }
+        }
+
+        Ok(events)
+    }
+
+    /// Fetch all of the events associated with this transaction. This succeeds whether
```
dvdplm, line 547: Do you think the docs should say if a call implies one or more network calls?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+            .block(Some(self.block_hash))
+            .await?
+            .ok_or(Error::Transaction(TransactionError::BlockHashNotFound))?;
+
+        let extrinsic_idx = block.block.extrinsics
+            .iter()
+            .position(|ext| {
+                let hash = T::Hashing::hash_of(ext);
+                hash == self.ext_hash
+            })
```
dvdplm, line 563: Are the extrinsics returned in any particular order? Can this be slow (even a DoS vector?) if the block has many extrinsics?

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+            // If we successfully obtain the block hash we think contains our
+            // extrinsic, the extrinsic should be in there somewhere..
+            .ok_or(Error::Transaction(TransactionError::BlockHashNotFound))?;
+
+        let raw_events = self
```
dvdplm, line 568: Could be batched with the first rpc call?

src/error.rs
```diff
@@ -158,3 +161,24 @@ pub struct PalletError {
     /// The error description.
     pub description: Vec<String>,
 }
+
+/// Transaction error.
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+pub enum TransactionError {
+    /// The finality subscription expired
```
dvdplm, line 168: State what the timeout is

src/error.rs
```diff
@@ -158,3 +161,24 @@ pub struct PalletError {
...
+    #[error("The finality subscription expired")]
+    FinalitySubscriptionTimeout,
+    /// The transaction was invalid
+    #[error("The transaction was invalid")]
+    Invalid,
+    /// The transaction was dropped from the pool
+    #[error("The transaction was dropped from the pool")]
+    Dropped,
```
dvdplm, line 176: safe to retry?

src/error.rs
```diff
@@ -158,3 +161,24 @@ pub struct PalletError {
...
+    /// The transaction was usurped
+    #[error("The transaction was usurped")]
+    Usurped,
```
dvdplm, line 179: safe to retry?

src/error.rs
```diff
@@ -158,3 +161,24 @@ pub struct PalletError {
...
+    /// The block hash that the tranaction was added to could not be found.
+    /// This is probably because the block was retracted before being finalized.
+    #[error("Block hash containing non-finalized transaction can no longer be found")]
```
dvdplm, line 182:
```suggestion
    #[error("Block hash containing non-finalized transaction can no longer be found. Retracted?")]
```

src/client.rs
```diff
@@ -280,3 +270,368 @@ where
...
+
+        let extrinsic_idx = block.block.extrinsics
+            .iter()
+            .position(|ext| {
+                let hash = T::Hashing::hash_of(ext);
+                hash == self.ext_hash
+            })
+            // If we successfully obtain the block hash we think contains our
+            // extrinsic, the extrinsic should be in there somewhere..
+            .ok_or(Error::Transaction(TransactionError::BlockHashNotFound))?;
```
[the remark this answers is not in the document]
jsdw: I ended up going for "The block containing the transaction can no longer be found (perhaps it was on a non-finalized fork?)"; what do you reckon?

examples/submit_and_watch.rs
```diff
@@ -42,14 +42,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         .balances()
         .transfer(dest, 10_000)
         .sign_and_submit_then_watch(&signer)
+        .await?
+        .wait_for_finalized_success()
         .await?;
 
-    if let Some(event) = result.find_event::<polkadot::balances::events::Transfer>()? {
+    let transfer_event =
+        balance_transfer.find_event::<polkadot::balances::events::Transfer>()?;
```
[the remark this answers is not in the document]
jsdw: You can have an arbitrary number of events that look the same actually. In many examples there's just one event to find. You can iterate and filter the events using a similar function youself, but this comment did get me thinking that perhaps the least-footgunny "simple" API to expose should return an iterator of found events. I'll have a think!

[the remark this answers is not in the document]
jsdw: (I changed `find_event` to `find_first_event` and added a helper to return the list, which should make it clearer to people what to expect!)

src/client.rs
jsdw: Not really; the `ext_hash` maybe at best. The `sub` debug is quite verbose and I haven't paid much attention to what it contains.

src/client.rs
jsdw: Yep, I reckon so! Buuttt, these variant names are ultimately copied verbatim out of substrate, and need to maintain compatibility with it, so while we _could_ rename the variants without an issue I think, it would probably confuse us when we need to check that they are in sync if we change the names, so I'd be hesitant unless it really helped users (and I reckon we can do that with good docs :)).

src/client.rs
jsdw: Transport level errors are returned by that `?` on the `.await` on this line. The blanket custom error (which I'm not really a fan of, but just copied from the previous approach) is just to handle the case that the subscription is dropped for some reason upstream without telling us any of the expected transaction progress states.

src/client.rs
jsdw: Oooh, I wanted to keep it separate from `TransactionStatus` which also exists in this crate, although I realise now that that enum isn't actually exposed to the public anywhere, so I'll make this name nicer and make that one uglier.

src/client.rs
jsdw: Judging by how the old code worked, if this event happens we expect that the transaction will still make its way into a block eventually.

I'll make a separate issue I think for documenting these things better; the comments all come straight out of substrate, and so I'm a bit hesitant to change them to what I assume they mean without double checking with somebody first (but they are pretty crap, and could def be improved)!

jsdw: So I dug into the substrate docs and I found a useful comment on the enum itself which I've adapted and will put into this, which I think helps to clear up some of the confusion (but also makes it more ambiguous in some ways...)

src/client.rs
dvdplm: Fair.

(But yeah, we can obviously open PRs to substrate for things we feel look odd too)

src/client.rs
jsdw: This comment makes no sense to me either; I re-wrote it!

src/client.rs
jsdw: Hopefully 5f9ce69e52a3987261d4ae14e6c9e736ba561664 explaing this more thoroughly!

src/client.rs
jsdw: Hopefully 5f9ce69e52a3987261d4ae14e6c9e736ba561664 helps, but I've still no idea when specifically a transaction can be deemed invalid; presumably the signature doesnt line up or some other thing like that, and it's possible that even if this is the case, once a runtime upgrade happens it might no longer be invalid and might still get included in a block!

src/client.rs
jsdw: At some point I looked for such a method, but offhand I haven't found a way to ask a node whether a specific block has been finalized (without getting the finalized head and iterating back!).

The best way to handle this use case is to use the "lower level" API exposed here and do something like:

```
while let Some(status) = transaction_progress.next().await? {
    if let Some(finalized) = status.as_in_block() {
        // made it into a block.. prepare some things
    }
    if let Some(finalized) = status.as_finalized() {
        // it's been finalized. we can check block hash is same as above and complete.
    }
}
```

src/client.rs
jsdw: That's a good idea!

src/client.rs
jsdw: I assume that the user is the person reading these docs, so to me, "you" feels better :)

src/client.rs
jsdw: Hmm; I tried to make the docs here imply that some network chatter was needed ("Fetch"). I might add a note that this is fairly expensive.

src/client.rs
jsdw: Extrinsics are returned in the order that they are added to the block, so we can rely on this position to tell us what the index of the extrinsic was (so that we can work out which events apply to it).

This is quite hideous and I'd like a better API for this personally, but it's the agreed-upon approach for doing this it seems :)

I'd assume that downloading block details can be slower than many calls, and so a possible attack vector for sure.

src/error.rs
jsdw: So, subxt has previously handled these errors as-is. However the docs I've read in substrate imply that even dropped/usurped etc things may still make it into blocks. If I was to follow that logic, there would be only one error ("finality timeout"), but equally, I imagine that most of the time dropped/usurped/invalid transactions will never again be included.

So, is it best to follow the substrate docs on this, which are super vague about when the transaction ends (finalized, or timed out waiting to finalize after a long period), orr, do I continue to assume that these cases are actually errors and that the transaction is probably scuppered.

I think the latter offhand, but do need to document this properly and will probably ask in dumb questions..

src/client.rs
jsdw: I reckon so (though not so easily at the JSONRPC level), although I like knowing that the block was successfully retrieved before trying to do anything else :)

src/error.rs
jsdw: ^ My current plan; stick with a similar approach to previous but add notes about things and add an issue to revisit

(I added a comment on the timeout!)

src/client.rs
dvdplm: I'm fine with "you". It is a bit unusual for docs to address the reader directly like that, but I don't mind it.

jsdw:
@gregdhill @sander2 you guys may both be interested in this PR, and it would def break some stuff :)

See the `examples/submit_and_watch.rs` example for the high level usage. I need to expand this to show the lower level usage as well.

src/error.rs
jsdw: ^ I didn't do that in the end; the higher level APIs now ignore the "error" statuses, knowing that the tx might still be finalized. That felt safer than potentially erroring but your tx still being included!

dvdplm (approved):
LGTM! A few nitpicks.

src/client.rs
```diff
@@ -434,42 +426,92 @@ impl<'client, T: Config> TransactionProgress<'client, T> {
...
+/// 3. `Invalid` transactions may become valid at some point in the future.
+///    (Note that runtimes are encouraged to use `UnknownValidity` to inform the
+///    pool about such cases).
+/// 4. `Retracted` transactions might be included in a future block.
+///
+/// The stream is considered finished only when either the `Finalized` or `FinalityTimeout`
+/// event is triggered. You are however free to unsubscribe from notifications at any point.
+/// The first one will be emitted when the block in which the transaction was included gets
+/// finalized. The `FinalityTimeout` event will be emitted when the block did not reach finality
+/// within 512 blocks. This either indicates that finality is not available for your chain,
```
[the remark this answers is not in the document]
jsdw: > I wonder how much `subxt` (and other tooling) breaks on chains without finality. :/

Mmm; I guess you have to `wait_for_in_block`, and manage to convince yourself using a separate subscription to block headers or something that the block it was in was the best one! I wonder how many chains opt not to have finality!

src/client.rs
```diff
@@ -434,42 +426,92 @@ impl<'client, T: Config> TransactionProgress<'client, T> {
...
+/// or that finality gadget is lagging behind.
 #[derive(Debug)]
-pub enum TransactionProgressStatus<'client, T: Config> {
-    /// Transaction is part of the future queue.
+pub enum TransactionStatus<'client, T: Config> {
+    /// The transaction is part of the "future" queue.
```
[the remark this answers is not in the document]
jsdw: I'm not sure enough on the terminology to really know! are "future" and "ready" transactions both part of the pool? 

The substrate glossary for transaction pool: "A collection of transactions that are not yet included in blocks but have been determined to be valid."

Maybe Future transactions haven't been validated yet? But in the big comment, they do mention "future and ready pool"...

Anyway, shrug, I'd lean towards playing it safe and leaving it as is for now :D

src/client.rs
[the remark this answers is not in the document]
dvdplm: > I wonder how many chains opt not to have finality!

All PoW chains... but not many of those in our ecosystem I guess.

src/client.rs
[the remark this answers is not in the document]
dvdplm: Maybe @ascjones knows? Either way, fine by me to leave it as is.

src/client.rs
niklasad1: fair enough, this is really if the subscription channel stream can't keep up with server.

as we use `mpsc::channel` in jsonrpsee we simply just drop the subscription channel if it gets full, maybe you have ideas to make it more ergonomic? :)

however, this `RpcError::Custom("RPC subscription dropped".into()` "should not during normal circumstances"

src/client.rs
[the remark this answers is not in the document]
ascjones: This distinction is new to me TBH

src/client.rs
```diff
@@ -280,3 +284,416 @@ where
...
+/// 3. Leaving the pool:
+///    - `InBlock`
+///    - `Invalid`
+///    - `Usurped`
+///    - `Dropped`
+/// 4. Re-entering the pool:
+///    - `Retracted`
+/// 5. Block finalized:
+///    - `Finalized`
+///    - `FinalityTimeout`
```
ascjones, line 466: Nice categorisation. Would it bring any value to bring this grouping to the type level, instead of the simple 1 => 1 mapping with `TransactionStatus`? i.e. split the groups into their own enums combined as variants in the top level enum.

src/client.rs
```diff
@@ -280,3 +284,416 @@ where
         Ok(signed)
     }
 }
+
+/// This struct represents a subscription to the progress of some transaction, and is
```
ascjones, line 288: I reckon all this stuff deserves it's own file, shouldn't be in `client.rs`

ascjones (approved):
LGTM

src/client.rs
jsdw: I actually did do this initially. I like the separation; I just wasn't keen on the circular dependency it would introduce (client makes a transactionprogress, transactionprogress et al have a reference to client) :)

src/client.rs
jsdw: I reckon there are few enough "groups", and things can move between them easily enough, that the flat enum is probably easier to work with (since one mode of use involves matching on variants), and has the nice benefit of aligning very closely with thr substrate version, and so easy to keep in sync going forwards :)

src/client.rs
jsdw: (I think I'll give it a try anyway though, because it would be nice breaking that file down!)

PR #361, opened 2021-12-09
Expand the `subscribe_and_watch` example

jsdw:
And make `has_event` take `&self`; it does not need or want ownership.

### DIFF
--- a/examples/submit_and_watch.rs
+++ b/examples/submit_and_watch.rs
@@ -35,6 +35,17 @@ pub mod polkadot {}
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     env_logger::init();
 
+    simple_transfer().await?;
+    simple_transfer_separate_events().await?;
+    handle_transfer_events().await?;
+
+    Ok(())
+}
+
+/// This is the highest level approach to using this API. We use `wait_for_finalized_success`
+/// to wait for the transaction to make it into a finalized block, and also ensure that the
+/// transaction was successful according to the associated events.
+async fn simple_transfer() -> Result<(), Box<dyn std::error::Error>> {
     let signer = PairSigner::new(AccountKeyring::Alice.pair());
     let dest = AccountKeyring::Bob.to_account_id().into();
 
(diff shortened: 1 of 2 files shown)


examples/submit_and_watch.rs
```diff
@@ -62,3 +73,124 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            println!("Balance transfer success: value: {:?}", event.2);
+        } else {
+            println!("Failed to find Balances::Transfer Event");
+        }
+    }
+
+    Ok(())
+}
+
+/// If we need more visibility into thestate of the transaction, we can also ditch
```
emostov, line 128:
```suggestion
/// If we need more visibility into the state of the transaction, we can also ditch
```

examples/submit_and_watch.rs
```diff
@@ -62,3 +73,124 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
     Ok(())
 }
+
+/// This is very similar to `simple_transfer`, except to show that we can handle
+/// waiting for the transaciton to be finalized separately from obtaining and checking
```
niklasad1, line 78:
```suggestion
/// waiting for the transaction to be finalized separately from obtaining and checking
```

examples/submit_and_watch.rs
```diff
@@ -62,3 +73,124 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                println!(
+                    "Balance transfer is now in block (but not finalized): value: {:?}",
+                    event.2
+                );
+            } else {
+                println!("Failed to find Balances::Transfer Event");
+            }
+        }
+        // Finalized!
+        else if let Finalized(details) = ev {
```
niklasad1, line 172: wasn't aware of this syntax but I prefer matching regardless :)

src/transaction.rs
```diff
@@ -436,7 +436,7 @@ impl<T: Config> TransactionEvents<T> {
     }
 
     /// Find an event. Returns true if it was found.
-    pub fn has_event<E: crate::Event>(self) -> Result<bool, Error> {
+    pub fn has_event<E: crate::Event>(&self) -> Result<bool, Error> {
```
niklasad1, line 439: :+1:

niklasad1 (approved):

examples/submit_and_watch.rs
jsdw: Aah, I kept flicking between a `match` and `if let` here, but since I only wanted to check for two states I thoughtI'd save a level of indentation with if let :)

examples/submit_and_watch.rs
```diff
@@ -62,3 +73,124 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    // details, including events. Calling `wait_for_finalized_success` is
+    // equivalent to calling `wait_for_finalized` and then `wait_for_success`:
+    let _events = balance_transfer.wait_for_success().await?;
+
+    // Alternately, we could just `fetch_events`, which grabs all of the events like
+    // the above, but does not check for success, and leaves it up to you:
+    let events = balance_transfer.fetch_events().await?;
+
+    let failed_event =
+        events.find_first_event::<polkadot::system::events::ExtrinsicFailed>()?;
```
dvdplm, line 108: Worth mentioning how `find_first_event` could return error?

dvdplm (approved):

examples/submit_and_watch.rs
jsdw: I don't think it's worth an explicit shout out in the example really, but I added this note to the docs for the relevant methods because it wasn't clear enough there :)

PR #362, opened 2021-12-10
Rebuild test-runtime if `substrate` binary is updated

### LINKED ISSUE #358: `test-runtime` does not re-run when the `substrate` binary on your PATH changes.
`test-runtime` talks to a substrate node and compiles a `subxt` runtime based on that nodes metadata. As it stands, it will cache builds far too aggressively, and so if you update the substrate binary for instance, it will not know to re-build itself, and so the metadata now will be out of sync with the substrate binary.

The workaround at present is to make some change to `build.rs` to trigger a rebuild.

Cargo exposes two ways to trigger a rebuild, which we can (and do) use like so:

```
// Re-build if we point to a different substrate binary:
println!("cargo:rerun-if-env-changed={}", SUBSTRATE_BIN_ENV_VAR);
// Re-build if this file changes:
println!("cargo:rerun-if-changed=build.rs");
```

Ideally, we'd test that the `substrate` binary we point to is identical (eg mathcing hash), and re-run if it changes.

Failing that, we'd want to tell the thing to always re-run, which would be annoying but safer.

Is there a cunning way to achieve either of these given the current machinery available to us?

jsdw:
Prior to this, one could update the substrate binary on their PATH, and the test-runtime would not rebuild in response to this, leading to a potentially out of date runtime being used (the fix was to tweak `test-runtime/build.rs` to force a rebuild).

This resolves the path to the substrate binary and tells cargo to keep an eye on that, too.

Tested by `touch`ing the substrate binary in use, which then caused a rebuild (and otherwise it would not rebuild).

Closes #358

### DIFF
--- a/test-runtime/Cargo.toml
+++ b/test-runtime/Cargo.toml
@@ -12,3 +12,4 @@ codec = { package = "parity-scale-codec", version = "2", default-features = fals
 subxt = { path = ".." }
 sp-core = { package = "sp-core", git = "https://github.com/paritytech/substrate/", branch = "master" }
 async-std = { version = "1.9.0", features = ["attributes", "tokio1"] }
+which = "4.2.2"
--- a/test-runtime/build.rs
+++ b/test-runtime/build.rs
@@ -113,6 +113,10 @@ async fn run() {
     fs::write(&runtime_path, runtime_api_contents)
         .expect("Couldn't write runtime rust output");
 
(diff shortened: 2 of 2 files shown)


dvdplm (approved):

ascjones (approved):
LGTM.

This reminds me, I think we need to look at the `subxt` macro to make sure it rebuilds when the contents of the static metadata file has changed. It probably won't rebuild atm.

PR #370, opened 2021-12-14
Implement `pre_dispatch` for SignedExtensions

ascjones:
`pre_dispatch` is a new method on the `SignedExtension` trait from substrate. We don't care about this internal implementation detail on the `subxt` side, but we require this trait to be implemented because we are using the `SignedPayload` and `UncheckedExtrinsic` types from `sp_runtime`.

The other option is of course to define our own local types and traits to remove this direct coupling, but that need some careful consideration.

In the meantime this makes things compile and continue working against the `substrate` master branch.

This also fixes an issue with duplicate substrate dependencies.

### DIFF
--- a/macro/Cargo.toml
+++ b/macro/Cargo.toml
@@ -34,4 +34,4 @@ pretty_assertions = "0.6.1"
 subxt = { path = ".." }
 trybuild = "1.0.38"
 
-sp-keyring = { package = "sp-keyring", git = "https://github.com/paritytech/substrate/" }
+sp-keyring = { package = "sp-keyring", git = "https://github.com/paritytech/substrate/", branch = "master" }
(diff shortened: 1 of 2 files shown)


dvdplm (approved):

macro/Cargo.toml
```diff
@@ -34,4 +34,4 @@ pretty_assertions = "0.6.1"
 subxt = { path = ".." }
 trybuild = "1.0.38"
 
-sp-keyring = { package = "sp-keyring", git = "https://github.com/paritytech/substrate/" }
+sp-keyring = { package = "sp-keyring", git = "https://github.com/paritytech/substrate/", branch = "master" }
```
niklasad1, line 37:
```suggestion
sp-keyring = { git = "https://github.com/paritytech/substrate/", branch = "master" }
```

niklasad1 (approved):

PR #371, opened 2021-12-14
Adds option to pass a mortality parameter to create_signed

conectado:
For my use case it'd be useful to add a parameter `mortality` to the `extrinsic::create_signed` function.

So that not every transaction has to be `Era::Immortal`. 

I was wondering if there's a chance to merge this into master?


conectado:
Nvm, just saw #204

PR #373, opened 2021-12-16
Configurable `Config` and `Extra` types

### LINKED ISSUE #331: Configurable extrinsic `Extra`
From #294

> Allow user to define default `Extra` implementation for additional signed data, or auto-generate from the metadata. At the moment this is hardcoded in the macro generated code. https://github.com/paritytech/subxt/blob/fc30e90282a2537b32e77a0330106f6979e0cb5a/codegen/src/api/mod.rs#L245

Replace hardcoded default extensions, these could be generated from the metadata, or the user could supply their own implementations.

Slight wrinkle is that a couple of the hardcoded ones require some data we need to supply up front, notably nonce and genesis hash. And currently the supplying of those values is done by the client, so need to come up with some solution there. It's probably safe to assume that most/all chains will support those two extensions.

Rel #330

ascjones:
Closes #331 

Currently the `Config`  and `Extra` trait implementations are hardcoded by the macro, which means that if the target runtime has different types or extrinsic signed extensions, then it will not work.

This PR enables the user to supply their own implementations of `Config` and `Extra` as parameters to the `RuntimeApi` type generated by the macro. E.g.

```rust
        .build()
        .await?
        .to_runtime_api::<runtime::RuntimeApi<subxt::DefaultConfig, subxt::DefaultExtra<subxt::DefaultConfig>>>();
```
Here we are using the built in default implementations as type parameters, but the user is now able to supply their own implementations in their place.

This may not be the final API, since we also need to consider how we can integrate the initialisation of some of the additional signed data (e.g. transaction payment), which currently have default values set. Also we will probably want to use metadata to have the macro generate the correct set of signed extensions, rather than requiring them to be hardcoded upfront.

However these things can be built upon this foundation, which at least allows the user the ability to customise these previously hardcoded types.


codegen/src/api/mod.rs
```diff
@@ -222,76 +223,55 @@ impl RuntimeGenerator {
...
                         result.nonce
                     }
-                    fn storage_entry(account_id: <DefaultConfig as ::subxt::Config>::AccountId) -> Self {
+                    fn storage_entry(account_id: <::subxt::DefaultConfig as ::subxt::Config>::AccountId) -> Self {
                         Self(account_id)
                     }
                 }
 
-                pub struct RuntimeApi<T: ::subxt::Config + ::subxt::ExtrinsicExtraData<T>> {
+                pub struct RuntimeApi<T: ::subxt::Config, E> {
```
jsdw, line 239: Is it worth defaulting these generic params eg

```
pub struct RuntimeApi<T: ::subxt::Config = ::subxt::DefaultConfig, E = ::subxt::DefaultExtra<::subxt::DefaultConfig>> {
```

so that we don't have to specify them when instantiating the `RuntimeApi` unless they differ? Or is the common case that you'll need to specify different ones, so it may as well be explicit?

jsdw (approved):
Looks great to me; basically a ton of threading a new generic param or two through things for that extra flexibility as far as I understand it. Nice one!

codegen/src/api/mod.rs
ascjones: I'm not sure TBH, on the one hand it is very convenient if you are using the defaults, on the other if they are hidden then there is not the clue that it is configurable.

I'd lean towards leaving it explicit for now, until the api has solidified and we have an idea of the common usage.

PR #375, opened 2021-12-20
Use released substrate dependencies

ascjones:
Use substrate crates from `crates.io`, which will allow us to do a release of this crate.

# todo

- [ ] Release `sp-keyring` which is still at an old version on `crates.io`. It's a dev dependency but it depends on `sp-core` and `sp-runtime`, so brings in conflicting versions which causes the tests to not compile.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -107,9 +107,6 @@ pub enum RuntimeError {
     /// There are no providers so the account cannot be created.
     #[error("There are no providers so the account cannot be created.")]
     NoProviders,
-    /// There are too many consumers so the account cannot be created.
-    #[error("There are too many consumers so the account cannot be created.")]
-    TooManyConsumers,
     /// Bad origin.
     #[error("Bad origin: throw by ensure_signed, ensure_root or ensure_none.")]
     BadOrigin,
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -144,7 +141,6 @@ impl RuntimeError {
             DispatchError::CannotLookup => Ok(Self::CannotLookup),
             DispatchError::ConsumerRemaining => Ok(Self::ConsumerRemaining),
             DispatchError::NoProviders => Ok(Self::NoProviders),
-            DispatchError::TooManyConsumers => Ok(Self::TooManyConsumers),
```
ascjones, line 147: We may want to parameterize `DispatchError` and allow the definition in the metadata to be used, in case of such a mismatch in the `DispatchError` type between the target node and the released version of `sp_runtime`

jsdw: Working on that in https://github.com/paritytech/subxt/pull/394

dvdplm (approved):
🎉

macro/Cargo.toml
```diff
@@ -33,5 +33,4 @@ subxt-codegen = { version = "0.2.0", path = "../codegen" }
 pretty_assertions = "0.6.1"
 subxt = { path = ".." }
 trybuild = "1.0.38"
-
-sp-keyring = { git = "https://github.com/paritytech/substrate/", branch = "master" }
+sp-keyring = { version = "4.0.0" }
```
niklasad1, line 36:
```suggestion
sp-keyring = "4.0.0"
```

niklasad1 (approved):

macro/Cargo.toml
jsdw: Ooh good point :) I'll fix both of those!

PR #382, opened 2022-01-04
fix: use index override when decoding enums in events

sander2:
Without this fix, if an event contains an enum with explicit indices, it fails to decode

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -187,13 +187,13 @@ where
             TypeDef::Variant(variant) => {
                 let variant_index = u8::decode(input)?;
                 variant_index.encode_to(output);
-                let variant =
-                    variant
-                        .variants()
-                        .get(variant_index as usize)
-                        .ok_or_else(|| {
-                            Error::Other(format!("Variant {} not found", variant_index))
-                        })?;
+                let variant = variant
+                    .variants()
+                    .iter()
+                    .find(|v| v.index() == variant_index)
+                    .ok_or_else(|| {
+                        Error::Other(format!("Variant {} not found", variant_index))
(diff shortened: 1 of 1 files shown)


jsdw (approved):
LGTM!

ascjones (commented):
I know there are already no tests for the events decoding, but I think we need some.

Should be trivial to add a test for this after https://github.com/paritytech/subxt/pull/384 is merged (which adds unit testing for the `EventsDecoder`).

ascjones:
@sander2 now https://github.com/paritytech/subxt/pull/384 is merged, you should be able to easily make a test for this now.

src/events.rs
```diff
@@ -522,4 +522,36 @@ mod tests {
         assert_eq!(events[0].1.variant_index, encoded_event[0]);
         assert_eq!(events[0].1.data.0, encoded_event[1..]);
     }
+
+    #[test]
+    fn event_containing_explicit_index() {
+        #[derive(Clone, Encode, TypeInfo)]
+        #[repr(u8)]
+        #[allow(trivial_numeric_casts)] // required because the Encode derive produces a warning otherwise
```
niklasad1, line 530:
```suggestion
        #[allow(trivial_numeric_casts, clippy::unnecessary_cast)] // required because the Encode derive produces a warning otherwise
```

niklasad1 (approved):
LGTM

src/events.rs
niklasad1: @sander2 just do this, looks like clippy is wrong it's just a test so should be ok :)

PR #385, opened 2022-01-06
fetch system chain, name and version

paulormart:
I find this additional information useful to be available from a client tool perspective. And goes in line with the methods available in `polkadot.js`.

### DIFF
--- a/tests/integration/client.rs
+++ b/tests/integration/client.rs
@@ -124,19 +124,10 @@ async fn test_iter() {
 }
 
 #[async_std::test]
-async fn fetch_system_chain() {
+async fn fetch_system_info() {
     let node_process = test_node_process().await;
-    node_process.client().rpc().system_chain().await.unwrap();
-}
-
-#[async_std::test]
-async fn fetch_system_name() {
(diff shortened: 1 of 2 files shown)


ascjones (commented):
Nice, please add some tests for these though.

Just something like https://github.com/paritytech/subxt/blob/415e1222d23ca5b00a68128f63c74e4554780a4e/tests/integration/client.rs#L51

tests/integration/client.rs
```diff
@@ -122,3 +122,21 @@ async fn test_iter() {
     }
     assert_eq!(i, 13);
 }
+
+#[async_std::test]
+async fn fetch_system_chain() {
+    let node_process = test_node_process().await;
+    node_process.client().rpc().system_chain().await.unwrap();
+}
```
ascjones, line 130: Actually could you combine these into one single test? Spinning up 3 separate nodes for this will unnecessarily increase CI times.

paulormart: Good point. done

ascjones (approved):
LGTM!

jsdw (approved):

PR #386, opened 2022-01-06
fix remote example

paulormart:
The port number is required in the `url` for the remote node. 

Also `client-ws-transport` feature needs to be specified for `wss` to be handled properly.


insipx (approved):
lgtm, thanks for the fix!

niklasad1 (approved):

PR #389, opened 2022-01-12
Remove bounds on Config trait that aren't strictly necessary

jsdw:
We use types implementing the `Config` trait in Subxt, but only at compile-time; we never instantiate an instance of them. And yet, the Config trait has various bounds on it (`Debug`, `Clone`, `Copy`, `Send`, `Sync` etc).

This PR removes those bounds (well, mostly; `'static` is still there for now and should be able to go away eventually, but that should never complicate anything in real use).

The reason I think that these bounds existed in the first place is that Rust's inbuilt "derive" mechanism is quite primitive (eg deriving `Clone` will derive something like `impl <T: Clone> Clone for MyType<T>`; the `T` also needs to be `Clone`). As a result, we ended up with a bunch of additional bounds on the `Config` type so that it could be passed around as a generic type `T` and would still satisfy these derives.

Removing the bounds on Config has pros and cons:

Pros:
- More technically correct; `Config` is never instantiated, and so it shouldn't need any of these bounds.
- Makes it a little easier for users to implement `Config` on their own types; they don't need to pointlessly derive anything.

Cons:
- Introduces `derivative` crate to allow ignoring the `T` when deriving various traits on things that use `Config`. (we could manually derive the traits everywhere, but that would be needlessly verbose). This additional complexity will apply to anything else that takes a `T: Config` type.
- Adds a `PhantomDataSendSync<T>` type (which is just `PhantomData + Send + Sync`) to ensure that the lack of `Send` + `Sync` on `Config` doesn't translate to a lack of `Send` + `Sync` on types using the config.

I think I lean towards being more "correct" at the cost of the additional complexity (and from a user facing POV, they won't see any of the added complexity, but will benefit a little from `Config` being easier to implement.

That said, I'm curious to hear people's opinions on the matter!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ sp-runtime = { git = "https://github.com/paritytech/substrate/", branch = "maste
 sp-version = { package = "sp-version", git = "https://github.com/paritytech/substrate/", branch = "master" }
 
 frame-metadata = "14.0.0"
+derivative = "2.2.0"
 
 [dev-dependencies]
 sp-arithmetic = { git = "https://github.com/paritytech/substrate/", branch = "master", default-features = false }
--- a/src/client.rs
+++ b/src/client.rs
@@ -40,6 +40,7 @@ use crate::{
     Config,
     Metadata,
 };
+use derivative::Derivative;
 use std::sync::Arc;
 
 /// ClientBuilder for constructing a Client.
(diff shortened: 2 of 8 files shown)


dvdplm (approved):
I think this is a bit of a hack and it adds a fair amount of line noise; the `derivative` crate seems well maintained and is stable. 
I do think this is better than before overall though and I think it's fair to err on the side of ergonomics here.

Code lgtm.

PR #394, opened 2022-01-14
Use the generated DispatchError instead of the hardcoded Substrate one

jsdw:
The intention of this PR is to use the version of DispatchError generated from the metadata, rather than using a baked in version which may not be compatible with the node beign targeted.


codegen/src/api/errors.rs
```diff
@@ -0,0 +1,171 @@
...
+        .iter()
+        .filter_map(|pallet| {
+            pallet.error.as_ref().map(|error| {
+                let type_def_variant = get_type_def_variant(error.ty.id())?;
+                Ok((pallet, type_def_variant))
+            })
+        })
+        .collect::<Result<Vec<(_, _)>, _>>()?;
+
+    let errors = pallet_errors
```
dvdplm, line 121: Maybe we can unify these two chains into a single? Like, why collect the `pallet_errors` first and then build the `errors` Vec?

codegen/src/api/errors.rs
```diff
@@ -0,0 +1,171 @@
...
+                }
+            })
+        })
+        .collect();
+
+    Ok(errors)
+}
+
+#[derive(Clone, Debug)]
+pub struct ErrorMetadata {
```
dvdplm, line 140: Missing docs here. Also, what do you think about moving this to the top?

codegen/src/api/errors.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub pallet_index: u8,
+    pub error_index: u8,
+    pub pallet: String,
+    pub error: String,
+    variant: scale_info::Variant<scale_info::form::PortableForm>,
```
dvdplm, line 145: dq: what is the reason for storing the variant for?

dvdplm: ah, for the `description()`?

codegen/src/api/errors.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl ErrorMetadata {
+    pub fn description(&self) -> String {
+        self.variant.docs().join("\n")
```
dvdplm, line 150: So if the node is compiled to strip the docs, this will be empty? Worth detecting and add some kind of default, "No docs available in the metadata"?

codegen/src/api/errors.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub enum InvalidMetadataError {
```
dvdplm, line 155: Missing docs.

examples/custom_type_derives.rs
```diff
@@ -16,7 +16,7 @@
 
 #[subxt::subxt(
     runtime_metadata_path = "examples/polkadot_metadata.scale",
-    generated_type_derives = "Clone, Debug"
+    generated_type_derives = "Clone"
```
dvdplm, line 19: A comment here and on `runtime_metadata_path` would be good.

dvdplm: Looking further, I think this whole example could use some attention. I am not sure what it actually does. :/
(Not a concern for this PR ofc)

dvdplm (approved):
Overall LGTM, nice work!

codegen/src/api/errors.rs
jsdw: It might be a bit tricky; We end up with `Result<_>` Items just prior to that collect, so the collecting I think is just there to give a chance to bail if we encounter any errors!

jsdw: I settled on standard for loops; it felt tidier:

```rust
let mut pallet_errors = vec![];
for pallet in &metadata.pallets {
    let error = match &pallet.error {
        Some(err) => err,
        None => continue
    };

    let type_def_variant = get_type_def_variant(error.ty.id())?;
    for var in type_def_variant.variants().iter() {
        pallet_errors.push(ErrorMetadata {
            pallet_index: pallet.index,
            error_index: var.index(),
            pallet: pallet.name.clone(),
            error: var.name().clone(),
            variant: var.clone(),
        });
    }
}
```

codegen/src/api/errors.rs
jsdw: Yup, that's it! I'm not sure why I did/left this now; it's just the doc string, which simplifies things a bit!

codegen/src/api/errors.rs
jsdw: In the empty case docs will just be "", which seems good enough (that can be detected and handled down the line in whichever way things prefer)

codegen/src/api/errors.rs
jsdw: Ah; doesn't need to be `pub` :)

examples/custom_type_derives.rs
jsdw: I gave the exampel a wee tidy; it's purpose appears solely to be showing off the `generated_type_derives` usage, so I made that a little clearer (and I didn't see a need for the strange way of calling `clone`.

niklasad1 (approved):
nice, LGTM

PR #395, opened 2022-01-17
Make own version of RuntimeVersion to avoid mismatches

jsdw:
This came about because `stateVersion` went away in a recentish version of substrate/polkadot, which started breaking `subxt` which was quite strict in what it expected to be decoded.

Here we make the decoding more lenient, only expecting the props we need and keeping props that exist but aren't needed in a `HashMap` incase they are useful to others.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -17,7 +17,6 @@
 use futures::future;
 use sp_runtime::traits::Hash;
 pub use sp_runtime::traits::SignedExtension;
-pub use sp_version::RuntimeVersion;
 
 use crate::{
     error::Error,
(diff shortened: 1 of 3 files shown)


niklasad1 (approved):

dvdplm (approved):
I don't think we ever serialize our flavour of `RuntimeVersion` right? We want to flatten when deserializing, but perhaps it would be safer to keep the `other` key when serializing a `subxt::RuntimeVersion` but not sure if that's a thing in serde.

jsdw:
Good points; I'm not actually aware of any need to Serialize that on the subxt side so I might just remove that bound and avoid thinking about it :)

Others can easily replicate and serialize what they need anyway, and we can think about adding it back if it's genuinely useful.

PR #407, opened 2022-01-25
Custom derive for cli

fewensa:
Add custom derive command for cli

Example

```
subxt codegen --url https://pangolin-rpc.darwinia.network --derive Clone --derive Other
```

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -33,6 +33,12 @@ use std::{
     path::PathBuf,
 };
 use structopt::StructOpt;
+use subxt_codegen::GeneratedTypeDerives;
+use syn::{
+    punctuated::Punctuated,
+    Ident,
+    __private::Span,
+};
 
 /// Utilities for working with substrate metadata for subxt.
 #[derive(Debug, StructOpt)]
@@ -70,6 +76,9 @@ enum Command {
         /// the path to the encoded metadata file.
         #[structopt(short, long, parse(from_os_str))]
         file: Option<PathBuf>,
+        /// Additional derives
+        #[structopt(long = "derive")]
+        derives: Vec<String>,
     },
 }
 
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -145,13 +154,24 @@ fn fetch_metadata(url: &url::Url) -> color_eyre::Result<(String, Vec<u8>)> {
...
+    raw_derives: Vec<String>,
+) -> color_eyre::Result<()> {
     let metadata = <RuntimeMetadataPrefixed as Decode>::decode(encoded)?;
     let generator = subxt_codegen::RuntimeGenerator::new(metadata);
     let item_mod = syn::parse_quote!(
         pub mod api {}
     );
-    let runtime_api = generator.generate_runtime(item_mod, Default::default());
+
+    let mut p: Punctuated<syn::Path, syn::Token![,]> = Punctuated::new();
```
jsdw, line 167: This could just be `Vec<syn::Path>` (as far as I can see, that's all it's being used for) and then below, `p.into_iter()` to avoid the `cloned()` :)

cli/src/main.rs
```diff
@@ -33,6 +33,11 @@ use std::{
     path::PathBuf,
 };
 use structopt::StructOpt;
+use subxt_codegen::GeneratedTypeDerives;
+use syn::{
+    Ident,
+    __private::Span,
```
ascjones, line 39: Prefer not to import this since it is marked as `__private`. So you can either use `proc_macro2::Span` or some other way of constructing the `syn::Path`. If you still want to construct an `Ident` you should be able to use `quote::format_ident`, or you can use `syn::parse_str` or similar to go straight to a `Path`.

ascjones (approved):
LGTM

jsdw (approved):

PR #409, opened 2022-01-26
Parameterize `AccountData`

### LINKED ISSUE #391: Cannot use my own `Config` type to instantiate an API
I had a go at trying to create a custom type which implements `Config` to instantiate an API with, like so:

```
#[derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct MyConfig;
impl Config for MyConfig {
    type Index = <DefaultConfig as Config>::Index;
    type BlockNumber = <DefaultConfig as Config>::BlockNumber;
    type Hash = <DefaultConfig as Config>::Hash;
    type Hashing = <DefaultConfig as Config>::Hashing;
    type AccountId = <DefaultConfig as Config>::AccountId;
    type Address = <DefaultConfig as Config>::Address;
    type Header = <DefaultConfig as Config>::Header;
    type Signature = <DefaultConfig as Config>::Signature;
    type Extrinsic = <DefaultConfig as Config>::Extrinsic;
}

    .build()
    .await?
    .to_runtime_api::<polkadot::RuntimeApi<MyConfig, DefaultExtra<MyConfig>>>();
```

(I tried this in the `examples/transfer_subscribe.rs` example).

My custom type uses the same types that `DefaultConfig` uses, but when I try to compile this example I run into:

```
error[E0599]: the method `balances` exists for struct `polkadot::TransactionApi<'_, MyConfig, DefaultExtraWithTxPayment<MyConfig, extrinsic::extra::ChargeTransactionPayment<MyConfig>>, system::storage::Account>`, but its trait bounds were not satisfied
  --> examples/transfer_subscribe.rs:70:10
   |
35 | #[subxt::subxt(runtime_metadata_path = "examples/polkadot_metadata.scale")]
   | ---------------------------------------------------------------------------
   | |
   | method `balances` not found for this
   | doesn't satisfy `_: subxt::AccountData<MyConfig>`
...
70 |         .balances()
   |          ^^^^^^^^ method cannot be called on `polkadot::TransactionApi<'_, MyConfig, DefaultExtraWithTxPayment<MyConfig, extrinsic::extra::ChargeTransactionPayment<MyConfig>>, system::storage::Account>` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `system::storage::Account: subxt::AccountData<MyConfig>`
```

I believe that the error exists because `DefaultConfig` is still hardcoded into a bit of the codegen:

```
pub type DefaultAccountData = self::system::storage::Account;

impl ::subxt::AccountData<::subxt::DefaultConfig> for DefaultAccountData {
    fn nonce(result: &<Self as ::subxt::StorageEntry>::Value) -> <::subxt::DefaultConfig as ::subxt::Config>::Index {
        result.nonce
    }
    fn storage_entry(account_id: <::subxt::DefaultConfig as ::subxt::Config>::AccountId) -> Self {
        Self(account_id)
    }
}
```

This `DefaultAccountData` type is passed as a type param to `TransactionApi` as the param `A`, which expects:

```
impl<'a, T, E, A> TransactionApi<'a, T, E, A>
    where
        T: ::subxt::Config,
        E: ::subxt::SignedExtra<T>,
        A: ::subxt::AccountData<T>,
```

But of course, `DefaultAccountData` does not implement `AccountData` for generic configs; only for the `DefaultConfig`, which prevents other configs being inserted in its place I believe.

`::subxt::AccountData` does however require that the types it knows about align with those generated from `system::storage::Account`, and so it's not a simple case of making the impl more generic. 

I'm not entirely sure what the best course of action is here; perhaps expose the type inside `system::storage::Account` and bound uses of `Config` such that the account ID types line up? But if that were the fix, then why have AccountId in the config at all if it's a hard requirement in the generated code? Likely I am not understanding the reasoning behind this code properly yet.

ascjones:
Fixes #391.

Previously the codegen hardcoded the `AccountData` parameter of `RuntimeApi` to `DefaultAccountData`, which in turn was hardcoded to using the `DefaultConfig`.

This meant that the user was not able to supply their custom `Config` implemetnation, nor a custom `AccountData` implementation.

This PR allows the user to supply their own custom `Config` impl, and optionally a custom `AccountData` impl.

## Update

The macro now autogenerates the `DefaultAccountData` impl against the `System::Account` storage entry, and hardcodes the associated key (`AccountId`) and value (`Nonce`) types by looking them under the known names in the metadata.

I really wanted to allow the type system to do as much as possible with no macro magic, but couldn't get it to work due to [this error.](https://github.com/paritytech/subxt/pull/409#issuecomment-1023019701) So now what it does is it e.g. looks up the type of `System::Account::nonce` and assigns it to `type Nonce`.


examples/custom_config.rs
```diff
@@ -0,0 +1,61 @@
...
+    type Index = <DefaultConfig as Config>::Index;
+    type BlockNumber = <DefaultConfig as Config>::BlockNumber;
+    type Hash = <DefaultConfig as Config>::Hash;
+    type Hashing = <DefaultConfig as Config>::Hashing;
+    type AccountId = <DefaultConfig as Config>::AccountId;
+    type Address = <DefaultConfig as Config>::Address;
+    type Header = <DefaultConfig as Config>::Header;
+    type Signature = <DefaultConfig as Config>::Signature;
+    type Extrinsic = <DefaultConfig as Config>::Extrinsic;
+}
```
jsdw, line 41: Just for the sake of the example, could we actually override one or more of these types perhaps (maybe then with a comment saying what's going on), just to make it clear that that's sortof the point of this impl? (I realise that doing so may prevent it from actually working below; something that would want to be made clear also.)

jsdw (approved):
Wee suggestion re the example but looks great!

trusch:
I tried using your example and ran into this:
```
error[E0277]: the trait bound `<T as Config>::Index: From<u64>` is not satisfied
  --> src/main.rs:26:1
   |
26 | #[subxt::subxt(runtime_metadata_path = "metadata.scale")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<u64>` is not implemented for `<T as Config>::Index`
   |
   = note: required because of the requirements on the impl of `Into<<T as Config>::Index>` for `u64`
```

examples/custom_config.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[subxt::subxt(runtime_metadata_path = "examples/polkadot_metadata.scale")]
+pub mod polkadot {}
+
+/// Custom [`Config`] impl where the default types for the target chain differ from the
+/// [`DefaultConfig`]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct MyConfig;
+impl Config for MyConfig {
+    type Index = u64; // this is different from the default `u32`
```
jsdw, line 34: I like this; worth noting that the example won't run unless this is set back to the default though I think.

jsdw:
> I tried using your example and ran into this:
> 
> ```
> error[E0277]: the trait bound `<T as Config>::Index: From<u64>` is not satisfied
>   --> src/main.rs:26:1
>    |
> 26 | #[subxt::subxt(runtime_metadata_path = "metadata.scale")]
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<u64>` is not implemented for `<T as Config>::Index`
>    |
>    = note: required because of the requirements on the impl of `Into<<T as Config>::Index>` for `u64`
> ```

That is deliberate (but should be emphasised); I think it's worth showing how you'd actually customise the config, but of course doing so does mean that it won't actually run against a default node any more.

examples/custom_config.rs
ascjones: The example does run though, I guess because `u64 : From<u32>`

jsdw: Ah; I assumed that @trusch's comment meant otherwise!

trusch:
Sorry for not providing enough information to reproduce this: I used the metadata from the kilt network and the only change to the config was that I set the `Index` to u64 exactly as in your example. I attached the metadata I used since its only available through the websocket endpoints and therefore not downloadable using subxt-cli. I had to add the `.md` extension so github allowed me to upload it. The code that I try to compile is excactly the same as the code from your example. 

Dependencies look look like this:
```
[dependencies]
subxt = { git = "https://github.com/paritytech/subxt.git", branch = "aj/custom-config" }
codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive", "full", "bit-vec"] }
sp-keyring = "4.0.0"
async-std = { version = "1.9.0", features = ["attributes", "tokio1"] }
```

[metadata.scale.md](https://github.com/paritytech/subxt/files/7949525/metadata.scale.md)

ascjones:
@trusch that file does not contain the metadata, just the `subxt-cli` usage docs :see_no_evil:

ascjones:
@trusch can you try now?

trusch:
@ascjones It worked! The particular error I experienced is gone now :+1: 

I still can't get it to fully compile because one of the types generated from the metadata doesn't properly implement `Decode`:

```
error[E0277]: the trait bound `BoundedBTreeSet<DidEncryptionKey>: Decode` is not satisfied
  --> src/main.rs:26:1
   |
26 | #[subxt::subxt(runtime_metadata_path = "metadata.scale")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Decode` is not implemented for `BoundedBTreeSet<DidEncryptionKey>`
   |
   = help: the following implementations were found:
             <BoundedBTreeSet<_0> as Decode>
```

But nevertheless, thanks for picking up and solving my initial issue so fast!

ascjones:
>  the trait bound `BoundedBTreeSet<DidEncryptionKey>: Decode` is not satisfied

I'm guessing that is because `DidEncryptionKey` does not implement `Ord`. You can work around that by using a type substitute (define your own `DidEncryptionKey` with `Ord` implemented), or possibly add the `Ord` derive to all types: not sure if that will work though,

examples/custom_config.rs
```diff
@@ -0,0 +1,70 @@
...
+    // This is different from the default `u32`.
+    //
+    // *Note* that in this example it does differ from the actual `Index` type in the
+    // polkadot runtime used, so some operations may fail. Normally when using a custom `Config`
+    // impl types MUST match exactly those used in the actual runtime.
+    type Index = u64;
+    type BlockNumber = <DefaultConfig as Config>::BlockNumber;
+    type Hash = <DefaultConfig as Config>::Hash;
+    type Hashing = <DefaultConfig as Config>::Hashing;
+    type AccountId = <DefaultConfig as Config>::AccountId;
```
dvdplm, line 43: I wonder if this example would be more useful if we made a more plausible change to the config, something a project might actually want to do. Not for this PR, it'd be more like a tutorial I reckon.

dvdplm (approved):
lgtm, left a few nits and doc suggestions.

examples/custom_config.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Custom [`Config`] impl where the default types for the target chain differ from the
+/// [`DefaultConfig`]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct MyConfig;
+impl Config for MyConfig {
+    // This is different from the default `u32`.
+    //
+    // *Note* that in this example it does differ from the actual `Index` type in the
+    // polkadot runtime used, so some operations may fail. Normally when using a custom `Config`
```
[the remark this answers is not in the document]
ascjones: Most likely

codegen/src/api/mod.rs
```diff
@@ -352,6 +356,72 @@ impl RuntimeGenerator {
     }
 }
 
+/// Most chains require a valid account nonce as part of the extrinsic, so the default behaviour of
+/// the client is to fetch the nonce for the current account (if it has not been specified).
+///
+/// The account nonce is commonly stored in the `System` pallet's `Account` storage item. This
```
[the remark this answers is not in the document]
ascjones: > I wonder if it's actually possible to build a chain that stores the nonce elsewhere?

It would involve a custom `frame_system` pallet. It's also possible that this will change across versions of `frame_system`, I know it has in the past

> Is "nonce" here the same as the "Account index"?

Yes indeed it is used interchangeably. Personally I find index slightly misleading, and the field is called nonce, but I can change the assoc type name here for consistency?

examples/custom_config.rs
ascjones: Yes, it would be best accompanied with an actual runtime with the different type

Issue #399, opened 2022-01-18
Scale codec error: `Not enough data to fill buffer`


shekohex:
Hello, I started to face this issue whenever I try to read from the Storage in the latest Substrate `master`. 

Here is a simple example to reproduce the issue:

Ok, I've made the minimal Example, please follow these steps:

> Tip: Clone everything in `/tmp` if you don't want to keep them on your drive.

1. Clone the node locally
```
git clone --depth 1  https://github.com/webb-tools/dkg-substrate.git
```
then check out the commit where we updated Substrate:
````
git checkout 1192013487c9580dcfb1656b504f486bb29c8592
````

2. Build the node into `release` mode (should take less than 10min).
```
cargo build --release -p dkg-standalone-node
```

3. Then, Run the script that will run (3 nodes, for bootstrapping)
```
./scripts/run-standalone.sh
```
4. once it is working, you can clone the SDK and run the test
```
git clone --depth 1 https://github.com/webb-tools/webb-rs.git
```
5. Running the `read_chain_nonce` test
```
cargo test read_chain_nonce
```

You will get that error, similar to this:
```
Codec(Error { cause: None, desc: "Not enough data to fill buffer" })
```

**Meta**

```
Substrate Version: https://github.com/paritytech/substrate/commit/9eddaec3b8f15eec0c458e7bb3a1c16db91d8199
Subxt Version: https://github.com/paritytech/subxt/commit/d99cb9f4c3b93fdf6a082f60993556016dabc90c
```

shekohex:
@ascjones do you think this is related to this? https://github.com/paritytech/substrate/issues/8629

ascjones:
Hi @shekohex, just back from holiday I can take a look tomorrow at this

ascjones:
The return type for `chain_nonces` is `u64` https://github.com/webb-tools/webb-rs/blob/main/src/substrate/dkg_runtime.rs#L5768, where it should be `u32`https://github.com/webb-tools/dkg-substrate/blob/fb480df7f7fa12602e3dc3cc005e2dd96907bfbf/dkg-runtime-primitives/src/proposal.rs#L10.

Regenerating `dkg_runtime.rs` should fix this.

shekohex:
Thank you very much @ascjones !

PR #415, opened 2022-01-28
Log debug message for JSON-RPC response

ascjones:
While troubleshooting #399, I wanted to look at the raw bytes returned from the storage query before attempting to decode. Previously this was logged by `jsonrpsee`, but no longer.

This is useful since we can just ask users to enable `RUST_LOG=debug` for their client/test and we can then easily check the decoding of the raw bytes..

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -235,6 +235,7 @@ impl RpcClient {
             RpcClient::WebSocket(inner) => inner.request(method, params).await,
             RpcClient::Http(inner) => inner.request(method, params).await,
         };
+        log::debug!("response: {:?}", data);
         data
     }
 


niklasad1:
It should be logged by `jsonrpsee` but might broke when we migrated to the tracing crate instead no sure....

niklasad1 (approved):

ascjones:
> It should be logged by `jsonrpsee` but might broke when we migrated to the tracing crate instead no sure....

Oh I didn't know that `jsonrpsee` had been migrated to `tracing`, I've not used that - do I need to init a `tracing_subscriber` to see those messages?

Anyway maybe we can do a follow up which converts messages here to `tracing` if it makes sense.

dvdplm (approved):

PR #418, opened 2022-02-01
Add release checklist

jsdw:
This is essentially a clone of our Soketto release checklist with a couple of tiny tweaks, so anybody with familiarity releasing Soketto will be familiar with this process too.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -28,7 +28,7 @@ We also assume that ongoing work done is being merged directly to the `master` b
 
 4.  Bump the crate version in `Cargo.toml` to whatever was decided in step 2 for `subxt-codegen`, `subxt-macro`, `subxt` and `subxt-cli`.
 
-5.  Update `CHANGELOG.md` to reflect the difference between this release and last. If you're unsure of
+5.  Update `CHANGELOG.md` to reflect the difference between this release and the last. If you're unsure of
     what to add, check with the Tools team. See the `CHANGELOG.md` file for details of the format it follows.
 
     Any [closed PRs](https://github.com/paritytech/subxt/pulls?q=is%3Apr+is%3Aclosed) between the last release and
@@ -52,18 +52,18 @@ We also assume that ongoing work done is being merged directly to the `master` b
 
     ```
     (cd codegen && cargo publish) && \
-        sleep 5 && \
+        sleep 10 && \
         (cd macro && cargo hack publish --no-dev-deps --allow-dirty) && \
-        sleep 5 && \
+        sleep 10 && \
         cargo publish && \
-        sleep 5 && \
+        sleep 10 && \
         (cd cli && cargo publish);
     ```
 
     If you run into any issues regarding crates not being able to find suitable versions of other `subxt-*` crates,
     you may just need to wait a little longer and then run the remaining portion of that command.
 
-9.  If the release was successful, then tag the commit that we released in the `master` branch with the
+9.  If the release was successful, tag the commit that we released in the `master` branch with the
     version that we just released, for example:
 
     ```
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -51,13 +51,13 @@ We also assume that ongoing work done is being merged directly to the `master` b
     with what we've published).
 
     ```
-    (cd macro && cargo hack publish --no-dev-deps --allow-dirty --dry-run) && \
+    (cd codegen && cargo publish) && \
         sleep 5 && \
-        (cd codegen && cargo publish --dry-run) && \
+        (cd macro && cargo hack publish --no-dev-deps --allow-dirty) && \
         sleep 5 && \
-        cargo publishd --dry-run && \
+        cargo publish && \
         sleep 5 && \
-        (cd cli && cargo publish --dry-run);
+        (cd cli && cargo publish);
     ```
 
     If you run into any issues regarding crates not being able to find suitable versions of other `subxt-*` crates,
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -31,7 +31,7 @@ We also assume that ongoing work done is being merged directly to the `master` b
 5.  Update `CHANGELOG.md` to reflect the difference between this release and last. If you're unsure of
     what to add, check with the Tools team. See the `CHANGELOG.md` file for details of the format it follows.
 
-    Any [closed PRs](https://github.com/paritytech/soketto/pulls?q=is%3Apr+is%3Aclosed) between the last release and
+    Any [closed PRs](https://github.com/paritytech/subxt/pulls?q=is%3Apr+is%3Aclosed) between the last release and
     this release branch should be noted.
 
 6.  Commit any of the above changes to the release branch and open a PR in GitHub with a base of `master`.
@@ -60,6 +60,6 @@ We also assume that ongoing work done is being merged directly to the `master` b
     git push --tags
     ```
 
-    Once this is pushed, go along to [the releases page on GitHub](https://github.com/paritytech/soketto/releases)
+    Once this is pushed, go along to [the releases page on GitHub](https://github.com/paritytech/subxt/releases)
     and draft a new release which points to the tag you just pushed to `master` above. Copy the changelog comments
     for the current release into the release description.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,6 +1,6 @@
 # Release Checklist
 
-These steps assume that you've checked out the Soketto repository and are in the root directory of it.
+These steps assume that you've checked out the Subxt repository and are in the root directory of it.
 
 We also assume that ongoing work done is being merged directly to the `master` branch.
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,65 @@
+# Release Checklist
+
+These steps assume that you've checked out the Soketto repository and are in the root directory of it.
+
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -0,0 +1,65 @@
+# Release Checklist
+
+These steps assume that you've checked out the Soketto repository and are in the root directory of it.
```
dvdplm:
```suggestion
These steps assume that you've checked out the subxt repository and are in the root directory of it.
```

RELEASING.md
```diff
@@ -0,0 +1,65 @@
...
+    are still valid like so:
+
+    ```
+    cargo install cargo-deadlinks
+    cargo deadlinks --check-http -- --all-features
+    ```
+
+    If there are minor issues with the documentation, they can be fixed in the release branch.
+
+4.  Bump the crate version in `Cargo.toml` to whatever was decided in step 2.
```
dvdplm, line 29:
```suggestion
4.  Bump the crate version in `Cargo.toml` to whatever was decided in step 2 for `subxt-codegen`, `subxt-macro`, `subxt` and `subxt-cli`.
```

RELEASING.md
```diff
@@ -0,0 +1,65 @@
...
+8.  Now, we're ready to publish the release to crates.io.
+
+    Checkout `master`, ensuring we're looking at that latest merge (`git pull`).
+
+    Next, do a dry run to make sure that things seem sane:
+    ```
+    cargo publish --dry-run
+    ```
+
+    If we're happy with everything, proceed with the release:
```
dvdplm, line 50: It's a bit more annoying than this, it has to be in the specific order:`macro`, `codegen`, `subxt`, `subxt-cli`. We should have some bash for this (pending `cargo unleash` we can take the bash script from `jsonrpsee`).
Also note that `macro` has a dev dependency on `subxt` so to avoid complaints from cargo you have to `cargo hack publish --no-dev-deps --allow-dirty` for that one. Not sure how to best capture that here though.

RELEASING.md
```diff
@@ -0,0 +1,65 @@
...
+    ```
+    cargo publish
+    ```
+
+9.  If the release was successful, then tag the commit that we released in the `master` branch with the
+    version that we just released, for example:
+
+    ```
+    git tag v0.17.0 # use the version number you've just published to crates.io, not this one
```
dvdplm, line 59: I actually used `subxt-v0.16.0` for the last one. Should we stick to the mere number you reckon?

RELEASING.md
```diff
@@ -0,0 +1,65 @@
...
```
niklasad1, line 59: require signed tags ^^

```suggestion
    git tag -s v0.17.0 # use the version number you've just published to crates.io, not this one
```

RELEASING.md
jsdw: I figued that any number would do, but just went for 0.17 knowing it would be the next number we actually release!

RELEASING.md
jsdw: Ooh, I've never really thought about signing tags before; interesting! I wonder whether we should make that change in other release checklists too (this came from Soketto)?

RELEASING.md
jsdw: Damn, I didn't fix them in time!

RELEASING.md
jsdw: Ah interesting! I guess for now we can just write our the code you'll want to use, so I guess something like:

```
(cd macro && cargo hack publish --no-dev-deps --allow-dirty)
(cd codegen && cargo publish)
cargo publish
(cd cli && cargo publish)
```

(with a preface about `cargo install cargo-hack`ing)

RELEASING.md
jsdw: Oh, I just realised I completely misread your comment. I guess I'd normally stick to just a number in the tag (the subxt bit is implicit anyway I guess?). If only just to align with what we do in other repos!

RELEASING.md
jsdw: I added some detail for this step and a command to run to publish things in the correct order

RELEASING.md
jsdw: (fyi I had a look and all previous tags were vx.xx.x so I tweaked this one and the release to match)

niklasad1 (approved):
great, LGTM

RELEASING.md
niklasad1: I prefer `vx.xx.x` too :)

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+
+7.  Once the branch has been reviewed and passes CI, merge it.
+
+8.  Now, we're ready to publish the release to crates.io.
+
+    Checkout `master`, ensuring we're looking at that latest merge (`git pull`).
+
+    The crates in this repository need publishing in a specific order, since they depend on each other.
+    Additionally, `subxt-macro` has a circular dev dependency on `subxt`, so we use `cargo hack` to remove
+    dev dependencies (and `--allow-dorty` to ignore the git changes as a result) to publish it.
```
ascjones, line 47:
```suggestion
    dev dependencies (and `--allow-dirty` to ignore the git changes as a result) to publish it.
```

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+    So, first install `cargo hack` with `cargo install cargo hack`. Next, you can run something like the following
+    command to publish each crate in the required order (allowing a little time inbetween each to let `crates.io` catch up)
+    with what we've published).
+
+    ```
+    (cd macro && cargo hack publish --no-dev-deps --allow-dirty --dry-run) && \
+        sleep 5 && \
+        (cd codegen && cargo publish --dry-run) && \
+        sleep 5 && \
+        cargo publishd --dry-run && \
```
ascjones, line 58:
```suggestion
        cargo publish --dry-run && \
```

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+    dev dependencies (and `--allow-dorty` to ignore the git changes as a result) to publish it.
+
+    So, first install `cargo hack` with `cargo install cargo hack`. Next, you can run something like the following
+    command to publish each crate in the required order (allowing a little time inbetween each to let `crates.io` catch up)
+    with what we've published).
+
+    ```
+    (cd macro && cargo hack publish --no-dev-deps --allow-dirty --dry-run) && \
+        sleep 5 && \
+        (cd codegen && cargo publish --dry-run) && \
```
ascjones, line 56: I believe codegen should be first since it `macro` depends on it?

Also I think only the first `--dry-run` will succeed since subsequent crates will depend on that crate having been published.

RELEASING.md
jsdw: Ah whoops; I deliberately broke that and added `--dry-run` to test that the script bailed appropriately and forgot to undo it!

RELEASING.md
jsdw: During my testing, the whole thing seemed to run successfully with the `--dry-runs` and without the `publishd` break, but I'll check the order now that you mention this and swap those two around as needed!

jsdw: I checked, swapped the order and undid my breaking bits!

RELEASING.md
dvdplm: Yes, I know all the old tags were numeric only. I was hesitating a bit because for a moment I wanted to tag all the released crates, then decided against it.

Numbers only is better. I'll re-tag.

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+    ```
+    cargo install cargo-deadlinks
+    cargo deadlinks --check-http -- --all-features
+    ```
+
+    If there are minor issues with the documentation, they can be fixed in the release branch.
+
+4.  Bump the crate version in `Cargo.toml` to whatever was decided in step 2 for `subxt-codegen`, `subxt-macro`, `subxt` and `subxt-cli`.
+
+5.  Update `CHANGELOG.md` to reflect the difference between this release and last. If you're unsure of
```
dvdplm, line 31:
```suggestion
5.  Update `CHANGELOG.md` to reflect the difference between this release and the last. If you're unsure of
```

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+
+    So, first install `cargo hack` with `cargo install cargo hack`. Next, you can run something like the following
+    command to publish each crate in the required order (allowing a little time inbetween each to let `crates.io` catch up)
+    with what we've published).
+
+    ```
+    (cd codegen && cargo publish) && \
+        sleep 5 && \
+        (cd macro && cargo hack publish --no-dev-deps --allow-dirty) && \
+        sleep 5 && \
```
dvdplm, line 57: In my experience it sometimes takes more than 5sec, so maybe bump this to 10?

RELEASING.md
```diff
@@ -0,0 +1,76 @@
...
+        cargo publish && \
+        sleep 5 && \
+        (cd cli && cargo publish);
+    ```
+
+    If you run into any issues regarding crates not being able to find suitable versions of other `subxt-*` crates,
+    you may just need to wait a little longer and then run the remaining portion of that command.
+
+9.  If the release was successful, then tag the commit that we released in the `master` branch with the
```
dvdplm, line 66:
```suggestion
9.  If the release was successful, tag the commit that we released in the `master` branch with the
```

dvdplm (approved):

PR #421, opened 2022-02-01
[Do Not Merge] Deprecate 0.15 crates

jsdw:
Some small changes to the last tagged 0.15 release so that we can publish a new 0.15.1 release of the old `substrate-` prefixed crates with a deprecation notice pointing to the new ones. 

This should never be merged (but can be tagged 0.15.1 to keep tabs on this code change).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "client", "proc-macro"]
 
 [package]
 name = "substrate-subxt"
-version = "0.15.0"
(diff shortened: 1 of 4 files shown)


jsdw:
I ended up publishing v0.16.0 of the old substrate-prefixed crates as it was the only way to get it to show up by default on crates.io. The crates contain no code owing to dependency issues, but the version bump should ensure nobody is automatically updated to them.

Issue #413, opened 2022-01-27
Simplify and Streamify `EventSubscription`


jsdw:
I reckon we could improve/simplify the `EventSubscription` struct. Some thoughts:

- Impl `Stream` (and so the `StreamExt` methods), so that the filtering logic can be (mostly?) moved out of this struct.
- Remove filtering on events from a specific block (if needed, add a separate method to obtain events from a single block which potentially returns the same event stream as we'd get here; this method could be reused in the `transaction.rs` stuff too if it existed).
- Remove filtering of events from a specific extrinsic (we get this now via the API in `transaction.rs`, and ideally it would be easy to manually filter the Stream for such events).
- Remove filtering of events by pallet/event index; make it easy to do so via the Stream.
- Update `transfer_subscribe` example to subscribe to events separately from making any transaction, and show off filtering and such. If needed, make transactions etc in a separate thread just to produce the desired events. (why? we shouldn't use this to get events for the current transaction as we have a proper API for that; but that's what the example seems to do).
- Possibly add other examples.

paulormart:
Just read this issue and PR #423 and i'am fully supportive for the implementation of this feature. This is definitely a must have for this library :)

I'm currently using a [custom events subscription](https://github.com/turboflakes/subxt/blob/bfeefda78a7ebc49ebb55fef530c66e5b45e3443/src/custom_subscription.rs#L150) so that _block number_ and _authority index_ could be easily accessible [here](https://github.com/turboflakes/subxt/blob/develop/examples/custom_events_subscription.rs#L49)

jsdw:
@paulormart , @lamafab and anybody else interested, feel free to check out #442 and leave feedback :)

PR #423, opened 2022-02-01
Get event context on EventSubscription

lamafab:
Currently, the `next` method on the `EventSubscription` type only returns the event itself, not clarifying where it occurred. This PR introduces a new method, `next_context`, that additionally returns the corresponding block hash and the index of the event.

Btw, I think it would be great to have a block number, too, but I think that'd require upstream substrate changes (or additional queries on `subxt`'s side).

Feedback/criticism appreciated.

EDIT: Will wait for your feedback before I'll resolve the conflicts.

### DIFF
--- a/subxt/src/subscription.rs
+++ b/subxt/src/subscription.rs
@@ -39,7 +39,8 @@ use sp_core::{
 use sp_runtime::traits::Header;
 use std::collections::VecDeque;
 
-
+/// Raw bytes for an Event, including the block hash where it occurred and its
+/// corresponding event index.
 #[derive(Debug)]
 #[cfg_attr(test, derive(PartialEq, Clone))]
 pub struct EventContext<T> {
@@ -92,8 +93,8 @@ impl<'a, T: Config> BlockReader<'a, T> {
                         x.into_iter()
                             .flatten()
                             .enumerate()
-                            .map(|(idx, (phase, raw))| {
-                                (phase, idx, raw)
+                            .map(|(event_idx, (phase, raw))| {
+                                (phase, event_idx, raw)
                             })
                             .collect()
                     });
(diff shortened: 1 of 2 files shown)


jsdw:
I'm happy with the general idea (though I think `next_context` should probably return a struct with named fields at this point), although please note that I'd like to work on https://github.com/paritytech/subxt/issues/413 in the coming weeks, which will probably lead to a bit of a revamp of all of this code. I'm definitely in favour of exposing all of the available information we have about events and will make sure that that is the case!

If you're keen on exposing these details sooner then I'd be happy for this to progress and get merged in the meantime.

src/subscription.rs
```diff
@@ -39,14 +39,24 @@ use sp_core::{
 use sp_runtime::traits::Header;
 use std::collections::VecDeque;
 
+/// Raw bytes for an Event, including the block hash where it occurred and its
+/// corresponding event index.
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq, Clone))]
+pub struct EventContext<T> {
+    pub block: T,
```
jsdw, line 47: How about:

```suggestion
pub struct EventContext<Hash> {
    pub block_hash: Hash,
```

lamafab: Yep, done! Also rebased to master.

lamafab:
Had to rebase twice, missed your second push to master :)

jsdw:
Awesome; CI runs approved so we can see how it goes.

Given this PR and comments on #413 I'll start work on simplfying this `EventSubscription` API quite soon now, and I'll take into account this use case. So I'm happy to merge this when it's ready, but it may not be around for too long before the whole thing is overhauled :)

lamafab:
@jsdw I would appreciate it if you could get this merged. I'm okay if you'll eventually replace it :)

jsdw:
@lamafab if you could resolve the small cargo fmt errors (run `cargo +nightly fmt`) and clippy then we should be good to go on this :)

jsdw (approved):

ascjones (approved):
LGTM

Issue #406, opened 2022-01-25
Add more dynamic decoding support to `subxt`


jsdw:
Currently `subxt` is optimised towards the compile-time oriented codegen usage; we generate an interface and then use that interface to query compatible nodes. This has many advantages (you benefit from compile-time errors, IDE support, performance and more). The main downside is that the compile-time generated interface is unable to work across different, differing nodes/chains in a general way (parts of it may continue to work, and others will not).

Having the ability to dynamically query nodes across different chains can be very useful in some cases, for example exploratory tooling which wants to interrogate the available interfaces, or tooling akin to polkagot.js which needs to be able to provide a dynamic interface to interact with arbitrary substrate based chains.

These two directions (compiletime, static codegen and dynamic, runtime interaction) are not mutually exclusive either. They would both benefit from the same client (JSONRPC transport/API) code for instance. Another overlap is that `subxt` is used to provide a static interface to a single pallet for the `cargo-contracts` tool, which means being able to dynamically decode `Events` emitted from other pallets that it does not know about (rather than erroring because it's unable to statically decode them).

So, my proposal is that we work towards something like the following:

- Three user facing library crates (names just placeholders); `subxt-static` (which itself uses `subxt-macro` and `subxt-codegen`), `subxt-dynamic` and `subxt-client`. 
    - `subxt-static` is just `subxt` as it stands (with a little runtime decoding moved out of it). 
    - `subxt-dynamic` provides similar functionality to the above, but at runtime. It contains the logic needed to dynamically encode and decode things.
    - `subxt-client` splits out the JSONRPC client API as it currently stands, so that it can be shared between the above.
    - (I wonder whether we would want to share the metadata wrapping in eg `subxt-metadata`, too; perhaps not initially)
- These crates should interop nicely; we can move back and forth between them where it makes sense. As an example, when we fetch the SCALE encoded Event blob, the `cargo-contracts` use case requires that we can dynamically decode (and throw away) any events we don't care about until we get to the one we're looking for (with eg `find_event::<MyEvent>()`).

ascjones:
Related, I also have some similar code in my `cargo-contract` PR which is dynamically decoding events for display on the command line: https://github.com/paritytech/cargo-contract/blob/9bb545d77c13c3812f988635c90e5ce04acc95e1/src/cmd/extrinsics/transcode/decode.rs

jsdw:
[...]

PR #424, opened 2022-02-02
Move Subxt crate into a subfolder

jsdw:
This paves the way for #406, given that for that we'll want to break out a couple of other crates.

~~An unfortunate side effect is that examples end up in the `subxt` subfolder; I've updated the README to point people there, but probably they should be placed into a separate crate in the workspace so that we get a top level `examples` folder back. Happy to leave that for a next PR though.~~

Relates to https://github.com/paritytech/frame-metadata/issues/34


jsdw:
I moved the examples into their own workspace crate as I didn't like that they weren't so visible in the root.

ascjones (approved):

examples/README.md
```diff
@@ -0,0 +1,3 @@
+# Subxt Examples
+
+Take a look in the [examples](./examples) subfolder for various `subxt` usage examples.
```
dvdplm, line 3: Maybe this can go into the main `README`?

Cargo.toml
```diff
@@ -1,53 +1,10 @@
 [workspace]
```
dvdplm, line 1:
```suggestion
[workspace]
resolver = "2"
```

dvdplm (approved):

Cargo.toml
ascjones: This is the default for Rust 2021 https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html#details

jsdw: Did you find any docs on this? I can't find anything suggesting that this is a valid addition offhand, but it kindof makes sense? 

All of the subcrates are "edition = 2021", which implies "resolver = 2" by default. But since resolution happens globally, is this important too?

jsdw: Ah, found it! I must have skimmed the relevant part of the docs before :)

jsdw: (ah sorry @ascjones; I hadn't seen your reply)

PR #431, opened 2022-02-03
[WIP]: jsonrpsee proc macros

niklasad1:
let's see if it's possible to get rid of those annoying required type params.


subxt/src/config.rs
```diff
@@ -78,7 +78,14 @@ pub trait Config: 'static {
     type Signature: Verify + Encode + Send + Sync + 'static;
 
     /// Extrinsic type within blocks.
-    type Extrinsic: Parameter + Extrinsic + Debug + MaybeSerializeDeserialize;
+    // TODO(niklasad1): I have no idea if this ok or not ^^ => won't work otherwise
+    // because `jsonrpsee` requires `Deserialize + Send + Sync` for the return types.
```
jsdw, line 82: Andrew would know better, but afaiu this is pretty much going to be some wrapped `Vec<u8>` I'd expect, so I don't see any reason why Send + Sync would be an issue if required!

jsdw:
Just to note; it seems likely that the plan will be to abstract over Rpc clients more, see https://github.com/paritytech/subxt/issues/596. So perhaps we don't want to commit to using generated jsonrpsee types too much?

niklasad1:
yeah, let's close this PR

It works now thanks to Alex's traits bound overrides but doesn't make things better for subxt anyway.

PR #447, opened 2022-02-14
Reference key storage api

### LINKED ISSUE #411: Change storage key arguments (and call arguments?) to being references
Currently we pass these by value:

https://github.com/paritytech/subxt/commit/84fe1a0fd19b8cf767a1b6067af739bbe5589744

But we can see in an example of the generated code that we only need a reference ultimately:

```
pub async fn pending_swap(
    &self,
    _0: runtime_types::polkadot_parachain::primitives::Id,
    hash: ::core::option::Option<T::Hash>,
) -> ::core::result::Result<
    ::core::option::Option<
        runtime_types::polkadot_parachain::primitives::Id,
    >,
    ::subxt::BasicError,
> {
    let entry = PendingSwap(_0);
    self.client.storage().fetch(&entry, hash).await
}
```

lexnv:
A `StorageEntryType::Plain` type requires no changes as it receives no arguments.

### Changes

1. Structs receive a named lifetime and take a reference to what they previously contained.
If they contain multiple fields, all are bounded to the same lifetime.
With the exception of the `storage::Account`, which is an associated type for `DefaultAccountData` trait's implementation. 

```
pub struct PendingSwap(pub runtime_types::polkadot_parachain::primitives::Id)

pub struct PendingSwap<'a>(pub &'a runtime_types::polkadot_parachain::primitives::Id);
```

2. `StorageEntry` trait is implemented for an anonymous lifetime.
```
impl ::subxt::StorageEntry for PendingSwap {

impl ::subxt::StorageEntry for PendingSwap<'_> {
```

3. Storage API uses reference for key arguments
```
pub async fn pending_swap(
  &self,
  _0: runtime_types::polkadot_parachain::primitives::Id,
  hash: ::core::option::Option<T::Hash>,
)

pub async fn pending_swap(
  &self,
  _0: &runtime_types::polkadot_parachain::primitives::Id,
  hash: ::core::option::Option<T::Hash>,
)
```

4. Vector references are replaced by slices
```
_0: &::std::vec::Vec<::core::primitive::u8>

0: &[::core::primitive::u8]
``` 

5. Lastly, the `iter` methods bound the lifetime of the generated structs to the named lifetime of the `StorageClient`.

```
pub async fn pending_swap_iter(
    &self,
    hash: ::core::option::Option<T::Hash>,
) -> ::core::result::Result<
    ::subxt::KeyIter<'a, T, PendingSwap>,
    ::subxt::BasicError,
> {

pub async fn pending_swap_iter(
    &self,
    hash: ::core::option::Option<T::Hash>,
) -> ::core::result::Result<
    ::subxt::KeyIter<'a, T, PendingSwap<'a>>,
    ::subxt::BasicError,
> {
```

closes #411.

### Example Output

```
Staking active era: index: 1, start: Some(1645114260002)
  Alice account id:        d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)
  Alice//stash account id: be5ddb1579b72e84524fc29e78609e3caf42e85aa118ebfe0b0ad404b5bdd25f (5GNJqTPy...)
    account controlled by: d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)
Era reward points: EraRewardPoints { total: 20740, individual: {be5ddb1579b72e84524fc29e78609e3caf42e85aa118ebfe0b0ad404b5bdd25f (5GNJqTPy...): 20740} }
```
---

### Next Steps

- [x] Fix clippy due to `&Vec<_>`
- [x] Add storage example


cli/src/main.rs
```diff
@@ -167,6 +167,26 @@ fn codegen<I: Input>(
...
+        // the Free Software Foundation, either version 3 of the License, or
+        // (at your option) any later version.
+        //
+        // subxt is distributed in the hope that it will be useful,
+        // but WITHOUT ANY WARRANTY; without even the implied warranty of
+        // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+        // GNU General Public License for more details.
+        //
+        // You should have received a copy of the GNU General Public License
+        // along with subxt.  If not, see <http://www.gnu.org/licenses/>.
```
jsdw, line 186: Actually, I don't reckon we can/should claim any copyright over the codegen output (it is all theirs, and it's up to them to then do whatever they like with it I'd say).

jsdw (commented):
Looking good overall! I'm wondering whether we can avoid the inconsistency/hardcoded string w.r.t that Account entry though :)

jsdw:
So it looks like the purpose of `AccountData` is basically to generate the correct storage key, and then decode a `nonce` from the storage value that comes back when node storage is queried with that key (the nonce being the number of transactions from the account, and is needed to be provided with each transaction).

See https://github.com/paritytech/subxt/blob/cb9177aaa91158df1e99d6d136b9a6d2eb3a96ad/subxt/src/client.rs#L268.

By default this trait is implemented for some default enum which points at the `Account` storage key (see https://github.com/paritytech/subxt/blob/cb9177aaa91158df1e99d6d136b9a6d2eb3a96ad/codegen/src/api/mod.rs#L452). This makes it hard or impossible to use an `Account` struct that has a lifetime at the mo, which is why you currently are not adding a lifetime to a struct that looks like "Account".

In order to have a consistent storage API where everything provided is a reference (rather than having some exceptional thing that is not a reference), one approach I think would be to generate a version of the "system::Account" type with a reference to be consistent with the rest of the API, but to also generate a version of it without a reference. This non-reference version could then be used by that default `AccountData` impl instead :)

You'll need to look for the correct type to do this with sortof like you do now, but it's worth making sure that the entire path of the type is correct (ie it is `["system", "Account"]`, and not for instance ["balances", "Account"]) (I had to do the same thing here: https://github.com/paritytech/subxt/blob/cb9177aaa91158df1e99d6d136b9a6d2eb3a96ad/codegen/src/api/errors.rs#L80).

@ascjones does that sound reasonable to you?

codegen/src/api/storage.rs
```diff
@@ -135,13 +138,28 @@ fn generate_storage_entry_fns(
...
-                    (fields, entry_struct, constructor, key_impl)
+                    (fields, entry_struct, constructor, key_impl, true)
                 }
                 _ => {
+                    let (lifetime_param, lifetime_ref) = (quote!(<'a>), quote!(&'a));
+
                     let ty_path = type_gen.resolve_type_path(key.id(), &[]);
                     let fields = vec![(format_ident!("_0"), ty_path.clone())];
+                    // `::system::storage::Account` was utilized as associated type `StorageEntry`
+                    // for `::subxt::AccountData` implementation of generated `DefaultAccountData`.
```
jsdw, line 149:
```suggestion
                    // for `::subxt::AccountData` implementation by the generated `DefaultAccountData`.
```

codegen/src/api/storage.rs
```diff
@@ -178,25 +196,47 @@ fn generate_storage_entry_fns(
...
+    let storage_entry_impl = quote! (
+        const PALLET: &'static str = #pallet_name;
+        const STORAGE: &'static str = #storage_name;
+        type Value = #storage_entry_value_ty;
+        fn key(&self) -> ::subxt::StorageEntryKey {
+            #key_impl
+        }
+    );
+
+    let wrapper_entry_impl = if is_account_wrapper {
```
jsdw, line 214: Perhaps worth a comment just to tie it to the above so that it's clear that these two pieces work together?

codegen/src/api/storage.rs
```diff
@@ -74,12 +74,15 @@ fn generate_storage_entry_fns(
     storage_entry: &StorageEntryMetadata<PortableForm>,
 ) -> (TokenStream2, TokenStream2) {
     let entry_struct_ident = format_ident!("{}", storage_entry.name);
-    let (fields, entry_struct, constructor, key_impl) = match storage_entry.ty {
+    let is_account_wrapper = pallet.name == "System" && storage_entry.name == "Account";
+    let wrapper_struct_ident = format_ident!("{}DefaultData", storage_entry.name);
```
jsdw, line 78: nitpicking: I'm not super fond of the name `AccountDefaultData`; I'd lean towards something like `AccountOwned` I guess.. but I'm not too fussed :D

lexnv: `AccountOwned` sounds better :D Thanks

codegen/src/types/type_path.rs
```diff
@@ -76,6 +76,27 @@ impl TypePath {
...
     }
+
+    /// Gets the vector type parameter if the data is represented as `TypeDef::Sequence`.
+    ///
+    /// **Note:** Utilized for transforming `std::vec::Vec<T>` into slices `&[T]` for the storage API.
+    pub fn vec_type_param(&self) -> Option<TokenStream> {
+        match self {
+            TypePath::Type(ty) => {
+                match ty.ty.type_def() {
+                    TypeDef::Sequence(_) => {
```
jsdw, line 87: I think things like HashMap and HashSet and such also encode to a `TypeDef::Sequence`, so you'll probably also need to check that the `.path() == &["Vec"]` :)

codegen/src/types/type_path.rs
```diff
@@ -76,6 +76,27 @@ impl TypePath {
...
+                        if ty.params.is_empty() {
```
jsdw, line 88: Looking at the `scale-info` crate it looks like you could more safely do:

```
TypeDef::Sequence(seq) => {
    let type_param = seq.type_param()
    // ...
```

I'd also lean towards not bothering to `quote!` the type_param before returning it (I think it'll work just fine if you don't), but I'm not too bothered either way!

Also as a matter of personal taste I'd probably avoid the indentation with some early returns, a bit like:

```
let ty = match self {
    TypePath::Type(ty) => ty,
    _ => return None
};

let sequence = match ty.ty.type_def() {
    TypeDef::Sequence(seq) => seq,
    _ => return None;
};

// ...
```

examples/examples/staking_details.rs
```diff
@@ -0,0 +1,85 @@
...
+//
+// subxt is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with subxt.  If not, see <http://www.gnu.org/licenses/>.
+
+//! To run this example, a local polkadot node should be running. Example verified against polkadot 0.9.13-82616422d0-aarch64-macos.
```
jsdw, line 17: Question: Is this "Example verified against $version" correct for you? I'd either remove it or, better, use `polkadot --version` to put whatever version you're using in (if it's different) :)

jsdw: Or, looking at the below, perhaps the exampel is using a `substrate` instance rather than a `polkadot` one (which the other examples are using)? I'd prolly stick to `polkadot` just to be the same as the other examples (and consistent with the comments) :)

examples/examples/staking_details.rs
```diff
@@ -0,0 +1,85 @@
...
+        .expect("Could not obtain stash signer pair")
+        .public()
+        .into();
+    println!("  Alice//stash account id: {:?}", alice_stash_id);
+
+    // Map from all locked "stash" accounts to the controller account.
+    let controller_acc = api
+        .storage()
+        .staking()
+        .bonded(&alice_stash_id, None)
```
jsdw, line 72: Cool!

I'm wondering whether it is worth renaming this ewxample to `fetch_staking_details.rs` so that it lines up with the other `fetch` example (both are about fetching stuff out of storage)?

subxt/tests/integration/frame/staking.rs
```diff
@@ -248,12 +248,13 @@ async fn storage_current_era() -> Result<(), Error<DispatchError>> {
 
 #[async_std::test]
 async fn storage_era_reward_points() -> Result<(), Error<DispatchError>> {
-    let ctx = test_context().await;
-    let current_era_result = ctx
+    let cxt = test_context().await;
+    let eras = 0;
```
jsdw, line 252: I think it's just `era` :)

subxt/tests/integration/storage.rs
```diff
@@ -79,7 +79,7 @@ async fn storage_n_mapish_key_is_properly_created(
     };
 
     // This is what the generated code hashes a `session().key_owner(..)` key into:
-    let actual_key_bytes = KeyOwner(KeyTypeId([1, 2, 3, 4]), vec![5u8, 6, 7, 8])
+    let actual_key_bytes = KeyOwner(&KeyTypeId([1, 2, 3, 4]), &[5u8, 6, 7, 8])
```
jsdw, line 82: Nice!

lexnv: Thanks! Having the API as reference does improve the ergonomics IMO :D

codegen/src/types/type_path.rs
lexnv: That makes a lot of sense, and indeed the code looks a lot better 😄 

I did give the `type_param()` a try, but it seems to return a `UntrackedSymbol<TypeId> ` which does not implement the `ToToken` trait.
Looking below in the code generation, it seems under `TypeDef::Sequence` there is the assumption of always having at least one type_param. I've removed the `is_empty()` check here for consistency.
Wondering if we could run into problems along the line due to this assumption. 🤔

codegen/src/types/type_path.rs
```diff
@@ -76,6 +76,21 @@ impl TypePath {
...
+    pub fn vec_type_param(&self) -> Option<&TypePath> {
+        let ty = match self {
+            TypePath::Type(ty) => ty,
+            _ => return None,
+        };
+
+        match ty.ty.type_def() {
+            TypeDef::Sequence(_) => Some(&ty.params[0]),
+            _ => None,
+        }
```
jsdw, line 92: Looking neater; nice :)

examples/examples/fetch_staking_details.rs
```diff
@@ -0,0 +1,85 @@
...
+        era.index, era.start
+    );
+
+    let alice_id = AccountKeyring::Alice.to_account_id();
+    println!("  Alice account id:        {:?}", alice_id);
+
+    // Get Alice' Stash account ID
+    //
+    // *Note*: This is pre-funded, as declared in the [genesis state]
+    // (https://github.com/substrate-developer-hub/substrate-node-template/blob/main/node/src/chain_spec.rs#L49)
```
jsdw, line 61: Because it's polkadot and not substrate, I think the link would point to something more like:

https://github.com/paritytech/polkadot/blob/dc784f9b47e4681897cfd477b4f0760330875a87/node/service/src/chain_spec.rs#L1277

But in other examples we've taken for granted that some test accounts have funds in, so I wouldn't bother with a comment about it myself :)

jsdw (approved):
You'll need to merge in the latest stuff on master and then re-generate that polkadot.rs file, but otherwise lgtm!

examples/examples/fetch_staking_details.rs
lexnv: Thanks, i ve removed it for now :D

ascjones (approved):
This is a huge usability improvement!

While I'm not a huge fan of the speciaI cased codegen for the `Account` storage item, I can see that the `AccountOwned` workaround is necessary to make this work with the current storage API, and I can't come up with anything better.

However we should keep an open mind as to how we can improve this in the future to make the codegen less convoluted. 

One avenue to explore would be to modify `fetch` and `fetch_or_default` to accept preconstructed `StorageKey`s, and have the macro construct those keys inline instead of generating and calling `StorageEntry::key`. Then could modify the `AccountData` trait to have a method which constructs the `StorageKey` from the `AccountId`.

This will all need a bit more thought, so not in the scope of this PR.

PR #453, opened 2022-02-16
Obtain DispatchError::Module info dynamically

### LINKED ISSUE #443: Dynamic `DispatchError::Module` details
https://github.com/paritytech/subxt/pull/394 [removes](https://github.com/paritytech/subxt/pull/394/files#diff-97e25e2a0e41c578875856e97b659be2719a65227c104b992e3144efa000c35eL127) the lookup for error details from metadata at runtime, instead using the static metadata. 

This will only work if the configuration of the target runtime modules is identical to that of the compile time metadata. So for tools which target a single pallet that might be present across many versions of many chains (e.g. `cargo-contract`, or `xcm-cli`) then the error resolution might not work.

This was one of the original design goals for this library, and indeed the resolution of indices for constructing extrinsics and decoding events is still dynamic. So there should also be a similar mechanism for resolving runtime error details.

jsdw:
Static error info is troublesome if the node you interact with uses a different pallet index for the pallet you're working with. This PR moves error info back to being obtained via runtime metadata, but keeps the `details()` fn in codegen for now to provide a nicer API for obtaining it.

closes #443.

### DIFF
--- a/subxt/src/transaction.rs
+++ b/subxt/src/transaction.rs
@@ -27,6 +27,8 @@ use crate::{
     error::{
         BasicError,
         Error,
+        HasModuleError,
+        ModuleError,
         RuntimeError,
         TransactionError,
     },
@@ -54,7 +56,7 @@ use jsonrpsee::core::{
 /// returned from [`crate::SubmittableExtrinsic::sign_and_submit_then_watch()`].
 #[derive(Derivative)]
 #[derivative(Debug(bound = ""))]
-pub struct TransactionProgress<'client, T: Config, E: Decode, Evs: Decode> {
+pub struct TransactionProgress<'client, T: Config, E, Evs> {
     sub: Option<RpcSubscription<SubstrateTransactionStatus<T::Hash, T::Hash>>>,
     ext_hash: T::Hash,
     client: &'client Client<T>,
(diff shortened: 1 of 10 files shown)


ascjones (commented):
Using this API is a bit clunky, e.g.

```rust
            .wait_for_finalized_success()
            .await
            .map_err(|e| match e {
                subxt::Error::Runtime(err) => {
                    let inner = err.inner();
                    match inner.details(self.api.client.metadata()) {
                        Ok(Some(details)) =>
                            eyre::eyre!(
                                "Runtime: {} -> {}: {:?}",
                                details.pallet(),
                                details.error(),
                                details.description(),
                            ),
                        Ok(None) => eyre::eyre!("Runtime: {:?}", inner),
                        Err(e) => e.into(),
                    }
                }
                err => err.into(),
            })?;
```
Perhaps we could integrate it directly into [`wait_for_success`](https://github.com/paritytech/subxt/blob/b1b717332eb44af3f992ad4486b0d43a8d1f6d79/subxt/src/transaction.rs#L389), by defining a trait with `fn module_error_indices(&self) -> (u8, u8)` which is implemented by the macro where at the moment you define `fn details(&metadata: Metadata)`. In this context we already have the `self.client` so can easily grab a reference to the metadata. Then we can construct a tuple `(E, Option<ErrorMetadata>)`.

The slight wrinkle is what to do if the error cannot be resolved from the metadata, perhaps it would be better to roll that in to a `None` on the right hand side rather than having a nested `Option<Result<ErrorMetadata>>` or `Option<Result<ErrorMetadata>>`. Because the original error is still present on the left hand side, so can still match that case with `(RuntimeError::ModuleError(_), None)`

ascjones:
> Oooh interesting! I definitely agree that the API is clunky!

It wouldn't be a problem, except that you have to use the same pattern for every extrinsic call, so would have to use a similar helper method everywhere.

> we could extend the GenericError enum and add a specific Module variant that will contain the ErrorMetadata (or something resolved from it, depending on how easily lifetimes can be worked with) if the error returned can be resolved to that. What do you reckon?

Yes that would be very nice, we are usually very interested in what the module error actually is.

jsdw:
Ok, I've added a Module variant, and any `DispatchError::Module`'s are now converted into that if possible. It does have to allocate some strings and such, but this may be preferable to having the error type carry a lifetime around (though something to ponder perhaps). If we fail to find the error details from the metadata we emit a metadata error rather than emitting the original module error (but this seems fair; the runtime metadata _should_ contain the details).

ascjones (approved):
Lovely :heart:

subxt/src/transaction.rs
```diff
@@ -64,12 +66,11 @@ pub struct TransactionProgress<'client, T: Config, E: Decode, Evs: Decode> {
...
 // so we manually make it clear that Unpin is actually fine regardless of `T`
 // (we don't care if this moves around in memory while it's "pinned").
-impl<'client, T: Config, E: Decode, Evs: Decode> Unpin
-    for TransactionProgress<'client, T, E, Evs>
-{
-}
+impl<'client, T: Config, E, Evs> Unpin for TransactionProgress<'client, T, E, Evs> {}
 
-impl<'client, T: Config, E: Decode, Evs: Decode> TransactionProgress<'client, T, E, Evs> {
+impl<'client, T: Config, E: Decode + HasModuleError, Evs: Decode>
```
dvdplm, line 71: Maybe worth moving to a `where`-clause at this point?

dvdplm (approved):

PR #455, opened 2022-02-16
Add hardcoded override to ElectionScore

### LINKED ISSUE #454: the trait `Decode` is not implemented for `BoundedBTreeMap<ElectionScore, u32>`
This error crops up when attempting to build subxt against the latest substrate master.

It occurs because `BoundedBTreeMap<K,V>` is a generated struct that contains a `BTreeMap<K,V>`, and `Decode`  is only implemented on `BTreeMap<K,V>` if `K: Ord`.

In the erroneous case, we generate a `BoundedBTreeMap<ElectionScore, u32>`, and `ElectionScore` is another generated type which gets the default derives that we apply to all generated types (which doesn't include `Ord`).

Adding `Ord` (and thus also adding `PartialOrd, Eq, PartialEq`) to all types doesn't work out (loads of things can't impl this).

The best idea I have offhand is to iterate through the type registry looking for anything called "BTreeMap", and then adding the type ID of the first parameter it takes to some list. Later, when we generate all of our types, make sure that those IDs get the extra traits that they need to allow Decoding when they are keys in a `BTreeMap`.

jsdw:
This is intended as an interim solution until we devise a better approach (see #454 for a suggestion on that).

Hardcode a type seen in the metadata so that we can add the derives it needs to be a key in a BTreeMap.

closes #454

### DIFF
--- a/subxt/src/lib.rs
+++ b/subxt/src/lib.rs
@@ -243,7 +243,7 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
 /// [`ElectionScore`] overrides any generated instance of `sp_npos_elections::ElectionScore` found
 /// in the metadata, so that we can add some extra derives required for it to be used as the key
 /// in a [`std::collections::BTreeMap`].
-#[derive(Encode, Decode, Debug, PartialOrd, Ord, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


lexnv (approved):

ascjones (approved):
LGTM

PR #458, opened 2022-02-18
Distinct handling for N fields + 1 hasher vs N fields + N hashers

### LINKED ISSUE #446: Issue querying storage with tuple key
I'm on subxt release 0.17, generated metadata on the same version, my runtime is based on node-template 4.0.0
The storage is defined as:
```
#[pallet::storage]
#[pallet::getter(fn my_storage_map)]
pub(super) type MyStorageMap<T: Config> =
        StorageMap<_, Twox64Concat, (T::AccountId, T::Index), MyStruct<T>, OptionQuery>;
```
In the substrate unit tests, inserting and retrieving from the same tuple key there is no problems, but in subxt the same query returns None, even though the insert is successful when I check by iterating all storage entries.
```
let entry = api
        .storage()
        .my_pallet()
        .my_storage_map(acc_id, idx, None)
        .await?;
println!("{entry:?}"); // -> None
``` 

Anything else I can provide to help debug?

jsdw:
~~Tests still needed before this merges (https://github.com/paritytech/substrate/blob/f318350b6410ef47e438aae24c0b4779373d0a7b/frame/assets/src/lib.rs#L274 is example of StorageNMap item).~~

I added a few general storage tests, and a specific one that is fixed by this PR and broken before it.

Closes #446

### DIFF
--- a/codegen/src/api/storage.rs
+++ b/codegen/src/api/storage.rs
@@ -126,9 +126,9 @@ fn generate_storage_entry_fns(
                         quote!( #entry_struct_ident( #( #field_names ),* ) );
 
                     let key_impl = if hashers.len() == fields.len() {
-                        // If the number of hashers matches the number of fields, we're dealing with a
-                        // StorageNMap, and each field should be hashed separately according to the
-                        // corresponding hasher.
+                        // If the number of hashers matches the number of fields, we're dealing with
+                        // something shaped like a StorageNMap, and each field should be hashed separately
+                        // according to the corresponding hasher.
                         let keys = (0..tuple.fields().len())
                             .into_iter()
                             .zip(hashers)
--- a/codegen/src/api/storage.rs
+++ b/codegen/src/api/storage.rs
@@ -145,7 +145,7 @@ fn generate_storage_entry_fns(
                         // If there is one hasher, then however many fields we have, we want to hash a
                         // tuple of them using the one hasher we're told about. This corresponds to a
                         // StorageMap.
-                        let hasher = hashers.get(0).expect("checked for 1 hashed");
+                        let hasher = hashers.get(0).expect("checked for 1 hasher");
                         let items =
                             (0..tuple.fields().len()).into_iter().map(|field_idx| {
                                 let index = syn::Index::from(field_idx);
--- a/codegen/src/api/storage.rs
+++ b/codegen/src/api/storage.rs
@@ -115,26 +115,58 @@ fn generate_storage_entry_fns(
                             (field_name, field_type)
                         })
                         .collect::<Vec<_>>();
-                    // toddo: [AJ] use unzip here?
-                    let tuple_struct_fields =
-                        fields.iter().map(|(_, field_type)| field_type);
-                    let field_names = fields.iter().map(|(field_name, _)| field_name);
+
+                    let field_names = fields.iter().map(|(n,_)| n);
+                    let field_types = fields.iter().map(|(_,t)| t);
+
                     let entry_struct = quote! {
-                        pub struct #entry_struct_ident( #( pub #tuple_struct_fields ),* );
+                        pub struct #entry_struct_ident( #( pub #field_types ),* );
                     };
                     let constructor =
                         quote!( #entry_struct_ident( #( #field_names ),* ) );
-                    let keys = (0..tuple.fields().len()).into_iter().zip(hashers).map(
-                        |(field, hasher)| {
-                            let index = syn::Index::from(field);
-                            quote!( ::subxt::StorageMapKey::new(&self.#index, #hasher) )
-                        },
-                    );
-                    let key_impl = quote! {
-                        ::subxt::StorageEntryKey::Map(
-                            vec![ #( #keys ),* ]
+
+                    let key_impl = if hashers.len() == fields.len() {
+                        // If the number of hashers matches the number of fields, we're dealing with a
+                        // StorageNMap, and each field should be hashed separately according to the
+                        // corresponding hasher.
+                        let keys = (0..tuple.fields().len())
+                            .into_iter()
+                            .zip(hashers)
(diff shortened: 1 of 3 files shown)


codegen/src/api/storage.rs
```diff
@@ -115,26 +115,58 @@ fn generate_storage_entry_fns(
...
-                            quote!( ::subxt::StorageMapKey::new(&self.#index, #hasher) )
-                        },
-                    );
-                    let key_impl = quote! {
-                        ::subxt::StorageEntryKey::Map(
-                            vec![ #( #keys ),* ]
+
+                    let key_impl = if hashers.len() == fields.len() {
+                        // If the number of hashers matches the number of fields, we're dealing with a
+                        // StorageNMap, and each field should be hashed separately according to the
```
ascjones, line 130: Comment not strictly accurate, as it would be possible (although useless) to have a single tuple key e.g. `(KeyType,)` for a regular `StorageMap`

jsdw: I tweaked the comment to "something shaped like a StorageNMap"; does that make more sense?

I guess in the 1-key version it doesn't really matter what branch we follow as both have the same effect anyway.

subxt/tests/integration/storage.rs
```diff
@@ -0,0 +1,137 @@
...
+        .await?;
+
+    // Look up the nonce for the user (we expect it to be 1).
+    let entry = ctx.api.storage().system().account(alice, None).await?;
+    assert_eq!(entry.nonce, 1);
+
+    Ok(())
+}
+
+// This fails until the fix in https://github.com/paritytech/subxt/pull/458 is introduced.
```
ascjones, line 62: Can remove this self referential comment now?

ascjones (approved):
LGTM!

subxt/tests/integration/storage.rs
jsdw: I wanted it there more as a record for why the slightly random looking test exists for future onlookers :)

codegen/src/api/storage.rs
```diff
@@ -115,26 +115,57 @@ fn generate_storage_entry_fns(
...
-                    let key_impl = quote! {
-                        ::subxt::StorageEntryKey::Map(
-                            vec![ #( #keys ),* ]
+
+                    let key_impl = if hashers.len() == fields.len() {
+                        // If the number of hashers matches the number of fields, we're dealing with
+                        // something shaped like a StorageNMap, and each field should be hashed separately
+                        // according to the corresponding hasher.
+                        let keys = (0..tuple.fields().len())
+                            .into_iter()
```
maciejhirsz, line 133: Don't think you need this, range should be an iterator already

codegen/src/api/storage.rs
```diff
@@ -115,26 +115,57 @@ fn generate_storage_entry_fns(
...
-                    );
-                    let key_impl = quote! {
-                        ::subxt::StorageEntryKey::Map(
-                            vec![ #( #keys ),* ]
+
+                    let key_impl = if hashers.len() == fields.len() {
+                        // If the number of hashers matches the number of fields, we're dealing with
+                        // something shaped like a StorageNMap, and each field should be hashed separately
+                        // according to the corresponding hasher.
+                        let keys = (0..tuple.fields().len())
```
maciejhirsz, line 132: I'd have probably gone with `hashers.into_iter().enumerate().take(fields.len())`, but that's just how my brain works with iterators, this is fine.

I'd change `tuple.fields().len()` to `fields.len()` though.

codegen/src/api/storage.rs
```diff
@@ -115,26 +115,57 @@ fn generate_storage_entry_fns(
...
+                                vec![ #( #keys ),* ]
+                            )
+                        }
+                    } else if hashers.len() == 1 {
+                        // If there is one hasher, then however many fields we have, we want to hash a
+                        // tuple of them using the one hasher we're told about. This corresponds to a
+                        // StorageMap.
+                        let hasher = hashers.get(0).expect("checked for 1 hasher");
+                        let items =
+                            (0..tuple.fields().len()).into_iter().map(|field_idx| {
```
maciejhirsz, line 150: Ditto re `fields.len()` and `into_iterator()`.

maciejhirsz (approved):
Minor nits

subxt/tests/integration/storage.rs
dvdplm, line 62:
```suggestion
// See https://github.com/paritytech/subxt/pull/458 for the motivation.
```

…maybe?

subxt/tests/integration/storage.rs
```diff
@@ -0,0 +1,137 @@
...
+    assert_eq!(actual_key_bytes, expected_key_bytes);
+    Ok(())
+}
+
+#[async_std::test]
+async fn storage_n_map_storage_lookup() -> Result<(), subxt::Error<DispatchError>> {
+    let ctx = test_context().await;
+
+    // Boilerplate; we create a new asset class with ID 99, and then
+    // we "approveTransfer" of some of this asset class. This Gives us an
```
dvdplm, line 106:
```suggestion
    // we "approveTransfer" of some of this asset class. This gives us an
```

dvdplm (approved):
lgtm

subxt/tests/integration/storage.rs
jsdw: I'm easy either way :)

codegen/src/api/storage.rs
jsdw: Good point; I think I'd have gone for the same sort of thing if I was approaching it fresh :)

codegen/src/api/storage.rs
jsdw: I always forget this (because they probably shouldn't be :D)

Issue #315, opened 2021-11-09
Codegen output not implements codec::Decode


fzzr-:
Some `codegen` output not implements `codec::Decode`.

To reproduce there's two runtimes metadata:
[pontem.scale.zip](https://github.com/paritytech/subxt/files/7506851/pontem.scale.zip)

```rust
#[subxt::subxt(runtime_metadata_path = "nodes/pontem.scale")]
// or similar for pontem-xcmp.scale
pub mod pontem {}
```

Error: the trait `Decode` is not implemented for `BTreeMap<pontem::runtime_types::polkadot_parachain::primitives::Id, MessageQueueChain>`

<details> 
	<summary type="button">build log errors</summary>
	<pre><code>...
 error[E0277]: the trait bound `BTreeMap<pontem::runtime_types::polkadot_parachain::primitives::Id, MessageQueueChain>: Decode` is not satisfied
	--> src/main.rs:31:1
	 |
 31 | #[subxt::subxt(runtime_metadata_path = "nodes/pontem-xcmp.scale")]
	 | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	 | |
	 | the trait `Decode` is not implemented for `BTreeMap<pontem::runtime_types::polkadot_parachain::primitives::Id, MessageQueueChain>`
	 | in this procedural macro expansion
	 |
	::: ~/.cargo/git/checkouts/subxt-d8146022e5587204/78fd590/macro/src/lib.rs:39:1
	 |
 39 | pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
	 | ------------------------------------------------------------------ in this expansion of `#[subxt::subxt]`
	 |
	 = help: the following implementations were found:
				  <BTreeMap<K, V> as Decode>
 note: required by a bound in `subxt::StorageEntry::Value`
	--> ~/.cargo/git/checkouts/subxt-d8146022e5587204/78fd590/src/storage.rs:50:17
	 |
 50 |     type Value: Decode;
	 |                 ^^^^^^ required by this bound in `subxt::StorageEntry::Value`
 
 error[E0277]: the trait bound `BTreeMap<pontem::runtime_types::polkadot_parachain::primitives::Id, Vec<pontem::runtime_types::polkadot_core_primitives::InboundHrmpMessage<u32>>>: Decode` is not satisfied
	 --> src/main.rs:31:1
	  |
[...]

PR #459, opened 2022-02-21
Substitute BTreeMap/BTreeSet generated types for Vec

### LINKED ISSUE #456: Decode BTreeMap/HashMap etc as Vec<(K,V)>
A recent change in substrate meant that we ran into #454, which was worked around in #455 by adding a hardcoded type with custom derives.

As a more permanent fix, and a simpler solution than working out what custom traits to add to certain types, @ascjones suggested instead decoding types like `HashMap<K,V>` and `BTreeMap<K,V>` into `Vec<(K,V)>`.

The advantage of doing this is that we don't need any particular traits to be implemented for `K` and `V`, and additionally any custom ordering and such is preserved, so that we don't need to think about custom impls on the Substrate side for such things.

Once this is implemented, remove the hack in #455.

jsdw:
Recently, we hit an issue (#454) where a key type given to a BTreeMap did not impl Ord, leading to compile errors attempting to use the generated code.

This PR impls @ascjones's suggestion in #456 and makes types that would have been decoded into BTreeMap/BTreeSet instead decode to `Vec<(K,V)>` or `Vec<K>`. Since Vecs don't require their contents to impl any particular traits, this avoids such issues, while still preserving the decode order.

Closes #456 and #315

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -147,11 +147,13 @@ impl RuntimeGenerator {
                 "frame_support::traits::misc::WrapperKeepOpaque",
                 parse_quote!(::subxt::WrapperKeepOpaque),
             ),
-            // We override this because it's used as a key in a BTreeMap, and so we
-            // need to implement some extra derives for it for that to compile.
             (
-                "sp_npos_elections::ElectionScore",
-                parse_quote!(::subxt::ElectionScore),
+                "BTreeMap",
+                parse_quote!(::subxt::KeyedVec)
+            ),
+            (
+                "BTreeSet",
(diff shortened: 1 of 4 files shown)


ascjones (commented):
That looks easier than I thought! I think this is worth documenting somewhere, at least in the codegen module docs.

jsdw:
> That looks easier than I thought! I think this is worth documenting somewhere, at least in the codegen module docs.

Good point! And yup if I thought it would be this straightforward I would have done it first and avoided the previous hack!

ascjones (approved):

dvdplm (approved):
:taco:

PR #461, opened 2022-02-22
Filter one or multiple events by type from an EventSubscription

### LINKED ISSUE #457: Make it easier to filter events to those of a specific type or types
The new evevnt API gives us back a stream of chunks of events, which we can filter and such as we please. A common desire is to be able to filter this stream of events to look for a specific type of events (or perhaps a couple of types.

Currently, you might need to do something like:

```rust
let mut transfer_events = api
    .events()
    .subscribe()
    .await?
    // Ignore errors returning events:
    .filter_map(|events| future::ready(events.ok()))
    // Map events to just the one we care about:
    .flat_map(|events| {
        let transfer_events = events
            .find::<polkadot::balances::events::Transfer>()
            .collect::<Vec<_>>();
        stream::iter(transfer_events)
    });
```

This results in a stream of events of type `polkadot::balances::events::Transfer`. Aside from being a little verbose, it allocates when it collects these events into a Vec, and it discards errors. And more, if we want to search for more than one type of event, it gets more complex.

It'd be lovely if we instead had an API that could be used like:

```rust
let mut transfer_events = api
    .events()
    .subscribe()
    .await?
    .filter_event::<polkadot::balances::events::Transfer>()
```

This would return something like `Result<polkadot::balances::events::Transfer, BasicError>`, where an error is returned if we fail to decode an error we expect to be able to decode, or if there is a problem obtaining events for a block.

Using a little trait magic, we may also be able to support something like:

```rust
let mut transfer_events = api
    .events()
    .subscribe()
    .await?
    .filter_events::<(polkadot::balances::events::Transfer, polkadot::balances::events::SomeOtherEvent)>()
```

Which allows us to simultaneously search for and decode multiple events. The return type here would likely be something like  `Result<(Option<polkadot::balances::events::Transfer>, Option<polkadot::balances::events::SomeOtherEvent>), BasicError>`, and this pattern could be extended to tuples of N length. We'd then match on our event tuple to work with them like:

```rust
match event {
    (Some(transfer), None) => println!("Transfer succeeded"),
    (None, Some(other)) => println!("Another event found")
}
```

(I'd guess that `filter_event` would just delegate to `filter_events` and call it with a 1-tuple).

This would make it easier to look for specific events, and could also avoid allocating and properly forward errors, unlike the obvious user-space approach.

jsdw:
In this PR, I split `events.rs` into several files (it was getting rather large) and add a `filtered_events` method/file, which allows you to extract one or more events from an event subscription stream.

See the examples and unit tests for usage!

Closes #457

### DIFF
--- a/examples/examples/subscribe_some_events.rs
+++ b/examples/examples/subscribe_some_events.rs
@@ -79,16 +79,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Our subscription will see all of the balance events we're filtering on:
     while let Some(ev) = balance_events.next().await {
-        let ev = ev?;
+        let event_details = ev?;
 
-        if let (Some(withdraw), _, _) = &ev {
-            println!("Withdraw event: {withdraw:?}");
+        let block_hash = event_details.block_hash;
+        let event = event_details.event;
+        println!("Event at {:?}:", block_hash);
+
+        if let (Some(withdraw), _, _) = &event {
+            println!("  Withdraw event: {withdraw:?}");
         }
-        if let (_, Some(transfer), _) = &ev {
-            println!("Transfer event: {transfer:?}");
+        if let (_, Some(transfer), _) = &event {
+            println!("  Transfer event: {transfer:?}");
         }
-        if let (_, _, Some(deposit)) = &ev {
-            println!("Deposit event: {deposit:?}");
+        if let (_, _, Some(deposit)) = &event {
+            println!("  Deposit event: {deposit:?}");
         }
     }
 
--- a/examples/examples/subscribe_some_events.rs
+++ b/examples/examples/subscribe_some_events.rs
@@ -79,19 +79,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Our subscription will see all of the balance events we're filtering on:
     while let Some(ev) = balance_events.next().await {
-        match ev? {
-            (Some(withdraw), _, _) => {
-                println!("Withdraw event: {withdraw:?}");
-            }
-            (_, Some(transfer), _) => {
-                println!("Transfer event: {transfer:?}");
-            }
-            (_, _, Some(deposit)) => {
-                println!("Deposit event: {deposit:?}");
-            }
-            _ => {
-                unreachable!();
-            }
+        let ev = ev?;
+
+        if let (Some(withdraw), _, _) = &ev {
+            println!("Withdraw event: {withdraw:?}");
+        }
+        if let (_, Some(transfer), _) = &ev {
+            println!("Transfer event: {transfer:?}");
+        }
+        if let (_, _, Some(deposit)) = &ev {
+            println!("Deposit event: {deposit:?}");
         }
     }
 
--- a/examples/examples/subscribe_some_events.rs
+++ b/examples/examples/subscribe_some_events.rs
@@ -50,15 +50,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Subscribe to several balance related events. If we ask for more than one event,
     // we'll be given a correpsonding tuple of `Option`'s, with exactly one
     // variant populated each time.
-    let mut balance_events = api
-        .events()
-        .subscribe()
-        .await?
-        .filter_events::<(
-            polkadot::balances::events::Withdraw,
-            polkadot::balances::events::Transfer,
-            polkadot::balances::events::Deposit,
-        )>();
+    let mut balance_events = api.events().subscribe().await?.filter_events::<(
+        polkadot::balances::events::Withdraw,
+        polkadot::balances::events::Transfer,
+        polkadot::balances::events::Deposit,
+    )>();
 
     // While this subscription is active, we imagine some balance transfers are made somewhere else:
     async_std::task::spawn(async {
@@ -86,13 +82,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         match ev? {
             (Some(withdraw), _, _) => {
                 println!("Withdraw event: {withdraw:?}");
-            },
+            }
             (_, Some(transfer), _) => {
                 println!("Transfer event: {transfer:?}");
-            },
+            }
             (_, _, Some(deposit)) => {
                 println!("Deposit event: {deposit:?}");
-            },
+            }
             _ => {
                 unreachable!();
             }
--- a/examples/examples/subscribe_some_events.rs
+++ b/examples/examples/subscribe_some_events.rs
@@ -47,14 +47,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .await?
         .to_runtime_api::<polkadot::RuntimeApi<DefaultConfig, DefaultExtra<DefaultConfig>>>();
 
-    // Subscribe to just balance transfer events, making use of `filter_events`
-    // to select a single event type (note the 1-tuple) to filter out and return.
-    let mut transfer_events = api
+    // Subscribe to several balance related events. If we ask for more than one event,
+    // we'll be given a correpsonding tuple of `Option`'s, with exactly one
+    // variant populated each time.
+    let mut balance_events = api
         .events()
         .subscribe()
         .await?
         .filter_events::<(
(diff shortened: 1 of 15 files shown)


jsdw:
I decided to wrap all filtered events in a `FilteredEventDetails` struct, so that we have access to the block hash and phase of each event being filtered over, too.

subxt/src/events/event_subscription.rs
```diff
@@ -0,0 +1,177 @@
...
+            at: None,
+            _event_type: std::marker::PhantomData,
+        }
+    }
+
+    /// Return only specific events matching the tuple of 1 or more event
+    /// types that has been provided as the `Filter` type parameter.
+    pub fn filter_events<Filter: EventFilter>(self) -> FilterEvents<'a, Self, T, Filter> {
+        FilterEvents::new(self)
+    }
```
jsdw, line 111: Added.

subxt/src/events/filter_events.rs
```diff
@@ -0,0 +1,418 @@
...
+//
+// subxt is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with subxt.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Filtering individual events from subscriptions.
```
jsdw, line 17: This file is new.

subxt/src/events/events_type.rs
```diff
@@ -480,222 +388,15 @@ fn decode_raw_event_details<T: Config>(
...
-    InvalidCompactType(String),
-    /// Invalid bit sequence type; bit store type or bit order type used aren't supported.
-    #[error("Invalid bit sequence type; bit store type {0} is not supported")]
-    InvalidBitSequenceType(String),
-}
-
+/// Event related test utilities made use of outside this module.
 #[cfg(test)]
-mod tests {
+pub(crate) mod test_utils {
```
jsdw, line 393: I moved some common event test bits to this module so that I could use them in filter_events tests, too.

subxt/src/events/events_type.rs
```diff
@@ -370,7 +274,7 @@ impl<'a, T: Config, Evs: Decode> Events<'a, T, Evs> {
     ///
     /// **Note:** This method internally uses [`Events::iter_raw()`], so it is safe to
     /// use even if you do not statically know about all of the possible events.
-    pub fn find_first_event<Ev: Event>(&self) -> Result<Option<Ev>, BasicError> {
+    pub fn find_first<Ev: Event>(&self) -> Result<Option<Ev>, BasicError> {
```
jsdw, line 277: renamed to be consistent with removal of "event" prefix from similar functions previously.

subxt/src/events/events_type.rs
```diff
@@ -352,6 +213,49 @@ impl<'a, T: Config, Evs: Decode> Events<'a, T, Evs> {
...
+    /// decode them as we go, and returning the raw bytes and other associated
+    /// details. If an error occurs, all subsequent iterations return `None`.
+    ///
+    /// This method is safe to use even if you do not statically know about
+    /// all of the possible events; it splits events up using the metadata
+    /// obtained at runtime, which does.
+    ///
+    /// Unlike [`Events::iter_raw()`] this consumes `self`, which can be useful
+    /// if you need to store the iterator somewhere and avoid lifetime issues.
+    pub fn into_iter_raw(
```
jsdw, line 226: Added to support `filter_events`; we need to consume `self` so that we can work with it in our FilterEvents Stream impl.

jsdw:
Since I was very unhelpful in combining the addition of the `filter_events` function with breaking out the `events.rs` file into smaller modules, I added a few pointers to the changes made as part of adding the `filter_events` function, since there's a lot of green that is just the result of breaking the file up!

examples/examples/subscribe_some_events.rs
```diff
@@ -0,0 +1,100 @@
...
+        .to_runtime_api::<polkadot::RuntimeApi<DefaultConfig, DefaultExtra<DefaultConfig>>>();
+
+    // Subscribe to several balance related events. If we ask for more than one event,
+    // we'll be given a correpsonding tuple of `Option`'s, with exactly one
+    // variant populated each time.
+    let mut balance_events = api.events().subscribe().await?.filter_events::<(
+        polkadot::balances::events::Withdraw,
+        polkadot::balances::events::Transfer,
+        polkadot::balances::events::Deposit,
+    )>();
```
jsdw, line 57: Added this new example to show off the "filter multiple events" case.

subxt/src/events/events_type.rs
dvdplm: It threw me off at first when reading the examples, but in the end I think it's the right call.

subxt/src/events/events_type.rs
```diff
@@ -480,222 +388,15 @@ fn decode_raw_event_details<T: Config>(
...
-    InvalidCompactPrimitive(TypeDefPrimitive),
-    /// Invalid compact type; error details in string.
-    #[error("Invalid compact composite type {0}")]
-    InvalidCompactType(String),
-    /// Invalid bit sequence type; bit store type or bit order type used aren't supported.
-    #[error("Invalid bit sequence type; bit store type {0} is not supported")]
-    InvalidBitSequenceType(String),
-}
-
+/// Event related test utilities made use of outside this module.
```
dvdplm, line 391:
```suggestion
/// Event related test utilities used outside this module.
```

subxt/src/events/filter_events.rs
```diff
@@ -0,0 +1,418 @@
...
+    events: Option<
+        Box<
+            dyn Iterator<
+                    Item = Result<
+                        FilteredEventDetails<T::Hash, Filter::ReturnType>,
+                        BasicError,
+                    >,
+                > + 'a,
+        >,
+    >,
```
dvdplm, line 56: This made me giggle, you can really tell `cargo fmt` was cussing and sweating here.

subxt/src/events/filter_events.rs
```diff
@@ -0,0 +1,418 @@
...
+use std::{
+    marker::Unpin,
+    task::Poll,
+};
+
+/// A stream which filters events based on the `Filter` param provided.
+/// If `Filter` is a 1-tuple of a single `Event` type, it will return every
+/// instance of that event as it's found. If `filter` is ` tuple of multiple
+/// `Event` types, it will return a corresponding tuple of `Option`s, where
+/// exactly one of these will be `Some(event)` each iteration.
```
dvdplm, line 40: Just curious, do you know of other APIs that adopt this approach to filtering? It's a tricky thing to get right (i.e. be ergonomic).

subxt/src/events/filter_events.rs
```diff
@@ -0,0 +1,418 @@
...
+    pub block_hash: BlockHash,
+    /// A type containing an event that we've filtered on.
+    /// Depending on the filter type, this may be a tuple
+    /// or a single event.
+    pub event: Evs,
+}
+
+/// This trait is implemented for tuples of Event types; any such tuple (up to size 8) can be
+/// used to filter an event subscription to return only the specified events.
+pub trait EventFilter: private::Sealed {
```
dvdplm, line 122: Maybe move this type higher up in the file, given it's used extensively by the rest of the code?

dvdplm (approved):
Nice.

subxt/src/events/filter_events.rs
jsdw: I don't! I suppose an alternative would be to generate/use an enum with the right number of variants for each case, which, come to think of it, might not be such a bad idea..

subxt/src/events/filter_events.rs
jsdw: I had a look, and `FilterEvents`/`FilteredEventDetails` both feel like they deserve to be higher to me, since they are the "api" we expose here, and `EventFilter` is just a part of how it works (albeit an important one!).

lexnv (approved):

PR #473, opened 2022-03-08
Return events from blocks skipped over during Finalization, too

### LINKED ISSUE #468: Fix skipped finalized blocks under EventSubscription
Hi, 

With the latest Runtime/9170 enacted already on Kusama not all the finalized events are emitted, see PR https://github.com/paritytech/substrate/pull/10639#issuecomment-1014763139.

I wonder if a workaround could be implemented inside this library rather than on the client side?

Ideally `let mut event_sub = api.events().subscribe_finalized().await?;` should always guarantee that the _last finalized block_ is the next one in the iterator.

Currently i'm playing with something around these lines to always get the latest events for all the finalized blocks:

```
use futures::StreamExt;
use subxt::{
    ClientBuilder,
    DefaultConfig,
    DefaultExtra,
};

#[subxt::subxt(runtime_metadata_path = "examples/kusama_metadata.scale")]
pub mod polkadot {}

/// Subscribe to all events, and then manually look through them and
/// pluck out the events that we care about.
#[async_std::main]
    env_logger::init();

        .set_url("wss://kusama-rpc.polkadot.io:443")
        .build()
        .await?
        .to_runtime_api::<polkadot::RuntimeApi<DefaultConfig, DefaultExtra<DefaultConfig>>>();

    // Subscribe to any events that occur:
    let mut event_sub = api.events().subscribe_finalized().await?;


    let finalized_hash = api.client.rpc().finalized_head().await?;

    let mut last_finalized_block_number = match api.client.rpc().block(Some(finalized_hash)).await? {
        Some(signed_block) => Some(signed_block.block.header.number),
        None => Some(11650722)
    };
    
    // Our subscription will see the events emitted as a result of this:
    while let Some(events) = event_sub.next().await {

        let events = events?;
        let block_hash = events.block_hash();

        if let Some(signed_block) = api.client.rpc().block(Some(block_hash)).await? {
            
            let block_number_finalized = signed_block.block.header.number;

            // Verify if any finalized block has been skipped
            while let Some(i) = last_finalized_block_number {
                let next_block_number = i + 1;
                if next_block_number > block_number_finalized {
                    println!("Wait for the last finalized block");
                    last_finalized_block_number = None;
                } else if next_block_number == block_number_finalized {
                    println!("Do whatever you want with the current events for block {}", block_number_finalized);
                    last_finalized_block_number = Some(block_number_finalized);
                } else {
                    println!("Fetch the events for the skipped block {}", next_block_number);
                    last_finalized_block_number = Some(next_block_number);
                }
            }

            // Set latest block processed and wait for a new finalized block
            last_finalized_block_number = Some(block_number_finalized);

        }        
        
    }

    Ok(())
}
```

You can see the skipped blocks below:
```
Wait for the last finalized block
Do whatever you want with the current events for block 11662409
Wait for the last finalized block
Do whatever you want with the current events for block 11662410
Wait for the last finalized block
Fetch the events for the skipped block 11662411
Do whatever you want with the current events for block 11662412
Wait for the last finalized block
Do whatever you want with the current events for block 11662413
Wait for the last finalized block
Do whatever you want with the current events for block 11662414
Wait for the last finalized block
Do whatever you want with the current events for block 11662415
Wait for the last finalized block
Fetch the events for the skipped block 11662416
Do whatever you want with the current events for block 11662417
Wait for the last finalized block
```

jsdw:
If you subscribe to receive finalized block headers, blocks may be skipped over whenever a bunch of them are finalized at the same time. This PR makes sure to return events from all of the skipped-over blocks too.

Closes #468

### DIFF
--- a/subxt/src/events/event_subscription.rs
+++ b/subxt/src/events/event_subscription.rs
@@ -35,7 +35,6 @@ use futures::{
     StreamExt,
 };
 use jsonrpsee::core::client::Subscription;
-use num_traits::One;
 use sp_runtime::traits::Header;
 use std::{
     marker::Unpin,
@@ -85,7 +84,7 @@ pub async fn subscribe_finalized<'a, T: Config, Evs: Decode + 'static>(
         .rpc()
         .header(Some(last_finalized_block_hash))
         .await?
-        .map(|h| *h.number());
+        .map(|h| (*h.number()).into());
 
     // The complexity here is because the finalized block subscription may skip over
     // some blocks, so here we attempt to ensure that we pay attention to every block
@@ -103,13 +102,13 @@ pub async fn subscribe_finalized<'a, T: Config, Evs: Decode + 'static>(
                     Err(e) => return Either::Left(stream::once(async { Err(e.into()) })),
                 };
 
+                // We want all previous details up to, but not including this current block num.
+                let end_block_number = (*header.number()).into();
+
                 // This is one after the last block we returned details for last time.
                 let start_block_num = last_finalized_block_number
-                    .map(|n| n + One::one())
-                    .unwrap_or(*header.number());
-
-                // We want all previous details up to, but not including this current block num.
-                let end_block_number = *header.number();
+                    .map(|n| n + 1)
+                    .unwrap_or(end_block_number);
 
                 // Iterate over all of the previous blocks we need headers for, ignoring the current block
                 // (which we already have the header info for):
@@ -129,8 +128,8 @@ pub async fn subscribe_finalized<'a, T: Config, Evs: Decode + 'static>(
 /// Return a Stream of all block headers starting from `current_block_num` and ending just before `end_num`.
 fn get_block_headers<'a, T: Config>(
     client: &'a Client<T>,
-    mut current_block_num: T::BlockNumber,
-    end_num: T::BlockNumber,
+    mut current_block_num: u128,
+    end_num: u128,
 ) -> impl Stream<Item = Result<T::Header, BasicError>> + Unpin + Send + 'a {
     // Iterate over all of the previous blocks we need headers for. We go from (start_num..end_num).
     // If start_num == end_num, return nothing.
@@ -141,7 +140,7 @@ fn get_block_headers<'a, T: Config>(
             Some(current_block_num)
         };
 
-        current_block_num = current_block_num + One::one();
+        current_block_num = current_block_num + 1;
         res
     });
 
@@ -149,7 +148,7 @@ fn get_block_headers<'a, T: Config>(
     let block_headers = stream::iter(block_numbers)
         .then(move |n| {
             async move {
-                let hash = client.rpc().block_hash_internal(n).await?;
+                let hash = client.rpc().block_hash(Some(n.into())).await?;
                 let header = client.rpc().header(hash).await?;
                 Ok::<_, BasicError>(header)
             }
--- a/subxt/src/events/event_subscription.rs
+++ b/subxt/src/events/event_subscription.rs
@@ -87,6 +87,9 @@ pub async fn subscribe_finalized<'a, T: Config, Evs: Decode + 'static>(
         .await?
         .map(|h| *h.number());
 
+    // The complexity here is because the finalized block subscription may skip over
+    // some blocks, so here we attempt to ensure that we pay attention to every block
+    // that was skipped over as well as the latest block we were told about.
     let block_subscription =
         client
             .rpc()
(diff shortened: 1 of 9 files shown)


jsdw:
@paulormart I wondeer whether you could give this change a go; I haven't given it much of a workout yet but it'd be great to get some initial impression of whether it seems to be doing the right thing :)

subxt/src/rpc.rs
```diff
@@ -355,6 +355,23 @@ impl<T: Config> Rpc<T> {
...
+        block_number: T::BlockNumber,
+    ) -> Result<Option<T::Hash>, BasicError> {
+        let block_number = ListOrValue::Value(block_number);
+        let params = rpc_params![block_number];
+        let list_or_value = self.client.request("chain_getBlockHash", params).await?;
+        match list_or_value {
+            ListOrValue::Value(hash) => Ok(hash),
+            ListOrValue::List(_) => Err("Expected a Value, got a List".into()),
+        }
+    }
```
jsdw, line 373: I'm not a fan of duplicating this `block_hash` function. The problem is that `Config::BlockNumber` isn't something that we can easily work with.

I could also require that `Config::BlockNumber`implements something like `Into<u128>`, or I could remove the "other" copy of this function, but it sounds like some consideration went into it, so I was hesitant. What do you think @ascjones?

jsdw: (the nice thing abour requiring `Into<u128>` or similar would be that I can get rid of `num_traits` again, and the other required traits I added to `BlockNumber`, which I was just using to add a minimal amount of information about the BlockNumber so that I could iterate over them.)

ascjones: I think it is reasonable to use `Into<u128>`

paulormart:
Awesome stuff @jsdw. 

I have just run my initial test against this branch and there is no skipped blocks. Seems to work fine, thank you.

```
Received 10 events from finalized block 11731282
Received 5 events from finalized block 11731283
Received 25 events from finalized block 11731284
Received 27 events from finalized block 11731285
Received 4 events from finalized block 11731286
Received 18 events from finalized block 11731287
Received 37 events from finalized block 11731288
Received 34 events from finalized block 11731289
Received 9 events from finalized block 11731290
Received 36 events from finalized block 11731291
Received 28 events from finalized block 11731292
Received 17 events from finalized block 11731293
Received 27 events from finalized block 11731294
Received 26 events from finalized block 11731295
Received 10 events from finalized block 11731296
Received 6 events from finalized block 11731297
Received 25 events from finalized block 11731298
Received 19 events from finalized block 11731299
Received 7 events from finalized block 11731300
Received 4 events from finalized block 11731301
Received 28 events from finalized block 11731302
Received 28 events from finalized block 11731303
Received 25 events from finalized block 11731304
Received 22 events from finalized block 11731305
Received 15 events from finalized block 11731306
Received 38 events from finalized block 11731307
Received 28 events from finalized block 11731308
Received 2 events from finalized block 11731309
Received 41 events from finalized block 11731310
Received 30 events from finalized block 11731311
Received 6 events from finalized block 11731312
Received 31 events from finalized block 11731313
```

jsdw:
Fantastic; thank you for doing that!

subxt/src/rpc.rs
jsdw: I went for `Into<u64>` in the end because u128's don't serialize to JSON so easily; I'm hoping `u64` is also a safe target here (I assume it's basically a choice between `u32` and `u64` that people would tend to make?)

dvdplm: `Into<u64>` should also be fine.

subxt/src/events/event_subscription.rs
```diff
@@ -51,36 +57,111 @@ pub use super::{
...
+/// Take a subscription that returns block headers, and if any block numbers are missed out
+/// betweem the block number provided and what's returned from the subscription, we fill in
+/// the gaps and get hold of all intermediate block headers.
+///
+/// **Note:** This is exposed so that we can run integration tests on it, but otherwise
+/// should not be used directly and may break between minor releases.
+#[doc(hidden)]
+pub fn subscribe_to_block_headers_filling_in_gaps<'a, S, E, T: Config>(
+    client: &'a Client<T>,
+    mut last_block_num: Option<u64>,
```
dvdplm, line 107: Is it ok to assume that the block number is `u64` here? Or would `Option<Into<u64>>` work?

dvdplm (approved):

subxt/src/events/event_subscription.rs
jsdw: since block numbers can always be converted into `u64`s now, it's just easier to do the conversion first and then pass in the u64's here (this function isn't "public" anyway).

ascjones (approved):
Brilliant

PR #476, opened 2022-03-14
Use RPC call to get account nonce

### LINKED ISSUE #474: Use of `system_accountNextIndex` as nonce
Hi guys,

I'm wondering, is there any reason you are using the latest nonce instead of the rpc call `system_accountNextIndex`on the `create_signed` function?

You are currently using;

```rust
    pub async fn fetch_nonce<A: AccountData>(
        &self,
        account: &T::AccountId,
    ) -> Result<T::Index, BasicError>
    where
        <A as AccountData>::AccountId: From<<T as Config>::AccountId>,
        <A as AccountData>::Index: Into<<T as Config>::Index>,
    {
        let account_storage_entry = A::storage_entry(account.clone().into());
        let account_data = self
            .storage()
            .fetch_or_default(&account_storage_entry, None)
            .await?;
        Ok(A::nonce(&account_data).into())
    }
```

Something like this seems appropriate as it allows submitting multiple extrinsic for the same account id on the same block more easily, and seems more aligned with the polkadot-js ecosystem;

```rust
    /// Fetch the next nonce for the given account id.
    pub async fn fetch_next_nonce<A: AccountData>(
        &self,
        account: &T::AccountId,
    ) -> Result<T::Index, BasicError>
    where
        <T as Config>::AccountId: MaybeSerializeDeserialize,
        <T as Config>::Index: MaybeSerializeDeserialize,
    {
        Ok(self
            .rpc()
            .client
            .request("system_accountNextIndex", rpc_params![account])
            .await?)
    }
```

Thanks!

jsdw:
This great suggestion from #474 allows us to remove a bunch of code around getting the account nonce from storage.

Closes #474


ascjones (approved):
Love it. :tada:

lexnv (approved):

lemarier:
Thanks! 🚀

PR #490, opened 2022-03-22
Simplify creating and signing extrinsics

jsdw:
- Reduce our reliance on substrate traits and move the transaction encoding logic into one place.
- Simplify the process of defining the additional and extrinsic parameters that are needed for your node, and allow them to be provided to the main entry point methods (`sign_and_submit`/`sign_and_submit_then_watch`), adding `_default` versions of these methods to handle the current case of falling back to default params.
- In turn, make it possible to provide a tip and mortality for Polkadot/Substrate like transactions at runtime, and add an example (`balance_transfer_with_params`) to do this.
- update examples/SCALE to be against Polkadot 0.9.18 to make sure they are working against latest Polkadot.

### DIFF
--- a/subxt/src/extrinsic/params.rs
+++ b/subxt/src/extrinsic/params.rs
@@ -32,7 +32,7 @@ pub use sp_runtime::generic::Era;
 /// see [`BaseExtrinsicParams`] for an implementation that is compatible with
 /// a Polkadot node.
 pub trait ExtrinsicParams<T: Config> {
-    /// Thexe parameters can be provided to the constructor along with
+    /// These parameters can be provided to the constructor along with
     /// some default parameters that `subxt` understands, in order to
     /// help construct your [`ExtrinsicParams`] object.
     type OtherParams;


subxt/src/client.rs
```diff
@@ -236,26 +267,22 @@ where
...
-        additional_params: X::Parameters,
-    ) -> Result<UncheckedExtrinsic<T, X>, BasicError>
-    where
-        <<X as SignedExtra<T>>::Extra as SignedExtension>::AdditionalSigned:
-            Send + Sync + 'static,
-    {
+        signer: &(dyn Signer<T> + Send + Sync),
+        other_params: X::OtherParams,
+    ) -> Result<Encoded, BasicError> {
+        // 1. get nonce
```
lexnv, line 285: Nice! This does make more sense when following the code logic!

subxt/src/extrinsic/extra.rs
```diff
@@ -1,481 +0,0 @@
...
-        _len: usize,
-    ) -> Result<Self::Pre, TransactionValidityError> {
-        Ok(())
-    }
-}
-
-/// A default `SignedExtra` configuration, with [`ChargeTransactionPayment`] for tipping.
-///
-/// Note that this must match the `SignedExtra` type in the target runtime's extrinsic definition.
-pub type DefaultExtra<T> = DefaultExtraWithTxPayment<T, ChargeTransactionPayment<T>>;
```
jsdw, line 481: I am wondering how this single `DefaultExtra` seems to work for both polkadot (in the examples) and substrate (in the tests), because the former just expects a `Compact<u128>` for the tip section in the extra params, and the latter expects a `(Compact<u128>, Option<u32>)`, which is what this provides via `ChargeTransactionPayment`. Am I missing something or did we get lucky somehow with that extra `0` byte for the `Option<u32> = None` being put into the transaction extras in our examples?

ascjones: `ChargeTransactionPayment` comes last, so I guess that the node will successfully decode the `Compact<u128>` and ignore the subsequent data. This data isn't included in `AdditionalSigned` so won't affect the signature.

jsdw: I'd have thought that decoding the call data which comes next would be naffed up in that case! It's a mystery to me at the moment; I'd have to experiment a bit!

jsdw: (Also, the "extra" data is a part of the payload that is signed, so the signature would be different too?)

ascjones: Not sure what's going on here then...

examples/examples/balance_transfer_with_params.rs
```diff
@@ -0,0 +1,69 @@
...
+    let tx_params = Params::new()
+        .tip(PlainTip::new(20_000_000_000))
+        .era(Era::Immortal, *api.client.genesis());
+
+    // Send the transaction:
+    let hash = api
+        .tx()
+        .balances()
+        .transfer(dest, 123_456_789_012_345)
+        .sign_and_submit(&signer, tx_params)
```
ascjones, line 63: Not sure whether this is possible/desirable, but did you consider adding a builder style method onto `SubmittableExtrinsic` which allows setting the params, and would use the default if not specified e.g.

```
let hash = api
        .tx()
        .balances()
        .transfer(dest, 123_456_789_012_345)
        .params(tx_params)
        .sign_and_submit(&signer)
```
It would avoid the requirement of the `_default` method equivalents, on the other hand would be less explicit if adding params is the common case.

jsdw: Ooh interesting! I didn't consider that offhand, no.

One thing I like about having the `_default` and the param-taking methods is that it's up to the user whether the params actually do implement `Default` at all or not; the `Default` constraint is only on those `_default` methods now, so if for some reason somebody could not or did not want a default set of params for their chain, that'd work just fine.

jsdw: I say that but actually I think I left a Default clause somewhere I shouldn't have, so this is a good pointer to remove that!

lexnv (approved):

subxt/src/client.rs
```diff
@@ -264,21 +294,63 @@ where
...
+            // from address for signature
+            signer.address().encode_to(&mut encoded_inner);
+            // the signature bytes
+            signature.encode_to(&mut encoded_inner);
+            // attach custom extra params
+            additional_and_extra_params.encode_extra_to(&mut encoded_inner);
+            // and now, call data
+            call_data.encode_to(&mut encoded_inner);
+            // now, prefix byte length:
+            let len = Compact(encoded_inner.len() as u64);
```
ascjones, line 345: Needs to be `u32` as per https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/generic/unchecked_extrinsic.rs#L244. So something like:

```suggestion
           let len = Compact(u32::try_from(encoded_inner.len())?);
```

Though you probably will need to `map_err`

subxt/src/extrinsic/params.rs
```diff
@@ -0,0 +1,237 @@
...
+
+// We require Era as a param below, so make it available from here.
+pub use sp_runtime::generic::Era;
+
+/// This trait allows you to configure the "signed extra" and
+/// "additional" parameters that are signed and used in transactions.
+/// see [`BaseExtrinsicParams`] for an implementation that is compatible with
+/// a Polkadot node.
+pub trait ExtrinsicParams<T: Config> {
+    /// Thexe parameters can be provided to the constructor along with
```
ascjones, line 35:
```suggestion
    /// These parameters can be provided to the constructor along with
```

subxt/src/client.rs
jsdw: Being compact encoded it doesn't really matter what the original type is, does it (u8, u16, u32, u64 etc) since they all encode to the same compact bit representation anyway?

jsdw: (but a length that didn't actually fit into a u32 wouldn't decode into a `Compact<u32>` of course)

subxt/src/extrinsic/params.rs
```diff
@@ -0,0 +1,237 @@
...
+        tx_version: u32,
+        nonce: T::Index,
+        genesis_hash: T::Hash,
+        other_params: Self::OtherParams,
+    ) -> Self;
+
+    /// This is expected to SCALE encode the "signed extra" parameters
+    /// to some buffer that has been provided. These are the parameters
+    /// which are sent along with the transaction, as well as taken into
+    /// account when signing the transaction.
```
ascjones, line 52: These are not taken into account when signing a transactions, the encoded extra data is part of the transaction payload. This topic is very confusing so it is important to be clear here.

subxt/src/client.rs
```diff
@@ -264,21 +294,63 @@ where
...
+            additional_and_extra_params.encode_additional_to(&mut bytes);
+            if bytes.len() > 256 {
+                signer.sign(&sp_core::blake2_256(&bytes))
+            } else {
+                signer.sign(&bytes)
+            }
+        };
+
+        // 5. Encode extrinsic, now that we have the parts we need.
+        // (this may not be 100% correct but should be close enough for this example)
```
ascjones, line 331: I assume this comment is a copy/paste error, because it should be 100% correct. 

We should replace it with an explanation that this is a compatible copy of the `Encode` implementation for `UncheckedExtrinsic` for version 4.

subxt/src/client.rs
```diff
@@ -264,21 +294,63 @@ where
...
-            signer,
-            additional_params,
-        )
-        .await?;
-        Ok(signed)
+            self.client.genesis_hash,
+            other_params,
+        );
+
+        // 4. construct signature
```
ascjones, line 317: Please provide explanation and maybe a link that this is a copy of the `Encode` impl for `SignedPayload`

subxt/src/client.rs
jsdw: Ah, indeed, it was copied from my original example :) I'll fix up the comments!

subxt/src/extrinsic/params.rs
jsdw: Perhaps I wasn't clear; what I meant here was that these are part of the payload of bytes that is signed (call, extra, additional) as well as being sent along with the transaction (whereas the additional params belowaren't sent but are equally taken into account, ie included i nthe signed payload)

ascjones: You are right, I misread.

ascjones (approved):
LGTM, a big improvement.

A next step refactoring might be to extract the extrinsic construction in `create_signed` out of `client`, which would also allow for "offline" extrinsic construction and also easier unit testing of this critical part of code.

PR #492, opened 2022-03-28
Handle `sp_runtime::ModuleError` substrate updates

### LINKED ISSUE #491: Subxt integration tests failed against latest Substrate build.

lexnv:
The following change was introduced in substrate via [Allow pallet error enum variants to contain fields #10242](https://github.com/paritytech/substrate/pull/10242).

```
DispatchError::Module ( sp_runtime::ModuleError { index: u8, error: u8 } 
```
->
```
DispatchError::Module ( sp_runtime::ModuleError { index: u8, error: [u8; 4] } )
```

This PR handles the error of type `[u8; 4]` while maintaining backward compatibility.

Closes #491 

### Testing Done

- new substrate binary

```
fn module_error_indices(&self) -> Option<::subxt::ModuleErrorRaw> {
    if let Self::Module(module_error) = self {
        Some(::subxt::ModuleErrorRaw {
            pallet_index: module_error.index,
            error: module_error.error,
        })
    } else {
        None
    }
}
```

- legacy format

```
fn module_error_indices(&self) -> Option<::subxt::ModuleErrorRaw> {
    if let Self::Module(module_error) = self {
        Some(::subxt::ModuleErrorRaw {
            pallet_index: module_error.index,
            error: [module_error.error, 0, 0, 0],
        })
    } else {
        None
    }
}
```

- named format

```
fn module_error_indices(&self) -> Option<::subxt::ModuleErrorRaw> {
    if let &Self::Module { index, error } = self {
        Some(::subxt::ModuleErrorRaw {
            pallet_index: index,
            error: [error, 0, 0, 0],
        })
    } else {
        None
    }
}
```


codegen/src/api/errors.rs
```diff
@@ -74,9 +106,16 @@ pub fn generate_has_module_error_impl(
             }
         }
     } else {
+        let error_conversion = if module_legacy_err {
+            quote! { module_error.error }
+        } else {
+            // Convert [u8; 4] errors to legacy format.
```
lexnv, line 112: Error number is obtained as per: https://github.com/paritytech/substrate/pull/10242/files#diff-a8382c7a11352cf83fe1e60f06e0a9059a00443155b0a8e3189de37bc9ac4fedR1586.
Wondering if we could do more here. What do you think?

codegen/src/api/errors.rs
```diff
@@ -38,31 +38,63 @@ pub fn generate_has_module_error_impl(
...
-            .unwrap_or_else(|| {
-                abort_call_site!(
-                    "DispatchError::Module expected to contain 1 or more fields"
-                )
-            })
-            .name()
-            .is_some();
-        are_fields_named
+
+        let module_field = module_variant.fields().get(0).unwrap_or_else(|| {
```
jsdw, line 66: I find it a little hard following the two-booleans and nested if chains; how about defining a small enum like

```rust
enum ModuleErrorType {
    Case1,
    Case2,
    Case3,
}
```

And then perhaps tweaking the if statements to get rid of some nesting. Perhaps we can move this check to a different function for instance so that early returns are easy? Something like:

```rust
fn module_error_type(...) -> ModuleErrorType {    
    // if the fields are named, this is case 1; the oldest.
    if !module_field.name().is_none() {
        return ModuleErrorType::Case1
    }

    // get the error struct
    let module_err = metadata
        .types
        .resolve(module_field.ty().id())
        .unwrap_or_else(|| {
            abort_call_site!("sp_runtime::ModuleError type expected in metadata")
        });

    let error_type_def = match module_err.type_def() {
        TypeDef::Composite(composite) => composite,
        _ => abort_call_site!("sp_runtime::ModuleError type should be a composite type")
    };

    let error_field_in_composite = error_type_def
        .fields()
        .iter()
        .find(|field| field.name() == Some(&"error".to_string()))
        .unwrap_or_else(|| {
            abort_call_site!(
                "sp_runtime::ModuleError expected to contain error field"
            )
        });

    // if error type is u8, case2, else assume it's [u8; 4]
    if error_field.type_name() == Some(&"u8".to_string()) {
        ModuleErrorType::Case2
    } else {
        ModuleErrorType::Case3
    }
}
```

(Code completely untested). What do you reckon?

codegen/src/api/errors.rs
jsdw: I'm not sure what the other 3 bytes in that new `[u8; 4]` will do yet; it's prolly worth trying to find out (perhaps a question in dumb questions to see who answers?). But offhand, I'm sortof happy to do either: 

- drop back to the legacy single-u8 and accept that information will be lost (whatever it is), or 
- always end up with a `[u8; 4]` so that we preserve whatever additional info is there (I get the impression that setting the other 3 bytes to 0 when porting from the legacy formats will be ok).

@ascjones any thoughts?

codegen/src/api/errors.rs
lexnv: Indeed, that makes more sense to have a different function. Especially for supporting multiple versions, better change it now and simplify maintainability going forward :D Thanks for feedback

codegen/src/api/errors.rs
ascjones: I think while we are doing this, we should at the minimum expose the raw bytes so the user can extract any information that might be in there now or in the future. 

One way would be to modify [`HasModuleError`:](https://github.com/paritytech/subxt/blob/e866d744de5e8c4d43b5673701dd6b5232a1b7d1/subxt/src/error.rs#L190) to give access to the `[u8; 4]` possibly by changing the return type of `(u8, u8)` to include that (maybe promote to a struct).

Better than throwing that away, and once usage patterns emerge of that `[u8; 4]` we can add some structure to it.

codegen/src/api/errors.rs
```diff
@@ -18,7 +18,104 @@ use frame_metadata::v14::RuntimeMetadataV14;
...
+    // Get the error field from the `sp_runtime::ModuleError` structure.
+    let error_field = error_type_def
+        .fields()
+        .iter()
+        .find(|field| field.name() == Some(&"error".to_string()))
+        .unwrap_or_else(|| {
+            abort_call_site!("sp_runtime::ModuleError expected to contain error field")
+        });
+
+    if error_field.type_name() == Some(&"u8".to_string()) {
```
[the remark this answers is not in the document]
lexnv: That makes sense, added to the latest revision. Thanks!

codegen/src/api/errors.rs
```diff
@@ -18,7 +18,104 @@ use frame_metadata::v14::RuntimeMetadataV14;
...
+        ModuleErrorType::LegacyError
+    } else {
```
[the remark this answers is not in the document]
lexnv: Added to latest, thanks!

codegen/src/api/errors.rs
lexnv: That makes sense, have updated it with a `ModuleErrorRaw` that is exposed to the customer on the error path, as part of the `ModuleError`. 
We still do the conversion to error string assuming the error index is indeed the first byte. We could extend it further once the XCM is introduced.
What do you think? 😄

subxt/src/error.rs
```diff
@@ -182,6 +182,20 @@ pub struct ModuleError {
...
+    /// A raw byte representation of the error.
+    pub raw: ModuleErrorRaw,
+}
+
+/// The raw error details about a module error that has occurred.
+///
+/// **Note**: Structure used to obtain the underlying bytes of a ModuleError.
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("Pallet index {pallet_index}: raw error: {error:?}")]
+pub struct ModuleErrorRaw {
```
ascjones, line 194: Could do with a more descriptive name e.g. `ModuleErrorLookup` or `ModuleErrorData`

subxt/src/error.rs
```diff
@@ -190,5 +204,5 @@ pub struct ModuleError {
 pub trait HasModuleError {
     /// If the error has a `Module` variant, return a tuple of the
     /// pallet index and error index. Else, return `None`.
-    fn module_error_indices(&self) -> Option<(u8, u8)>;
+    fn module_error_indices(&self) -> Option<ModuleErrorRaw>;
```
ascjones, line 207: This method should be renamed depending on what name you choose for `ModuleErrorRaw`

subxt/src/transaction.rs
```diff
@@ -389,14 +389,17 @@ impl<'client, T: Config, E: Decode + HasModuleError, Evs: Decode>
...
-                if let Some((pallet_idx, error_idx)) =
-                    dispatch_error.module_error_indices()
-                {
-                    let details = self.client.metadata().error(pallet_idx, error_idx)?;
+                if let Some(raw_error) = dispatch_error.module_error_indices() {
+                    // Error index is utilized as the first byte from the error array.
+                    let details = self
+                        .client
+                        .metadata()
+                        .error(raw_error.pallet_index, raw_error.error[0])?;
```
ascjones, line 397: It would be better to encapsulate `raw_error.error[0])` in a method e.g. `raw_error.error_index()`

codegen/src/api/errors.rs
ascjones: Yes looks good I have added some nitpicks see below

ascjones (approved):
LGTM

PR #494, opened 2022-03-29
Read constants from metadata at runtime

### LINKED ISSUE #493: Fetch constants from runtime metadata?
Constants are currently quite literally constant, since they are embedded in the code generated from the metadata provided at compile time. 

However, I believe it is more useful to read the constants from the metadata fetched at runtime from the node. This is for the case where the constant values differ in the runtime targeted node, which is legitimate when building an app which can target different compatible chains.

I came across this when debugging `povstats` which [reads the maximum block weight](https://github.com/athei/povstats/blob/master/src/lib.rs#L57), and I was adjusting the block weight on the node side which wasn't being reflected on the client side.

Rel https://github.com/paritytech/subxt/pull/402

ascjones:
Closes #493


lexnv (approved):

lexnv:
LGTM, the build error "expected `u8`, found array `[u8; 4]`" is related to Issue #491 that is fixed in PR #492

jsdw (approved):

PR #495, opened 2022-04-01
Use tokio instead of async-std

### LINKED ISSUE #485: Use tokio instead of async-std.
Jsonrpsee (a dependency) uses `tokio` under the hood. Would it make sense for `subxt` to align on this and also use `tokio` instead of `async-std` (which is currently only a dev dependency anyway) to avoid pulling in both?

From a quick glacne it looks like it would be a fairly trivial swap; `#[async_std::test]` tp `#[tokio::test]` and a few other similarly easy changes.

lexnv:
Remove the `async-std` dependency and utilize `tokio` instead.
This is motivated by the fact that `jsonrpsee` utilizes under the hood `tokio`.

Closes #485.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -13,7 +13,7 @@ description = "Subxt example usage"
 
 [dev-dependencies]
 subxt = { path = "../subxt" }
-async-std = { version = "1.9.0", features = ["attributes", "tokio1"] }
+tokio = { version = "1.8", features = ["full"] }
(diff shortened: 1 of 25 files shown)


niklasad1 (approved):
nice with a smaller dependency tree, LGTM

ascjones (approved):
Nice :+1:

jsdw (approved):
Lovely stuff!

PR #498, opened 2022-04-05
Rely on the kernel for port allocation

### LINKED ISSUE #497: Test failing when having a `polkadot` binary started
The testing framework utilizes the `substrate` binary.
However, when having a `polkadot` binary started (version 4542a603cc) the `cargo test` is failing.

```
> cargo test
...
test frame::balances::constant_existential_deposit ... FAILED

failures:

---- frame::balances::constant_existential_deposit stdout ----
thread 'frame::balances::constant_existential_deposit' panicked at 'assertion failed: `(left == right)`
  left: `10000000000`,
 right: `100000000000000`', subxt/tests/integration/frame/balances.rs:272:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

lexnv:
The testing infrastructure requires that a `substrate` binary is present for individual testing.
However, having a different `polkadot` binary running resulted in failing tests.

### Root Cause

The function `next_open_port()` in `subxt` crate produced 3 port numbers by inspecting
a range of ports. 
To validate that a given port is opened the `TcpListener::bind(("0.0.0.0", 0))` bind call is executed.
However, the address "0.0.0.0" is not representative of the localhost, but of a non-routable unknown address.

This observation is validated using the `polkadot` binary which starts by default on `127.0.0.1:9944`. Any `TcpListener::bind(("0.0.0.0", 0))` will result in success even with the binary started.

### Port Allocation

While at it, remove the scanning logic and rely entirely on the Kernel to provide 3 available ports.
The port allocation is performed by the [inet_csk_get_port()](https://elixir.bootlin.com/linux/v5.8/source/net/ipv4/inet_connection_sock.c#L303) function.
When providing a zero port, asking for allocation, the code path goes to [inet_csk_find_open_port()](https://elixir.bootlin.com/linux/v5.8/source/net/ipv4/inet_connection_sock.c#L182).
A port scanning is performed that uses a spinlock for concurrent access with the addition of checking for bind conflicts.

The kernel should provide 3 different ports if they are available. 

### Testing Done
Started 100 threads, each requesting for 100 new ports. Multiple runs of the binary presented no duplicates.
However, when requesting more ports than available (ie from 1000 threads), the ports are reutilized and values eventually wrap from `net.inet.ip.portrange.last` to `net.inet.ip.portrange.first`.

Closes #497.

### DIFF
--- a/subxt/tests/integration/utils/node_proc.rs
+++ b/subxt/tests/integration/utils/node_proc.rs
@@ -22,10 +22,6 @@ use std::{
     },
     net::TcpListener,
     process,
-    sync::atomic::{
-        AtomicU16,
-        Ordering,
-    },
     thread,
     time,
 };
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):
nice :+1:

PR #501, opened 2022-04-06
Tests: parse port from substrate binary output to avoid races

jsdw:
Start the binary with port=0 and parse out the actual port number it starts on from its output. This (hopefully) avoids any races between finding and makign use of available ports. It also has the nice side effect of avoiding the need for any arbitrary sleeping to wait for the RPC server to start up.

### DIFF
--- a/subxt/tests/integration/utils/node_proc.rs
+++ b/subxt/tests/integration/utils/node_proc.rs
@@ -20,16 +20,14 @@ use std::{
         OsStr,
         OsString,
     },
-    net::TcpListener,
     process,
-    thread,
-    time,
 };
 use subxt::{
     Client,
     ClientBuilder,
     Config,
 };
+use std::io::{ Read, BufRead, BufReader };
 
 /// Spawn a local substrate node for testing subxt.
 pub struct TestNodeProcess<R: Config> {
@@ -79,7 +77,6 @@ where
 pub struct TestNodeProcessBuilder {
     node_path: OsString,
     authority: Option<AccountKeyring>,
-    scan_port_range: bool,
 }
 
 impl TestNodeProcessBuilder {
@@ -90,7 +87,6 @@ impl TestNodeProcessBuilder {
         Self {
             node_path: node_path.as_ref().into(),
             authority: None,
-            scan_port_range: false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


lexnv (approved):

subxt/tests/integration/utils/node_proc.rs
```diff
@@ -185,25 +150,30 @@ impl TestNodeProcessBuilder {
...
+// locate the port number that is logged out to it.
+fn find_substrate_port_from_output(r: impl Read + Send + 'static) -> u16 {
+    BufReader::new(r)
+        .lines()
+        .find_map(|line| {
+            let line = line
+                .expect("failed to obtain next line from stdout for port discovery");
+
+            // does the line contain our port (we expect this specific output from substrate).
+            let line_end = match line.rsplit_once("Listening for new connections on 127.0.0.1:") {
```
niklasad1, line 163: this will not work once we update to jsonrpsee in substrate but a later question I guess.

test-runtime/Cargo.toml
```diff
@@ -11,5 +11,5 @@ codec = { package = "parity-scale-codec", version = "3.0.0", default-features =
 [build-dependencies]
 subxt = { path = "../subxt" }
 sp-core = "6.0.0"
-tokio = { version = "1.8", features = ["rt", "macros"] }
+tokio = { version = "1.8", features = ["rt", "macros", "rt-multi-thread"] }
```
niklasad1, line 14:
```suggestion
tokio = { version = "1.8", features = ["macros", "rt-multi-thread"] }
```

niklasad1 (approved):

subxt/tests/integration/utils/node_proc.rs
TarikGul: DQ: Why wont this work with the update? Just out of curiousity.

TarikGul (approved):
👍

subxt/tests/integration/utils/node_proc.rs
jsdw: Aah, different log output for the port I guess?

(I wonder what the log output will be? it should be easy enough to look for both when the time comes, to continue to support older nodes :))

(I do wish there was a better way to find the port that a thing is running on; maybe I should look into this a bit more!)

ascjones: Never tried it, but maybe you could use something like https://docs.rs/netstat/latest/netstat/ to find the open ports of the spawned process, then might be easy enough to find the websockets port?

jsdw: Ooh thanks for the pointer; I might have a look into this!

niklasad1: > Aah, different log output for the port I guess?

Yeah, but we could change to have this log in jsonrpsee it's quite useful.

PR #507, opened 2022-04-07
[static metadata] validate storage, calls and constants per call

jsdw:
This is to be merged into the statric metadata PR. It baiscally does per-call/constant/storage validation, so that we only pay for what we use, and any validation errors we get back are specific to the call being used.

Some things yet to be done, that can be picked up when this is merged:
- ~~cache the per-call/event/storage stuff~~
- ~~remove the pallet hashing code? (If we aren't using it now, let's ditch it to keep things simple)~~
- ~~simplify what `Metadata.metadata_hash` returns? (again; if no more per-pallet stuff then maybe this should just return a hash, and we can separately cache the per call/storage/constant stuff?)~~

Edit: I had a pass over and added this stuff :)


metadata/src/lib.rs
```diff
@@ -98,13 +98,7 @@ fn get_type_def_hash(
...
         TypeDef::Variant(variant) => {
             let mut bytes = Vec::new();
-            // The type at path `node_template_runtime::Call` contains variants of the pallets
-            // registered in order. Swapping the order between two pallets would result
-            // in a different hash, but the functionality is still identical.
-            // Sort by variant name to result in deterministic hashing.
-            let mut variants: Vec<_> = variant.variants().iter().collect();
-            variants.sort_by_key(|variant| variant.name());
-            for var in variants {
+            for var in variant.variants().iter() {
```
jsdw, line 101: I removed this logic; it seems like we sort pallets (or should sort them) at the level of getting an entire-metadata hash, which I think will satisfy why we are sorting them?

Also, we can't sort _any_ variant type in this way, because variant types will be SCALE encoded/decoded taking into account the variant index, and so if some of the variants move around normally, it's a breaking change. We can _only_ do this at th pallet level as a special one-off to allow pallets to move around (and probably at the call level as well because our runtime code allows these to be shuffled around too I think, but would need to check!)

Removing this didn't seem to break any tests, but def worth making sure we aren't breaking any use cases and don't need to have this sort of logic somewhere else!

metadata/src/lib.rs
```diff
@@ -216,6 +210,120 @@ fn get_extrinsic_hash(
...
+        .find(|v| v.name() == call_name)
+        .ok_or(NotFound::Item)?;
+
+    // hash the specific variant representing the call we are interested in.
+    let hash = get_variant_hash(&metadata.types, variant, &mut HashSet::new());
+    Ok(hash)
+}
+
 /// Obtain the hash representation of a `frame_metadata::PalletMetadata`.
 pub fn get_pallet_hash(
```
jsdw, line 328: perhaps we can remove this function if we no longer need it?

jsdw:
Although CI doesn't show it, this passes all of the tests locally and such. @lexnv feel free to have a look over this and merge it into your static metadata PR whenever you're ready!

lexnv (approved):

PR #511, opened 2022-04-18
Propagate documentation to runtime API

### LINKED ISSUE #503: Rustdoc for the generated API?
Is there rustdoc (or any documentation) for the generated API?

lexnv:
This PR propagates the documentation from the metadata blob to the generated API.

The documentation is propagated to either TransactionApi or StorageApi if the inspected type is a call or storage, instead of being placed directly on the generated object.

This is done mainly to document the paths by which the user interacts with the API, as opposed to documenting the inner details of the API. For other types, the metadata documentation is set in place for the generated object.

#### Inherited from generated object

```rust
impl<'a, T: ::subxt::Config> StorageApi<'a, T> {
...
    #[doc = " The full account information for a particular account ID."]
    pub async fn account(
```

```rust

pub struct TransactionApi<'a, T: ::subxt::Config, X> {
...
    #[doc = "Anonymously schedule a task."]
    pub fn schedule(
```

```rust
pub struct ConstantsApi<'a, T: ::subxt::Config> {
...
    #[doc = " The maximum weight that may be scheduled per block for any dispatchables of less"]
    #[doc = " priority than `schedule::HARD_DEADLINE`."]
    pub fn maximum_weight(
```

#### Placed directly on the generated object

```rust
pub type Event = runtime_types::polkadot_runtime_common::crowdloan::pallet::Event;
  pub mod events {
    use super::runtime_types;
    #[derive(:: subxt :: codec :: Encode, :: subxt :: codec :: Decode, Debug)]
    #[doc = "Create a new crowdloaning campaign. `[fund_index]`"]
    pub struct Created(pub runtime_types::polkadot_parachain::primitives::Id);
```

```rust
pub type Event = runtime_types::pallet_preimage::pallet::Event;
  pub mod events {
    use super::runtime_types;
    #[derive(:: subxt :: codec :: Encode, :: subxt :: codec :: Decode, Debug)]
    #[doc = "A preimage has been noted."]
    pub struct Noted {
        pub hash: ::subxt::sp_core::H256,
    }
```

Closes #503 .

### DIFF
--- a/codegen/src/api/calls.rs
+++ b/codegen/src/api/calls.rs
@@ -40,14 +40,14 @@ pub fn generate_calls(
     call: &PalletCallMetadata<PortableForm>,
     types_mod_ident: &syn::Ident,
 ) -> TokenStream2 {
-    let struct_defs = super::generate_structs_from_variants(
+    let mut struct_defs = super::generate_structs_from_variants(
         type_gen,
         call.ty.id(),
         |name| name.to_upper_camel_case().into(),
         "Call",
     );
     let (call_structs, call_fns): (Vec<_>, Vec<_>) = struct_defs
-        .iter()
+        .iter_mut()
         .map(|struct_def| {
             let (call_fn_args, call_args): (Vec<_>, Vec<_>) =
                 match struct_def.fields {
@@ -78,6 +78,10 @@ pub fn generate_calls(
             let function_name = struct_def.name.to_string().to_snake_case();
             let fn_name = format_ident!("{}", function_name);
 
+            // Propagate the documentation just to `TransactionApi` methods, while
+            // draining the documentation of inner call structures.
+            let docs = struct_def.docs.take();
+            // The call structure's documentation was stripped above.
             let call_struct = quote! {
                 #struct_def
 
@@ -87,6 +91,7 @@ pub fn generate_calls(
                 }
             };
             let client_fn = quote! {
+                #docs
                 pub fn #fn_name(
                     &self,
                     #( #call_fn_args, )*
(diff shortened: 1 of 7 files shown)


codegen/src/api/calls.rs
```diff
@@ -78,6 +78,10 @@ pub fn generate_calls(
             let function_name = struct_def.name.to_string().to_snake_case();
             let fn_name = format_ident!("{}", function_name);
 
+            // Propagate the documentation just to `TransactionApi` methods, while
```
niklasad1, line 81: can you explain the why the docs for the inner call structures are "drained" and generated for the wrappers only?

niklasad1 (approved):
Looks clean, is there a way to test that the docs are generated with a compile-time test such as `#[deny(missing_docs)]` or something?

codegen/src/api/calls.rs
lexnv: My initial intuition was to keep the documentation on both the calling methods and inner structures.
The user should call the exposed API and not necessarily care about the inner structures used to construct the call. 

```rust
  pub struct Transfer {
   ...
      pub value: ::core::primitive::u128,
  }

  /// Doc goes here
  pub fn transfer( ... ) {
      let call = Transfer { dest, value };
      ::subxt::SubmittableExtrinsic::new(self.client, call)
  }
```

Although someone could potentially submit a manually constructed extrinsic (as the `pub fn transfer` implementation), I think this is quite uncommon. 

Would you think having documentation on both would also satisfy those manually constructed cases? 🤔

lexnv:
@niklasad1 Thanks for the feedback and review :D I don't think we could do a compile-time check with simply `#[deny(missing_docs)]`, some Types simply would not provide documentation. However, I do believe a small script/test could be added to: 
- Get metadata from runtime node
- Obtain recursively the documentation from the metadata
- Read `subxt/codegen/../polkadot.rs` line by line and ensure all lines from metadata are present there

codegen/src/api/calls.rs
niklasad1: Ok, I agree the outer docs should be sufficient then.

I think someone constructs their own extrinsics I guess they are on their own anyway.

ascjones (approved):
Awesome :+1:

PR #513, opened 2022-04-20
Subscribe to Runtime upgrades for proper extrinsic construction

### LINKED ISSUE #509: Gracefully handle runtime upgrades while connected to a node
During a runtime upgrade, things like the spec/runtime version will potentially change. The subxt `Client` downloads these when it first connects to a node, and after an update they might no longer be valid. This would lead to things like transactions (which rely on these numbers lining up between subxt and node) no longer working. The current workaround would be to create a new `Client`.

Can/should we subscribe to something like `state_subscribeRuntimeVersion` when the Client connects to a node, to make sure that any changes to these values (we care about `transaction_version` and `spec_version`) are kept in sync?

The metadata might also change; could also spot this and update our stored metadata? With things like the static validation work, this would make sure that a long running client can properly warn about differences (as well as make sure that constants and such as obtained from the metadata are still correct).

lexnv:
There are cases when the subxt `Client` is connected to a substrate node
that performs a [Runtime update (forkless update)](https://docs.substrate.io/tutorials/v3/forkless-upgrades/).

The `RuntimeVersion` is populated when the `Client` first connects to a node, but it was never
updated to keep in sync with the node. The `spec_version` and `transaction_version` of the
`RuntimeVersion` are included in the transaction parameters. Having the values out of sync means
that the subxt `Client` can no longer submit any extrinsics.

This PR exposes the ability to subscribe to the runtime updates via `subscribe_runtime_version` and
receive notifications when the node performs certain updates.

Customers can easily subscribe to runtime updates utilizing a few lines of code
```rust
    let update_client = api.client.updates();
    tokio::spawn(async move {
        let result = update_client.perform_runtime_updates().await;
        println!("Runtime update failed with result={:?}", result);
    });
```

To keep `subxt` agnostic of the async runtime preferred by the customer, and to simplify the
code needed for providing the updates, the runtime update is handled entirely by the customer.
Furthermore, when a runtime update is detected the customer has the ability to set on the
subxt `Client` the new `RuntimeVersion` and `Metadata` which would allow for further extrinsics to be submitted.

An example to illustrate the runtime update is introduced:

```bash
> RUST_LOG=debug cargo run --example subscribe_runtime_updates                                                            

Balance transfer extrinsic submitted: 0xce5c…003a
[2022-05-02T16:31:00Z INFO  subxt::updates] Performing runtime update from 100 to 101
[2022-05-02T16:31:00Z DEBUG subxt::updates] Performing metadata update
[2022-05-02T16:31:00Z DEBUG subxt::updates] Runtime update completed

Balance transfer extrinsic submitted: 0x444e…a06c
[2022-05-02T16:31:54Z INFO  subxt::updates] Performing runtime update from 101 to 102
[2022-05-02T16:31:54Z DEBUG subxt::updates] Performing metadata update
[2022-05-02T16:31:54Z DEBUG subxt::updates] Runtime update completed

Balance transfer extrinsic submitted: 0xca0e…df24
```

Performing the same operations of updating the node, but without updating subxt `Client` itself:

```bash
Balance transfer extrinsic submitted: 0xfbef…0230
Balance transfer extrinsic submitted: 0x6790…a08f
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Rpc(Call(Custom { code: 1010, message: "Invalid Transaction", data: Some(RawValue("Transaction has a bad signature")) }))', examples/examples/subscribe_runtime_updates.rs:55:14
```

### Next Steps
- [x] Synchronize the client metadata with the runtime metadata to detect changes using #478 (however it would be counterproductive to do so until the static metadata PR is merged due to conflicts).

Closes #509.


subxt/src/client.rs
```diff
@@ -179,6 +182,18 @@ impl<T: Config> Client<T> {
...
+
+    /// Returns a snapshot of the client Runtime Version.
+    pub fn runtime_version(&self) -> RuntimeVersion {
+        let runtime = self.runtime_version.read().unwrap();
+        runtime.clone()
+    }
+
+    /// Set the given Runtime Version on the client.
+    pub fn set_runtime_version(&self, runtime: RuntimeVersion) {
+        let mut actual = self.runtime_version.write().unwrap();
```
lexnv, line 194: One other option would be:
```rust
if actual.spec_version < runtime.spec_version {
    *actual = runtime;
}
```

And therefore eliminate the need for cloning the runtime whenever inspecting the code.

examples/examples/subscribe_runtime_updates.rs
```diff
@@ -0,0 +1,130 @@
...
+            .unwrap();
+
+        println!("Balance transfer extrinsic submitted: {}", hash);
+        tokio::time::sleep(Duration::from_secs(30)).await;
+    }
+}
+
+/// This function handles runtime updates via subscribing to the
+/// node's RuntimeVersion.
+async fn runtime_update(api: &RuntimeApi) {
```
lexnv, line 64: I do believe that this should exist by default on the subxt `Client`. I would move this when we handle `Metadata` injection to the `Client`.

What do you think?

niklasad1 (commented):
I found the API a little hard to use for end-users.

Basically, I was kind of expecting subxt to spawn a separate task that does the runtime subscription for the user and update the runtime version under the hood. Is that too opinionated?

After the a runtime upgrade the `subxt client` is mostly useless anyway (e.g. the spec version gets bumped)

Otherwise, the code looks good

jsdw:
This is an interesting point! I was also expecting subxt to spawn something under the hood and do this transparently, but as @lexnv  points out, `subxt` is currently runtime agnostic, which is handy, and it would have to give that up to spawn tasks.

Let's have a think about this and try to find the simplest API possible for this that keeps subxt agnostic of the runtime used.

I note that we are pulling in Tokio from `jsonrpsee` at the mo, but thinking ahead to things like supporting wasm, (which would no longer bring that in) means I would still bve reluctant to committing to it here, for the sake of future simplicity!

jsdw:
A random stab at what an API could look like:

```rust
// spawn a background task to keep subxt up to date w.r.t node runtime updates.
tokio::spawn(async move {
    api.client.perform_automatic_runtime_updates().await // awaits forever.
});

// use the api however you like in the meantime, and know it'll be kept uptodate in the background...
api.tx().balances()...
```

How this would work roughly is that the background update task would subscribe to runtime updates, and when it receives one, do something like:
- acquire a lock on the (async) mutex containing runtime + metadata,
- go and download the metadata
- update metadata + runtime version
- unlock the mutex. 

The idea is that as soon as we know an update has happened, block any pending requests until we have new runtime + metadata and can send them with the correct versions and validate them against the correct metadata. It'll require a bit of care to do properly (need to handle errors while locked and such).

codegen/src/api/constants.rs
```diff
@@ -48,9 +48,9 @@ pub fn generate_constants(
 
         quote! {
             #( #[doc = #docs ] )*
-            pub async fn #fn_name(&self) -> ::core::result::Result<#return_ty, ::subxt::BasicError> {
+            pub fn #fn_name(&self) -> ::core::result::Result<#return_ty, ::subxt::BasicError> {
                 let locked_metadata = self.client.metadata();
-                let metadata = locked_metadata.lock().await;
+                let metadata = locked_metadata.read();
                 if metadata.constant_hash(#pallet_name, #constant_name)? == [#(#constant_hash,)*] {
```
jsdw, line 54: I like the pattern you did in `calls.rs`; I reckon it'd be nice to repeat that here and in `validate_metdata` below for consistency :)

examples/examples/subscribe_runtime_updates.rs
```diff
@@ -0,0 +1,82 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with subxt.  If not, see <http://www.gnu.org/licenses/>.
+
+//! To run this example, a local polkadot node should be running. Example verified against polkadot 0.9.18-f6d6ab005d-aarch64-macos.
+//!
+//! E.g.
+//! ```bash
+//! curl "https://github.com/paritytech/polkadot/releases/download/v0.9.13/polkadot" --output /usr/local/bin/polkadot --location
```
jsdw, line 21: We need to update all of these URLs; they all point to v0.9.13 but should at least point to 0.9.18 or whatever version we've verified examples against to date (ideally 0.9.21 in the next couple of daye).

examples/examples/subscribe_runtime_updates.rs
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    // Make multiple transfers to simulate a long running `subxt::Client` use-case.
+    //
+    // Meanwhile, the tokio task will perform any necessary updates to ensure
+    // that submitted extrinsics are still valid.
+    //
+    // Ideally, the polkadot node should perform a few runtime updates
+    // For more details on how to perform updates on a node, please follow:
+    // https://docs.substrate.io/tutorials/v3/forkless-upgrades/
```
jsdw, line 61:
```suggestion
    // Make multiple transfers to simulate a long running `subxt::Client` use-case.
    //
    // Meanwhile, the tokio task above will perform any necessary updates to keep in sync
    // with the node we've connected to. Transactions submitted in the vicinity of a runtime
    // update may still fail, however, owing to a race between the update happening and
    // subxt synchronising its internal state with it.
```

subxt/src/events/events_type.rs
```diff
@@ -759,7 +771,7 @@ mod tests {
         }
 
         // Create fake metadata that knows about our single event, above:
-        let metadata = metadata::<Event>();
+        let metadata = Arc::new(RwLock::new(metadata::<Event>()));
```
jsdw, line 774: This pattern of creating metadata in an Arc + RwLock is copied a lot; perhaps a helper function (`metadata()`?) should handle the arc/rwlocking too?

subxt/src/updates.rs
```diff
@@ -0,0 +1,104 @@
...
+        while let Some(update_runtime_version) = update_subscription.next().await {
+            // The Runtime Version obtained via subscription.
+            let update_runtime_version = update_runtime_version?;
+
+            // Ensure that the provided Runtime Version can be applied to the current
+            // version of the client. There are cases when the subscription to the
+            // Runtime Version of the node would produce spurious update events.
+            // In those cases, set the Runtime Version on the client if and only if
+            // the provided runtime version is bigger than what the client currently
+            // has stored.
```
jsdw, line 68: I'd assume that the new runtimeversion would only ever be either identical to the one we currently have (no need to do anything), or different, which implies we should update to match. I'd guess we never need to worry about it being older than what we know about, and if for some reason it is, then we should probably just assume it's true and update to match it anyway?

(basically, maybe we should just do an equality check?)

jsdw:
Left a couple of very minor comments, but otherwise I think this is great stuff, and am happy to see it merge :)

codegen/src/api/constants.rs
lexnv: Have updated for `validate_metdata` thanks! For `constants.rs` the metadata needs to live a bit longer to extract the constant value from runtime.

subxt/src/events/events_type.rs
lexnv: Have changed it! Thanks!

subxt/src/updates.rs
lexnv: Indeed, that model makes more sense. Have updated the comment as well to describe the spurious event a bit better.
Also, the equality check is something we do for: https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Init.ts#L277. Thanks for pointing this out! :D

niklasad1 (approved):
Looks great,

I like the improved API

jsdw (approved):
Awesome stuff :)

PR #515, opened 2022-04-22
Put integration tests behind feature flag

### LINKED ISSUE #351: Put integration tests behind feature flag to avoid need for substrate on PATH
Currently, running `cargo test` requires a `substrate` binary on your PATH. This used to be a runtime error (and only the affected tests would fail), but now it is a compile-time error because we now generate the runtime details at build time against the `substrate` binary we'll point at for tests.

I'd suggest hiding the integration tests behind an `integration-tests` feature flag, so that this is not the case.

If any files are generated (`cargo c --tests` is reported to have left behind a `client/dev-chain.json` file for instance), put generation of them behind the feature flag if applicable, and at least write them to a temp fodlere if possible.

lexnv:
The `substrate` binary is used internally for integration testing purposes.

This PR ensures that building/running tests do not rely on the existence of the
`SUBSTRATE_NODE_PATH` in the environment, which is guarded under the
`integration-tests` crate.

The `subxt` tests that rely on the substrate binary are moved to a dedicated crate.
At the same time, any dependency on the `test-runtime` is removed to avoid any
cyclic dependencies. The only crate that requires the substrate binary to exist in the
environment is the `integration-tests` crate.

### Testing Done
```bash
cd subxt
SUBSTRATE_NODE_PATH="" cargo build --release
SUBSTRATE_NODE_PATH="" cargo test
```

Closes #351.

### Next Steps
- [x] Move integration tests to their own crate to avoid substrate binary dependency on other crates
- [x] Construct manual metadata for subxt caching tests
- [x] Add metadata.scale blob for examples and metadata crate testing
- [x] Ensure `#[cfg]` is valid with feature flag


ascjones (commented):
> An alternative to the test-runtime feature-flag approach would be having the
dependency declared as optional = true and included for the feature flag only.
Unfortunately, this option does not work for dev-dependencies. Having the
test-runtime as simply dependency would cause cyclic reference.

Since `tests/integration` is really it's own crate, this could be moved to the top level and made into an explicit crate which would allow using a normal dependency and make it `optional`?

subxt/tests/integration/main.rs
```diff
@@ -15,19 +15,26 @@
 // along with subxt.  If not, see <http://www.gnu.org/licenses/>.
 
 mod codegen;
+#[cfg(feature = "integration-tests")]
```
jsdw, line 18: Could we just have a single

`#![cfg(feature = "integration-tests")]`

at the top of the file, to cover it all?

lexnv: Yep, I think we could remove those entirely from here when we move it to its own `integration-tests` crate

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
```
jsdw, line 1: Could we remove a layer of directories and just have everything in `src/tests` rather than `src/tests/integration`?

(I think it's a given that any tests in the separate `tests` folder are already integration tests anyway, really.

subxt/src/metadata/metadata_type.rs
```diff
@@ -478,13 +478,64 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
...
+                signed_extensions: vec![],
+            },
+            meta_type::<()>(),
+        );
+        let prefixed = RuntimeMetadataPrefixed::from(metadata);
 
-        Metadata::try_from(meta)
+        Metadata::try_from(prefixed)
             .expect("Cannot translate runtime metadata to internal Metadata")
     }
```
jsdw, line 541:
```
// Currently the caching does not take into account different pallets
// as the intended behavior is to use this method only once.
// Enforce this behavior into testing.
let hash_old = metadata.metadata_hash(&["Balances"]);
assert_eq!(hash_old, hash);
```

I don't really understand what this line does? (also doesn't look like there is a balances pallet in the fake metadata?)

jsdw:
Looks good to me offhand, and I think the separate crate is a nice organisational improvement, and gives people the opportunity to avoid the integration tests if they don't have a suitable substrate binary to hand while still making it the default, which I think is right.

Just a couple of small comments and it's all good from me :)

ascjones (commented):
How about moving the contents of `integration-tests/tests/` (which at the moment is an implicit sub-crate) into `integration-tests/src` (renaming `main.rs` to `lib.rs`?

lexnv:
@ascjones Moving the`/tests` directory directly under `/src` makes the crate more ergonomic for me. Thanks for the feedback!

ascjones (approved):
LGTM

subxt/src/metadata/metadata_type.rs
jsdw: Thanks, that is clearer! :)

jsdw (approved):
Looking good to me, nice one!

PR #524, opened 2022-04-29
Use minimum amount of dependencies for crates 

lexnv:
Reduce the number of dependencies for isolated crates.
This will greatly benefit developers that want to work with a subset of
`subxt` crates, as redundant crates are no longer included in the `Cargo.toml`.

One real benefit of the trimming of dependencies is the `chameleon` crate.


jsdw (approved):
Lovely stuff! Looks good to me :)

PR #528, opened 2022-05-09
Bump jsonrpsee

ascjones:


subxt/src/rpc.rs
```diff
@@ -564,7 +564,7 @@ pub async fn ws_client(url: &str) -> Result<RpcClient, RpcError> {
     let (sender, receiver) = ws_transport(url).await?;
     Ok(RpcClientBuilder::default()
         .max_notifs_per_subscription(4096)
-        .build(sender, receiver))
```
ascjones, line 567: Would be fun if someone has some time to try `build_with_wasm` in a wasm environment...

jsdw (approved):

PR #532, opened 2022-05-11
[WIP] Add crate for runtime SCALE encoding/decoding into a Value type.

jsdw:
This lays some of the foundations that we need to be able to submit dynamic queries and dynamically decode results in Subxt.

The static interface will always be preferable, but there are times where somebody needs to be able to connect a subxt client to nodes across different chains, and be able to dynamically learn about and query the interface each node provides (think Polkadot.js for instance). Having a dynamic value type also means that we have something we can target when parsing arguments from a command line tool, or are able to provide more structured information about things like unknown events.


value/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    ) -> Result<T, DeserializeError> {
+        T::deserialize(value)
+    }
+
+    // TODO: Eventually let's implement Serializer on Value so that we can convert from some type into a Value:
+    //
+    // /// Attempt to serialize some type into a [`crate::Value`].
+    // pub fn to_value<'de, Ctx, T: serde::Serialize>(ty: T) -> Result<crate::Value<()>, SerializeError> {
+    //     ty.serialize(serializer)
+    // }
```
jsdw, line 52: I think we can merge this PR without adding this Serialize impl, but it would be a nice to have to round out the interface (it's the only serde trait not implemented, and being able to cast from structs to intermediate Value types may be useful (for example in allowing more flexible encodings if there is a specific known type you'd like to encode to).

jsdw:
I've moved this code (and PR) to its own repository, `scale-value`, since I think it is more general than `subxt` and doesn't really make sense to be published alongside it at the same version as other subxt crates:

https://github.com/paritytech/scale-value/pull/1

So I'll close it here.

PR #535, opened 2022-05-12
Replace `log` with `tracing` and record extrinsic info

ascjones:
Migrate to `tracing` crate in line with `jsonrpsee` and `substrate`.

Also add some extra logs for inspecting extrinsic construction.


niklasad1 (approved):
LGTM, but you need to fix `log!` in a couple of more places :P

subxt/src/extrinsic/params.rs
```diff
@@ -31,7 +32,7 @@ pub use sp_runtime::generic::Era;
 /// "additional" parameters that are signed and used in transactions.
 /// see [`BaseExtrinsicParams`] for an implementation that is compatible with
 /// a Polkadot node.
-pub trait ExtrinsicParams<T: Config> {
+pub trait ExtrinsicParams<T: Config>: Debug {
```
jsdw, line 35: Just wodnering why this is needed?

subxt/src/client.rs
```diff
@@ -352,6 +354,11 @@ where
             )
         };
 
+        tracing::debug!(
+            "additional_and_extra_params: {:?}",
+            additional_and_extra_params
+        );
```
ascjones, line 360: @jsdw because of this

subxt/src/extrinsic/params.rs
ascjones: See comment below :point_down: 

Chose to do on the trait itself rather than the method for a better error message.

subxt/src/client.rs
jsdw: Aha!

jsdw (approved):

dvdplm (approved):

PR #536, opened 2022-05-12
fix: Find substrate port on different log lines

lexnv:
Updating `substrate` node to the latest version would result in an infinite loop for our integration testing.

The root cause of this issue is that the `substrate` switched to json-rpsee.
This included changing the log message we are looking for when finding the port of the substrate.

Ensure that the port can be obtained from both new and old versions of the substrate node.

#### Before
```
test client::chain_subscribe_blocks has been running for over 60 seconds
test client::chain_subscribe_finalized_blocks has been running for over 60 seconds
test client::fetch_block has been running for over 60 seconds
```

### After

```
test client::chain_subscribe_finalized_blocks ... ok
test client::fetch_block ... ok
test client::chain_subscribe_blocks ... ok
test client::fetch_read_proof ... ok
```

### DIFF
--- a/integration-tests/src/utils/node_proc.rs
+++ b/integration-tests/src/utils/node_proc.rs
@@ -168,22 +168,30 @@ fn find_substrate_port_from_output(r: impl Read + Send + 'static) -> u16 {
     BufReader::new(r)
         .lines()
         .find_map(|line| {
-            let line = line
-                .expect("failed to obtain next line from stdout for port discovery");
+            let line =
+                line.expect("failed to obtain next line from stdout for port discovery");
 
             // does the line contain our port (we expect this specific output from substrate).
-            let line_end = match line.rsplit_once("Listening for new connections on 127.0.0.1:") {
-                None => return None,
-                Some((_, after)) => after
+            let line_end = match line
+                .rsplit_once("Listening for new connections on 127.0.0.1:")
+            {
+                None => {
+                    match line.rsplit_once("Running JSON-RPC WS server: addr=127.0.0.1:")
+                    {
+                        None => return None,
+                        Some((_, after)) => after,
+                    }
+                }
+                Some((_, after)) => after,
             };
 
             // trim non-numeric chars from the end of the port part of the line.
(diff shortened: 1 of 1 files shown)


integration-tests/src/utils/node_proc.rs
```diff
@@ -168,22 +168,30 @@ fn find_substrate_port_from_output(r: impl Read + Send + 'static) -> u16 {
...
 
             // does the line contain our port (we expect this specific output from substrate).
-            let line_end = match line.rsplit_once("Listening for new connections on 127.0.0.1:") {
-                None => return None,
-                Some((_, after)) => after
+            let line_end = match line
+                .rsplit_once("Listening for new connections on 127.0.0.1:")
+            {
+                None => {
+                    match line.rsplit_once("Running JSON-RPC WS server: addr=127.0.0.1:")
```
niklasad1, line 179: you could also do: https://github.com/paritytech/substrate/blob/master/bin/node/cli/tests/common.rs#L164-#L167

then just parse is as `SocketAddr` then call port on it.

niklasad1: sorry, I didn't think about making the log easy to parse :)

niklasad1 (approved):

integration-tests/src/utils/node_proc.rs
jsdw: For the testing pov I guess it's ok assuming 127.0.0.1 though (ie you could start it on 0.0.0.0, but you couldn't then connect to 0.0.0.0)?

jsdw (approved):
Other than the suggestion LGTM

integration-tests/src/utils/node_proc.rs
niklasad1: it's fair assumption, I was thinking that some weird platforms could actually have some other loopback address than 127.0.0.1. Picking hair here just ignore me but I was more thinking about having to avoid parsing the port manually :P

I think you can connect to 0.0.0.0 but it means all addresses IIRC.

PR #544, opened 2022-05-13
Implement `Clone` for the generated `RuntimeApi`

lexnv:
This PR offers customers the ability to clone the `RuntimeAPI`.

### DIFF
--- a/subxt/src/lib.rs
+++ b/subxt/src/lib.rs
@@ -54,7 +54,7 @@ use codec::{
     Encode,
 };
 use core::fmt::Debug;
-pub use derivative::Derivative;
+use derivative::Derivative;
 
 mod client;
 mod config;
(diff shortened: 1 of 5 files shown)


subxt/src/lib.rs
```diff
@@ -54,7 +54,7 @@ use codec::{
     Encode,
 };
 use core::fmt::Debug;
-use derivative::Derivative;
+pub use derivative::Derivative;
```
jsdw, line 57: I'd lean towards just manually implementing clone instead of exposing Derivative in any way myself :)

lexnv:
Good point, i ll manually implement `Clone` to avoid introducing any bounds

jsdw (approved):
Good stuff :)

niklasad1 (approved):

PR #547, opened 2022-05-19
feat: refactor signing in order to more easily be able to dryrun

savudani8:
Co-authored-by: Sander Bosma <sanderbosma@gmail.com>
Co-authored-by: Daniel Savu <savudani04@yahoo.ro>

`create_signed` is refactored, adding support for `dry_run`ning an extrinsic before sending it.


integration-tests/src/frame/balances.rs
```diff
@@ -95,6 +97,87 @@ async fn tx_basic_transfer() -> Result<(), subxt::Error<DispatchError>> {
...
+    let cxt = test_context().await;
+    let api = &cxt.api;
+
+    let signed_extrinsic = api
+        .tx()
+        .balances()
+        .transfer(bob_address, 10_000)?
+        .create_signed(&alice, Default::default())
+        .await?;
+
```
savudani8, line 114: This test fails unless a `sleep` is added on this line: `tokio::time::sleep(Duration::from_secs(2)).await;`

I'm not sure how to investigate why this might happen, do you have any suggestions? The error is: `TransactionValidityError::Invalid(InvalidTransaction::BadProof`

savudani8:
Looks like the clippy error is unrelated to this PR. Maybe it has to do with the used toolchain?

jsdw:
> Looks like the clippy error is unrelated to this PR. Maybe it has to do with the used toolchain?

Yup, I suspect owing to a new version of rust and such being released yesterday. I'll look into it!

integration-tests/src/frame/balances.rs
jsdw: I wonder whether adding a sleep above the `let signed_extrinsic` line also works?

(I'm wondering whether it's something like the test_context needs to produce a block first or something before it accepts the extrinsic?!)

subxt/src/client.rs
```diff
@@ -404,6 +406,68 @@ where
...
+    ///
+    /// Returns `Ok` with an [`ApplyExtrinsicResult`], which is the result of applying of an extrinsic.
+    pub async fn dry_run(&self) -> Result<ApplyExtrinsicResult, BasicError> {
+        let params = rpc_params![Bytes::from(self.encoded.0.clone())];
+        let result_bytes: Bytes = self
+            .client
+            .rpc()
+            .client
+            .request("system_dryRun", params)
+            .await?;
```
jsdw, line 468: Just for consistency, I'm wondering whether this should be a function in rpc.rs

jsdw:
A fix for the clippy error is on it's way! 

I need to stare at the code a little more but offhand it looks good to me; thankyou for the PR!

jsdw:
If you merge up from master, the clippy error should go away now!

integration-tests/src/frame/balances.rs
savudani8: Managed to fix it on my local, submitting a prior extrinsic seemed to somehow interfere with the dryrun rpc so I removed the extrinsic.

subxt/src/rpc.rs
```diff
@@ -570,6 +576,28 @@ impl<T: Config> Rpc<T> {
         let params = rpc_params![public_key, key_type];
         Ok(self.client.request("author_hasKey", params).await?)
     }
+
+    /// Submits the extrinsic to the dry_run RPC, to test if it would succeed.
+    ///
+    /// Returns `Ok` with an [`ApplyExtrinsicResult`], which is the result of applying of an extrinsic.
+    pub async fn dry_run<'client, Y, X, E, Evs>(
+        &self,
+        signed_submittable_extrinsic: &SignedSubmittableExtrinsic<'client, Y, X, E, Evs>,
```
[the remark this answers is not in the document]
jsdw: In fact, I think if we go with `&[u8]` here, we can use `hex::encode` to turn that into hex for the RPC params.

subxt/src/client.rs
```diff
@@ -404,6 +398,56 @@ where
...
+    ///
+    /// Success does not mean the extrinsic has been included in the block, just that it is valid
+    /// and has been included in the transaction pool.
+    pub async fn submit(&self) -> Result<T::Hash, BasicError> {
+        self.client.rpc().submit_extrinsic(&self.encoded).await
+    }
+
+    pub fn encoded(&self) -> Encoded {
+        self.encoded.clone()
     }
```
jsdw, line 452: As suggested in the other comment, I reckon this can return a `&[u8]`, giving users the option to clone or not as needed. I'd add some docs too; something like:

```
/// Returns the SCALE encoded extrinsic bytes.
```

jsdw:
Looking great; a couple of small suggestions but otherwise I am really liking this separation; thank you for the contribution!

sander2:
I haven't followed this pr too closely but in the initial commit it was possible to do `create_signed().dry_run()`. Now instead you need to do 
```
let signed = create_signed();
client.rpc().dry_run(signed)
```
which in my opinion is a little bit less elegant. @jsdw would you be opposed to adding `dry_run` back into SignedSubmittableExtrinsic just as a wrapper to the one in rpc.rs?

jsdw:
I'd be perfectly up for having a dry_run function on `SignedSubmittableExtrinsic` that calls the RPC method with the bytes it has available!

jsdw:
One thing to do to fix the CI tests is to prepend "0x" to the hex string generated by `hex::encode()` (I always forget about that).

As for local tests not working, my only guess is that the substrate binary isn't on your PATH or something like that, so all of the tests are waiting for it to run (I assume no integration tests complete).

jsdw (approved):
Looks good to me!

lexnv (approved):

PR #561, opened 2022-06-07
Make storage futures only borrow client, not self, for better ergonomics

### LINKED ISSUE #558: futures::join! two Futures from api
Trying to understand how to make parallel calls , just like I can do via polkadotjs/api via Promises.all()

```
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>>{

        .set_url("wss://rpc.polkadot.io:443")
        .build()
        .await?
        .to_runtime_api::<polkadot::api::RuntimeApi<DefaultConfig, SubstrateExtrinsicParams<DefaultConfig>>>();


    let addr = AccountId32::from_str("13Ybj8CPEArUee78DxUAP9yX3ABmFNVQME1ZH4w8HVncHGzc").unwrap();

    let controller = api.storage().staking().bonded(&addr, None);
    let ledger = api.storage().staking().ledger(&addr, None);

    let (a, b) = join!(controller, ledger);

    Ok(())
}
```

What I get is the following:

```
22 |     let controller = api.storage().staking().bonded(&addr, None);
   |                      ^^^^^^^^^^^^^^^^^^^^^^^                    - temporary value is freed at the end of this statement
   |                      |
   |                      creates a temporary which is freed while still in use
...
25 |     let (a, b) = join!(controller, ledger);
   |                        ---------- borrow later used here
```

I can extract staking and ledger as variables, but maybe I do something wrong? probably I do not understand the concept.

jsdw:
Just a small qol improvement that allows one to get a future back from a storage query and await it later without needing to store any intermediate variables due to unnecessarily large lifetimes being captured by said futures.

Without this you'd need to write:

```rust
let staking = api.storage().staking();
let a_fut = staking.bonded(&addr, None);
let b_fut = staking.ledger(&addr, None);
let (a, b) = join!(a_fut, b_fut);
```

But with it you can chain all the way up to the future like so:

```rust
let a_fut = api.storage().staking().bonded(&addr, None);
let b_fut = api.storage().staking().ledger(&addr, None);
let (a, b) = join!(a_fut, b_fut);
```

Closes #558

### DIFF
--- a/codegen/src/api/storage.rs
+++ b/codegen/src/api/storage.rs
@@ -265,7 +265,7 @@ fn generate_storage_entry_fns(
 
     let lifetime_param = match should_ref {
         true => quote!(<'a>),
-        false => quote!()
+        false => quote!(),
     };
     let client_iter_fn = if matches!(storage_entry.ty, StorageEntryType::Map { .. }) {
         quote! (
@@ -303,7 +303,7 @@ fn generate_storage_entry_fns(
 
     let key_args_ref = match should_ref {
         true => quote!(&'a),
-        false => quote!()
+        false => quote!(),
     };
     let key_args = fields.iter().map(|(field_name, field_type)| {
         // The field type is translated from `std::vec::Vec<T>` to `[T]`, if the
--- a/codegen/src/api/storage.rs
+++ b/codegen/src/api/storage.rs
@@ -240,12 +240,6 @@ fn generate_storage_entry_fns(
         }
     };
 
-    let (lifetime_param, reference) = if should_ref {
-        (quote!(<'a>), quote!(&'a))
-    } else {
-        (quote!(), quote!())
-    };
-
     let storage_entry_impl = quote! (
         const PALLET: &'static str = #pallet_name;
         const STORAGE: &'static str = #storage_name;
@@ -268,6 +262,11 @@ fn generate_storage_entry_fns(
 
     let docs = &storage_entry.docs;
     let docs_token = quote! { #( #[doc = #docs ] )* };
+
+    let lifetime_param = match should_ref {
+        true => quote!(<'a>),
+        false => quote!()
+    };
     let client_iter_fn = if matches!(storage_entry.ty, StorageEntryType::Map { .. }) {
         quote! (
             #docs_token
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):
:+1:

codegen/src/api/storage.rs
```diff
@@ -300,21 +315,32 @@ fn generate_storage_entry_fns(
 
     let client_fns = quote! {
         #docs_token
-        pub async fn #fn_name(
+        pub fn #fn_name(
             &self,
             #( #key_args, )*
```
jsdw, line 320: Note to self; key args have a lifetime of 'a, which is the same as the lifetime of the client struct.

Should key args have a lifetime of 'b, and the returned future rely on both 'a and 'b, to avoid any unnecessary restrictions on how the future is used in conjunction with these references?

jsdw: I did a little testing and it seems OK to use the `'a` lifetime for key_args and client; from reading around I think Rust will pick the correct smallest lifetime that all params with `'a` use anyway, so no need to add other lifetimes.

codegen/src/api/storage.rs
```diff
@@ -267,19 +271,30 @@ fn generate_storage_entry_fns(
...
-                    metadata.storage_hash::<#entry_struct_ident>()?
-                };
-                if runtime_storage_hash == [#(#storage_hash,)*] {
-                    self.client.storage().iter(block_hash).await
-                } else {
-                    Err(::subxt::MetadataError::IncompatibleMetadata.into())
+            ) -> impl ::core::future::Future<
+                Output = ::core::result::Result<::subxt::KeyIter<'a, T, #entry_struct_ident #lifetime_param>, ::subxt::BasicError>
+            > + 'a {
+                // instead of an async fn which borrows all of self,
```
TarikGul, line 280:
```suggestion
                // Instead of an async fn which borrows all of self,
```

codegen/src/api/storage.rs
```diff
@@ -300,21 +315,32 @@ fn generate_storage_entry_fns(
...
-            };
-            if runtime_storage_hash == [#(#storage_hash,)*] {
-                let entry = #constructor;
-                self.client.storage().#fetch(&entry, block_hash).await
-            } else {
-                Err(::subxt::MetadataError::IncompatibleMetadata.into())
+        ) -> impl ::core::future::Future<
+            Output = ::core::result::Result<#return_ty, ::subxt::BasicError>
+        > + 'a {
+            // instead of an async fn which borrows all of self,
```
TarikGul, line 325:
```suggestion
            // Instead of an async fn which borrows all of self,
```

TarikGul (approved):

PR #565, opened 2022-06-14
Handle `StorageEntry` empty keys

lexnv:
There are cases when the `StorageEntry` has an empty key.

For example, defining a custom pallet 
```rust
#[pallet::config]
pub trait Config: frame_system::Config {
   type AssetNativeLocation: Parameter + Member + Default;
}

#[pallet::storage]
#[pallet::getter(fn location_assets)]
/// Local asset for native location.
pub type LocationAssets<T: Config> =
    StorageMap<_, Blake2_128Concat, T::AssetNativeLocation, bool, OptionQuery>;
```
and implementing it

```
impl pallet_asset_registry::Config for Runtime {
    type AssetNativeLocation = ();
```
results in `subxt` generating an extra lifetime parameter for a parameter that does not exist.
For more details see #552.

This PR ensures that no lifetimes are added when dealing with empty keys.

### Testing Done

Added the following patch to the [substrate-node-template](https://github.com/substrate-developer-hub/substrate-node-template).

```git
+++ b/pallets/template/src/lib.rs
@@ -24,6 +24,9 @@ pub mod pallet {
        pub trait Config: frame_system::Config {
                /// Because this pallet emits events, it depends on the runtime's definition of an event.
                type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+               /// AssetNativeLocation for () storage.
+               type AssetNativeLocation: Parameter + Member + Default + MaxEncodedLen;
        }

        #[pallet::pallet]
@@ -38,6 +41,12 @@ pub mod pallet {
        // https://docs.substrate.io/v3/runtime/storage#declaring-storage-items
        pub type Something<T> = StorageValue<_, u32>;

+       #[pallet::storage]
+       #[pallet::getter(fn location_assets)]
+       /// Local asset for native location.
+       pub type LocationAssets<T: Config> =
+               StorageMap<_, Blake2_128Concat, T::AssetNativeLocation, bool, OptionQuery>;
+
        // Pallets use events to inform users when important changes are made.
        // https://docs.substrate.io/v3/runtime/events-and-errors
        #[pallet::event]
diff --git a/runtime/src/lib.rs b/runtime/src/lib.rs
index 8a00054..4781e95 100644
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -266,6 +266,7 @@ impl pallet_sudo::Config for Runtime {
 /// Configure the pallet-template in pallets/template.
 impl pallet_template::Config for Runtime {
[...]

PR #567, opened 2022-06-15
Move test crates into a "testing" folder and add a ui (trybuild) test and ui-test helpers

jsdw:
- move test things into a `testing` folder to avoid polluting the root too much with test bits.
- add a `ui-tests` crate with utils to make it easier to generate metadata and use `trybuild` to check that it compiles properly (or not).
- add a few UI tests to prove the concept and test a couple of obvious bits.


jsdw:
This will fail until #565 is merged in (as one of the tests is for that failure), but is good to go otherwise.

testing/ui-tests/src/storage.rs
```diff
@@ -0,0 +1,42 @@
...
+    StorageEntryModifier,
+    StorageEntryType,
+};
+use scale_info::meta_type;
+
+use crate::utils::generate_metadata_from_storage_entries;
+
+/// Generate metadata which contains a `Map` storage entry with no hashers/values.
+/// This is a bit of an odd case, but it was raised in https://github.com/paritytech/subxt/issues/552,
+/// and this test will fail before the fix and should pass once the fix is applied.
```
TarikGul, line 29: Can we remove this comment, now that the fix is in?

testing/ui-tests/src/utils/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use scale_info::{
+    meta_type,
+    IntoPortable,
+    TypeInfo,
+};
+
+pub use metadata_test_runner::MetadataTestRunner;
+
+/// Given some pallet metadata, generate a [`RuntimeMetadataPrefixed`] struct.
+/// We default to a duff extrinsic type, and register a fake `DispatchError` type
```
TarikGul, line 37: What is "duff"?

TarikGul (approved):
👍

lexnv (approved):

testing/ui-tests/src/storage.rs
jsdw: I'm sortof keen to keep it, just to provide the historical reason for adding the test (ie "this test is actually useful because there was a bug in that area in the past")

PR #569, opened 2022-06-16
Add subscribe_beefy_justification api

DaviRain-Su:


niklasad1 (commented):
hey, 

The RPC API for `beefy_subscribeJustifications` looks good but can you please remove the octopus specific stuff in the example?

`beefy` should available on substrate anyway.

jsdw:
On the current live polkadot nodes I can only see a beefy.getFinalizedHead call, but if it's a thing, I don't hate the RPC call existing really as long as the other bits are removed (though since the RPC call just returns bytes by the looks of it, it doesn't offer much value (or help subxt internally) IMO, and using the raw client would suffice for now so that we don't commit to it in a public API without thinking throug the types better.

DaviRain-Su:
> On the current live polkadot nodes I can only see a beefy.getFinalizedHead call, but if it's a thing, I don't _hate_ the RPC call existing really as long as the other bits are removed.
> 
> That said, since the RPC call just returns bytes by the looks of it, it doesn't offer much value (or help subxt internally) IMO, and using the raw client would suffice for now so that we don't commit to it in a public API without coming up with more useful types.

I understand what you're saying, but providing a signedcommitment fetch api makes sense to me. beefy is also being used by the current light client authentication

DaviRain-Su:
> hey,
> 
> The RPC API for `beefy_subscribeJustifications` looks good but can you please remove the octopus specific stuff in the example?
> 
> `beefy` should available on substrate anyway.

Now there is no substrate or polkadot where we can see beefy being integrated, so I provided a template we modified to test beefy's functiona

jsdw:
I'll close this for now as it's lingered around a whille, and especially given the big subxt refactoring that has happened lately. Perhaps create an issue to discuss what you'd want and we can go from there? :)

PR #573, opened 2022-06-20
Expose the SCALE encoded call data of an extrinsic

### LINKED ISSUE #571: how can I get encoded but not signed call?
I want to send XCM with Transact. For that I need scale encoded call like that:
```rust
// 2. SCALE encode call data to bytes (pallet u8, call u8, call params).
        let call_data = {
            let mut bytes = Vec::new();
            let pallet = self.client.metadata().pallet(C::PALLET)?;
            bytes.push(pallet.index());
            bytes.push(pallet.call_index::<C>()?);
            self.call.encode_to(&mut bytes);
            Encoded(bytes)
        };
```

But there is no API to get it, nor call and client are public to allow to write my own trait

jsdw:
Closes #571

### DIFF
--- a/subxt/src/client.rs
+++ b/subxt/src/client.rs
@@ -355,9 +355,7 @@ where
     }
 
     /// Return the SCALE encoded bytes representing the call data of the transaction.
-    pub fn call_data(
-        &self
-    ) -> Result<Vec<u8>, BasicError> {
+    pub fn call_data(&self) -> Result<Vec<u8>, BasicError> {
         let mut bytes = Vec::new();
         let locked_metadata = self.client.metadata();
         let metadata = locked_metadata.read();
--- a/subxt/src/client.rs
+++ b/subxt/src/client.rs
@@ -354,6 +354,20 @@ where
             .await
     }
 
+    /// Return the SCALE encoded bytes representing the call data of the transaction.
+    pub fn call_data(
+        &self
+    ) -> Result<Vec<u8>, BasicError> {
+        let mut bytes = Vec::new();
+        let locked_metadata = self.client.metadata();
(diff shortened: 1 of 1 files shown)


subxt/src/client.rs
```diff
@@ -354,6 +354,20 @@ where
             .await
     }
 
+    /// Return the SCALE encoded bytes representing the call data of the transaction.
+    pub fn call_data(
+        &self
+    ) -> Result<Vec<u8>, BasicError> {
```
lexnv, line 360:
```suggestion
    pub fn call_data(&self) -> Result<Vec<u8>, BasicError> {
```

lexnv (approved):

subxt/src/client.rs
```diff
@@ -354,6 +354,18 @@ where
             .await
     }
 
+    /// Return the SCALE encoded bytes representing the call data of the transaction.
+    pub fn call_data(&self) -> Result<Vec<u8>, BasicError> {
+        let mut bytes = Vec::new();
+        let locked_metadata = self.client.metadata();
+        let metadata = locked_metadata.read();
```
ascjones, line 361: This pattern is very common to use the read only locked metadata, I wonder whether it is worth adding a method for that.

ascjones (approved):

subxt/src/client.rs
jsdw: Mmm yeah, quite possibly; it might also be worth just cloning it out of the lock immediately too; no need to actually hold the lock for any appreciable amount of time either I think!

ascjones: Cloning the metadata seems a bit wasteful in most cases since as you say it is usually short lived

PR #575, opened 2022-06-20
Change `subxt` license to MIT

lexnv:
This PR changes the `subxt`'s license to MIT.


jsdw:
Let's close this for now. I think if we do a change, it'll be to Apache 2, but I'm in discussion about this :)

PR #576, opened 2022-06-21
Decode raw events using scale_value and return the decoded Values, too

jsdw:
Remove the decode logic and rely on scale-value instead. Since that returns Value types which are more descriptive than just the event bytes, we forward those on in the RawEventDetails.

There will be a small hit to performance as a result of using scale_value's decode, since it doesn't just throw the bytes away but instead builds and returns Value types. I think the tradeoff of not having to have duplicate decode logic (and the scale_value decode stuff I think does the best job so far of catering for edge cases and bitvec handling and such) is worth it though.


ascjones (approved):
Awesome!

lexnv (approved):

PR #579, opened 2022-06-22
Add support for `ws` in `subxt-cli`

### LINKED ISSUE #512: wss support for subxt metadata
It would be nice if we could supply a websocket address rather than a url as it's easy to get the ws address of parachain nodes using polkadot-js. (we can just select any parachain and then switch to custom to see what ws address we're looking at)

```
subxt metadata --url wss://kusama.public.curie.radiumblock.co/ws
```

Figuring out what the equivalent http url is seem more hit and miss... it would be great if subxt metadata worked wtih ws/wss as well as http. (This is assuming that we can get the metadata from a websocket - I think we can?)

lexnv:
This PR adds support for the `cli` tool to connect to substrate binaries via WebSockets,
while keeping the compatibility with `http`.

Closes #512 .

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -27,10 +27,7 @@ use frame_metadata::{
     META_RESERVED,
 };
 use jsonrpsee::{
-    async_client::{
-        Client,
-        ClientBuilder,
-    },
+    async_client::ClientBuilder,
     client_transport::ws::{
         Uri,
         WsTransportClientBuilder,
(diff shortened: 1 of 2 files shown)


niklasad1:
any numbers how much the dependency tree grew by this change?

cli/src/main.rs
```diff
@@ -173,28 +191,26 @@ fn handle_pallet_metadata(nodes: &[url::Url], name: &str) -> color_eyre::Result<
...
+        let metadata = fetch_runtime_metadata(node).await?;
 
         match metadata.pallets.iter().find(|pallet| pallet.name == name) {
             Some(pallet_metadata) => {
                 let hash = get_pallet_hash(&metadata.types, pallet_metadata);
                 let hex_hash = hex::encode(hash);
                 println!(
                     "Node {:?} has pallet metadata hash {:?}",
-                    node.as_str(),
+                    node.to_string(),
```
niklasad1, line 202: why?

niklasad1 (approved):

jsdw (approved):

Issue #591, opened 2022-06-29
Stability of metadata.scale format


alistair-singh:
We have an issue where by the same build of our parachain on two different machines produced `metadata.scale` files with different hashes. This has caused our application built with `subxt-codegen` to fail due to `Error: Metadata(IncompatibleMetadata)`. I understand this error is produced when the hash of the metadata that was used to generate the client is not the same as the metadata on the parachain node. Getting the latest metadata from the parachain indeed shows a difference between the hashes.

Originally we thought that the metadata was changed so we updated the `metadata.scale` file. This fixed the issue on our side, but on some machines the metadata was still out of date.

I have attached a diff of the metadata files and the generated rust sources. 
https://drive.google.com/file/d/1-TtoQkEZCTCoNWqtz1rkVv3K3GEZd7K3/view?usp=sharing

The parachain code was identical in both builds.
* ours.metadata.scale - built on Ubuntu 20.04 AWS instance
* theirs.metadata.scale - built on Ubuntu 22.04 AWS instance

Code used:
[Snowbridge](https://github.com/Snowfork/snowbridge) -> `a65d9118dd4b2277eb7a8513c6f9d7273f277fc2`
Polkadot -> `a7e188cd9665c735f4b9d5a58cdbc4dd1850eae0` - `release-v0.9.23`

Looking at the diff file `diff.xxd.1col.patch` it does not look like new metadata was added. It looks like it was scale encoded differently.

Taking a look at the first difference:
```diff
--- ours.1col.xxd       2022-06-23 13:38:08.405761839 +0000
+++ theirs.1col.xxd     2022-06-23 13:38:21.021826697 +0000
@@ -500,28 +500,14 @@
 00
 04
 01
-58
+20
 5b
 75
 38
 3b
 20
-2f
-2a
-c2
-ab
-2a
-2f
-20
 33
 32
-20
-2f
-2a
-c2
-bb
-2a
-2f
 5d
 00
 00
```

ours.xxd:
```
000001c0: 5400 0124 6d61 6e64 6174 6f72 7920 0104  T..$mandatory ..
000001d0: 5400 0020 0000 0506 0024 083c 7072 696d  T.. .....$.<prim
000001e0: 6974 6976 655f 7479 7065 7310 4832 3536  itive_types.H256
000001f0: 0000 0400 0401 585b 7538 3b20 2f2a c2ab  ......X[u8; /*..
[...]

PR #595, opened 2022-07-06
metadata: Exclude `field::type_name` from metadata validation

lexnv:
The #591 presented a metadata validation issue detected on the [Snowbridge](https://github.com/Snowfork/snowbridge) parachain.

The metadata validation would reject extrinsics due to the following difference in metadata:

| Source|  Metadata |
| --- | --- |
| snowbridge | "typeName": "/*«*/ u32 /*»*/" |
| polkadot | "typeName": "u32"|

This PR removes the hashing of the `typeName` as the unique deterministic representation of the field is deduced from its `type` and `name`. 

### Testing
Utilizing the metadata provided in the issue:

```bash
subxt-cli -- codegen --file ours.metadata.scale > ours.rs
subxt-cli -- codegen --file theirs.metadata.scale > theirs.rs

diff ours.rs theirs.rs
#ok - identical hashes for validation
```

### DIFF
--- a/codegen/src/types/type_def_params.rs
+++ b/codegen/src/types/type_def_params.rs
@@ -65,7 +65,7 @@ impl TypeDefParameters {
     }
 }
 
-impl<'a> quote::ToTokens for TypeDefParameters {
+impl quote::ToTokens for TypeDefParameters {
     fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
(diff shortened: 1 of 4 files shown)


TarikGul (approved):
Interesting, I never really thought about it but omitting the check for `type_name` makes sense since you can do validation through `type`, and `name` itself as @lexnv mentioned. 

Nice job 👍

PR #607, opened 2022-08-02
Use scale-decode to help optimise event decoding

jsdw:
Skip over event bytes using the `IgnoreVisitor` from `scale_decode`, and only decode fields into `Value`s if we ask for them explicitly.

Also avoid cloning event bytes around the place by reusing the same Vec (`Arc<[u8]>`) everywhere and avoiding any string cloning (again, pushing the work until the user asks for pallet/variant name), and tidy the API around working with events a little.

This builds off the big API change of #593.


lexnv (approved):

ascjones (approved):
LGTM

PR #623, opened 2022-08-17
use 'sp_core::Hxxx' for all hash types

### LINKED ISSUE #619: Autogenerated  code uses different libraries for `H160` vs `H256`
When autogenerating bindings for pallet structs that use `primitive_types::H160` and `primitive_types::H256` subxt codegen ends up using different libraries for `H160` vs `H256` in the generated code.

`subxt` generated code shows the following when running `cargo expand`:
- H160: `runtime_types::primitive_types::H160`
- H256: `::subxt::sp_core::H256`

The fact that codegen uses `primitive_types` for `H160` and `sp_core` for `H256` is causing type conflicts. Would like for both to use `sp_core` instead.

jsdw:
Since we have a type override for H256 already, also add one for H160 and H512

Closes #619

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -120,8 +120,16 @@ impl RuntimeGenerator {
                 parse_quote!(::subxt::ext::sp_core::crypto::AccountId32),
             ),
             (
+                "primitive_types::H160",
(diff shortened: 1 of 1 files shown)


lexnv (approved):
LGTM!

niklasad1 (approved):

PR #633, opened 2022-08-23
Specialize incompatible metadata errors

### LINKED ISSUE #584: Better error message for incompatible call metadata
User reported an [error](https://github.com/paritytech/cargo-contract/issues/616) with what looks like an incompatible call, for which we receive the following generic error message:

`Node metadata is not fully compatible`

From this error variant: https://github.com/paritytech/subxt/blob/e3732e354f846a9aa807a9d971b1412c902c43ed/subxt/src/metadata/metadata_type.rs#L81, raised here https://github.com/paritytech/subxt/blob/0d67b80cc5cb758c8e3e089a1b77b4207ef0e77c/codegen/src/api/calls.rs#L149.

What would be better is something with more context e.g:

`Metadata for the "instantiate" call is incompatible with the target node` or similar. 

Possibly we already discussed this issue in the original PR but I can't find anything to link to.

lexnv:
This PR specializes the metadata errors to contain information
about the incompatible components, for static validation purposes.

Closes #584.

### DIFF
--- a/subxt/src/constants/constants_client.rs
+++ b/subxt/src/constants/constants_client.rs
@@ -47,7 +47,11 @@ impl<T: Config, Client: OfflineClientT<T>> ConstantsClient<T, Client> {
                 .metadata()
                 .constant_hash(address.pallet_name(), address.constant_name())?;
             if actual_hash != expected_hash {
-                return Err(MetadataError::IncompatibleMetadata.into())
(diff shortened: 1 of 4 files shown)


ascjones (approved):
Yes! :heart:

jsdw (approved):
Whoop, nice!

niklasad1 (approved):

PR #634, opened 2022-08-24
Allow generalising over RPC implementation

### LINKED ISSUE #596: Light client support / generalizing over RPC provider
I'm playing around with [smoldot](https://github.com/paritytech/smoldot) and trying to combine it with subxt. The smoldot light-base library gives access to RPCs against the node.

In principle, this should work fine with subxt except for the fact that subxt directly uses the `jsonrpsee::Client` and `ClientBuilder` which has some deeply-baked assumption about using websockets (as seen in the [`TransportSenderT`](https://docs.rs/jsonrpsee-core/latest/jsonrpsee_core/client/trait.TransportSenderT.html)).

If not for that, it would be relatively straightforward to build the `jsonrpsee::Client` with custom transport types. However, I think this may be better addressed at the `subxt` level by having the `subxt::rpc::RpcClient` be generic over an `R: jsonrpsee::ClientT + jsonrpsee::SubscriptionClientT` which would allow us to implement a shim over a smoldot light-client.

This unfortunately does need to add a generic parameter instead of a trait object because the `ClientT` traits aren't object-safe as they contain generic functions. There is a possibility to do an upstream contribution to `jsonrpsee` to introduce `UntypedClientT` and `UntypedSubscriptionClientT` traits which are object-safe and over which you can build a decoding wrapper. That would avoid the extra generic parameter at the cost of virtual calls, which would be dominated by network utilization anyway.

jsdw:
Make jsonrpsee optional, and generalise the underlying RPC mechanism over any clients that implement the new `RpcClientT` trait we expose.

The `RpcClientT` trait has been made to be "low level", as in order to be object-safe it cannot have generic parameters and such. We want it to be object-safe to avoid the proliferation of some `Rpc` trait bound everywhere (it turns out that even things like `OfflineClientT`, which should not care at all, would need it, because it can consturct other types which would need it).

`jsonrpsee` is still used by default (unless the feature is disabled). Currently it's impl of  this trait is somewhat clunky, but https://github.com/paritytech/jsonrpsee/issues/862 should resolve that.

Closes #596


subxt/src/client/online_client.rs
```diff
@@ -68,16 +70,20 @@ impl<T: Config> OnlineClient<T> {
...
     }
+}
 
-    /// Construct a new [`OnlineClient`] by providing the underlying [`RpcClient`]
-    /// to use to drive the connection.
-    pub async fn from_rpc_client(
-        rpc_client: impl Into<RpcClient>,
+impl<T: Config> OnlineClient<T> {
+    /// Construct a new [`OnlineClient`] by providing an underlying [`RpcClientT`]
+    /// implementation to use to drive the connection.
```
lexnv, line 82:
```suggestion
    /// implementation to drive the connection.
```

subxt/src/rpc/rpc_client_t.rs
```diff
@@ -0,0 +1,46 @@
...
+/// the caller. This is the case because we want the methods to be object-safe (which prohibits
+/// generics), and want to avoid any unnecessary allocations in serializing/deserializing
+/// parameters.
+pub trait RpcClientT: Send + Sync + 'static {
+    /// Make a raw request for which we expect a single response back from. The params will
+    /// be provided in the form of a pre-encoded JSON array.
+    fn request_raw<'a>(
+        &'a self,
+        method: &'a str,
+        params: Box<RawValue>,
```
lexnv, line 27: nit: Can we also receive no parameters here?

lexnv (approved):

subxt/src/rpc/rpc_client_t.rs
niklasad1: I think the intention is that you are supposed to use the `rpc_params` macro that James wrote which takes care of that for you.

However, I agree that creating a `RawValue` from the an empty `JSON String` is quite awkward because you need to allocate a String for it.

subxt/src/rpc/rpc_client_t.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::{
+    future::Future,
+    pin::Pin,
+};
+
+/// Any RPC client which implements this can be used in our [`super::Rpc`] type
+/// to talk to a node.
+///
+/// This is a low level interface which expects an already-serialized set of params,
+/// and returns an owned but still-serialized [`RawValue`], deferring deserialization to
```
niklasad1, line 17: RawValue is quite hard to deal with in general and it would be good if you could point to `RpcParams` for an example implementation of this.

My noob eyes find it hard to look at this trait along with the docs of `RawValue` to understand how to do it

subxt/Cargo.toml
```diff
@@ -13,11 +13,17 @@ description = "Submit extrinsics (transactions) to a substrate node via RPC"
...
+default = ["jsonrpsee"]
+
 # Activate this to expose functionality only used for integration testing.
 # The exposed functionality is subject to breaking changes at any point,
 # and should not be relied upon.
 integration-tests = []
 
+# Jsonrpsee if the default RPC provider used in Subxt. However, it can be
+# swapped out for an alternative implementation, and so is optional.
+jsonrpsee = ["dep:jsonrpsee"]
```
niklasad1, line 25: why is this feature needed at all?

`jsonrpsee` is already an optional dep?

subxt/src/rpc/rpc_client_t.rs
jsdw: I have just tweaked these things to accept `Option<Box<RawValue>>`, to allow subxt to avoid the unnecessary allocations when there are no params :)

subxt/src/rpc/rpc_client_t.rs
jsdw: That makes sense; I'll add some more examples to show how these methods can be called manually!

subxt/Cargo.toml
jsdw: I literally just added it so that there was a place to add a comment about the feature above it (and also I quite like when all of the features are clearly listed in one place in the file, rather than needing to spot which dependencies are optional; it feels a bit clearer to me :)

subxt/src/rpc/rpc_client_t.rs
jsdw: (I just ended up pointing to RpcParams in the docs to show examples of building params)

niklasad1 (approved):

PR #636, opened 2022-08-25
Add conversion and default functions for `NumberOrHex`

cmichi:
Needed for https://github.com/paritytech/ink/issues/1234.

Shamelessly copied from `sp-rc`, which I guess you wanted to avoid as a dep.

### DIFF
--- a/subxt/src/rpc.rs
+++ b/subxt/src/rpc.rs
@@ -177,7 +177,8 @@ impl From<U256> for NumberOrHex {
 
 /// An error type that signals an out-of-range conversion attempt.
 #[derive(Debug, thiserror::Error)]
-pub struct TryFromIntError(pub(crate) ());
+#[error("Out-of-range conversion attempt")]
+pub struct TryFromIntError();
 
 impl TryFrom<NumberOrHex> for u32 {
     type Error = TryFromIntError;
@@ -185,7 +186,7 @@ impl TryFrom<NumberOrHex> for u32 {
         num_or_hex
             .into_u256()
             .try_into()
-            .map_err(|_| TryFromIntError(()))
+            .map_err(|_| TryFromIntError())
     }
 }
 
@@ -195,7 +196,7 @@ impl TryFrom<NumberOrHex> for u64 {
         num_or_hex
             .into_u256()
             .try_into()
-            .map_err(|_| TryFromIntError(()))
+            .map_err(|_| TryFromIntError())
     }
 }
 
@@ -205,7 +206,7 @@ impl TryFrom<NumberOrHex> for u128 {
         num_or_hex
             .into_u256()
             .try_into()
-            .map_err(|_| TryFromIntError(()))
+            .map_err(|_| TryFromIntError())
     }
 }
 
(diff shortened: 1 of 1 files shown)


subxt/src/rpc.rs
```diff
@@ -135,6 +135,85 @@ impl From<NumberOrHex> for BlockNumber {
...
+}
+
+impl From<U256> for NumberOrHex {
+    fn from(n: U256) -> Self {
+        NumberOrHex::Hex(n)
+    }
+}
+
+/// An error type that signals an out-of-range conversion attempt.
+pub struct TryFromIntError(pub(crate) ());
```
ascjones, line 179:
```suggestion
#[derive(thiserror::Error)]
pub struct TryFromIntError(pub(crate) ());
```
Then will integrate nicely with error libs and `?`. Also maybe some way to not swallow the actual inner error here, which I assumes would be an overflow.

subxt/src/rpc.rs
```diff
@@ -135,6 +135,86 @@ impl From<NumberOrHex> for BlockNumber {
...
+
+impl From<U256> for NumberOrHex {
+    fn from(n: U256) -> Self {
+        NumberOrHex::Hex(n)
+    }
+}
+
+/// An error type that signals an out-of-range conversion attempt.
+#[derive(thiserror::Error)]
+pub struct TryFromIntError(pub(crate) ());
```
jsdw, line 180: Is the `pub(crate) ()` inner type just intended to allow constructing it within subxt but prevent people creating it externally? Personally I'd scrap that if so; I don't see why preventing others from making the error would help?

subxt/src/rpc.rs
```diff
@@ -639,6 +719,19 @@ fn to_hex(bytes: impl AsRef<[u8]>) -> String {
 mod test {
     use super::*;
 
+    /// A util function to assert the result of serialization and deserialization is the same.
+    #[cfg(test)]
```
jsdw, line 723: No need for this since it's in a module with `#[cfg(test)]` already :)

jsdw:
Aside from a couple of small niggles it looks good to me!

subxt/src/rpc.rs
```diff
@@ -176,16 +176,17 @@ impl From<U256> for NumberOrHex {
 }
 
 /// An error type that signals an out-of-range conversion attempt.
-#[derive(thiserror::Error)]
-pub struct TryFromIntError(pub(crate) ());
+#[derive(Debug, thiserror::Error)]
+#[error("Out-of-range conversion attempt")]
+pub struct TryFromIntError();
```
ascjones, line 181:
```suggestion
pub struct TryFromIntError;
```
Redundant `()`, would need to fix usages

ascjones (approved):

subxt/src/rpc.rs
cmichi: Thx, interesting that `clippy` doesn't flag it.

jsdw (approved):
LGTM!

Issue #637, opened 2022-08-25
Support serializing/deserializing StaticTxPayload


trusch:
I'd like to create an extrinsic, serialize and transmit to another computer and then deserialize it again to submit it to the chain.

If I take for example this `0x000110deadbeef` (which is what polkadot.js gives me for system::remark(0xdeadbeef)) I'd like to parse it into a StaticTxPayload so I can submit it.

Is this possible?

trusch:
I tried this, but it gives me an error when submitting the extrinsic:
```rust
struct RawCall(Vec<u8>);

impl TxPayload for RawCall {
    fn pallet_name(&self) -> &str {
        ""
    }

    fn call_name(&self) -> &str {
        ""
    }

    fn encode_call_data(
            &self,
            _: &subxt::Metadata,
            out: &mut Vec<u8>,
        ) -> Result<(), subxt::Error> {
        out.extend_from_slice(&self.0);
        Ok(())
    }
}

...

let call_data = "0x000110deadbeef";
let call_bs = hex::decode(call_data.trim_start_matches("0x"))?;    
let payload = RawCall(call_bs);
let progress = cli.tx().sign_and_submit_then_watch_default(&payload, &signer).await.map_err(|err|Error::Chain(err))?;
```

Error is this:
```
Chain(Rpc(Call(Custom(ErrorObject { code: ServerError(1002), message: "Verification Error: Runtime error: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\n\n    0: 0x3ef3c - <unknown>!rust_begin_unwind\n    1: 0x3995 - <unknown>!core::panicking::panic_fmt::h5118e89563022e7e\n    2: 0x259f9a - <unknown>!TaggedTransactionQueue_validate_transaction\n", data: Some(RawValue("Runtime error: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\n\n    0: 0x3ef3c - <unknown>!rust_begin_unwind\n    1: 0x3995 - <unknown>!core::panicking::panic_fmt::h5118e89563022e7e\n    2: 0x259f9a - <unknown>!TaggedTransactionQueue_validate_transaction\n")) }))))
```

jsdw:
I'll have a look into this, because it makes me wonder whether the `TxPayload` interface could be tweaked. Lemme get back to you!

jsdw:
[...]

PR #638, opened 2022-08-25
Simplify the TxPayload trait a little

jsdw:
Most custom implementations will not do static validation, and for these, there's no need to require that a pallet and call name can be handed back. This tweak to the trait signature makes that clear.

I also exposed a way to create a `SubmittableExtrinsic` without going through the previous steps, in case somebody has bytes for a pre-signed extrinsic that they'd like to submit.

In response to #637 (though I think their issue is unrelated, it'll help with this sort of thing)

### DIFF
--- a/subxt/src/tx/tx_client.rs
+++ b/subxt/src/tx/tx_client.rs
@@ -114,11 +114,10 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
         };
 
         // Wrap in Encoded to ensure that any more "encode" calls leave it in the right state.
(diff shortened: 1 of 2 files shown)


TarikGul (approved):
LGTM 👍 , From what I see this PR replaces using `call` to abstract information with `call.validation_details` so that all information is in a single struct (which i definitely like).

jsdw:
Yeah; I realised that the call and pallet name were only ever used in validation so wanted to make it clear that they didn't need implementing/werent used otherwise :)

lexnv (approved):

PR #642, opened 2022-08-30
feat: add `subscribe_all_blocks` and rename `subscribe_blocks` to `subscribe_new_blocks`

yjhmelody:


jsdw:
I need to look up what the difference is between `chain_subscribeNewHeads` vs `chain_subscribeAllHeads` is?

Looks ok otherwise (though I may cause a merge conflict with an Rpc PR I'll be merging shortly!)

I would note though that the RPC interface will eventually be reworked in Substrate, and so I'm growing hesitant to keep adding RPC methods to this underlying Rpc interface that Subxt doesn't actually use itself. Thus, a better approach I reckon would be that you use the underlying Rpc client yourself to make whatever underlying calls you want.

Future versions of Subxt will expose an interfacte to the Runtime APIs directly and automatically, and have a much smaller and more static set of RPC calls to expose, which will be a bunch nicer I reckon :)

lexnv:
The substrate's [chain_subscribeNewHeads](https://github.com/paritytech/substrate/blob/e77cbe39c4d1bfd978bb03c686fc9f60d86a3d06/client/rpc/src/chain/chain_full.rs#L86) returns the header of the block if the block is considered the new best block. While the [chain_subscribeAllHeads](https://github.com/paritytech/substrate/blob/e77cbe39c4d1bfd978bb03c686fc9f60d86a3d06/client/rpc/src/chain/chain_full.rs#L72) returns the header of new blocks, regardless if they are the best blocks or not. 

We are working towards having a new unified RPC API for the substrate's node, that will simplify all the `subxt/src/rpc` methods 😄 While this PR looks solid, I would postpone it for a bit until we have the new API landing in substrate.

jsdw:
Perhaps your `BlocksClient` PR can address this stuff in part anyway.

But @yjhmelody I'd like to know what you need this for?

But yeah; in general I'd rather not add RPC methods at the moment if Subxt doesn't need them because eventually we'll move over to a new RPC interface anyway and then all of these methods are liable to changing.

jsdw:
This will be addressed in #691 so I'll close this now; the result will be that a new `blocks` API will support subscribing to finalized, all or best block (headers), from which yo ucan then get events/extrinsics etc. The udnerlying RPc method to support this was also added and names tidied.

PR #651, opened 2022-09-14
Fix codegen for `codec::Compact` as type parameters

ascjones:
Fixes https://github.com/paritytech/cargo-contract/issues/734.

When a `codec::Compact` type appears as a type parameter e.g. `Option<Compact<u128>>` as it does in the contracts dispatchables e.g. https://github.com/paritytech/substrate/blob/4b7911607b48a5ca7ebf1d655f12d91cdd241866/frame/contracts/src/lib.rs#L498, then this will incorrectly result in a `Option<u128>` type being generated by `subxt-codegen`.

This is because the current codegen assumes that the compact value will be the type of the field directly and can generate the following:

`struct S { #[codec(compact)] field: u128 }`

The `compact` attribute will cause the conversion to be handled during encoding and decoding of the struct. This means that the "inner" type (in this case `u128`) can be used directly.

However when the compact type is a type parameter then it is required to use the `codec::Compact<T>` wrapper type directly instead of the field level attribute.

## Refactoring

This PR also incorporates some refactoring of the type path codegen, specifically expanding the `TypePathType` to a strongly typed enum which maps from the `scale_info::TypeDef` but with all the information necessary to generate the hierarchical type paths.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -57,8 +57,6 @@ skip_children = false
 hide_parse_errors = false
 error_on_line_overflow = false
 error_on_unformatted = false
-report_todo = "Always"
-report_fixme = "Always"
 ignore = []
 
 # Below are `rustfmt` internal settings
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -9,7 +9,6 @@ format_code_in_doc_comments = false
 comment_width = 80
 normalize_comments = true # changed
 normalize_doc_attributes = false
-license_template_path = "FILE_TEMPLATE" # changed
 format_strings = false
 format_macro_matchers = false
 format_macro_bodies = true
--- a/codegen/src/types/tests.rs
+++ b/codegen/src/types/tests.rs
@@ -372,6 +372,44 @@ fn compact_fields() {
 }
 
 #[test]
+fn compact_generic_parameter() {
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    struct S {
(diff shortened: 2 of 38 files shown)


.rustfmt.toml
```diff
@@ -9,7 +9,6 @@ format_code_in_doc_comments = false
 comment_width = 80
 normalize_comments = true # changed
 normalize_doc_attributes = false
-license_template_path = "FILE_TEMPLATE" # changed
```
jsdw, line 12: I never really noticed but does removing this mean no checking that the license header is at the top of files now?

codegen/src/types/tests.rs
```diff
@@ -371,6 +371,43 @@ fn compact_fields() {
     )
 }
 
+#[test]
+fn compact_generic_parameter() {
+    #[allow(unused)]
+    #[derive(TypeInfo)]
+    struct S {
+        a: Option<<u128 as codec::HasCompact>::Type>,
+    }
```
jsdw, line 380: Is it worth adding tests to check that some of the other TypePath variants lead to the output you expect?

jsdw (approved):
Look good; tidy!

.rustfmt.toml
ascjones: No it has been removed now so it doesn't do anything other than give a warning https://github.com/rust-lang/rustfmt/issues/5103

lexnv (approved):

PR #654, opened 2022-09-15
Store type name of a field in event metadata

SkymanOne:
Fields in `EventMetadata` now also store name of a type specified in the pallet's source code.

Required for [#673](https://github.com/paritytech/cargo-contract/issues/673) in `cargo-contract`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.23.1] - 2022-09-15
-
-Fields in `EventMetadata` now also store name of a type specified in the pallet's source code.
-
 ## [0.23.0] - 2022-08-11
 
 This is one of the most significant releases to date in Subxt, and carries with it a number of significant breaking changes, but in exchange, a number of significant improvements. The most significant PR is [#593](https://github.com/paritytech/subxt/pull/593); the fundamental change that this makes is to separate creating a query/transaction/address from submitting it. This gives us flexibility when creating queries; they can be either dynamically or statically generated, but also flexibility in our client, enabling methods to be exposed for online or offline use.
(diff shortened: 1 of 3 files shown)


subxt/src/metadata/metadata_type.rs
```diff
@@ -304,7 +304,7 @@ pub struct EventMetadata {
     // behind an Arc to avoid lots of needless clones of it existing.
     pallet: Arc<str>,
     event: String,
-    fields: Vec<(Option<String>, u32)>,
+    fields: Vec<(Option<String>, Option<String>, u32)>,
```
ascjones, line 307: I think the introduction of a `struct` here e.g. `EventFieldMetadata` to disambiguate the fields would be good.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.23.1] - 2022-09-15
+
+Fields in `EventMetadata` now also store name of a type specified in the pallet's source code.
+
```
jsdw, line 10:
```suggestion
```

Let's remove these; Love the general idea, but there are a bunch of commits that aren't listed, and we don't want to commit to a release date :)

jsdw:
LGTM aside from my tiny nit and Andrew's suggestion!

CHANGELOG.md
ascjones: Could add an `[unnreleased]` section at the top as per convention

jsdw: I pondered that briefly, but it only makes sense if the other commits are also listed, else it could potentially lead to things being missed!

Might be worth starting to do that at some point though for sure!

subxt/src/metadata/metadata_type.rs
```diff
@@ -457,7 +490,13 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
                             fields: variant
                                 .fields()
                                 .iter()
-                                .map(|f| (f.name().map(|n| n.to_owned()), f.ty().id()))
+                                .map(|f| {
+                                    EventFieldMetadata::new(
```
lexnv, line 494: nit: Maybe we could use something like `impl From<Field<PortableForm>>` here

lexnv (approved):
LGTM! Thanks for contributing! 👍

ascjones (approved):

Issue #652, opened 2022-09-14
runtime upgrades API: provide low-level API


niklasad1:
Currently, it's only possible perform runtime upgrades as follows:

```rust
tokio::spawn(async move {
	let update_client = api2.subscribe_to_updates();

	// This will run runtime updates in a loop which only returns if the RPC subscription is terminated.
	if let Err(e) = match update_client.perform_runtime_updates().await {
	    log::error!(target: LOG_TARGET, "Runtime update failed with result: {:?}", e);
		let _ = tx_runtime.send(e);
	}
});
```

subxt spawns a loop performs upgrades under the hood but I would want to know when a runtime upgrade occurred to provide some kind of metrics for it.

I think the current API is good but having a low-level API or some `impl Stream` wrapper that fires every time a runtime upgrade occurs.

jsdw:
Makes sense to me to have a Stream that maybe has the version info in and that `perform_runtime_updates` can be a super simple wrapper around!

PR #657, opened 2022-09-20
feat: add low-level `runtime upgrade API`

niklasad1:
Closes https://github.com/paritytech/subxt/issues/652

### DIFF
--- a/examples/examples/subscribe_runtime_updates.rs
+++ b/examples/examples/subscribe_runtime_updates.rs
@@ -12,7 +12,6 @@
 
 use std::time::Duration;
 use subxt::{
-    client::UpgradeResult,
     OnlineClient,
     PolkadotConfig,
 };
@@ -31,21 +30,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         let mut update_stream = updater.runtime_updates().await.unwrap();
 
         while let Some(Ok(update)) = update_stream.next().await {
-            let version = update.runtime_version.spec_version;
+            let version = update.runtime_version().spec_version;
 
-            match updater.apply_update(update).await {
-                Ok(UpgradeResult::Success) => {
+            match updater.apply_update(update) {
+                Ok(()) => {
                     println!("Upgrade to version: {} successful", version)
                 }
-                Ok(reason) => {
-                    println!("Upgrade to version: {} failed {:?}", version, reason)
-                }
                 Err(e) => {
-                    println!(
-                        "Upgrade failed {:?} (the websocket connection is probably gone)",
-                        e
-                    );
-                    return
+                    println!("Upgrade to version {} failed {:?}", version, e);
                 }
             };
         }
--- a/examples/examples/subscribe_runtime_updates.rs
+++ b/examples/examples/subscribe_runtime_updates.rs
@@ -11,10 +11,7 @@
 //! ```
 
 use std::time::Duration;
-use subxt::{
-    OnlineClient,
-    PolkadotConfig,
-};
+use subxt::{client::UpgradeResult, OnlineClient, PolkadotConfig};
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -27,12 +24,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // utilizing the API for other use cases.
     let updater = api.subscribe_to_updates();
     tokio::spawn(async move {
-        let mut update_stream = updater.stream().await.unwrap();
+        let mut update_stream = updater.runtime_updates().await.unwrap();
 
-        while let Some(Ok(new_runtime_version)) = update_stream.next().await {
-            if updater.is_runtime_version_different(&new_runtime_version) {
-                updater.do_update(new_runtime_version).await.unwrap();
-            }
+        while let Some(Ok(update)) = update_stream.next().await {
+            let version = update.runtime_version.spec_version;
+
+            match updater.apply_update(update).await {
+                Ok(UpgradeResult::Success) => {
+                    println!("Upgrade to version: {} successful", version)
+                }
+                Ok(reason) => {
+                    println!("Upgrade to version: {} failed {:?}", version, reason)
+                }
+                Err(e) => {
+                    println!(
+                        "Upgrade failed {:?} (the websocket connection is probably gone)",
+                        e
+                    );
+                    return;
+                }
+            };
         }
     });
 
--- a/examples/examples/subscribe_runtime_updates.rs
+++ b/examples/examples/subscribe_runtime_updates.rs
@@ -25,10 +25,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Start a new tokio task to perform the runtime updates while
     // utilizing the API for other use cases.
-    let update_client = api.subscribe_to_updates();
+    let updater = api.subscribe_to_updates();
     tokio::spawn(async move {
-        let result = update_client.perform_runtime_updates().await;
-        println!("Runtime update failed with result={:?}", result);
+        let mut update_stream = updater.stream().await.unwrap();
+
+        while let Some(Ok(new_runtime_version)) = update_stream.next().await {
+            if updater.is_runtime_version_different(&new_runtime_version) {
+                updater.do_update(new_runtime_version).await.unwrap();
+            }
+        }
     });
 
     // If this client is kept in use a while, it'll update its metadata and such
--- a/examples/examples/subscribe_runtime_updates.rs
+++ b/examples/examples/subscribe_runtime_updates.rs
@@ -11,7 +11,11 @@
 //! ```
 
 use std::time::Duration;
-use subxt::{client::UpgradeResult, OnlineClient, PolkadotConfig};
+use subxt::{
+    client::UpgradeResult,
+    OnlineClient,
+    PolkadotConfig,
+};
 
 #[tokio::main]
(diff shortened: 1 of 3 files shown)


examples/examples/subscribe_runtime_updates.rs
```diff
@@ -25,10 +25,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Start a new tokio task to perform the runtime updates while
     // utilizing the API for other use cases.
-    let update_client = api.subscribe_to_updates();
+    let updater = api.subscribe_to_updates();
```
niklasad1, line 28: I'll revert this later but just an example so you get the point :)

subxt/src/client/online_client.rs
```diff
@@ -204,20 +222,35 @@ impl<T: Config> OnlineClientT<T> for OnlineClient<T> {
...
         let curr = self.0.inner.read();
         &curr.runtime_version != new
     }
 
+    pub async fn do_update(
+        &self,
+        new_runtime_version: RuntimeVersion,
+    ) -> Result<(), Error> {
+        // Fetch new metadata.
+        let new_metadata = self.0.rpc.metadata().await?;
```
jsdw, line 235: I guess it's possible that the RuntimeVersion passed in becomes out of sync with the metadata that's obtained, since you can hold onto some RuntimeVersion for a while and then later on you can call `do_update` with that, and it goes and sets it without really checking anything, and fetches the _latest_ metadata to go along with it.

What would you think about fetching metadata in `perform_runtime_updates` so eg:

```rust
struct Update {
    version: RuntimeVersion,
    metadata: Metadata
}

// as now; the "easy" thing.
async fn perform_runtime_updates(&self) -> result<(), Error>;
// return a subscription which returns update payloads as updates are required.
async fn runtime_updates(&self) -> Result<Subscription<Update>, Error>;
// apply some Update payload.
fn apply_update(&self, update: Update) -> Result<(), Error>;
 ```

ie keep metadata and version bundled to try and ensure things will always be consistent?

Comment (jsdw) [reply subxt/src/client/online_client.rs:235]: Example usage might then just be:

```rust
while let Some(Ok(update)) = updater.runtime_updates() {
    updater.apply_update(update)
}
```

(and I guess instead of exposing "is_update_different" you could filter the returned Subscription to do that for you perhaps? Hmm. Because if you've already obtained metadata then doing this check is a bit late. So I see why you obtained it separately.. hmmm)

niklasad1: > It's only ever not different for the first result though I think

Yes, because on startup the runtime version is fetched by `subxt` and then when the subscription is created then the node will send out the initial value 

I think 3) is my cup of tea here your proposed API looks a bit nicer IMO

jsdw: > Yes, because on startup the runtime version is fetched by subxt and then when the subscription is created then the node will send out the initial value

Oh I mean as in, when you subscribe to runtime updates you always get the current version as the first message that comes back, so it's almost always equal :)

3) sounds good to me :)

niklasad1: > Oh I mean as in, when you subscribe to runtime updates you always get the current version as the first message that comes back, so it's almost always equal :)

Yes, that was my point as subxt fetches the current runtime version on startup and once you subscribe to the runtime version then Substrate sends out the initial value of the runtime version on each subscription and those will be equal "all the time in practice" when a new subscription is created. After that I don't expect these numbers to be equal.

I ended up with 1) anyway I don't like having to use the `RwLock` in the `stream impl` let's be naive for now.

subxt/src/client/online_client.rs
```diff
@@ -209,34 +227,120 @@ impl<T: Config> ClientRuntimeUpdater<T> {
...
+    ///
+    /// Returns `Err(Error)` if the metadata couldn't be fetched.
+    pub async fn apply_update(&self, update: Update) -> Result<UpgradeResult, Error> {
+        if !self.is_runtime_version_different(&update.runtime_version) {
+            return Ok(UpgradeResult::SameVersion)
+        }
+
+        let metadata = self.0.rpc.metadata().await?;
+
+        if update.metadata.runtime_metadata() != metadata.runtime_metadata() {
```
niklasad1, line 249: not sure if this is correct...

subxt/src/client/online_client.rs
jsdw: Sounds good to me :)

subxt/src/client/online_client.rs
jsdw: It's quite an expensive check to do relatively speaking. Anyway, if the RuntimeVersion is different then we need to call `do_update` regardless to set that new RuntimeVersion regardless.

Maybe just have `SameVersion` and `Success` and ignore this possibility?

niklasad1: ah, I think it's fine to remove that check as the subscription items should anyway come in order so we would apply the updates in the same order. Got it, your comment was good misunderstood that

subxt/src/client/online_client.rs
```diff
@@ -114,10 +115,34 @@ impl<T: Config> OnlineClient<T> {
     ///
     /// let client = OnlineClient::<PolkadotConfig>::new().await.unwrap();
     ///
+    /// // high level API.
+    /// // no possible to know when an upgrade occurs just that it occurs.
```
jsdw, line 119:
```suggestion
```

Just a nit; prolly not necessary :)

subxt/src/client/online_client.rs
```diff
@@ -209,34 +234,119 @@ impl<T: Config> ClientRuntimeUpdater<T> {
...
+    }
+
     /// Performs runtime updates indefinitely unless encountering an error.
     ///
     /// *Note:* This will run indefinitely until it errors, so the typical usage
     /// would be to run it in a separate background task.
     pub async fn perform_runtime_updates(&self) -> Result<(), Error> {
         // Obtain an update subscription to further detect changes in the runtime version of the node.
-        let mut update_subscription = self.0.rpc.subscribe_runtime_version().await?;
+        let mut runtime_version_stream = self.0.rpc().subscribe_runtime_version().await?;
```
jsdw, line 260: I guess you could scrap this function body and swap it with something like:

```rust
let mut update_stream = updater.runtime_updates().await.unwrap();

while let Some(Ok(update)) = update_stream.next().await {
    let version = update.runtime_version().spec_version;

    if let Err(e) = updater.apply_update(update) {
        return Err(e)
    };
}
```

So that the actual update logic is all in one place, just using the low level bits you've added :)

subxt/src/client/online_client.rs
```diff
@@ -114,10 +106,34 @@ impl<T: Config> OnlineClient<T> {
     ///
     /// let client = OnlineClient::<PolkadotConfig>::new().await.unwrap();
     ///
+    /// // high level API.
+    /// // no possible to know when an upgrade occurs just that it occurs.
```
niklasad1, line 110:
```suggestion
```

subxt/src/client/online_client.rs
```diff
@@ -209,34 +233,107 @@ impl<T: Config> ClientRuntimeUpdater<T> {
...
 
-            // Fetch new metadata.
-            let new_metadata = self.0.rpc.metadata().await?;
-
-            // Do the update.
-            let mut writable = self.0.inner.write();
-            writable.metadata = new_metadata;
-            writable.runtime_version = new_runtime_version;
+            // Ignore if the update fails.
+            let _ = self.apply_update(update);
```
jsdw, line 265: Nit: I guess the comment should be tweaked; this function will not "run until it encounters an error", it'll run forever (unless it hits an error subscribing to updates in the first place) :)

jsdw (approved):
Looks great! Just a small nit regarding a comment that needs tweaking!

_Edit: nope, my bad, all is good :)_

subxt/src/client/online_client.rs
```diff
@@ -209,34 +233,107 @@ impl<T: Config> ClientRuntimeUpdater<T> {
...
+            let update = update?;
 
-            // Fetch new metadata.
-            let new_metadata = self.0.rpc.metadata().await?;
-
-            // Do the update.
-            let mut writable = self.0.inner.write();
-            writable.metadata = new_metadata;
-            writable.runtime_version = new_runtime_version;
+            // Ignore if the update fails.
```
niklasad1, line 264:
```suggestion
            // This only fails if received the runtime version is the same the current runtime version
            // which might occur because that runtime subscriptions in substrate sends out the initial
            // value when they created and not only when runtime upgrades occurs.
            // Thus, fine to ignore here as it strictly speaking isn't really an error
```

subxt/src/client/online_client.rs
```diff
@@ -209,34 +233,110 @@ impl<T: Config> ClientRuntimeUpdater<T> {
...
+
+        self.do_update(update);
+
+        Ok(())
+    }
+
     /// Performs runtime updates indefinitely unless encountering an error.
     ///
     /// *Note:* This will run indefinitely until it errors, so the typical usage
     /// would be to run it in a separate background task.
```
jsdw, line 256: Sorry; I meant this comment that was now a bit out of date :)

niklasad1: yeah but it's still true.

I mean it's depends whether we regard `failed runtime upgrades` as error before we didn't :P

lexnv (approved):
LGTM!

subxt/src/client/online_client.rs
jsdw: Oh actually I completely missed a `?` in the while loop in the new one, so I see what you mean; my bad :)

PR #660, opened 2022-09-21
Add accessor for `StaticTxPayload::call_data`

cmichi:
This might be required e.g. for `state_call` RPCs where the encoded `call_data` is without pallet and call index prefixes.


ascjones (approved):
Please leave a comment in the description as to why this might be required e.g. for `state_call` RPCs where the  encoded `call_data` is (without pallet and call index prefixes)

jsdw (approved):

Issue #655, opened 2022-09-20
Introduce path prefix for generated `::subxt` code


cmichi:
I'm running into a problem using `subxt-codegen`:

- `subxt-codegen` generates code like this for certain code paths: `::subxt::tx::StaticTxPayload::new`.
- If `subxt` is not a root dependency of the invoking crate this results  in a compiler error `could not find 'subxt' in the list of imported crates`.

I suggest to introduce an optional prefix which can be passed to `subxt-codegen`.

jsdw:
That makes sense, and it's the same sort of thing other proc macro crates in the ecosystem do to tackle the issue :)

PR #664, opened 2022-09-22
Allow specifying the `subxt` crate path for generated code

cmichi:
Closes https://github.com/paritytech/subxt/issues/655.

We need this in ink! for https://github.com/paritytech/ink/issues/1234.

I'm not particularly fond of the testing story for this feature, but I would also not duplicate all the existing tests. Hmm.

### DIFF
--- a/codegen/src/types/composite_def.rs
+++ b/codegen/src/types/composite_def.rs
@@ -76,7 +76,7 @@ impl CompositeDef {
                             | TypeDefPrimitive::U128
                     )
                 ) {
-                    derives.insert_codec_compact_as(&crate_path.clone())
+                    derives.insert_codec_compact_as(crate_path)
                 }
             }
         }
--- a/codegen/src/types/composite_def.rs
+++ b/codegen/src/types/composite_def.rs
@@ -3,6 +3,7 @@
 // see LICENSE for license details.
 
 use super::{
+    CratePath,
     Derives,
     Field,
     TypeDefParameters,
@@ -10,7 +11,6 @@ use super::{
     TypeParameter,
     TypePath,
 };
-use crate::CratePath;
 use proc_macro2::TokenStream;
 use proc_macro_error::abort_call_site;
 use quote::{
--- a/codegen/src/types/composite_def.rs
+++ b/codegen/src/types/composite_def.rs
@@ -10,6 +10,7 @@ use super::{
     TypeParameter,
     TypePath,
 };
+use crate::CratePath;
 use proc_macro2::TokenStream;
 use proc_macro_error::abort_call_site;
 use quote::{
@@ -41,6 +42,7 @@ pub struct CompositeDef {
 
 impl CompositeDef {
     /// Construct a definition which will generate code for a standalone `struct`.
+    #[allow(clippy::too_many_arguments)]
     pub fn struct_def(
         ty: &Type<PortableForm>,
         ident: &str,
@@ -49,6 +51,7 @@ impl CompositeDef {
         field_visibility: Option<syn::Visibility>,
         type_gen: &TypeGenerator,
         docs: &[String],
+        crate_path: &CratePath,
     ) -> Self {
         let mut derives = type_gen.type_derives(ty);
         let fields: Vec<_> = fields_def.field_types().collect();
@@ -73,7 +76,7 @@ impl CompositeDef {
                             | TypeDefPrimitive::U128
                     )
                 ) {
-                    derives.insert_codec_compact_as()
+                    derives.insert_codec_compact_as(&crate_path.clone())
                 }
             }
         }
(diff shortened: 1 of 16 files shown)


codegen/src/types/composite_def.rs
```diff
@@ -41,6 +42,7 @@ pub struct CompositeDef {
 
 impl CompositeDef {
     /// Construct a definition which will generate code for a standalone `struct`.
+    #[allow(clippy::too_many_arguments)]
```
cmichi, line 45: Possibly create a follow-up ticket to refactor with a config struct instead of the params.

jsdw:
I haven't looked properly at the test you've added yet. One quicker thing you could do is to do all of the codegen with a custom path and check that the string "subxt" doesn't appear at all in it. Otherwise I wonder about tweaking the subxt crate import in the tests to be a custom path; at least if using acustom path is the default thing in the tests it'll get fairly well covered!

codegen/src/types/mod.rs
```diff
@@ -358,3 +369,52 @@ impl Module {
...
+
+impl From<syn::Path> for CratePath {
+    fn from(path: syn::Path) -> Self {
+        CratePath::new(path)
+    }
+}
+
+impl From<String> for CratePath {
+    fn from(crate_path: String) -> Self {
+        Self(syn::Path::from_string(&crate_path).unwrap_or_else(|err| {
```
ascjones, line 402: Could remove this small duplication by delegating to the `From<&str>` impl

codegen/src/types/tests.rs
```diff
@@ -455,6 +455,45 @@ fn generate_array_field() {
...
+        Default::default(),
+    );
+    let types = type_gen.generate_types_mod(&"::foo".into());
+    let tests_mod = get_mod(&types, MOD_PATH).unwrap();
+
+    assert_eq!(
+        tests_mod
+            .into_token_stream()
+            .to_string()
+            .replace(":: subxt", ":: foo"),
```
ascjones, line 483: I don't understand this, surely the `type_gen.generate_types_mod(&"::foo".into());` should take care of the derive prefixes?

codegen/src/types/mod.rs
```diff
@@ -83,7 +84,7 @@ impl<'a> TypeGenerator<'a> {
     }
 
     /// Generate a module containing all types defined in the supplied type registry.
-    pub fn generate_types_mod(&self) -> Module {
+    pub fn generate_types_mod(&self, crate_path: &CratePath) -> Module {
```
ascjones, line 87: Perhaps the `crate_path` is better as a property to avoid having to pass around?

codegen/src/types/derives.rs
```diff
@@ -54,6 +56,13 @@ impl DerivesRegistry {
         }
         Derives { derives: defaults }
     }
+
+    pub fn default_with_crate_path(crate_path: &CratePath) -> Self {
+        Self {
+            default_derives: Derives::default_with_crate_path(crate_path),
```
jsdw, line 62: How about removing the `Default` impls because we never want to be able to forget to provide a crate path?

This could then just be `new` or `new_with_crate_path`

codegen/src/types/derives.rs
```diff
@@ -83,18 +92,24 @@ impl Derives {
...
         derives.insert(syn::parse_quote!(Debug));
         Self { derives }
     }
 }
 
+impl Default for Derives {
+    fn default() -> Self {
+        Derives::default_with_crate_path(&CratePath::default())
+    }
+}
```
jsdw, line 111: Should this exist? I think we should make sure crate path is always explicitly provided so we are forced to pass it through always.

cmichi: So to be on the same page: would this go through until the user-facing side? I.e. would each user always be required to specify `::subxt` as the crate path to use?

jsdw:
Looks great overall! 

I am happy with the tests (defaulting to checking that a custom path works everywhere I think is good enough for now and will almost certainly catch any issues there).

Just a couple of wee comments; I think we should remove Default impls for things that rely on a crate path wherever possible just to make it hard for future us to miss threading the crate path through somewhere it's needed.

cmichi:
@jsdw @ascjones All comments have been implemented.

codegen/src/types/mod.rs
```diff
@@ -358,3 +366,58 @@ impl Module {
...
+impl From<syn::Path> for CratePath {
+    fn from(path: syn::Path) -> Self {
+        CratePath::new(path)
+    }
+}
+
+impl ToTokens for CratePath {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let crate_path = &self.0;
+        tokens.extend(quote!(#crate_path))
```
ascjones, line 395:
```suggestion
        self.0.to_tokens(tokens)
```

ascjones (approved):
LGTM

jsdw (approved):
Looks great, thankyou @cmichi!

PR #671, opened 2022-09-29
Implement `BlocksClient` for working with blocks

### LINKED ISSUE #630: provide `finalized headers abstraction` that fetches "skipped" blocks when subscribing to `chain_finalizedHeads`
> Currently, follow-chain executes only these blocks, that come from subscription for finalizedHeads. Although this works as expected in Polkadot, for some chains some blocks may lack justification (although their successor have). Such block will be ignored, which might lead to a corrupted local state. This is the case that happens at e.g. AlephZero-based chains and possibly other.

I reckon that useful for many users of subxt to not have to implement this themselves

See https://github.com/paritytech/substrate/pull/12048/files for more informatuon

lexnv:
The filling in of missing finalized blocks can be applied to other use cases than that of finding events.

Move the core functionality of filling to the RPC client to make it more generic.

Blocked by #673.
Closes #630.


subxt/src/rpc/rpc.rs
```diff
@@ -569,6 +576,97 @@ impl<T: Config> Rpc<T> {
...
+
+        let sub = self.subscribe_finalized_blocks().await?;
+
+        // Adjust the subscription stream to fill in any missing blocks.
+        Ok(Rpc::subscribe_to_block_headers_filling_in_gaps(
+            self.clone(),
+            last_finalized_block_num,
+            sub,
+        ))
+    }
```
jsdw, line 620: I def like the idea of this, but I'm not suepr fond of _where_ it is; it's more than just an RPC call at this point, which is reall what this RPC module is all about. Eventually we'll also be transitioning to the new RPCs too. For these reasons I don't think we should add more stuff here if we can avoid it.

We have talked previously about a `api.blocks()` API to go alongside `api.constants()`, `api.events()` and so on. Perhaps this is the perfect time to add something there?

So just like we currently have `api.events().subscribe_finalized()`, we can have an `api.blocks().subscribe_finalized()` which will be this guy.

For now we don't have to add any other functionality to this new `BlocksClient` but it'll be a good place to start building stuff much like the EventsClient (eg `api.blocks().at()` to get some Block details, etc).

What do you think?

lexnv: Sounds like a plan! It helps us build the foundation for fetching blocks that we can then reutilize all over the place, like events! Thanks for the suggestion!

subxt/src/blocks/blocks_client.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// client         | Block A | ---             --- | Block C |
+    /// ```
+    ///
+    /// This function mitigates this by filling the gaps of missing blocks.
+    ///
+    /// ```sh
+    /// substrate      | Block A | --- | Block B | --- | Block C |
+    /// client         | Block A | --- | Block B | --- | Block C |
+    ///                                ^^^ filled
+    /// ```
```
jsdw, line 80: I wonder whether all of this detail is useful (it feels like an implementation detail that people would just "expect" anyway most of the time). I feel like it would be enough to say something like:

Subscribe to finalized block headers.

While the Substrate RPC method does not guarantee that all finalized block headers are provided, this function does.

subxt/src/blocks/blocks_client.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+}
+
+impl<T, Client> BlocksClient<T, Client>
+where
+    T: Config,
+    Client: OnlineClientT<T>,
+{
+    /// Subscribe to blocks headers.
```
jsdw, line 44:
```suggestion
    /// Subscribe to block headers.
```

subxt/src/events/events_client.rs
```diff
@@ -96,7 +89,11 @@ where
     ) -> impl Future<
         Output = Result<EventSubscription<T, Client, EventSub<T::Header>>, Error>,
     > + Send
-           + 'static {
+           + 'static
+    where
+        Client: Send + Sync + 'static,
+        T: Send + Sync,
```
jsdw, line 95: I wonder why these additional bounds are needed?

(it's always a shame if we need any bounds on T, since T is a phantom type and it shouldn't matter it all. Perhaps we need to use PhantomDataSendSync in more places, which does a better job of ignoring T?)

jsdw:
Looks awesome, good job! Just a couple of minor nits :)

subxt/src/blocks/blocks_client.rs
lexnv: It makes sense, thanks for the suggestion!

subxt/src/events/events_client.rs
lexnv: I compiler complained about the future returned by the event's client not being Send / Sync due to unsatisfied trait bounds on the market type `T`. `PhanthomDataSendSync` sounds like a viable option here, thanks!

subxt/src/blocks/blocks_client.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// These blocks haven't necessarily been finalised yet. Prefer
+    /// [`BlocksClient::subscribe_finalized_headers()`] if that is important.
+    pub fn subscribe_headers(
+        &self,
+    ) -> impl Future<Output = Result<impl Stream<Item = Result<T::Header, Error>>, Error>>
+           + Send
+           + 'static {
+        let client = self.client.clone();
+        async move { client.rpc().subscribe_blocks().await }
+    }
```
jsdw, line 57: Just thinking about that other PR; does this return _all_ new blocks or all new "best" blocks?

Might be worth clarifying this in the docs either way?

I wonder whether if we only subscribe to all new best blocks, we might miss some blocks? 
But I also guess that the block number might be the same for more than one block we get back via this call, if the "best" chain changes (or if we are asking for all blocks anyway).

So maybe w can't fill things in, but we can document precisely what is returned either way :)

lexnv: > Just thinking about that other PR; does this return all new blocks or all new "best" blocks?

This does subscribe to all new "best" blocks indeed.

> Might be worth clarifying this in the docs either way?

It's a good idea to document this properly. 

> But I also guess that the block number might be the same for more than one block we get back via this call, if the "best" chain changes (or if we are asking for all blocks anyway).

Yep, I think we can end up in a situation where we have two blocks with the same block number. In this case, we can't guarantee that we can fill in missing blocks.

jsdw: > Yep, I think we can end up in a situation where we have two blocks with the same block number. In this case, we can't guarantee that we can fill in missing blocks.

Yup, for that reason I don't think we should even try; as long as we document what will be returned it's all good.

I do wonder whether we should be subscribing to all new blocks too. Or whether just to add both "subscribe_best_headers" and "subscribe_all_headers" to go with "subscribe_finalized_headers". But as you said inthe other PR, let's just do what the new RPC API is going to support for now (I can't remember!)

lexnv: The new API will support all events: blocks, best blocks, and finalized blocks.
We don't support the simple block events (non best and non finalized) in subxt yet.

We can easily extend this to add all types of events, tho I think we can wait for a bit and implement it on top of the substrate's new API once it's released.

subxt/src/blocks/blocks_client.rs
```diff
@@ -0,0 +1,151 @@
...
+{
+    /// Subscribe to new best block headers.
+    ///
+    /// # Note
+    ///
+    /// The best block is selected by the consensus algorithm.
+    ///
+    /// These blocks haven't necessarily been finalised yet. Prefer
+    /// [`BlocksClient::subscribe_finalized_headers()`] if that is important.
+    pub fn subscribe_headers(
```
niklasad1, line 53: this can't be an async fn because of some lifetime issue or what's it the reason having the the `-> impl Future<_>`?

testing/integration-tests/src/blocks/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        use sp_runtime::traits::Header;
+        let block_number: u128 = (*header.number()).into();
+
+        if let Some(last) = last_block_number {
+            assert_eq!(last + 1, block_number);
+        }
+        last_block_number = Some(block_number);
+    }
+
+    Ok(())
```
niklasad1, line 87: is there any way to know whether the gaps were filled accordingly or failed i.e, is the stream returns `None` is that a bug or possible?

```suggestion
    if last_block_number.is_some() {
        Ok(())
    } else {
        panic!("We must have gaps??");
    }
```

testing/integration-tests/src/blocks/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    // (this can be a bit slow as we have to wait for finalization)
+    let header = sub.next().await.unwrap()?;
+    let finalized_hash = api.rpc().finalized_head().await?;
+
+    assert_eq!(header.hash(), finalized_hash);
+    Ok(())
+}
+
+#[tokio::test]
+async fn missing_block_headers_will_be_filled_in() -> Result<(), subxt::Error> {
```
niklasad1, line 46: why even return result here?

subxt/src/blocks/blocks_client.rs
jsdw: The reason it's used in events, and I think useful here, is so as to not tie the thing you get back to the lifetime of `self`. This just gives it more flexibility to be moved around and such.

subxt/src/blocks/blocks_client.rs
```diff
@@ -0,0 +1,151 @@
...
+    mut last_block_num: Option<u64>,
+    sub: S,
+) -> impl Stream<Item = Result<T::Header, Error>> + Send
+where
+    T: Config,
+    S: Stream<Item = Result<T::Header, E>> + Send,
+    E: Into<Error> + Send + 'static,
+{
+    sub.flat_map(move |s| {
+        let rpc = client.clone();
```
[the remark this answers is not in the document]
jsdw: I don't recall off the top of my head why there is a clone here. My guess is removing the move stuff would introduce a lifetime on `client` which wouldn't work out since we want to return that thing. Perhaps we can avoid a little cloning somewhere though?

testing/integration-tests/src/blocks/mod.rs
jsdw: I think tokio::test handles this and fails the test if an error is returned. It just allows you to use `?`'s more liberally. Don't think it's really very different from not having this and unwrapping everywhere.

subxt/src/blocks/blocks_client.rs
jsdw: As long as the API and naming here means that we can migrate easily enough and add whatever else we need, I'm happy with it! And the current choice of function is what we use in events, so it makes sense to me how it is!

Let's just make sure to clearly document it :)

subxt/src/blocks/blocks_client.rs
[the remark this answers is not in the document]
lexnv: Couldn't remove the clone from here, but removed one clone from above `client.rpc().clone()` when calling this function. The clone was circumvented by passing in a `Client` trait directly and then using the `.rpc()` inside the fn.

niklasad1 (approved):
LGTM

jsdw (approved):

PR #672, opened 2022-09-29
expose jsonrpc-core client

seunlanlege:
We depend on the jsonrpc-core client so we can make typed rpc requests

eg: https://github.com/ComposableFi/hyperspace/blob/311bc18ac6d4b24ed4586ba98552e9e145b0b068/parachain/src/provider.rs#L335-L342

### DIFF
--- a/examples/examples/custom_rpc_client.rs
+++ b/examples/examples/custom_rpc_client.rs
@@ -88,7 +88,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Pass this into our OnlineClient to instantiate it. This will lead to some
     // RPC calls being made to fetch chain details/metadata, which will immediately
     // fail..
-    let _ = OnlineClient::<PolkadotConfig>::from_rpc_client(rpc_client).await;
+    let _ = OnlineClient::<PolkadotConfig>::from_rpc_client(Arc::new(rpc_client)).await;
 
     // But, we can see that the calls were made via our custom RPC client:
     println!("Log of calls made:\n\n{}", log.lock().unwrap().as_str());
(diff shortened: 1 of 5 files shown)


niklasad1:
There is already an `Deref impl` for the rpc client, why can't you use that?

jsdw:
How about instead of this, the existing Deref impl returns `&Arc<dyn RpcClientT>` rather than `dyn rpcClientT` as it does at the mo? Arc has a deref anyway so it shouldn't alter anything, and it'll give you access to the arc you can clone or whatever without exposing any "internals"? Feels a bit cleaner at least to me :)

jsdw:
Though another approach for you may just be to clone and keep hold of a copy of the original `Client` itself rather than having to cast a `dyn RpcClientT`? This may mean constructing it explicitly and then initialising the `OnlineClient` with it, but if you want access to the same client subxt is using, that is probably the safest option.

seunlanlege:
> Another, safer, approach for you may just be to clone and keep hold of a copy of the original jsonrpsee Client itself rather than having to cast a dyn RpcClientT? This means constructing it explicitly and then passing it to the OnlineClient::from_rpc_client method (as well as cloning it to keep separately), but if you want access to the same client subxt is using, that is probably the safest option and avoids any casting (which may break in the future) or whatever

I've modified the PR with this approach

niklasad1 (approved):
LGTM, just to run `cargo fmt` to make the CI happy.

jsdw (approved):
LGTM once fmt is ran and we've fixed our tests.

seunlanlege:
> LGTM, just to run cargo fmt to make the CI happy.

Done 🙏🏿

PR #679, opened 2022-10-03
Fix contracts integration test

ascjones:
Prior to https://github.com/paritytech/substrate/pull/12383, setting `proof_size` to `0` worked for all extrinsics. Following that a proper value needed to be supplied. However `u64::MAX` was too high as it breached the limit for normal class transactions weight (I think about `0.75 * MAX_BLOCK_WEIGHT`).

Interestingly `0` still worked for `instantiate_with_code` and `call`, but the fix was to set it to `u64::MAX / 2`.

### DIFF
--- a/testing/integration-tests/src/frame/contracts.rs
+++ b/testing/integration-tests/src/frame/contracts.rs
@@ -61,7 +61,7 @@ impl ContractsTestContext {
             100_000_000_000_000_000, // endowment
             Weight {
                 ref_time: 500_000_000_000,
-                proof_size: u64::MAX,
+                proof_size: 0,
             }, // gas_limit
             None,                    // storage_deposit_limit
             code,
@@ -106,7 +106,7 @@ impl ContractsTestContext {
             100_000_000_000_000_000, // endowment
             Weight {
                 ref_time: 500_000_000_000,
-                proof_size: u64::MAX,
+                proof_size: 0,
             }, // gas_limit
             None,                    // storage_deposit_limit
             code_hash,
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

lexnv:
I was trying the same thing, then decided to adjust the `ref_time` to fit the contract:

`1_250_000_000_000` out of gas

`1_280_000_000_000` out of gas

`1_290_000_000_000` out of gas

`1_295_000_000_000` out of gas

`1_299_000_000_000` block limit

`1_300_000_000_000` block limit

`1_310_000_000_000` block limit 

`1_375_000_000_000` block limit

But couldn't find any appropriate values that would accept our dummy contract, which is strange.

I also waited for 2 blocks between the `instantiate_with_code` call and `instantiate` without any effect, based on the comment from substrate about exhausted transactions: _Invalidity might also be temporary. In case of `ExhaustsResources` the transaction does not fit to the current block, but it might be okay for the next one._

lexnv:
Replicating the test using the old API resulted in the same behavior:

```rust
        use crate::node_runtime::runtime_types::sp_weights::OldWeight;

        let instantiate_tx = node_runtime::tx().contracts().instantiate_old_weight(
            100_000_000_000_000_000, // endowment
            OldWeight(
                1_295_000_000_000,
            ), // gas_limit
            None,                    // storage_deposit_limit
            code_hash,
            data,
            salt,
        );

```

lexnv (approved):
Thanks for fixing this! LGTM!

PR #680, opened 2022-10-03
Make it possible to get raw bytes back from dynamic constant/storage queries

jsdw:
When you make a dynamic storage or constant query at the moment, you get back a `Value`. In some cases, you might just want the raw SCALE bytes back to do as you wish with (eg here: https://github.com/paritytech/staking-miner-v2/pull/387/files#diff-6e448cdd99756118d32afe57b3d7d260cb1626cdbdbace46a91f1f9d20aabd4eR118.).

This PR makes dynamic things decode into this intermediate `DecodedValueThunk` type, from which you can either finish decoding into a value with `.to_value()` or access the bytes with `.to_encoded()` to do as you wish with instead.

### DIFF
--- a/examples/examples/dynamic_queries.rs
+++ b/examples/examples/dynamic_queries.rs
@@ -51,7 +51,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // 2. Dynamic constant access (the dynamic equivalent to the fetch_constants example).
 
     let constant_address = subxt::dynamic::constant("Balances", "ExistentialDeposit");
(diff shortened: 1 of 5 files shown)


niklasad1 (approved):
nice, LGTM

I don't understand why `full-md.scale` and `stripped-md.scale` were committed though

jsdw:
Ah whoops; I didn't notice those! I'll reset and get rid of them!

ascjones (approved):
Good stuff, I was missing something like this when experimenting with it.

PR #681, opened 2022-10-06
Make Config::Extrinsic `Send`

Wizdave97:
This fixes the `future is not Send as this value is used across an await
` error  when calling `TxInBlock::wait_for_success`
Current Error I get when I call that function
image


Wizdave97:
@ascjones  Please take a look

jsdw (approved):
I'm ok with this; basically every other type is `Send` anyway, and it shouldn't be an added burden!

ascjones (approved):

PR #683, opened 2022-10-06
Allow taking out raw bytes from a SubmittableExtrinsic

MOZGIII:
This is useful for something like this:

```rust
    let balances_transfer_tx = ...;

    let signed_balances_transfer_tx = api
        .tx()
        .create_signed(&balances_transfer_tx, signer, Default::default())
        .await?;

    let extrinsic_hash = Hashing::hash(signed_balances_transfer_tx.encoded());

    // We need `extrinsic_hash` because we want to log it, before we try to submit and watch the extrinsic.
    // We can't use `signed_balances_transfer_tx.submit_and_watch()` as we want to avoid doing the work twice.
    debug!(
        message = "Balance transfer extrinsic signed",
        ?address,
        ?extrinsic_hash
    );

    // Manually do the `watch_extrinsic`.
    let sub = api
      .rpc()
      .watch_extrinsic(Encoded(signed_balances_transfer_tx.into_encoded())) // <---- this is where we need the new API
      .await?

     let mut progress =
        TxProgress::<chain::HumanodeConfig, _>::new(sub, api.clone(), extrinsic_hash);

    // Continue as if we called `signed_balances_transfer_tx.submit_and_watch()`...
```

### DIFF
--- a/subxt/src/tx/tx_client.rs
+++ b/subxt/src/tx/tx_client.rs
@@ -350,6 +350,12 @@ where
     pub fn encoded(&self) -> &[u8] {
         &self.encoded.0
     }
+
+    /// Consumes [`SubmittableExtrinsic`] and returns the SCALE encoded
+    /// extrinsic bytes.
+    pub fn into_encoded(self) -> Vec<u8> {
+        self.encoded.0
+    }
 }
 
 impl<T, C> SubmittableExtrinsic<T, C>


niklasad1 (approved):

subxt/src/tx/tx_client.rs
```diff
@@ -350,6 +350,12 @@ where
     pub fn encoded(&self) -> &[u8] {
         &self.encoded.0
     }
+
+    /// Consumes [`SubmittableExtrinsic`] and returns the SCALE encoded
+    /// extrinsic bytes.
+    pub fn into_encoded(self) -> Vec<u8> {
```
ascjones, line 356: Note there is a similar method on [DecodedValueThunk](https://github.com/paritytech/subxt/blob/master/subxt/src/dynamic.rs#L66), but it's called `to_encoded`. Not sure which is more idiomatic, but I think they should be the same for the sake of consistency.

niklasad1: I think generally in rust `into_` implies taking ownership of the actual value and `to_` implies borrowing and copying/cloning the value for non-copy types.

so my two cents on `into_encoded`.

MOZGIII: `into_` here implies that we deconstruct ourselves into parts. Intuition here is the value is just taken as is without transformation, allowing for very efficient underlying machine code.

lexnv: The `DecodedValueThunk::to_encoded(self) -> Vec<u8>` presents the same functionality as we would like here.

It's a good idea to have consistency with that to avoid confusion.

We should then decide if we want to name it `into_encoded` or `to_encoded` and modify both places.

TIL: On one hand we have the [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) trait which consumes self, on the other we [to_vec](https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec) taking a self reference and [to_be](https://doc.rust-lang.org/std/primitive.i64.html#method.to_be) consuming self again.

That being said, I think we can't go wrong with either and I'd lean towards the `into_` as I find it resonating with the `Into` trait

niklasad1: slice and u64 is copy right then it could add up?  but yeah a bit confusing maybe my comment is wrong... I still prefer into_encoded though :)

jsdw: `into_encoded` makes more sense to me; let's change the `to_encoded` in DecodeValueThunk to match :)

MOZGIII: Let's do it in a separate PR. I don't want this to take longer than it needs to be.

MOZGIII: We could also do a `From` impl, but then why don't we just make the fields public at this point?

What we could also add is `into_parts` where we return both hash and encoded data.

We can add `AsRef`s too.

But all that I'd suggest we do in a bit, as it will likely be a breaking change, and will take more time, whereas this PR can be merged without any breakage now.

jsdw: There are already breaking changes in master anyway (and the block stuff hasn't been released yet anyway iirc) so I'd just throw in the trivial `to_encoded` => `into_encoded` rename on the `DecodeValueThunk` for consistency sake and nothing more. If not, I'll throw in a quick PR for the same.

jsdw (approved):

subxt/src/tx/tx_client.rs
MOZGIII: Please do, it'll probably take a bit longer for me atm

PR #689, opened 2022-10-14
Add runtime_metadata_url to pull metadata directly from a node

### LINKED ISSUE #640: Add option for automatic metadata download to macro
## Problem Statement

For compile-time generated types `subxt` currently requires having the metadata as a fixture locally:

```rust
#[subxt::subxt(runtime_metadata_path = "metadata/contracts-node.scale")]
pub mod api {}
```

We are using `subxt` as part of [an E2E testing framework](https://github.com/paritytech/ink/issues/1234) that we'll ship with the next major ink! release. Developers can then write tests for their contracts in the usual `#[test]` style, with an E2E test being executed in the background.

As ink! needs to support all kind of Substrate runtimes this would currently require us to make `subxt-cli` a dep of ink!, and devs would have to execute `subxt metadata > contracts-node.scale` before running their E2E tests.

## Solution Ideas

Could we make the `runtime_metadata_path` optional and if omitted download the metadata from the node for each macro invocation?

jsdw:
And reuse this fetch functionality in the CLI crate to avoid duplication.

Closes #640

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -19,6 +19,10 @@ use crate::{
         CompositeDefFields,
         TypeGenerator,
     },
+    utils::{
+        fetch_metadata_bytes_blocking,
+        Uri,
+    },
     CratePath,
 };
 use codec::Decode;
@@ -50,9 +54,10 @@ use syn::parse_quote;
 /// * `item_mod` - The module declaration for which the API is implemented.
 /// * `path` - The path to the scale encoded metadata of the runtime node.
 /// * `derives` - Provide custom derives for the generated types.
+/// * `crate_path` - Path to the `subxt` crate.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
-pub fn generate_runtime_api<P>(
+pub fn generate_runtime_api_from_path<P>(
     item_mod: syn::ItemMod,
     path: P,
     derives: DerivesRegistry,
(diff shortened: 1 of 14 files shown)


codegen/src/api/mod.rs
```diff
@@ -50,9 +54,10 @@ use syn::parse_quote;
 /// * `item_mod` - The module declaration for which the API is implemented.
 /// * `path` - The path to the scale encoded metadata of the runtime node.
 /// * `derives` - Provide custom derives for the generated types.
+/// * `crate_path` - Path to the `subxt` crate.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
-pub fn generate_runtime_api<P>(
+pub fn generate_runtime_api_from_path<P>(
```
lexnv, line 60: Nit: we could also update the `RuntimeGenerator` doc to reflect this change

lexnv (approved):

codegen/src/api/mod.rs
jsdw: Ooh good point; our docs CI thing caught that also :)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+pub async fn fetch_metadata_bytes(url: &Uri) -> Result<Vec<u8>, FetchMetadataError> {
+    let hex = fetch_metadata_hex(url).await?;
+    let bytes = hex::decode(hex.trim_start_matches("0x"))?;
+    Ok(bytes)
+}
+
+/// Returns the raw, 0x prefixed metadata hex from the provided URL.
+pub async fn fetch_metadata_hex(url: &Uri) -> Result<String, FetchMetadataError> {
+    let hex_data = match url.scheme_str() {
+        Some("http") => fetch_metadata_http(url).await,
```
niklasad1, line 43: what about `wss` and `https`?

niklasad1: as you are using `jsonrpsee` both should work AFAIK

niklasad1 (approved):
this is a nice feature to have, LGTM

I have a question why `https` is not supported?

codegen/src/utils/fetch_metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+        client::ClientT,
+        Error,
+    },
+    http_client::HttpClientBuilder,
+    rpc_params,
+};
+
+/// Returns the metadata bytes from the provided URL, blocking the current thread.
+pub fn fetch_metadata_bytes_blocking(url: &Uri) -> Result<Vec<u8>, FetchMetadataError> {
+    tokio::runtime::Builder::new_multi_thread()
```
niklasad1, line 17: right, tokio is not initialized here because it's a compile time however you could have another helper function to avoid repeating the tokio init:

```rust
fn tokio_block_on<T, Fut: Future<Output = T>>(fut: Fut) -> Result<T, FetchMetadataError> {
     tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .unwrap()
        .block_on(fut)
}
```

jsdw:
> I have a question why https is not supported?

I'd have to check; for some reason it wasn't supported by the looks of it in the previous iteration, and so I didn't think too much about it. I'll check and see whether that restriction can be removed (I mean, if WSS is supported I don't see why not!)

codegen/src/utils/fetch_metadata.rs
jsdw: Good idea!

jsdw:
Yeah, HTTPS does indeed work fine, I guess there was no reason it wasn't always supported!

PR #691, opened 2022-10-18
Extend the new `api.blocks()` to be the primary way to subscribe and fetch blocks/extrinsics/events

jsdw:
Also:
- Remove the `Extrinsic` type from config (We just treat every extrinsic as a collection of bytes at the moment, and if/when we want to decode them properly we can treat them like events and ask for a type/shape at decode time).
- Tie tx submission in to this; both extrinsics in blocks and tx submission have associated events.
- Add an example subscribing to finalized blocks and printing some details of them.
- Make `api.blocks().subscribe_*()` the only way to subscribe to streams of things (removing similar functionality from Events to avoid lots of repetition and simplify the API surface).
- Remove the `FilterEvents` stuff; it makes much more sense to ask for events via this new blocks API (either for the entire block or for specific extrinsics), and it was all rather complicated.


subxt/src/blocks/blocks_client.rs
```diff
@@ -41,6 +47,72 @@ where
...
+                        .rpc()
+                        .block_hash(None)
+                        .await?
+                        .expect("didn't pass a block number; qed")
+                }
+            };
+
+            let res = match client.rpc().block(Some(block_hash)).await? {
+                Some(block) => block,
+                None => return Err(BlockError::BlockHashNotFound(hex::encode(block_hash)).into())
```
lexnv, line 73: DQ: should we prefix this with `0x`?

jsdw: That would be nicer wouldn't it!

subxt/src/blocks/block_types.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl <'a, T, C> Extrinsic<'a, T, C>
+where
+    T: Config,
+    C: OnlineClientT<T>
+{
+    /// The events associated with the extrinsic.
+    pub async fn events(&self) -> Result<ExtrinsicEvents<T>, Error> {
+        let ext_hash = T::Hashing::hash_of(&self.bytes);
+        let events = events::EventsClient::new(self.client.clone())
```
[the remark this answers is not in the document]
jsdw: Great idea!

lexnv (approved):

subxt/src/blocks/block_types.rs
```diff
@@ -0,0 +1,284 @@
...
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""))]
+pub struct ExtrinsicEvents<T: Config> {
+    // The hash of the extrinsic (handy to expose here because
+    // this type is returned from TxProgress things in the most
+    // basic flows, so it's the only place people can access it
+    // without complicating things for themselves).
+    ext_hash: T::Hash,
+    // The index of the extrinsic:
+    idx: u32,
```
mep-jerry, line 188: Could you add the idx getting function, because we want a combination of blockhash and ext index

jsdw: I don't see any harm in adding that (though note that you can get it from the extrinsic already)

subxt/src/blocks/block_types.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+/// A single extrinsic in a block.
+pub struct Extrinsic<'a, T: Config, C> {
+    index: u32,
+    bytes: &'a [u8],
+    client: C,
+    block_hash: T::Hash,
+    cached_events: Arc<AsyncMutex<Option<events::Events<T>>>>,
```
ascjones, line 144:
```suggestion
    cached_events: CachedEvents<T>,
```

ascjones (approved):
LGTM

ascjones:
New `scale-info` `2.3` has broken the CI: see https://github.com/paritytech/subxt/pull/704

PR #697, opened 2022-10-20
Add getters for `Module`

cmichi:
Needed for https://github.com/paritytech/ink/issues/1234.

### DIFF
--- a/codegen/src/types/mod.rs
+++ b/codegen/src/types/mod.rs
@@ -365,6 +365,21 @@ impl Module {
     pub fn ident(&self) -> &Ident {
         &self.name
     }
+
+    /// Returns this `Module`s child `mod`s.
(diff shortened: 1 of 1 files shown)


codegen/src/types/mod.rs
```diff
@@ -365,6 +365,21 @@ impl Module {
     pub fn ident(&self) -> &Ident {
         &self.name
     }
+
+    /// Returns this `Module`s child `mod`s.
+    pub fn children(&self) -> &BTreeMap<Ident, Module> {
```
ascjones, line 370: Should this (and `types` below) instead return `impl Iterator<Item = (&Ident, &Module)`? After all the `BTreeMap` usage is an implementation detail which could change.

ascjones (approved):

jsdw (approved):

PR #700, opened 2022-10-25
add wasm support

niklasad1:
Co-authored by @Xanewok


subxt/src/lib.rs
```diff
@@ -135,6 +135,10 @@
 
 pub use subxt_macro::subxt;
 
+// Used to enable js feature for wasm.
```
niklasad1, line 138:
```suggestion
// Used to enable the js feature for wasm.
```

subxt/src/rpc/jsonrpsee_impl.rs
```diff
@@ -62,21 +66,3 @@ impl RpcClientT for Client {
...
-    };
-    let val = serde_json::to_value(&params).expect("RawValue guarantees valid JSON");
-    let arr = match val {
-        Value::Array(arr) => arr,
-        _ => {
-            panic!("RPC Params are expected to be an array but got {params}");
-        }
-    };
-    ParamsSer::Array(arr)
-}
```
jsdw, line 82: Wooo, awesome that this gets to go away :)

wasm-tests/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
jsdw, line 1: can these wasm-tests bits be moved into the `testing` folder?

wasm-tests/tests/wasm.rs
```diff
@@ -0,0 +1,26 @@
...
+    init_tracing();
+
+    let client =
+        subxt::client::OnlineClient::<PolkadotConfig>::from_url("ws://127.0.0.1:9944")
+            .await
+            .unwrap();
+
+    let chain = client.rpc().system_chain().await.unwrap();
+    assert_eq!(&chain, "Development");
+}
```
jsdw, line 26: 🎉

subxt/src/client/online_client.rs
```diff
@@ -61,7 +61,10 @@ impl<T: Config> std::fmt::Debug for OnlineClient<T> {
 }
 
 // The default constructors assume Jsonrpsee.
-#[cfg(feature = "jsonrpsee")]
+#[cfg(any(
+    all(feature = "jsonrpsee-ws", not(target_arch = "wasm32")),
+    all(feature = "jsonrpsee-web", target_arch = "wasm32")
+))]
```
jsdw, line 67: Should we have a check at the root lib.rs to make sure that somebody cannot enable both of these features?

What are the pros and cons for also having this `target_arch` check as well as just checking the features? It'd be ncie if we could just check for a feature and not worry about the rest, for simplicity?

niklasad1: > Should we have a check at the root lib.rs to make sure that somebody cannot enable both of these jsonrpsee features simultaneously?

Sure

> What are the pros and cons for also having this target_arch check as well as just checking the features? It'd be nice if we could just check for a feature and not worry about the rest, for simplicity

It's my own fault in [jsonrpsee](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/async_client/mod.rs#L196) i.e, it won't compile with the target_arch

wasm-tests/Cargo.toml
niklasad1: sure

subxt/src/client/online_client.rs
jsdw: Ok that makes sense! Is it worth filing an issue/PR in jsonrpsee to tweak that so we can simplify here?

niklasad1: yeah, I can take a look if it's possible to remove those

jsdw (approved):
Looks great! Would be ncie if we could slightly simplify the feature flags but it's not a blocker to getting this merged!

subxt/Cargo.toml
```diff
@@ -22,19 +22,20 @@ integration-tests = []
...
 bitvec = { version = "1.0.0", default-features = false, features = ["alloc"] }
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full", "bit-vec"] }
 scale-info = { version = "2.0.0", features = ["bit-vec"] }
 scale-value = "0.6.0"
 scale-decode = "0.4.0"
-futures = "0.3.13"
+futures = { version = "0.3.13", default-features = false }
 hex = "0.4.3"
-jsonrpsee = { version = "0.15.1", features = ["async-client", "client-ws-transport", "jsonrpsee-types"], optional = true }
+jsonrpsee = { version = "0.16", optional = true, features = ["jsonrpsee-types"] }
```
ascjones, line 36: Is it possible to still keep `jsonrpsee` as an optional dependency? For anyone implementing a custom client who doesn't need it.

ascjones (approved):
Awesome! Can't wait to build a toy web app with this :sunglasses:

subxt/Cargo.toml
niklasad1: yes, good catch

ascjones: Oh yes sorry the optional moved before the features :grin:

PR #701, opened 2022-10-27
Cargo.lock to be part of git repo to have deterministics builds

dzmitry-lahoda:
- crates generate lock on build
- we need commit not on crates yet
- so making git to be same as crates dep
- we add lock
- which works well with nix
- other options way harder to maintain


niklasad1:
Hey @dzmitry-lahoda 

I don't understand what you are trying to accomplish by commiting `Cargo.lock` for the entire `subxt` repo. 
I suppose committing `Cargo.lock`  for `subxt-cli` would be required for "determinstic builds"?

The rest of crates are just libraries and committing `Cargo.lock` is just ignored then anyway...

jsdw:
Since the workspace has both library crates (which don't want or need a cargo.lock) and binary crate which does, I'd propose for now that we follow https://github.com/BurntSushi/ripgrep and do save the cargo.lock.

we should also look at their release checklist and make sure we add a step to ours to ensure the lock file is updated each time we do a release.

jsdw:
That all said, for my own sanity I'd like to generate and commit the lock file on our side (since it's so hard to review it for correctness) so I'll leave this PR open as a reminder and then open a new one with a lock file I've generated and close this :)

jsdw:
We'll take care of this in #795 :)

PR #702, opened 2022-10-28
No need to entangle Signer and nonce now

jsdw:
We can create transactions and provide separate nonces now, so there is no need to expose this nonce functionality in the Signer trait, which shouldn't need to care.

### DIFF
--- a/subxt/src/tx/tx_client.rs
+++ b/subxt/src/tx/tx_client.rs
@@ -224,14 +224,10 @@ where
         Call: TxPayload,
     {
         // Get nonce from the node.
-        let account_nonce = if let Some(nonce) = signer.nonce() {
-            nonce
-        } else {
-            self.client
-                .rpc()
-                .system_account_next_index(signer.account_id())
-                .await?
-        };
+        let account_nonce = self.client
+            .rpc()
+            .system_account_next_index(signer.account_id())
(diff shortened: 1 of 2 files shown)


subxt/src/tx/tx_client.rs
```diff
@@ -224,14 +224,11 @@ where
...
-        } else {
-            self.client
-                .rpc()
-                .system_account_next_index(signer.account_id())
-                .await?
-        };
+        let account_nonce = self
+            .client
+            .rpc()
+            .system_account_next_index(signer.account_id())
```
ascjones, line 230: There still should be some way to set the nonce locally, this will return the next index from the most recent block.

ascjones: e.g. https://github.com/paritytech/smart-bench/blob/361da7782d2eeb72e056b98ce34a2438678e245d/src/wasm/runner.rs#L142

ascjones: Ok didn't know about https://github.com/paritytech/subxt/blob/563afdaded26f954812646f2e163f6e940d50145/subxt/src/tx/tx_client.rs#L128

ascjones (approved):

niklasad1 (approved):

PR #703, opened 2022-10-31
Use git instead of crates.io for Substrate deps

ggwpez:
Changes:  
- Use git instead of crates.io for Substrate

There are two reasons for this as far as I am concerned:  
- Parity is actually not in control of all Substrate crates on `crates.io`. So we risk pulling in something that is not Substrate at all.   
- Substrate is not being released regularly on `crates.io`, so the versions here are out-dated which prevents downstream projects like https://github.com/paritytech/polkadot-stps (cc @bernardoaraujor) from updating their Substrate versions.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,7 +14,7 @@ description = "Subxt example usage"
 [dev-dependencies]
 subxt = { path = "../subxt" }
 tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
-sp-keyring = "6.0.0"
+sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "master" }
 futures = "0.3.13"
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full", "bit-vec"] }
 hex = "0.4.3"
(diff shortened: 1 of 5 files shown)


jsdw:
We spoke in person but just for the info here:

- subxt can only point to things on crates.io; if it points to GitHub it can't be published to crates.io
- all of the substrate crates subxt uses are in our control im pretty certain; I've published them before at least!
- there is work ongoing by Mira and the release people to start publishing substrate crates to crates.io and at that point subxt can keep up to date easily enough.

we will do an interim publish of substrate crates though as it's been a while and I'm aware that there is a frustrating issue with something in the dependency tree not being happy to exist at two versions in a tree.

niklasad1:
Meanwhile you could hack around this and patch `sp-core`

ggwpez:
> Meanwhile you could hack around this and patch sp-core

Yea even if I patch the stuff then there is something with the `parity-util-mem` which breaks.

> we will do an interim publish of substrate crates though as it's been a while and I'm aware that there is a frustrating issue with something in the dependency tree not being happy to exist at two versions in a tree.

Many thanks! That should most likely fix it for me.

niklasad1:
> Yea even if I patch the stuff then there is something with the parity-util-mem which breaks.

Did you also patch `sp-runtime`? I think it should work if patch everything because `parity-util-mem` has this guard against duplicate versions of it. Try `cargo-tree` or something other tool to check what brings in `parity-util-mem 0.11` besides subxt ^^

Yeah, we should fix it I have run into it as well.

PR #709, opened 2022-11-09
Update substrate dependencies, removing `parity-util-mem` transitive dependency

ascjones:
*WIP* pending release of substrate dependencies, pointing at my PR https://github.com/paritytech/substrate/pull/12657.

image

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,7 +14,7 @@ description = "Subxt example usage"
 [dev-dependencies]
 subxt = { path = "../subxt" }
 tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
-sp-keyring = "6.0.0"
+sp-keyring = { version = "6.0.0", git = "https://github.com/paritytech/substrate", branch = "aj/sp-runtime-parity-util-mem", package = "sp-keyring", default-features = false }
 futures = "0.3.13"
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full", "bit-vec"] }
 hex = "0.4.3"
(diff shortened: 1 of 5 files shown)


niklasad1:
@ascjones 

use default features for sp-runtime again implies that we will have `parity-util-mem` in the dependency tree? :face_with_spiral_eyes:

ascjones:
> use default features for sp-runtime again implies that we will have `parity-util-mem` in the dependency tree? face_with_spiral_eyes

Sadly it seems that way. The issue seems to be that the `std` feature of `sp-runtime` ensures `sp-io/std` is used. If it is not then you get all the `no-std` wasm stuff.

So we need to use a dependency feature for `parity-util-mem` instead of `std`, or go the whole hog and remove the `parity-util-mem` entirely: https://github.com/paritytech/substrate/issues/12658

ascjones:
> All right, I guess it's worth waiting for [paritytech/substrate#12658](https://github.com/paritytech/substrate/issues/12658) before doing a new `sp-core` release then?

I think so. 

From my side we are planning a new `ink 4.0-beta` releases this week which would require a new `cargo-contract` release which depends on both `sp-runtime/core` and `subxt`. However for the sake of expediency we may have to do require to install from `github` rather than `crates.io`.

examples/Cargo.toml
```diff
@@ -14,7 +14,7 @@ description = "Subxt example usage"
 [dev-dependencies]
 subxt = { path = "../subxt" }
 tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
-sp-keyring = "6.0.0"
+sp-keyring = { version = "7.0.0", git = "https://github.com/paritytech/substrate", default-features = false }
```
niklasad1, line 17: why git version on this?

niklasad1 (approved):

athei (approved):

Issue #712, opened 2022-11-10
Example: Obtain call type id and submit a `DynamicTxPayload`


lexnv:
Add an example that uses the new dynamic API exposed by subxt
that constructs the tx payload dynamically. 

To ensure the tx payload can work across different chains, fetch
the metadata of the node and extract the call type ID of a given
pallet.

One suggestion could be from `Multisig::as_multi`

jsdw:
In coming up with https://github.com/paritytech/subxt/pull/714 I actually realised that you don't need to prefix the call length etc and it actually turned out to be quite nice and avoids needing call type IDs etc (which we'd much rather people never needed to mess abotu with explicitly anyway).

niklasad1:
Closed by #714

PR #713, opened 2022-11-10
Add multisig static example

lexnv:
This example executes a `multisig::as_multi` extrinsic
containing a `Assets:freeze_asset` call.

While at it, also add the substrate scale metadata to facilitate
working with the substrate nodes.

Part of https://github.com/paritytech/subxt/issues/712.


examples/examples/multisig.rs
```diff
@@ -0,0 +1,66 @@
...
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! To run this example, a local polkadot node should be running. Example verified against substrate latest-38a955ba4a0.
+//!
+//! E.g.
+//! ```bash
+//! curl "https://releases.parity.io/substrate/x86_64-debian:stretch/latest/substrate/substrate" --output /usr/local/bin/substrate --location
+//! substrate --dev --tmp
+//! ```
```
jsdw, line 11: Is this the only example using Substrate? I think just for consistency with other examples maybe it's best to stick to Polkadot?

lexnv: The Polkadot node does not contain the `Assets` pallet, at least I didn't see it in the metadata of `wss://rpc.polkadot.io:443`

jsdw: Oh, that's ok; we don't need to specifically use this Assets call as the example; in https://github.com/paritytech/subxt/pull/714/files I used a balance transfer so perhaps we could standardise and both just do that?

jsdw: Also I think I'll rename my example to `dynamic_multisig.rs` to align with yours :)

examples/examples/multisig.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    // Create a client to use:
+    let api = OnlineClient::<SubstrateConfig>::new().await?;
+
+    // The call data for the `Multisig::as_multi` extrinsic.
+    // This must be of type `RuntimeCall` and represents the `Assets::freeze_asset` method.
+    // Otherwise, could have been obtained using `substrate::tx().assets().freeze_asset(100)`
+    let call = substrate::runtime_types::kitchensink_runtime::RuntimeCall::Assets(
+        substrate::runtime_types::pallet_assets::pallet::Call::freeze_asset { id: 100 },
+    );
```
jsdw, line 37: what about:

```rust
substrate::tx().assets().freeze_asset(100).call_data().clone()
```

does that work instead?

And if so, perhaps we could add an `into_call_data()` to mirror the `into_value() I added on the dynamic side, and avoid the clone here?

jsdw:
Since Alex is away today I'll combine this with my PR and fix anything up there to make the two consistent :)

PR #714, opened 2022-11-10
Make working with nested queries a touch easier

jsdw:
- Add example of dynamic and static nested query showing a multisig thing.
- `.into_value()` on a dynamic TX to turn it into a `Value` that we can pass into other calls.
- Allow dynamix tx args to be named or unnamed for better readability.

This does bring to light, again, a problem with constructing nested calls in static code; the outer call will have a dynamic pallet/call index according to the node we're sending it to, but the inner call won't, and so the code won't be portable across different chains. The dynamic stuff acts as a workaround for this at the mo, but would be good to address it in static code so I'll raise an issue.

Co-Authored by @lexnv.


examples/examples/multisig.rs
```diff
@@ -0,0 +1,76 @@
...
+        polkadot::runtime_types::pallet_balances::pallet::Call::transfer {
+            dest: dest.into(),
+            value: 123_456_789_012_345,
+        },
+    );
+
+    // Now, build an outer call which this inner call will be a part of.
+    // This sets up the multisig arrangement.
+    let tx = polkadot::tx().multisig().as_multi(
+        // threashold
```
niklasad1, line 54:
```suggestion
        // threshold
```

niklasad1 (approved):
neat, LGTM

ascjones (approved):

PR #715, opened 2022-11-11
fix clippy

niklasad1:


codegen/src/types/mod.rs
```diff
@@ -109,6 +109,7 @@ impl<'a> TypeGenerator<'a> {
         root_mod
     }
 
+    #[allow(clippy::only_used_in_recursion)]
```
niklasad1, line 112: https://rust-lang.github.io/rust-clippy/master/index.html#only_used_in_recursion

`id: u32` below is used directly in the recursive call not modified such as `id + 1` or something
I think that is fine because the recursion stuff is based on `path: Vec<String>`

subxt/src/tx/tx_payload.rs
```diff
@@ -138,7 +138,7 @@ impl<'a> TxPayload for DynamicTxPayload<'a> {
         let pallet = metadata.pallet(&self.pallet_name)?;
         let call_id = pallet.call_ty_id().ok_or(MetadataError::CallNotFound)?;
         let call_value =
-            Value::unnamed_variant(self.call_name.to_owned(), self.fields.clone());
+            Value::unnamed_variant(self.call_name.clone(), self.fields.clone());
```
niklasad1, line 141: clippy didn't like that because `to_owned` is called on `Cow`, thus clippy suggests either `into_owned` or `clone (which should be cheaper)`

jsdw (approved):

jsdw:
Given Alex and igor are off today I'll just merge this since it's not controversial

PR #718, opened 2022-11-17
Add bitvec-like generic support to the scale-bits type for use in codegen

### LINKED ISSUE #693: Remove bitvec from subxt-codegen
Currently, we still use `bitvec` in the subxt codegen stuff (we generate specific store/order things when we encounter bitvec types). As BitVec can't be compiled to wasm with every store type we support, this means we may occasionally generate code that doesn't compile to WASM. And anyway, we don't use bitvec elsewhere now so it'd be good to get rid of it from everywhere.

So let's figure out how to do this! Possibly something like:

```rust
pub enum U8 {}
pub enum U16 {}
// ...

pub enum Lsb0 {}
pub enum Msb0 {}

#[derive(Debug, Clone)]
pub struct DecodedBits<Store, Order>(scale_bits::Bits)

impl Decode for DecodedBits<U8, Lsb0> { // ... use scale_bits to do the heavy lifting }
impl Decode for DecodedBits<U16, Lsb0> { // ... use scale_bits to do the heavy lifting }
// ...

// and then in the codegen, use those types in substitute, instead of swapping out the 
// bitvec store/order types to make valid bitvecs
```

This is a bit verbose of an approach; perhaps we can think of something simpler?

Xanewok:
Fixes #693 

Hopefully this is in line with what was expected from #693 cc @jsdw

### DIFF
--- a/subxt/src/utils.rs
+++ b/subxt/src/utils.rs
@@ -96,3 +96,124 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
 /// with collections like BTreeMap. This has the same type params
 /// as `BTreeMap` which allows us to easily swap the two during codegen.
 pub type KeyedVec<K, V> = Vec<(K, V)>;
+
+/// Generic `scale_bits` over `bitvec`-like `BitOrder` and `BitFormat` types.
+pub mod bits {
+    use codec::{Compact, Input};
+    use scale_bits::{
+        scale::format::{Format, OrderFormat, StoreFormat},
+        Bits,
+    };
+    use std::marker::PhantomData;
+
+    macro_rules! store {
+        ($ident: ident; $(($ty: ident, $wrapped: ty)),*) => {
+            /// Associates `bitvec::store::BitStore` trait with corresponding, type-erased `scale_bits::StoreFormat` enum.
+            ///
+            /// Used to decode bit sequences in runtime by providing `scale_bits::StoreFormat` using
+            /// `bitvec`-like type type parameters.
+            pub trait $ident {
+                /// Corresponding `scale_bits::StoreFormat` value.
+                const FORMAT: StoreFormat;
+                /// Number of bits that the backing store types holds.
+                const BITS: u32;
+            }
+
+            $(
+                impl $ident for $wrapped {
+                    const FORMAT: StoreFormat = StoreFormat::$ty;
+                    const BITS: u32 = <$wrapped>::BITS;
+                }
+            )*
+        };
+    }
+
+    macro_rules! order {
+        ($ident: ident; $($ty: ident),*) => {
+            /// Associates `bitvec::order::BitOrder` trait with corresponding, type-erased `scale_bits::OrderFormat` enum.
+            ///
+            /// Used to decode bit sequences in runtime by providing `scale_bits::OrderFormat` using
+            /// `bitvec`-like type type parameters.
+            pub trait $ident {
+                /// Corresponding `scale_bits::OrderFormat` value.
+                const FORMAT: OrderFormat;
+            }
+
+            $(
+                #[doc = concat!("Type-level value that corresponds to `scale_bits::OrderFormat::", stringify!($ty), "` at run-time")]
+                #[doc = concat!(" and `bitvec::order::BitOrder::", stringify!($ty), "` at the type level.")]
+                #[derive(Clone, Debug, PartialEq, Eq)]
+                pub enum $ty {}
+                impl $ident for $ty {
+                    const FORMAT: OrderFormat = OrderFormat::$ty;
+                }
+            )*
+        };
+    }
+
+    store!(BitStore; (U8, u8), (U16, u16), (U32, u32), (U64, u64));
+    order!(BitOrder; Lsb0, Msb0);
+
+    /// Constructs a run-time format parameters based on the corresponding type-level parameters.
+    pub fn bit_format<Store: BitStore, Order: BitOrder>() -> Format {
+        Format {
+            order: Order::FORMAT,
+            store: Store::FORMAT,
+        }
+    }
+
+    /// `scale_bits::Bits` generic over the bit store (`u8`/`u16`/`u32`/`u64`) and bit order (LSB, MSB) types.
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct DecodedBits<Store: BitStore, Order: BitOrder>(
+        pub Bits,
+        pub PhantomData<Store>,
+        pub PhantomData<Order>,
+    );
+
+    impl<Store: BitStore, Order: BitOrder> codec::Decode for DecodedBits<Store, Order> {
+        fn decode<I: Input>(input: &mut I) -> Result<Self, codec::Error> {
+            /// Equivalent of `BitSlice::MAX_BITS` on 32bit machine.
+            const ARCH32BIT_BITSLICE_MAX_BITS: usize = 0x1fff_ffff;
+
+            let Compact(bits) = <Compact<u32>>::decode(input)?;
+            // Otherwise it is impossible to store it on 32bit machine.
+            if bits as usize > ARCH32BIT_BITSLICE_MAX_BITS {
+                return Err("Attempt to decode a BitVec with too many bits".into());
+            }
+            // NOTE: Replace with `bits.div_ceil(Store::BITS)` if `int_roundings` is stabilised
+            let bytes_needed =
+                (bits / Store::BITS) as usize + (bits % Store::BITS != 0) as usize;
+
+            // NOTE: We could reduce allocations if it would be possible to directly
+            // decode from an `Input` type using a custom format (rather than default <u8, Lsb0>)
+            // for the `Bits` type.
+            let mut storage = Vec::with_capacity(bytes_needed);
+            input.read(&mut storage)?;
+
+            let decoder = scale_bits::decode_using_format_from(
+                &storage,
+                bit_format::<Store, Order>(),
+            )?;
+            let bits = decoder.collect::<Result<Vec<_>, _>>()?;
+            let bits = Bits::from_iter(bits);
+
+            Ok(DecodedBits(bits, PhantomData, PhantomData))
+        }
+    }
+
+    impl<Store: BitStore, Order: BitOrder> codec::Encode for DecodedBits<Store, Order> {
+        fn size_hint(&self) -> usize {
+            self.0.size_hint()
+        }
+
+        fn encoded_size(&self) -> usize {
+            self.0.encoded_size()
+        }
+
+        fn encode(&self) -> Vec<u8> {
+            scale_bits::encode_using_format(self.0.iter(), bit_format::<Store, Order>())
+        }
+    }
+}
+
+pub use bits::{BitOrder, BitStore, DecodedBits, Lsb0, Msb0};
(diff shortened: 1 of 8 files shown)


subxt/src/utils.rs
```diff
@@ -96,3 +96,137 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
...
+            if bits as usize > ARCH32BIT_BITSLICE_MAX_BITS {
+                return Err("Attempt to decode a BitVec with too many bits".into())
+            }
+            // NOTE: Replace with `bits.div_ceil(Store::BITS)` if `int_roundings` is stabilised
+            let bytes_needed =
+                (bits / Store::BITS) as usize + (bits % Store::BITS != 0) as usize;
+
+            // NOTE: We could reduce allocations if it would be possible to directly
+            // decode from an `Input` type using a custom format (rather than default <u8, Lsb0>)
+            // for the `Bits` type.
```
jsdw, line 196: It definitely feels from this PR that we could improve `scale_bits::decode_using_format_from` to work with that `Input` trait rather than accepting an array of bytes; it'd be nice to be able to condense all of this decode logic into a few lines much like the `encode` version below, and the only reason that you can't is that the interface in scale-bits isn't quite right :)

subxt/src/utils.rs
```diff
@@ -96,3 +96,137 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
...
+    store!(BitStore; (U8, u8), (U16, u16), (U32, u32), (U64, u64));
+    order!(BitOrder; Lsb0, Msb0);
+
+    /// Constructs a run-time format parameters based on the corresponding type-level parameters.
+    pub fn bit_format<Store: BitStore, Order: BitOrder>() -> Format {
+        Format {
+            order: Order::FORMAT,
+            store: Store::FORMAT,
+        }
+    }
```
jsdw, line 170: does this need to be `pub`? (I don't think we need to expose it anywhere?)

subxt/src/utils.rs
```diff
@@ -96,3 +96,137 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
...
+}
+
+pub use bits::{
+    BitOrder,
+    BitStore,
+    DecodedBits,
+    Lsb0,
+    Msb0,
+};
```
jsdw, line 232: Probably it'd be good to have a few tests which assert that the decoded/encoded output lines up vs the BitVec equivalents (it's fine including bitvec as a dev dependency anyway).

Maybe something like a function which:
- takes an iterator of bools
- turn them into the various types above + their bitvec equivalents
- SCALE encode those all; make sure they line up
- decode again and make sure you get the expected output back

And then call that for some different bool arrays?

jsdw (commented):
Looks great; love the approach (shame that `scale-bits` doesn't provide the right interface to help with decoding, so we should fix that). Just a wee tweak and perhaps some tests to make sure that our encoding/decoding lines up with BitVec still to check compatibility.

Maybe with those tests especially it becomes worth turning `utils` into a folder and moving this `bits` to a separate file?

subxt/src/utils.rs
Xanewok: We certainly don't need to but if we're exposing bridge types for both `bitvec` and `scale-bits`, I thought it doesn't hurt to expose this convenience function as well. Should I hide it?

jsdw: I guess I just can't think of a use for it offhand; people can encode the decoded bits again back into the same format, or they can iterate over the Bits and encode to any other format, or what have you :)

Plus, I tend to lean towards keeping the API surface small I suppose just to keep our options open!

subxt/src/utils.rs
```diff
@@ -96,3 +96,137 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
...
+            store: Store::FORMAT,
+        }
+    }
+
+    /// `scale_bits::Bits` generic over the bit store (`u8`/`u16`/`u32`/`u64`) and bit order (LSB, MSB) types.
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct DecodedBits<Store: BitStore, Order: BitOrder>(
+        pub Bits,
+        pub PhantomData<Store>,
+        pub PhantomData<Order>,
```
jsdw, line 177: One thing I forgot to add a comment for; instead of those all being `pub`, how about having an `into_bits()` method or something to pull the `Bits` out, and otherwise hiding the internals?

Xanewok: Hiding the internals means hiding the default constructor and not being able to extract the wrapped value using a match pattern. Not saying it's greatly useful but I thought we wanted to be more transparent about this being a simple wrapper around `scale_bits::Bits`. Maybe we should add the `into_bits` convenience method but not hide the internals?

jsdw: That's true, but when would somebody need to match on it? I guess I think the `PhantomData` markers are a bit ugly to have as part of a public interface is all; perhaps they could be made not `pub` so that you can match on the `bits` while hiding those things?

subxt/src/utils.rs
jsdw: Opened an issue for this here: https://github.com/paritytech/scale-decode/issues/10

subxt/src/utils.rs
Xanewok: Unfortunately, hiding any member hides the constructor, so you can't do it, even with skipping the rest like so `let DecodedBits(bits, ..) = ...`. Do you mean matching on `value.0`?

Do you want me to only keep the functions from https://github.com/paritytech/subxt/pull/718/commits/2772f360aaee7139c789c7ef7cdecc6550326a10 and hide the internals?

subxt/src/utils.rs
Xanewok: Good call with tests! Turns out I had a bug or two... :sweat_smile: Added those in https://github.com/paritytech/subxt/pull/718/commits/f973fbc9e674e8816092e14ef094d82117f9044f

subxt/src/utils.rs
Xanewok: Sure, makes sense! Done in https://github.com/paritytech/subxt/pull/718/commits/9bdcd73ac129bcf86f3b5363c9821c8ec2826700

subxt/src/utils.rs
jsdw: Ah ok! I think personally I prefer the `into_bits()` and `to_bits()` approach then (my pov I guess is that I prefer to hide internals and limit the public API surface so that we are freer to make changes that don't break, plus `into_bits()` et al just feels consistent with various other parts of the subxt APIs) :)

Xanewok: Makes sense! Done in https://github.com/paritytech/subxt/pull/718/commits/8e58032c9e50e5b4696f9998b5516fa7feea3eed

subxt/src/utils/bits.rs
```diff
@@ -0,0 +1,227 @@
...
+        for test_case in &test_cases {
+            scales_like_bitvec_and_roundtrips::<u8, Lsb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u16, Lsb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u32, Lsb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u64, Lsb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u8, Msb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u16, Msb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u32, Msb0>(test_case);
+            scales_like_bitvec_and_roundtrips::<u64, Msb0>(test_case);
+        }
```
jsdw, line 225: Lovely stuff :)

subxt/src/utils/mod.rs
```diff
@@ -4,13 +4,23 @@
...
 };
 use derivative::Derivative;
 
+pub use bits::{
+    BitOrder,
+    BitStore,
+    DecodedBits,
+    Lsb0,
+    Msb0,
+};
```
jsdw, line 22: Nitpick: imo I wouldn't expose these here; make people go via `utils::bits` for clarity :)

jsdw (approved):
Looks great! One tiny nit about `util::bits` imports but great stuff!

subxt/src/utils/bits.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! Generic `scale_bits` over `bitvec`-like `BitOrder` and `BitFormat` types.
+use codec::{
```
niklasad1, line 6:
```suggestion

use codec::{
```

subxt/src/utils/bits.rs
```diff
@@ -0,0 +1,227 @@
...
+    },
+    Bits,
+};
+use std::marker::PhantomData;
+
+macro_rules! store {
+        ($ident: ident; $(($ty: ident, $wrapped: ty)),*) => {
+            /// Associates `bitvec::store::BitStore` trait with corresponding, type-erased `scale_bits::StoreFormat` enum.
+            ///
+            /// Used to decode bit sequences in runtime by providing `scale_bits::StoreFormat` using
```
niklasad1, line 24: I don't really understand this sentence, can runtime be omitted? 

```suggestion
            /// Used to decode bit sequences by providing `scale_bits::StoreFormat` using
```

subxt/src/utils/bits.rs
```diff
@@ -0,0 +1,227 @@
...
+);
+
+impl<Store: BitStore, Order: BitOrder> DecodedBits<Store, Order> {
+    /// Extracts the underlying `scale_bits::Bits` value.
+    pub fn into_bits(self) -> Bits {
+        self.0
+    }
+
+    /// References the underlying `scale_bits::Bits` value.
+    pub fn to_bits(&self) -> &Bits {
```
niklasad1, line 92: I would like `as_bits` better :)

subxt/src/utils/bits.rs
```diff
@@ -0,0 +1,227 @@
...
+{
+    fn from_iter<T: IntoIterator<Item = bool>>(iter: T) -> Self {
+        DecodedBits(Bits::from_iter(iter), PhantomData, PhantomData)
+    }
+}
+
+impl<Store: BitStore, Order: BitOrder> codec::Decode for DecodedBits<Store, Order> {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, codec::Error> {
+        /// Equivalent of `BitSlice::MAX_BITS` on 32bit machine.
+        const ARCH32BIT_BITSLICE_MAX_BITS: usize = 0x1fff_ffff;
```
niklasad1, line 108: I think `bits` below is already u32 and coukd remove some `as` casts I guess

```suggestion
        const ARCH32BIT_BITSLICE_MAX_BITS: u32 = 0x1fff_ffff;
```

niklasad1 (approved):
nice work, looks good 

I have some minor comments which are optional to fix :P

subxt/src/utils/bits.rs
Xanewok: I wanted to emphasize that StoreFormat is a dynamic configuration whereas the type parameters are compile-time but you're right, sticking runtime in here doesn't fit that much, since all the decoding is done in runtime anyway, regardless of the type of the decoding parameters. Will remove it, thanks!

subxt/src/utils/bits.rs
Xanewok: That's fair, simplified a bit with this turned into `u32`. Thanks!

PR #723, opened 2022-11-21
Add and adapt a publish.sh utility script from jsonrpsee

### LINKED ISSUE #428: Adapt script from jsonrpsee to help automate publishing steps
jsonrpsee has a nice script to help publish each of the crates:

https://github.com/paritytech/jsonrpsee/blob/master/.scripts/publish.sh

Perhaps we should adapt and adopt it for publishing here also?

Xanewok:
Closes #428

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -3,6 +3,7 @@ name = "subxt-cli"
 version = "0.25.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
+publish = true
 
 license = "Apache-2.0 OR GPL-3.0"
 repository = "https://github.com/paritytech/subxt"
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -12,8 +12,8 @@ ORDER=(metadata codegen macro subxt cli)
 function read_toml () {
 	NAME=""
 	VERSION=""
-	# There can be multiple "name = ..." lines, so make sure to only catch the first one after the "[package]" section
-	NAME=$(grep -Pzo "\[package\]\\n([^name].*\\n)*name\ *=\ *\"(.*)\"" ./Cargo.toml | tr -d '\000' | grep -a name | sed -e 's/.*"\(.*\)"/\1/')
+	# Extract and parse the "name = ..." line that belongs to the [package] section
+	NAME=$(grep -e "\[package\]" -e 'name*=*' ./Cargo.toml | grep -A1 "\[package\]" | tail -n 1 |  sed -e 's/.*"\(.*\)"/\1/')
 	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
 }
 function remote_version () {
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -93,7 +93,7 @@ for CRATE_DIR in ${ORDER[@]}; do
 	cd - > /dev/null
 done
 
-echo "Tagging jsonrpsee@$VERSION"
+echo "Tagging subxt@$VERSION"
 set -x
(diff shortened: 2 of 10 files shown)


scripts/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Find publishable crates by running something like this below and figure out the topological order:
+# $ find . -iname 'Cargo.toml' | xargs grep "publish\ *=\ *true" | grep "^.*\.toml" -o
+ORDER=(metadata codegen macro subxt cli)
+
+function read_toml () {
+	NAME=""
+	VERSION=""
+	# There can be multiple "name = ..." lines, so make sure to only catch the first one after the "[package]" section
+	NAME=$(grep -Pzo "\[package\]\\n([^name].*\\n)*name\ *=\ *\"(.*)\"" ./Cargo.toml | tr -d '\000' | grep -a name | sed -e 's/.*"\(.*\)"/\1/')
```
Xanewok, line 16: This was specifically done because there's a `name = ...` section under `[[bench]]` that this script caught, so I tried to apply the heuristic outlined in the comment to filter out any unwanted names.

cli/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "subxt-cli"
 version = "0.25.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
+publish = true
```
niklasad1, line 6: :+1:

scripts/publish.sh
niklasad1: it is not allowed to have multiple `name` fields in a package:

```toml
[package]
name = "foo"
fields = "..."

# package name = "foo" ends here 
[dependency]
bar = "1"
```

I guess the regex would be even more complicated than that and it might even be better to re-write in rust

scripts/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
```
niklasad1, line 17: I reckon that this doesn't work for `alpha.1`, `rc` and similar

niklasad1: I was wrong it works

niklasad1 (approved):

scripts/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+			sleep 3
+			break
+		else
+			echo "#### Got $NAME@$REMOTE_VERSION but expected $NAME@$VERSION. Retrying..."
+		fi
+	done
+	cd - > /dev/null
+done
+
+echo "Tagging jsonrpsee@$VERSION"
```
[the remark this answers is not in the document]
Xanewok: Oh, whoops! Good catch!

jsdw (approved):
Has the script been tested on mac, linux or both? I can see it's def pushing my internal "bash complexity budget", but happy that we can cross that bridge if it gives us grief. On the other hand it looks like it covers all of the bases offhand; nice one! :)

scripts/publish.sh
```diff
@@ -0,0 +1,100 @@
...
+set -x
+git tag -a v$VERSION -m "Version $VERSION"
```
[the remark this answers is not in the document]
Xanewok: Makes sense, thanks!

lexnv (approved):

Xanewok:
I was testing this on my Linux setup but turns out that the grep `-P` was not portable; fixed and tested on both Linux and macOS

scripts/publish.sh
Xanewok: I meant multiple `name = ...` lines in the file in general, not in the package section. I rephrased that, hopefully it's more clear now

Issue #725, opened 2022-11-22
Fetching events from blocks previous than the runtime upgrade Kusama/9320 is broken


paulormart:
After the last runtime upgrade on Kusama/9320 is no longer possible to fetch events from blocks before the runtime upgrade.

When fetching for the events at block [15426015](https://kusama.subscan.io/block/15426015?tab=log) where the runtime upgrade was enacted we get the following error:

```
Error: Codec(Error { cause: None, desc: "out of range decoding Compact<u32>" })
```

For any block number smaller than 15426015 we get the following error:

```
Error: DecodeValue(VariantNotFound(161, TypeDefVariant { variants: [Variant { name: "Normal", fields: [], index: 0, docs: [] }, Variant { name: "Operational", fields: [], index: 1, docs: [] }, Variant { name: "Mandatory", fields: [], index: 2, docs: [] }] }))
```

Below is just the example to demonstrate the error when fetching a block:

```
use futures::StreamExt;
use sp_keyring::AccountKeyring;
use std::time::Duration;
use subxt::{rpc::BlockNumber, tx::PairSigner, OnlineClient, PolkadotConfig};

#[subxt::subxt(
    runtime_metadata_url = "wss://kusama-rpc.polkadot.io:443",
    derive_for_all_types = "PartialEq, Clone"
)]
pub mod polkadot {}

/// Subscribe to all events, and then manually look through them and
/// pluck out the events that we care about.
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    tracing_subscriber::fmt::init();

    // Create a client to use:
    // let api = OnlineClient::<PolkadotConfig>::new().await?;
    let api =
        OnlineClient::<PolkadotConfig>::from_url("wss://kusama-rpc.polkadot.io:443")
            .await?;

    let block_number: u32 = 15426014;
    let block_hash = api.rpc().block_hash(Some(block_number.into())).await?;
    let events = api.events().at(block_hash).await?;
    if let Some(event) =
    events.find_first::<polkadot::system::events::ExtrinsicSuccess>()?
    {
        println!("{:?}", event);
    }

    Ok(())
}
```
Output: 
```
Error: DecodeValue(VariantNotFound(161, TypeDefVariant { variants: [Variant { name: "Normal", fields: [], index: 0, docs: [] }, Variant { name: "Operational", fields: [], index: 1, docs: [] }, Variant { name: "Mandatory", fields: [], index: 2, docs: [] }] }))
```

[...]

PR #727, opened 2022-11-22
events: Fetch metadata at arbitrary blocks

lexnv:
The events are using the client metadata for dynamic decoding.
However, the users are allowed to inspect the events at an arbitrary block.
If the arbitrary block is a block where a runtime upgrade happened that
changed the metadata content, the dynamic decoding of the events
would fail.

To mitigate this scenario, expose the ability to query the metadata at
an arbitrary block, and fetch the metadata of a block when fetching the
block events.

This ensures that dynamic decoding of events will succeed if an arbitrary
block hash is provided.

The client is encouraged to keep the metadata in sync via subscriptions
to the runtime upgrade.

Closes: https://github.com/paritytech/subxt/issues/725

### DIFF
--- a/subxt/src/events/events_client.rs
+++ b/subxt/src/events/events_client.rs
@@ -55,14 +55,26 @@ where
         async move {
             // If block hash is not provided, get the hash
             // for the latest block and use that.
-            let block_hash = match block_hash {
-                Some(hash) => hash,
+            let (block_hash, metadata) = match block_hash {
+                Some(hash) => {
+                    // If the block hash is provided, the hash can be of an older
+                    // block where the metadata was different. Fetch the metadata
+                    // at the specific block to ensure proper dynamic decoding of
+                    // events.
+                    (hash, client.rpc().metadata(Some(hash)).await?)
+                }
                 None => {
-                    client
-                        .rpc()
-                        .block_hash(None)
-                        .await?
-                        .expect("didn't pass a block number; qed")
+                    // If the block hash was not provided, the metadata is
+                    // extracted from the client and is presumed to be up to
+                    // date (ie client should subscribe to the runtime upgrades).
+                    (
+                        client
+                            .rpc()
+                            .block_hash(None)
+                            .await?
+                            .expect("didn't pass a block number; qed"),
+                        client.metadata(),
+                    )
                 }
             };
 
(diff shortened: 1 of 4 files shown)


niklasad1 (approved):

jsdw:
One reason I've been pretty hesitant about introducing this sort of change (and instead just warning users) is that it's quite a lot of overhead to download a big metadata for every single block requested in this way. It's also v hard to cache (and I wouldn't want to cache a ton of metadata blobs per block or something).

Given the new RPC API that you're working on too, I wonder whether this sort of change would be compatible or not; I can see us eventually going the way of removing the ability to query previous block numbers at all, and providing either "archive" methods to "get anything you want" or "chainHead" things to follow and be able to work with anything at the head of the chain.

jsdw (changes requested):
Will mark as "request changes" because I'm quite hesitant about this being merged for the above reason (I posted on the issue also to elaborate) :)

subxt/src/events/events_client.rs
```diff
@@ -73,7 +85,7 @@ where
                 .map(|e| e.0)
                 .unwrap_or_else(Vec::new);
 
-            Ok(Events::new(client.metadata(), block_hash, event_bytes))
+            Ok(Events::new(metadata, block_hash, event_bytes))
```
jsdw, line 88: Maybe the solution really is that we allow the user to manually do the above and obtain metadata for the older block if they are in doubt, and ensure that they can construct this `Events` object.

It might be nicer then to move the `event_bytes` stuff into `Events::new` and make that async or something, so that the interface is simpler (they provide metadata and block hash and get events back for it).

niklasad1: I recall that comparing metadata is super expensive as well, maybe there is some simple way to find which runtime version a block belongs to without using some block explorer API. So how would I know if a block is "old" i.e. need to call `metadata`? I suppose one could do some simple calculation based on that 14 400 blocks per day is produced * 30 or something.

My point is that comparing metadata is superslow as well then fetching metadata could make sense to just be naive and always fetch the metadata regardless.....  so folks doesn't end up with decoding errors.

However, I agree it's probably better to have different APIs or take `Option<Metadata>` as input in `fn at` i.e, if None just use the latest.

subxt/src/events/events_type.rs
```diff
@@ -64,6 +66,50 @@ impl<T: Config> Events<T> {
         }
     }
 
+    /// Obtain the events from a block hash, custom metadata and provided client.
```
jsdw, line 69:
```suggestion
    /// Obtain the events from a block hash given custom metadata and a client.
```

jsdw (approved):
Nice one, looks good to me :)

Issue #732, opened 2022-11-24
Trial support for new RPC API


lexnv:
Create a PoC for integrating the new [RPC Spec V2 API](https://github.com/paritytech/json-rpc-interface-spec/) into subxt.

The integration will provide further user feedback on the new API, and identify its strengths and weaknesses.

The implementation should leverage the newly exposed APIs from the [chainHead](https://github.com/paritytech/substrate/pull/12544).

Provide a user ergonomic API for subscribing to [chainHead_unstable_follow](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_follow.html#chainhead_unstable_follow) notifications.

The follow subscription must be extended with the ability to:
- query storage - [chainHead_unstable_storage](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_storage.html#chainhead_unstable_storage)
- fetch body - [chainHead_unstable_body](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_body.html#chainhead_unstable_body)
- fetch header - [chainHead_unstable_header](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_header.html#chainhead_unstable_header)
- call into the runtime API on the provided blocks -  [chainHead_unstable_call](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_call.html#chainhead_unstable_call)

As well as dropping the provided blocks using the [chainHead_unstable_unpin](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_unpin.html#chainhead_unstable_unpin) method to avoid suspending the follow subscription upon exceeding the number of pinned blocks.

jsdw:
Just to add more detail for how I think we should go about this:

1. In `rpc.rs`, remove all references to the "old" RPC methods, and only expose new methods (don't have an "unstable" prtefix in thier names.
2. The interfaces for these will try to match that of the new RPC API as closely as possible.
3. The implementations of these will:
    - if an `unstable-rpc-v2` feature flag is enabled, directly call the new RPC methods
    - if not, we will "shim" them by calling the old methods instead and converting the results into the format expected for the new APIs. 
    - Things like "pinning" can be exposed but will simply be a no-op when `unstable-rpc-v2` is disabled
    - We may run into the odd thing that the new API makes possible which is not possible via the old API. For now, we will hide such things and only expose behaviours we can shim using the old APIs.
[...]

PR #733, opened 2022-11-24
rpc: Expose the `subscription ID` for `RpcClientT`

lexnv:
Extent the `RpcClientT` to obtain the ID associated with subscriptions.

To ensure that this extension is not impacting users that do not want to
implement/return it, the subscription ID is optional.

The subscription ID needs to be exposed for implementing https://github.com/paritytech/subxt/issues/732.

### DIFF
--- a/subxt/src/rpc/rpc_client_t.rs
+++ b/subxt/src/rpc/rpc_client_t.rs
@@ -52,15 +52,23 @@ pub trait RpcClientT: Send + Sync + 'static {
         sub: &'a str,
         params: Option<Box<RawValue>>,
         unsub: &'a str,
-    ) -> RpcFuture<'a, (RpcSubscription, Option<RpcSubscriptionId>)>;
+    ) -> RpcFuture<'a, RpcSubscription>;
 }
 
 /// A boxed future that is returned from the [`RpcClientT`] methods.
 pub type RpcFuture<'a, T> =
     Pin<Box<dyn Future<Output = Result<T, RpcError>> + Send + 'a>>;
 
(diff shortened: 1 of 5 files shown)


jsdw (approved):
LGTM!

subxt/src/rpc/rpc_client_t.rs
```diff
@@ -62,3 +62,6 @@ pub type RpcFuture<'a, T> =
 /// The inner subscription stream returned from our [`RpcClientT`]'s `subscription` method.
 pub type RpcSubscription =
     Pin<Box<dyn Stream<Item = Result<Box<RawValue>, RpcError>> + Send + 'static>>;
+
+/// The ID associated with the [`RpcClientT`]'s `subscription`.
+pub type RpcSubscriptionId = String;
```
niklasad1, line 67: This is configurable in substrate via the `SubscriptionIdProvider` so it could something different than String. Should be String 99% of cases anyway but just saying.

Does the new `rpc api v2` specify the format of the subscription ID? :)

niklasad1 (approved):

subxt/src/rpc/rpc_client_t.rs
lexnv: The RPC layer should take the `subscriptionID` as a parameter in string format. From [chainHead_unstable_body](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_body.html#chainhead_unstable_body): 

> followSubscription: An opaque string that was returned by chainHead_unstable_follow.

jsdw (approved):

Xanewok (approved):

PR #735, opened 2022-12-01
Allow for remapping type parameters in type substitutions

### LINKED ISSUE #666: Allow substitute types to take on 0, 1 or all of the generic params from the types they stand in for
See https://github.com/paritytech/subxt/pull/627 for initial details.

Basically, right now we can do something like:

```rust
#[subxt::subxt(runtime_metadata_path = "./metadata.scale")]
pub mod node {
    #[subxt(substitute_type = "interbtc_primitives::VaultId")]
    use crate::DefaultVaultId;
}
```

The effect of this is that instead of generating/using an `interbtc_primitives::VaultId` type, we'd use the `crate::DefaultValueId` instead.

If the `interbtc_primitives::VaultId` type has generic params, so it would look like eg `struct VaultId<A, B, C>`, then all of those generic params are assumed to exist on the substitute type; we'd expect a `struct DefaultValueId<A, B, C>`.

It would be useful to have more flexibility over how the generic parameters are mapped to the substitute type though (see #627 for one such issue statement).

Suggested syntax ideas:

## 1. pattern matching on generics.

```rust
#[subxt::subxt(runtime_metadata_path = "./metadata.scale")]
pub mod node {
    // The default case; any generic parameters are applied as-is to the substitute type:
    #[subxt(substitute_type = "interbtc_primitives::VaultId")]
    use crate::DefaultVaultId;

    // Assert that `interbtc_primitives::VaultId` has 3 generic params. We want to ignore them all:
    #[subxt(substitute_type = "interbtc_primitives::VaultId<_, _, _>")]
    use crate::DefaultVaultId;

    // Assert that `interbtc_primitives::VaultId` has 3 generic params. Perform the mapping
    // `interbtc_primitives::VaultId<_2, X, _1>` => `DefaultVaultId<_1, _2>`
    // Ie, pattern match on VauldId generics and use numbers to map them to the output position.
    #[subxt(substitute_type = "interbtc_primitives::VaultId<_2, _, _1>")]
    use crate::DefaultVaultId;
}
```

Advantages:
- Close to current syntax.
- Rust-like generic syntax might feel familiar.
- Built-in assertion that the input type has some number of generic params; we can error if this expectation changes.

Disadvantages:
- Personally I find it hard to parse/read the third case. 
- Not obvious that numbers mean something.

## 2. Add a "params" attr.

```rust
#[subxt::subxt(runtime_metadata_path = "./metadata.scale")]
pub mod node {
    // The default case; any generic parameters are applied as-is to the substitute type:
    #[subxt(substitute_type = "interbtc_primitives::VaultId")]
    use crate::DefaultVaultId;

    // We want to ignore all of the generic params and apply none of them to DefaultVaultId
    #[subxt(substitute_type = "interbtc_primitives::VaultId", substitute_params = "")]
    use crate::DefaultVaultId;

    // Use the "params" attr to map by position in VaultId to substitute param order.
    #[subxt(substitute_type = "interbtc_primitives::VaultId", substitute_params = "2, 1")]
    use crate::DefaultVaultId;
}
```

Advantages:
- Perhaps a little easier to read than the above?

## 3. use a new attribute to perform substitution

```rust
#[subxt::subxt(runtime_metadata_path = "./metadata.scale")]
#[subxt(substitute(input = "interbtc_primitives::VaultId", output = "crate::DefaultVaultId"))]
#[subxt(substitute(input = "interbtc_primitives::VaultId<A,B,C>", output = "crate::DefaultVaultId"))]
#[subxt(substitute(input = "interbtc_primitives::VaultId<A,B,C>", output = "crate::DefaultVaultId<B, A>"))]
pub mod node { }
```

Advantages:
- IMO the clearest sort of approach; we can be clear and obivous with how we are mapping since we control input and output paths fully.

Disadvantages:
- Strays from the current approach, which there may have been a good reason for using in the first place.

@ascjones @gregdhill I'd love to get your feedback; do you like the idea, and if so, do you have a preference on which syntax to go for, or reasons against any?

Xanewok:
Closes #666 

For now, this moves type substitution annotations from the use statements outside, to the `subxt` macro + separates some substitution logic. Opening as a draft for now to get feedback; next thing to work on is to implement the generic wrangling/reordering mentioned in #666.

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -18,6 +18,7 @@ use crate::{
         CompositeDef,
         CompositeDefFields,
         TypeGenerator,
+        TypeSubstitutes,
     },
     utils::{
         fetch_metadata_bytes_blocking,
@@ -39,7 +40,6 @@ use quote::{
     quote,
 };
 use std::{
-    collections::HashMap,
     fs,
     io::Read,
     path,
@@ -54,6 +54,7 @@ use syn::parse_quote;
 /// * `item_mod` - The module declaration for which the API is implemented.
 /// * `path` - The path to the scale encoded metadata of the runtime node.
 /// * `derives` - Provide custom derives for the generated types.
+/// * `type_substitutes` - Provide custom type substitutes.
 /// * `crate_path` - Path to the `subxt` crate.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
@@ -61,6 +62,7 @@ pub fn generate_runtime_api_from_path<P>(
     item_mod: syn::ItemMod,
     path: P,
     derives: DerivesRegistry,
+    type_substitutes: TypeSubstitutes,
     crate_path: CratePath,
 ) -> TokenStream2
 where
@@ -74,7 +76,13 @@ where
     file.read_to_end(&mut bytes)
         .unwrap_or_else(|e| abort_call_site!("Failed to read metadata file: {}", e));
 
-    generate_runtime_api_from_bytes(item_mod, &bytes, derives, crate_path)
+    generate_runtime_api_from_bytes(
+        item_mod,
+        &bytes,
+        derives,
+        type_substitutes,
+        crate_path,
+    )
 }
 
 /// Generates the API for interacting with a substrate runtime, using metadata
@@ -86,6 +94,7 @@ where
 /// * `item_mod` - The module declaration for which the API is implemented.
 /// * `url` - HTTP/WS URL to the substrate node you'd like to pull metadata from.
 /// * `derives` - Provide custom derives for the generated types.
+/// * `type_substitutes` - Provide custom type substitutes.
 /// * `crate_path` - Path to the `subxt` crate.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
@@ -93,12 +102,19 @@ pub fn generate_runtime_api_from_url(
     item_mod: syn::ItemMod,
     url: &Uri,
     derives: DerivesRegistry,
+    type_substitutes: TypeSubstitutes,
     crate_path: CratePath,
 ) -> TokenStream2 {
     let bytes = fetch_metadata_bytes_blocking(url)
         .unwrap_or_else(|e| abort_call_site!("Failed to obtain metadata: {}", e));
 
-    generate_runtime_api_from_bytes(item_mod, &bytes, derives, crate_path)
+    generate_runtime_api_from_bytes(
+        item_mod,
+        &bytes,
+        derives,
+        type_substitutes,
+        crate_path,
+    )
 }
 
 /// Generates the API for interacting with a substrate runtime, using metadata bytes.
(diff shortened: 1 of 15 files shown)


macro/src/lib.rs
```diff
@@ -153,6 +167,19 @@ pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
         )
     }
 
+    let mut type_substitutes = TypeSubstitutes::new(&crate_path);
+    type_substitutes.extend(args.substitute_type.into_iter().map(
+        |SubstituteType { ty, with }| {
+            (
+                ty,
+                with.try_into()
```
Xanewok, line 175: This can be quite cryptic, but I didn't want to import the `AbsoluteTypePath` newtype wrapper that I introduced for the validation purpose (rather than it being a more well-rounded/useful type on its own), to name it directly here.

Xanewok:
Squashed and force-pushed now after #760 landed.

Implements variant specified in https://github.com/paritytech/subxt/issues/666#issuecomment-1259302669 (so uses new syntax as well). This now includes the ability to strip generics altogether for a source type or taking any combination of parameters from the source type.

This lacks the ability to specify concrete types becase we'd need to plug into type resolution as part of the substitution but since this PR refactored type substitution and made it more self-contained and introduced the new mapping feature, I figured we could do that in a follow-up PR.

codegen/src/types/substitutes.rs
```diff
@@ -0,0 +1,237 @@
...
+    parse_quote,
+    spanned::Spanned as _,
+};
+
+use super::TypePath;
+
+#[derive(Debug)]
+pub struct TypeSubstitutes {
+    inner: HashMap<PathSegments, syn::Path>,
+    params: HashMap<PathSegments, TypeParamMapping>,
```
jsdw, line 21: So I think at the moment, for some type substitute, we can look up the param mapping in the `params` entry. So either `params.get(path) == None` or `params.get(path) == Some(TypeParamMapping::None)` I guess mean the same?

I wonder whether it makes sense to merge these so that we have more like:

```rust
inner: HashMap<PathSegments, (syn::Path, TypeParamMapping)>
```

And then for any path, we always have a type param mapping which is either `TypeParamMapping::None` or `Specified`. I geuss all of the "default" type substitutes would then have `None` to mean "copy over any types as is"

codegen/src/api/mod.rs
```diff
@@ -115,13 +131,14 @@ pub fn generate_runtime_api_from_bytes(
     item_mod: syn::ItemMod,
     bytes: &[u8],
     derives: DerivesRegistry,
+    type_substitutes: TypeSubstitutes,
```
lexnv, line 134: Super small nit: missing 
```/// * `type_substitutes` - Provide custom type substitutes.```
 on the doc.

Tho, I'm wondering if we should extend this documentation style to our repo, or just remove those `# Arguments` in the future. I guess the easiest would be later since we already have some comprehensive docs.

lexnv (approved):

codegen/src/types/substitutes.rs
Xanewok: Yeah, this makes sense; I forgot to merge those two in the final PR form. Thanks for applying the fix yourself!

codegen/src/api/mod.rs
jsdw: Good spot!

I added the extra param for now, though I'd note that these comments aren't exposed publically anywhere and so in general I'd be inclined to keep them simple as they are only for us devs who work on it anyway :)

jsdw (approved):

PR #748, opened 2022-12-15
cargo: Update `sp-*` crates from v7.0.0 to v9.0.0

lexnv:
Update the `sp-core`, `sp-keyring` and `sp-runtime` crates to the latest v8.0.0 version.

The crates need a manual update to properly sync all dependencies, as noted by
dependabot in https://github.com/paritytech/subxt/pull/747, https://github.com/paritytech/subxt/pull/746 and https://github.com/paritytech/subxt/pull/745.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -185,8 +185,8 @@ jobs:
     - name: Install chrome
       uses: browser-actions/setup-chrome@latest
 
-    # - name: Rust Cache
-    #   uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+    - name: Rust Cache
+      uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
 
     - name: Download Substrate
       run: |
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -185,8 +185,8 @@ jobs:
     - name: Install chrome
       uses: browser-actions/setup-chrome@latest
 
-    - name: Rust Cache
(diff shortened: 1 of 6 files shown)


niklasad1:
@lexnv looks like the build script in `wasmtime-runtime` fails, is there something weird going with our C toolchain in the CI?

lexnv:
Yep, there seems to be an issue with the `wasmtime-runtime` CI step. It seems that there is one file missing needed for compilation, it might be possible that is not included in the environment path: 

```bash
warning: src/helpers.c:1:10: fatal error: 'setjmp.h' file not found
warning: #include <setjmp.h>
```

niklasad1:
I just checked that `wasmtime-runtime` wasn't part of the dependency tree before and it seems not compile for WASM.

https://crates.io/crates/sp-wasm-interface/8.0.0/dependencies is the cause that `wasmtime-runtime` is brought into the dependency tree but I don't follow we are using `no-default-features = true` is it a bug in substrate?

niklasad1:
wasmtime is hard-dependency in substrate now which causes the wasm stuff to not compile anymore... 

see https://github.com/paritytech/substrate/pull/12684/files#diff-c2654579a98c27b848ec78860417cbec4375ed7c1dc8df990d77c6ea6f68428cR26

jsdw:
Ok, so summary so far:

- Removing wasmtime means disabling the "std" feature (ie `default-features = false`).
- But disabling this features in Substrate makes it assume we want a WASM build so a bunch of other stuff goes wrong.
- I'll experiment with removing our dependency on substrate crates and see how much of a pain (or not) that'll be.
- I'm sortof keen on this anyway if we can do it without copying too much code (ie just rely on hashing libs directly and make some traits on our side that it's fairly easy to impl for substrate things etc). It may let us streamline and simplify stuff.

jsdw:
I'll close this now since it's been superceded by #760

PR #758, opened 2022-12-26
Make verbose rpc error display

SabaunT:
Just a small PR with changes very need for my tiny tool.

### DIFF
--- a/subxt/src/error.rs
+++ b/subxt/src/error.rs
@@ -114,7 +114,7 @@ pub enum RpcError {
     #[error("RPC error: {0}")]
     ClientError(Box<dyn std::error::Error + Send + Sync + 'static>),
     /// The RPC subscription dropped.
-    #[error("RPC error: Subscription dropped.")]
+    #[error("RPC error: subscription dropped.")]
     SubscriptionDropped,
 }
 
--- a/subxt/src/error.rs
+++ b/subxt/src/error.rs
@@ -111,10 +111,10 @@ pub enum RpcError {
     // Dev note: We need the error to be safely sent between threads
     // for `subscribe_to_block_headers_filling_in_gaps` and friends.
     /// Error related to the RPC client.
-    #[error("{0}")]
+    #[error("RPC error: {0}")]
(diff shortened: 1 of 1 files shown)


SabaunT:
@jsdw @niklasad1 hey guys, can I ask you for a quick review?

subxt/src/error.rs
```diff
@@ -107,13 +107,14 @@ impl From<DispatchError> for Error {
 /// An RPC error. Since we are generic over the RPC client that is used,
 /// the error is boxed and could be casted.
 #[derive(Debug, thiserror::Error)]
-#[error("RPC error")]
 pub enum RpcError {
     // Dev note: We need the error to be safely sent between threads
     // for `subscribe_to_block_headers_filling_in_gaps` and friends.
     /// Error related to the RPC client.
+    #[error("{0}")]
```
niklasad1, line 114: I don't remember if all these errors actually states that it's a `RpcError`.

Might be better to add  `#[error("RPC error: {0}")]` or something?

subxt/src/error.rs
```diff
@@ -107,13 +107,14 @@ impl From<DispatchError> for Error {
...
     ClientError(Box<dyn std::error::Error + Send + Sync + 'static>),
     /// The RPC subscription dropped.
+    #[error("Subscription dropped.")]
```
niklasad1, line 117:
```suggestion
    #[error("RPC error: subscription dropped.")]
```

subxt/src/error.rs
jsdw: Since the error is very generic and could be anything, I'd agree with soomething like the above :)

SabaunT: done

subxt/src/error.rs
SabaunT: done

niklasad1 (approved):
LGTM, thank you

lexnv (approved):

Xanewok:
This is miniscule and improves error display, so I'm merging this.

PR #760, opened 2023-01-06
Make sp_core and sp_runtime dependencies optional, and bump to latest

jsdw:
This PR makes `sp_core` and `sp_runtime` dependencies optional (opted-in by default). The reasons for doing this are:

- Supporting compile-to-WASM is simpler without these dependencies (currently they have introduced a hard dependency on wasmtime (under the "std" feature) which cannot be compiled to WASM, but disabling the "std" feature hits a bunch of assumptions about no_std which breaks other things.
- A smaller dependency graph: for the most part we only rely on certain types from these packages to describe the shape of some responses and such, so taking these types into Subxt allows us to not pull in a load of other stuff that we don't need. 
- Reducing the need to keep up to date with Substrate packlage versions: in the past, Substrate dependencies like parity-util-mem have prevented Subxt from being used in conjunction with Substrate packages unless the two used the same versions. We aren't there yet, but I'd like to get to the point where Subxt will happily co-exist with a wide range of Substrate versions, as long as the actual types sent back and forth via the RPC APIs remain compatible.

At the time of writing, the signer stuff still lives in Substrate and so we still depend on sp_core and sp_runtime to provide a `PairSigner` which allows transactions to be signed (which won't be available for now in the WASM builds as a result).  This isn't ideal, but porting that logic over seemed more involved offhand and so I opted to avoid it in this first pass. Providing a simple set of signing primitives feels like a good next step though, and would allow us to completely shed our dependencies on sp_core and sp_runtime (and everything that they bring in).


jsdw:
Ok so there is a feature flag, "substrate-compat", which when enabled will bring in `sp_core` and `sp_runtime` in order that we can provide a `PairSigner` to help sign transactions and be compatible with things like `AccountKeyring` and such from Substrate. The feature flag is enabled by default to try and minimise breakage (perhaps we'll want to make it opt-in in future).

The default `SubstrateConfig` and `PolkadotConfig` use our own versions of `AccountId32` etc. The `PairSigner` will work so long as we can convert from the substrate AccountId32 etc into the ones we configure (this is true for our new built-in copies of `AccountId32` etc).

Removing Substrate deps seems to shrink the total deps quite a lot:

Using a command like `cargo tree --no-default-features -e no-dev,no-build | grep -v '*' | grep -v 'subxt v0.25.0' | wc -l`:

Deps with no default features: ~237
Deps with sp_core/sp_runtime included: ~454

subxt/src/tx/signer.rs
```diff
@@ -29,55 +24,82 @@ pub trait Signer<T: Config> {
...
 
-    /// Return the account ID.
-    pub fn account_id(&self) -> &T::AccountId {
-        &self.account_id
-    }
-}
+    impl<T, Pair> PairSigner<T, Pair>
+    where
+        T: Config,
+        T::AccountId: From<SpAccountId32>,
```
jsdw, line 56: You create a PairSigner by passing a Pair implementation whose signature type can provide an `sp_runtime::MultiSignature`, where that has a Signer type that can provide back an `sp_runtime::AccountId32`. You must be able to convert `sp_runtime::AccountId32` into our configured account type, `T::AccountId`.

Below you can see that you also need to be able to convert a `Pair::Signature`into an `sp_runtime::MultiSignature`, and take that into our configured `T::Signature` type.

In other words, we use the substrate types as a go-between. I'm hoping that this won't limit the use cases for this type, but I'm not sure offhand so would love to hear any feedback or thoughts here!

jsdw: I'll see how hard it is to impl the substrate signer stuff on MultiSignature etc and then maybe we can avoid any intermediate types.

niklasad1:
> Reducing the need to keep up to date with Substrate packlage versions: in the past, Substrate dependencies like parity-util-mem have prevented Subxt from being used in conjunction with Substrate packages unless the two used the same versions. We aren't there yet, but I'd like to get to the point where Subxt will happily co-exist with a wide range of Substrate versions, as long as the actual types sent back and forth via the RPC APIs remain compatible

I don't understand this comment, you bumped the substrate dependencies so `parity-util-mem` isn't part of the dependency tree anymore. Are you referring to "old substrate dependencies" or what do you mean?

subxt/src/config/mod.rs
```diff
@@ -34,7 +35,6 @@ pub trait Config: 'static {
         + Member
         + serde::de::DeserializeOwned
         + Default
-        + AtLeast32Bit
```
niklasad1, line 37: why removed? I guess that trait comes from substrate

subxt/src/rpc/rpc.rs
```diff
@@ -56,25 +56,28 @@ use codec::{
...
 };
 
+/// Signal what the result of doing a dry run of an extrinsic is.
+pub type DryRunResult = Result<(), DryRunError>;
+
+/// An error dry running an extrinsic.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum DryRunError {
+    /// The extrinsic will not be included in the block
+    TransactionValidityError,
```
niklasad1, line 76: it's a bummer not to get back the reason why the transaction was rejected... but perhaps a PITA to get

subxt/src/utils/multi_address.rs
```diff
@@ -0,0 +1,65 @@
...
+//! The "default" Substrate/Polkadot Address type. This is used in codegen, as well as signing related bits.
+//! This doesn't contain much functionality itself, but is easy to convert to/from an `sp_runtime::MultiAddress`
+//! for instance, to gain functionality without forcing a dependency on Substrate crates here.
+
+use codec::{
+    Decode,
+    Encode,
+};
+
+/// A multi-format address wrapper for on-chain accounts. This is a simplified version ob Substrate's
```
niklasad1, line 14:
```suggestion
/// A multi-format address wrapper for on-chain accounts. This is a simplified version of Substrate's
```

subxt/src/utils/multi_address.rs
```diff
@@ -0,0 +1,65 @@
...
+/// `sp_runtime::MultiAddress`. To obtain more functionality, convert this into that type.
```
niklasad1, line 15: perhaps explain than `#[cfg(feature = "substrate-compat")]` is required for converting it `sp_runtime::MultiAddress`

subxt/src/rpc/rpc.rs
jsdw: Yeah I agree; what I wasn't sure about when I did this was whether anybody is currently using the full details. The reason I didn't include them all right here was just because it involved bringing a load more types over, including DispatchError, which we already have to handle when submitting transactions and we have special logic to decode it from the metadata, so having a "hard copy" felt wrong.

But maybe we do need more details to be returned here?

subxt/src/utils/multi_address.rs
jsdw: When I wrote this I was assuming that people could manually convert it if they wanted (because it has the same structure so is easy to see how), and then I added some From things behind `substrate-compat` becasue I needed them in the PairSigner impl to keep things simple!

Def worth mentioning substrate-compat here now though yup :)

subxt/src/config/mod.rs
jsdw: Yeah; I basically removed various bits that we didn't actually make use of or care about on the Subxt side. Like, if somebody configures the type wrong (whether it's because it's not 32bit or simply just not what the chain is configured with) it won't work either way, so I didn't see the point of the extra check anyway

jsdw:
> > Reducing the need to keep up to date with Substrate packlage versions: in the past, Substrate dependencies like parity-util-mem have prevented Subxt from being used in conjunction with Substrate packages unless the two used the same versions. We aren't there yet, but I'd like to get to the point where Subxt will happily co-exist with a wide range of Substrate versions, as long as the actual types sent back and forth via the RPC APIs remain compatible
> 
> I don't understand this comment, you bumped the substrate dependencies so `parity-util-mem` isn't part of the dependency tree anymore. Are you referring to "old substrate dependencies" or what do you mean?

Ah yeah; I just meant that in the past we had issues like `parity-util-mem`, and avoiding substrate deps avoids potential future issues like that from cropping up too :)

subxt/src/tx/signer.rs
jsdw: I ended up simplifying them a bit and adding a couple of extra `From` impls to hopefully make the different signature types in Substrate "just work".

subxt/src/utils/mod.rs
```diff
@@ -4,7 +4,10 @@
 
 //! Miscellaneous utility helpers.
 
+pub mod account_id;
```
ascjones, line 7: Not a big fan of these being grouped together in a module called `utils`, I think a more precise name would be better to indicate that these are primitive types from `sp` (substrate primitives) crates. Could just be `primitives`?

codegen/src/api/mod.rs
```diff
@@ -169,23 +169,23 @@ impl RuntimeGenerator {
             ),
             (
                 "sp_core::crypto::AccountId32",
-                parse_quote!(#crate_path::ext::sp_core::crypto::AccountId32),
+                parse_quote!(#crate_path::utils::account_id::AccountId32),
```
ascjones, line 172: Better if `AccountId32` is reexported at `utils`, robust to refactoring of namespace.

metadata/Cargo.toml
```diff
@@ -16,7 +16,7 @@ description = "Command line utilities for checking metadata compatibility betwee
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full"] }
 frame-metadata = "15.0.0"
 scale-info = "2.0.0"
-sp-core = "7.0.0"
+sp-core-hashing = "6.0.0"
```
ascjones, line 19: You could go one step further and use https://crates.io/crates/twox-hash directly here

subxt/Cargo.toml
```diff
@@ -32,23 +39,32 @@ scale-info = "2.0.0"
...
 
 subxt-macro = { version = "0.25.0", path = "../macro" }
 subxt-metadata = { version = "0.25.0", path = "../metadata" }
 
-sp-core = { version = "7.0.0", default-features = false  }
-sp-runtime = "7.0.0"
+# Provides some deserialization, types like U256/H256 and hashing impls like twox/blake256:
+impl-serde = { version = "0.4.0" }
+primitive-types = { version = "0.12.0", default-features = false, features = ["codec", "scale-info", "serde"] }
+sp-core-hashing = "6.0.0"
```
ascjones, line 59: Also here if you import `twox-hash`, you already have `blake2` below and this can go. Would mean no more `sp-std`...

subxt/src/config/mod.rs
```diff
@@ -44,14 +44,15 @@ pub trait Config: 'static {
         + Member
         + Default
         + Copy
-        + core::hash::Hash
-        + core::str::FromStr
+        + std::hash::Hash
+        + std::str::FromStr
```
ascjones, line 48: I know we are not targeting `no_std` yet but why not stick with `core`?

subxt/src/config/substrate.rs
```diff
@@ -0,0 +1,294 @@
...
+#[derive(Encode, Decode)]
+enum DigestItemType {
+    Other = 0u32,
+    Consensus = 4u32,
+    Seal = 5u32,
+    PreRuntime = 6u32,
+    RuntimeEnvironmentUpdated = 8u32,
+}
+impl Encode for DigestItem {
+    fn encode(&self) -> Vec<u8> {
```
ascjones, line 200: Maybe we could add some tests which perform encoding roundtrips of locally defined types with the substrate equivalents (as dev-dependencies). That way we will catch breaking changes to the custom encoding of these types.

subxt/src/utils/account_id.rs
```diff
@@ -0,0 +1,165 @@
...
+        v.extend(&r[0..2]);
+        // then encode to base58.
+        use base58::ToBase58;
+        v.to_base58()
+    }
+
+    // This isn't strictly needed, but to give our AccountId32 a little more usefulness, we also
+    // implement the logic needed to decode an AccountId32 from an SS58 encoded string. This is exposed
+    // via a `FromStr` impl.
+    fn from_ss58check(s: &str) -> Result<Self, FromSs58Error> {
```
ascjones, line 62: This could do with at least one basic unit test to check it works.

ascjones (approved):
LGTM!

Would be nice to have some basic regression tests for the duplicated types so we can catch any changes as early as possible.

subxt/src/config/extrinsic_params.rs
```diff
@@ -2,19 +2,26 @@
 // This file is dual-licensed as Apache-2.0 or GPL-3.0.
 // see LICENSE for license details.
 
+//! This module contains a trait which controls the parameters that can must
```
lexnv, line 5:
```suggestion
//! This module contains a trait which controls the parameters that must
```

lexnv (approved):

Xanewok (approved):
I agree with @ascjones that grouping `sp-*` related things under a dedicated module rather than utils would be great and it'd be also great to have regression tests using Substrate crates as part of dev-dependencies.

Otherwise, looks good - great job!

subxt/src/utils/mod.rs
jsdw: Yeah, I was considering where to put them too. Really, they are sortof just "optional types that you can use in Config if you like", so they aren't explicitly a part of any API interface etc.

Perhaps they should be exposed in `config::substrate` for that reason?

subxt/Cargo.toml
jsdw: Ooh I like it! Thanks for the idea!

subxt/src/config/mod.rs
jsdw: Yeah I guess at the moment I wasn't a fan of the mnix of std and core imports and figured for now to just standardise to using std

subxt/src/config/substrate.rs
jsdw: I was actually just mulling over adding this now, so yup I'm def on board with this!

metadata/Cargo.toml
jsdw: Ooh I remember now; I did do this initially but I had to copy a little code from sp-core-hashing to hash in the same way (offhand at least; I didn't dig into whether it could be simplified), and being aware that the hash functions (this and blake2) are needed in a couple of subxt crates I just ended up sticking with sp-core-hashing to provide them)

subxt/src/utils/mod.rs
ascjones: Yeah it's tricky, I think it's okay for misc stuff to go in `helpers`/`utils` namespace but in this case there is a clear category. Maybe `config::types`?

jsdw: If it helps, they are actually already exposed as `config::substrate::AccountId32` etc since they are used for substrate config, so they could also just be exposed from both places? And since they are used in codegen (in a way that I suppose isn't so directly related to the config?), maybe it's justifiable having them exposed in utils also?

But yeah if you prefer them living entirely in config::types or similar I'm happy with that too!

jsdw:
I've added some tests to cover types (but haven't covered everything; there were various existing types that I was reluctant to pull in more substrate crates to write such tests for, though ultimately that may be exactly what we want to do here!

AccountId/MultiAddress/MultiSignature still live in utils, but are exported from `config::substrate` too since they are used there. I do agree that it'd be nice to organise the types a little better.

I originally thought of keepign all of the "sp" types together, but decided that it would be better to group them not according to whether they are defined in substrate crates, but according to how they are used in subxt (ie some types are duplicated because they are part of the RPC API, some types duplicated because they have some useful functionality that we use internally, some types because they are used in config etc. Ultimately we are aiming for "interface compatibility" anyway, and it just happens that duplicating some types helps us achieve that!

Anyway, I hope that for now this is a good place to start! The next step is to look into adding a "native" signer API so we don't need substrate crates at all for basic transaction submitting workflows. Even if such an API exists just as code to show others how to go about signing things, I'd be happy I think. I'll write an issue for this.

PR #761, opened 2023-01-09
ci: Appease Clippy

Xanewok:
This should make the CI green again, as the lints have been introduced in a new stable release, outside of the changes made in this repository.


lexnv (approved):
LGTM! Thanks for having a look at it! 
I think should also be handled by https://github.com/paritytech/subxt/pull/760, which ever gets merged first.

niklasad1 (approved):

PR #766, opened 2023-01-10
Add `chainHead` RPC methods

lexnv:
This PR adds support for the low level `chainHead` interface exposed by the substrate nodes.

The `chianHead` family of functions is added to subxt's RPC methods, together with the corresponding events required for interpreting the data of the RPC methods.

This PR takes the essential parts of #762 PR to aim towards a mergeable form.

Part of https://github.com/paritytech/subxt/issues/732.


subxt/src/rpc/rpc.rs
```diff
@@ -446,6 +450,163 @@ impl<T: Config> Rpc<T> {
             self.client.request("system_dryRun", params).await?;
         Ok(types::decode_dry_run_result(&mut &*result_bytes.0)?)
     }
+
+    /// Subscribe to `chainHead_unstable_follow` to obtain all reported blocks by the chain.
+    ///
+    /// The subscription ID can be used to make queries for the
+    /// block's body [`chainhead_unstable_body`](Rpc::chainhead_unstable_follow),
```
jsdw, line 457:
```suggestion
    /// block's body ([`chainhead_unstable_body`](Rpc::chainhead_unstable_follow)),
```

For consistency with the others :)

jsdw (commented):
This looks great to me offhand!

One thing I'd suggest I think is to add a few integration tests to call these methods and ensure that they actually work: given that they may be around a while until we are fully using them, it'd be good to ensure that they keep working incase anything changes in the (unstable) API in the meantime :)

jsdw (approved):
Love the tests too; great work!

subxt/src/rpc/rpc.rs
```diff
@@ -446,6 +450,163 @@ impl<T: Config> Rpc<T> {
...
+    /// block's body ([`chainhead_unstable_body`](Rpc::chainhead_unstable_follow)),
+    /// block's header ([`chainhead_unstable_header`](Rpc::chainhead_unstable_header)),
+    /// block's storage ([`chainhead_unstable_storage`](Rpc::chainhead_unstable_storage)) and submitting
+    /// runtime API calls at this block ([`chainhead_unstable_call`](Rpc::chainhead_unstable_call)).
+    ///
+    /// # Note
+    ///
+    /// When the user is no longer interested in a block, the user is responsible
+    /// for calling the [`chainhead_unstable_unpin`](Rpc::chainhead_unstable_unpin) method on each.
+    /// Failure to do so will result in the subscription being dropped by the chain.
```
niklasad1, line 466: on failures will the `stop notification` be sent?

subxt/src/rpc/rpc.rs
```diff
@@ -446,6 +450,163 @@ impl<T: Config> Rpc<T> {
...
+    pub async fn chainhead_unstable_follow(
```
niklasad1, line 467: could be nice to have a separate module or something for the v2 impls but perhaps `_unstable` is sufficient

niklasad1 (approved):

subxt/src/rpc/rpc.rs
lexnv: Yep, nice catch thanks! Have changed to commit to reflect that the subscription is stopped with the `stop` event

PR #767, opened 2023-01-10
Fix decoding events via `.as_root_event()` and add test

### LINKED ISSUE #743: Runtime event parsing
It is currently not possible to parse the `runtime::Event` type. To hack around this it is possible to skip the `Phase` like such:  
```rust
let mut data = event.bytes();
let event = <(subxt::events::Phase, kusama::Event)>::decode(&mut data).map(|r| r.1);
```
But it would be nice to have something like `as_runtime_event`. I am not sure if `as_root_event` is supposed to do that, but it does not do.  
The advantage is that it would enable one to easily parse all events of a runtime, instead of just pallet specific ones.

jsdw:
Closes #743

### DIFF
--- a/subxt/src/events/events_type.rs
+++ b/subxt/src/events/events_type.rs
@@ -593,7 +593,7 @@ mod tests {
         // construst an Events object to iterate them:
         let event = Event::A(1, true, vec!["Hi".into()]);
         let events = events::<Event>(
-            metadata.clone(),
+            metadata,
             vec![event_record(Phase::ApplyExtrinsic(123), event.clone())],
         );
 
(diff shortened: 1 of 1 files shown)


subxt/src/events/events_type.rs
```diff
@@ -201,10 +201,12 @@ pub struct EventDetails {
     all_bytes: Arc<[u8]>,
     // start of the bytes (phase, pallet/variant index and then fields and then topic to follow).
     start_idx: usize,
+    // start of the event (ie pallet/variant index and then the fields and topic after).
+    event_start_idx: usize,
     // start of the fields (ie after phase nad pallet/variant index).
```
lexnv, line 206:
```suggestion
    // start of the fields (ie after phase and pallet/variant index).
```

lexnv (approved):

ascjones (approved):

Xanewok (approved):

PR #771, opened 2023-01-11
Make dependabot less aggressive

jsdw:
And ignore sp- crates, since they need to be updated together, and dependabot can't yet do grouped PRs.

We should periodically just keep on top of this anyway, but I don't think we need to be constantly updating them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,16 @@ updates:
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: monthly
+    ignore:
+    # these need to be updated together, so dependabot PRs 
+    # are just noise. So, ignore them:
+    - package-name: sp-core
+    - package-name: sp-keyring
+    - package-name: sp-runtime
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,8 +3,16 @@ updates:
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: monthly
```
niklasad1, line 6: monthly seems a bit too long, how about weekly? :)

jsdw: I guess for me, monthly seemed like a decent interval to try and keep packages uptodate; weekly feels too noisy offhand, ie I'd just end up ignoring them (though I don't mind trying it first!)

lexnv (approved):

.github/dependabot.yml
jsdw: I'll put it to weekly for now :)

Xanewok (approved):

sergejparity (approved):

PR #773, opened 2023-01-12
Implement deserialize

simonsso:
With some exposed functions `AccountId32` can replace `AccountId32` from substrate

### DIFF
--- a/subxt/src/utils/account_id.rs
+++ b/subxt/src/utils/account_id.rs
@@ -10,7 +10,10 @@ use codec::{
     Decode,
     Encode,
 };
-use serde::Serialize;
+use serde::{
+    Deserialize,
+    Serialize,
+};
 
 /// A 32-byte cryptographic identifier. This is a simplified version of Substrate's
 /// `sp_core::crypto::AccountId32`. To obtain more functionality, convert this into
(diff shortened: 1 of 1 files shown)


subxt/src/utils/account_id.rs
```diff
@@ -37,9 +40,9 @@ impl From<[u8; 32]> for AccountId32 {
 }
 
 impl AccountId32 {
-    // Return the ss58-check string for this key. Adapted from `sp_core::crypto`. We need this to
-    // serialize our account appropriately but otherwise don't care.
-    fn to_ss58check(&self) -> String {
+    /// Return the ss58-check string for this key. Adapted from `sp_core::crypto`. We need this to
+    /// serialize our account appropriately but otherwise don't care.
+    pub fn to_ss58check(&self) -> String {
```
jsdw, line 45: This and `from_ss58_check` are already exposed via `.to_string()` and `.parse()`.

One reason I'm a little hesitant to expose the actual methods is because you can't set the chain prefix; it's always "Substrate" (which is fine for display/FromStr and fine for serializing them to get the account nonce, since it doesn't matter there, but not so good as a public API).

jsdw: (that all said, perhaps it is worth copying across the "full" to_ss58 function so that it's possible to do this; in this original cut I was just adding the bare minimum needed to make things work, really..)

simonsso: I will remove this here and keep the serialize/deserialize which is most critical for my use case.

simonsso:
@jsdw I have split this code into two parts now.

jsdw (approved):
Looks great, thank you!

jsdw:
Merging `master` should fix any CI issues taht crop up :)

lx38:
How do you get a ss58check encoded format of the AccountId32? to_string() always encodes into 42 format. (not polkadot's).

Issue #739, opened 2022-12-10
add ` subxt --version`


dzmitry-lahoda:
so it would be easier to do bug reports

dzmitry-lahoda:
weird, crates version has no version

jsdw:
Sounds sensible to me!

PR #776, opened 2023-01-13
Add version subcommand

DaviRain-Su:
close: https://github.com/paritytech/subxt/issues/739

### DIFF
--- a/cli/src/commands/version.rs
+++ b/cli/src/commands/version.rs
@@ -29,7 +29,7 @@ pub fn run(_opts: Opts) {
         }
     };
 
-    println!("subxt-cli {}", get_version(&commit));
+    println!("{} {}", clap::crate_name!(), get_version(&commit));
 }
 
 fn get_version(impl_commit: &str) -> String {
--- a/cli/src/commands/version.rs
+++ b/cli/src/commands/version.rs
@@ -1,0 +1,13 @@
+use clap::Parser as ClapParser;
+
+/// Version subcommand
+///
+/// This subcommand is implemented for backward compatibility reasons.
+/// Its behavior should be the same as that of the `--version` flag which
+/// is handled internally by clap.
+#[derive(Debug, ClapParser)]
+pub struct Opts {}
+
+pub fn run(_opts: Opts) {
+    println!("subxt-cli {}", clap::crate_version!());
+}
--- a/cli/src/commands/version.rs
+++ b/cli/src/commands/version.rs
@@ -1,6 +1,6 @@
 use clap::Parser as ClapParser;
 
-/// Version subcommand
+/// Prints version information
 ///
 /// This subcommand is implemented for backward compatibility reasons.
 /// Its behavior should be the same as that of the `--version` flag which
--- a/cli/src/commands/version.rs
+++ b/cli/src/commands/version.rs
@@ -1,3 +1,5 @@
+use std::process::Command;
+
 use clap::Parser as ClapParser;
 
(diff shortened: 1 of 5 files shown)


cli/src/commands/version.rs
```diff
@@ -0,0 +1,13 @@
...
+/// Prints version information
+///
+/// This subcommand is implemented for backward compatibility reasons.
+/// Its behavior should be the same as that of the `--version` flag which
+/// is handled internally by clap.
+#[derive(Debug, ClapParser)]
+pub struct Opts {}
+
+pub fn run(_opts: Opts) {
+    println!("subxt-cli {}", clap::crate_version!());
```
lexnv, line 12: Thanks for having a go at this, the PR looks good! 
I'm wondering if this information would suffice to identify subxt versions for debugging purposes. 
Substrate provides a similar [versioning](https://github.com/paritytech/substrate/blob/e9b0facf70eeb08032cc7e83548c62f0b4a24bb1/utils/build-script-utils/src/version.rs#L52) including the commit hash as well.

DaviRain-Su: done

cli/src/commands/version.rs
```diff
@@ -0,0 +1,39 @@
...
+    let commit = match Command::new("git")
+        .args(["rev-parse", "--short=11", "HEAD"])
+        .output()
+    {
+        Ok(o) if o.status.success() => {
+            let sha = String::from_utf8_lossy(&o.stdout).trim().to_owned();
+            sha
+        }
+        Ok(o) => {
+            println!("cargo:warning=Git command failed with status: {}", o.status);
```
niklasad1, line 23: these git errors are intended to be compile warnings as these are copy-pasted from a build script...

we can keep them but I think these would be better suited to be logged instead of the "ordinary" CLI output as it might be confusing.

DaviRain-Su: Done

niklasad1 (approved):
LGTM, thank you

lexnv (approved):
Thanks for fixing this! 😄

cli/src/commands/version.rs
```diff
@@ -0,0 +1,40 @@
...
+/// This subcommand is implemented for backward compatibility reasons.
+/// Its behavior should be the same as that of the `--version` flag which
+/// is handled internally by clap.
+#[derive(Debug, ClapParser)]
+pub struct Opts {}
+
+pub fn run(_opts: Opts) {
+    let commit = match Command::new("git")
+        .args(["rev-parse", "--short=11", "HEAD"])
+        .output()
```
jsdw, line 17: The issue I have here is that this will only work if you're within the right directory.

Having the igt hash is v useful though; Perhaps we should have a `build.rs` thing to get it compiled into the output binary as a static string that this can reference?

jsdw: Actually https://stackoverflow.com/questions/43753491/include-git-commit-hash-as-string-into-rust-program looks interesting.

Inparticular the https://github.com/rustyhorde/vergen crate, which ensures we'll re-run in the right cases. Though I also wouldn't be opposed to a `build.rs` that manually emits a rerun-if-changed for `.git/HEAD` and uses `.git/HEAD` and then whatever file that points to to find the current git commit hash and put it into an env variable for the program to pull out.

jsdw:
I tweaked the version to be obtained at compile-time and so I'll close this in favour of #787. Thanks!

PR #777, opened 2023-01-13
Add Runtime API to execute runtime calls

lexnv:
This PR adds the ability to call into the runtime API of the subxt.

At this moment, the subxt utilises the `state_call` method.
This work lays the ground for transitioning to the `chainHead` and `archive` methods of the RPC SpecV2.

The API is exposed both at the root level via, as well at the blocks level (block-centric approach).

Part of #732 .

### DIFF
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -117,9 +117,7 @@ impl<T: Config> OnlineClient<T> {
 
     /// Fetch the metadata from substrate using the runtime API.
     async fn fetch_metadata(rpc: &Rpc<T>) -> Result<Metadata, Error> {
-        let bytes = rpc
-            .state_call("Metadata_metadata".into(), None, None)
-            .await?;
+        let bytes = rpc.state_call("Metadata_metadata", None, None).await?;
         let cursor = &mut &*bytes;
         let _ = <Compact<u32>>::decode(cursor)?;
         let meta: RuntimeMetadataPrefixed = Decode::decode(cursor)?;
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -19,6 +19,7 @@ use crate::{
         Rpc,
         RpcClientT,
     },
+    runtime_api::RuntimeApiClient,
     storage::StorageClient,
     tx::TxClient,
     Config,
@@ -213,6 +214,11 @@ impl<T: Config> OnlineClient<T> {
     /// Work with blocks.
     pub fn blocks(&self) -> BlocksClient<T, Self> {
         <Self as OfflineClientT<T>>::blocks(self)
+    }
+
+    /// Work with runtime API.
+    pub fn runtime_api(&self) -> RuntimeApiClient<T, Self> {
+        <Self as OfflineClientT<T>>::runtime_api(self)
     }
 }
 
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -25,7 +25,12 @@ use crate::{
     Config,
     Metadata,
 };
+use codec::{
+    Compact,
+    Decode,
+};
 use derivative::Derivative;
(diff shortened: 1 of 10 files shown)


subxt/src/client/online_client.rs
```diff
@@ -109,6 +115,17 @@ impl<T: Config> OnlineClient<T> {
...
+    /// Fetch the metadata from substrate using the runtime API.
+    async fn fetch_metadata(rpc: &Rpc<T>) -> Result<Metadata, Error> {
+        let bytes = rpc
+            .state_call("Metadata_metadata".into(), None, None)
+            .await?;
+        let cursor = &mut &*bytes;
+        let _ = <Compact<u32>>::decode(cursor)?;
+        let meta: RuntimeMetadataPrefixed = Decode::decode(cursor)?;
+        Ok(meta.try_into()?)
+    }
```
jsdw, line 127: Cool! :)

subxt/src/runtime_api/runtime_types.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl<T, Client> RuntimeApi<T, Client>
+where
+    T: Config,
+    Client: OnlineClientT<T>,
+{
+    /// Execute a raw runtime API call.
+    pub fn call_raw<'a>(
+        &self,
+        function: String,
```
jsdw, line 44: Can this be a &str instead?

jsdw (approved):
Looks awesome! Just one wee comment :)

subxt/src/runtime_api/runtime_client.rs
```diff
@@ -0,0 +1,66 @@
...
+            // If block hash is not provided, get the hash
+            // for the latest block and use that.
+            let block_hash = match block_hash {
+                Some(hash) => hash,
+                None => {
+                    client
+                        .rpc()
+                        .block_hash(None)
+                        .await?
+                        .expect("didn't pass a block number; qed")
```
niklasad1, line 59: I don't get this `expect` the rpc call could fail if the connection is closed, or am I misunderstanding this?

niklasad1 (approved):
looks clean, one question regarding an expect :P

subxt/src/runtime_api/runtime_client.rs
niklasad1: are you sure it looks like we call https://github.com/paritytech/subxt/blob/master/subxt/src/rpc/rpc.rs#L220-#L227?

lexnv: For the runtime API, as well as storage, I mimicked the events API that we have in place.

Yep, in our use-case the substrate node must provide the `best_hash` of the chain. And in all cases, the RPC should return a concrete `Some` value for the best hash (may it be genesis). 

The client backend must always provide a value for the `.info()` from [here](https://github.com/paritytech/substrate/blob/a447f03727ffd84020682f3e0e4044d5d282d274/primitives/blockchain/src/backend.rs#L37-L38).

I'd imagine that the initial thought was around the lines: if the substrate node misbehaves for this RPC query, we should not trust it anymore to submit further RPC calls; and if the substrate behavior does indeed change we'll capture it in our testing scenarios.

jsdw: > are you sure it looks like we call https://github.com/paritytech/subxt/blob/master/subxt/src/rpc/rpc.rs#L220-#L227?

I was showing the Substrate impl.

But yeah, I think the reason is simply that a node shoiuld always be able to give back a "latest" block hash, and the option only exists because it may not be able to give a block hash back at some explicit block number.

niklasad1: ah, I missed the `await?.expect(..)`.

got it then I get why you where talking about the substrate impl :)

niklasad1, line 59:
```suggestion
                        .expect("substrate return the best block when no block number is provided; qed")
```

PR #786, opened 2023-01-17
Update CI

simonsso:


simonsso:
This was not the intended destination yet

PR #794, opened 2023-01-18
Add ability to set custom metadata etc on OnlineClient

### LINKED ISSUE #793: Allow online client to initialise at a particular block height
It's not necessarily always the case that we only want to query things from the point after the latest runtime upgrade. 

One use-case would be to scan all blocks from a particular block height until the latest. Let's take events in particular.

Currently the way to do this would be to use something like: https://github.com/paritytech/subxt/pull/727 and pass in the metadata to the events query. In order to use this to scan from a particular block, you would probably wrap the `OnlineClient` and store the metadata alongside it, updating it when you hit a runtime upgrade boundary.

However, this design is a little clunkier than it needs to be. For one, we're already storing metadata in the OnlineClient. Metadata we aren't using because we're behind in block number. And in the case our scanning reaches the latest runtime, we are now duplicating storage of the metadata. 

Instead of doing this, simply allowing the initiation of the online client from a particular provided block height, means we can use `apply_upgrade` on the `ClientRuntimeUpdater` to bump the runtime version only in one place (and no need to introduce external state in the wrapper). It would also be necessary to introduce some constructor for `Update`, `new_from_raw()` as the name perhaps, to allude to the fact it's a little less safe than using the subscription util, which will always pair the `runtime_version` and `metadata` for you. See https://github.com/paritytech/subxt/issues/789 for the beginning of that discussion.

Code changes would be pretty simple. Just another method that can construct an `OnlineClient` that would wrap `from_rpc_client`. Orrrr could just have the one method, that takes an `Option<BlockHash>` (open to ideas here).

I'd be happy to do this PR, once there's buy in.

jsdw:
This allows one to instantiate clients capable of decoding older (as long as still V14) blocks and such, with suitable warnings that such clients may not be able to interact with the head of the chain and such.

Hopefully this closes #793

### DIFF
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -71,16 +71,43 @@ impl<T: Config> OnlineClient<T> {
     /// Construct a new [`OnlineClient`] using default settings which
     /// point to a locally running node on `ws://127.0.0.1:9944`.
     pub async fn new() -> Result<OnlineClient<T>, Error> {
+        OnlineClient::new_at(None).await
+    }
+
+    /// Construct a new [`OnlineClient`] using default settings which
+    /// point to a locally running node on `ws://127.0.0.1:9944`.
+    ///
+    /// # Warning
+    ///
+    /// If you use this function, you take responsibility for knowing which blocks this
+    /// client will be capable of interacting with successfully. For example, subscriptions
+    /// to recent blocks and submitting transactions may fail.
+    pub async fn new_at(block_hash: Option<T::Hash>) -> Result<OnlineClient<T>, Error> {
         let url = "ws://127.0.0.1:9944";
-        OnlineClient::from_url(url).await
+        OnlineClient::from_url_at(url, block_hash).await
     }
 
     /// Construct a new [`OnlineClient`], providing a URL to connect to.
     pub async fn from_url(url: impl AsRef<str>) -> Result<OnlineClient<T>, Error> {
+        OnlineClient::from_url_at(url, None).await
+    }
+
+    /// Construct a new [`OnlineClient`], providing a URL to connect to and a block
+    /// hash from which to load metadata and runtime information.
+    ///
+    /// # Warning
+    ///
+    /// If you use this function, you take responsibility for knowing which blocks this
+    /// client will be capable of interacting with successfully. For example, subscriptions
+    /// to recent blocks and submitting transactions may fail.
+    pub async fn from_url_at(
+        url: impl AsRef<str>,
+        block_hash: Option<T::Hash>,
+    ) -> Result<OnlineClient<T>, Error> {
         let client = jsonrpsee_helpers::client(url.as_ref())
             .await
             .map_err(|e| crate::error::RpcError::ClientError(Box::new(e)))?;
-        OnlineClient::from_rpc_client(Arc::new(client)).await
+        OnlineClient::from_rpc_client_at(Arc::new(client), block_hash).await
     }
 }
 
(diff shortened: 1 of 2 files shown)


subxt/src/client/online_client.rs
```diff
@@ -90,22 +123,69 @@ impl<T: Config> OnlineClient<T> {
...
+    pub async fn from_rpc_client_at<R: RpcClientT>(
+        rpc_client: Arc<R>,
+        block_hash: Option<T::Hash>,
+    ) -> Result<OnlineClient<T>, Error> {
+        let rpc = Rpc::<T>::new(rpc_client.clone());
         let (genesis_hash, runtime_version, metadata) = future::join3(
             rpc.genesis_hash(),
-            rpc.runtime_version(None),
-            rpc.metadata(None),
+            // This _could_ always be the latest runtime version, but offhand it makes sense to keep
```
kylezs, line 148: not sure what you mean here? Are you saying that it could be the latest because it doesn't impact anything downstream? like event decoding?

lexnv: I personally interpret the comment as the block hash (if provided) could already contain the latest metadata

lexnv (approved):

subxt/src/client/online_client.rs
kylezs: Hmm I would almost argue the comment makes things more confusing than no comment in that case

niklasad1: yeah, I think you can remove this comment.

The docs and the function name itself makes this code obvious IMO

subxt/src/client/online_client.rs
```diff
@@ -90,22 +123,69 @@ impl<T: Config> OnlineClient<T> {
...
+    ///
+    /// - If you use this function, subxt will be unable to work with blocks that aren't
+    ///   compatible with the metadata at the provided block hash, and if the block hash is
+    ///   not recent, things like subscribing to the head of the chain and submitting
+    ///   transactions will likely fail or encounter errors.
+    /// - Subxt does not support blocks using pre-V14 metadata and will error if you attempt
+    ///   to instantiate a client at such a block.
+    pub async fn from_rpc_client_at<R: RpcClientT>(
+        rpc_client: Arc<R>,
+        block_hash: Option<T::Hash>,
```
niklasad1, line 143: it's almost like we should remove this `Option stuff` with a enum or something which it's much more obvious than to have `None` which means the latest state/block

niklasad1: or perhaps document that `None` implies the latest

niklasad1 (approved):

subxt/src/client/online_client.rs
jsdw: Yeah I can remove it; what I meant was just that even if the metadata was at some specific block and the runtime version was always the latest, I don't think it would break anything related to working with older blocks (and would increase the chance that submitting transactions would work). But in the end I prefer to keep the two aligned always.

subxt/src/client/online_client.rs
jsdw: I don't care too much about Substrate "compatibility" here, so maybe an enum is the way to go, especially with all of the other `at()` methods we have around (my thinking there was to provide `.at()` and `.at_current()` in those places, but maybe it's easier to just do `.at(Block::Current)`/`.at(Block::Hash)` (cc @lexnv).

Anyway, I think it's all good for now but def food for thought :)

subxt/src/client/online_client.rs
```diff
@@ -237,18 +195,51 @@ impl<T: Config> OnlineClient<T> {
...
         inner.runtime_version.clone()
     }
 
+    /// Change the [`RuntimeVersion`] used in this client.
+    ///
+    /// # Warning
+    ///
+    /// Setting a custom runtime version may leave Subxt unable to
+    /// submit valid transactions.
+    pub fn set_runtime_version(&self, runtime_version: RuntimeVersion) {
```
kylezs, line 238: Personally I would have the runtime_version and metadata setting in one method, rather than one each. Makes it less likely they are put out of sync, and also means we only need to acquire the lock once when we are updating both runtime the metadata (which we should be doing). But this would work

jsdw:
As I said in the associated issue, I added `set_` methods so that one can alter the metadata etc used by an `OnlineClient`, but as a tradeoff I removed the `_at()` constructors so that there is only one way to construct a "non happy path" `OnlineClient` now.

kylezs (approved):

subxt/src/client/online_client.rs
jsdw: I did contemplate this, as I'm def aware of the extra locking and synchronisation things, but from an API perspective it just felt nicer to keep them separate and ignore the fact that it happens to lock under the hood (and the locking should be very trivial anyway in terms of time cost), and for your sake you probably only really ever need to update the `Metadata` and can ignore the rest :)

PR #795, opened 2023-01-18
Add `Cargo.lock` for deterministic builds

lexnv:
This PR adds a `Cargo.lock` file to the repository to ensure we have a deterministic build.

While at it, add an extra step in our releasing process to ensure we are always releasing with an up-to-date lock file.

The lock file has been generated with `cagro clean && cargo build --release` and double checked with `cargo generate-lockfile`


jsdw (approved):
Nice!

RELEASING.md
```diff
@@ -28,7 +28,13 @@ We also assume that ongoing work done is being merged directly to the `master` b
 
 4.  Bump the crate version in `Cargo.toml` to whatever was decided in step 2 for `subxt-metadata`, `subxt-cli`, `subxt-codegen`, `subxt-examples`, `subxt-macro` ,`subxt`, `integration-tests`, `test-runtime`, `ui-tests`.
 
-5.  Update `CHANGELOG.md` to reflect the difference between this release and the last. If you're unsure of
+5.  Ensure the `Cargo.lock` file is up to date.
+
+    ```
+    cargo generate-lockfile
```
niklasad1, line 34: > This command will create the Cargo.lock lockfile for the current package or workspace. If the lockfile already exists, it will be rebuilt with the latest available version of every package.

should we really bump all dependencies to the latest on each release?
I guess our tests are sufficient but I'm a little bit scared bumping all.

niklasad1 (approved):

RELEASING.md
dzmitry-lahoda: bump to latest compatible with substrate.

jsdw: > should we really bump all dependencies to the latest on each release?
> I guess our tests are sufficient but I'm a little bit scared bumping all.

Ah I didn't realise that also bumped dependencies. I guess I'm not too worried overall since it'll just be thel ate3st that our toml files are asking for anyway, which is what a fresh install of everything would bring in?

PR #796, opened 2023-01-19
Fix bad signature when the extrinsic size is over 256 bytes

kvinwang:
This is a bug introduced in [#760](https://github.com/paritytech/subxt/pull/760/files#diff-2fb361b08d8543a8a513a3f16b7d44ed1d1105716427e00006c44e763cef8290R172).


jsdw (approved):
Great catch, thank you! Happy to merge this and we should also add a test to cover this case

lexnv (approved):

PR #821, opened 2023-02-10
Add `find_last` for events_types

lean-apple:
In order to optimise the code present in this [PR](https://github.com/paritytech/cargo-contract/pull/965), used for this [issue](https://github.com/paritytech/cargo-contract/issues/777) fixing the wrong return of contract account id with `cargo contract instantiate`

At the moment, only the method was added, yet more integration tests should be. 

As @ascjones suggests, finding 'a pallet where we know that two of the same type of event will be raised by the same extrinsic' should be more relevant than just adding the same integration tests where `find_first` is used,

What do you think would be the best @jsdw ?

### DIFF
--- a/subxt/src/events/events_type.rs
+++ b/subxt/src/events/events_type.rs
@@ -187,6 +187,12 @@ impl<T: Config> Events<T> {
         self.find::<Ev>().next().transpose()
     }
 
+    /// Iterate through the events using metadata to dynamically decode and skip
+    /// them, and return the last event found which decodes to the provided `Ev` type.
+    pub fn find_last<Ev: StaticEvent>(&self) -> Result<Option<Ev>, Error> {
+        self.find::<Ev>().last().transpose()
+    }
+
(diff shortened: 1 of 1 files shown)


ascjones (commented):
Looks good. Regarding testing, it looks like you can add a relatively simple unit test: see the tests in this file.

jsdw:
There aren't any tests for `find` or `find_first` yet either (just the iterating on which they are based). It would be amazing to have a short test for each if you are able, but I know it's more than you would have been bargaining for with this addition so it's up to you :)

lean-apple:
In fact we need this function quickly to fix this [issue](https://github.com/paritytech/cargo-contract/issues/777,) I will keep in mind to add these tests for the 3 find methods/open an issue for it, can we merge it in the meantime ?

jsdw (approved):

jsdw:
This is a tiny addition so I'm happy to approve for now!

lexnv (approved):

PR #833, opened 2023-02-20
export missing config

yjhmelody:
At least should export `BaseExtrinsicParams` and `Tip` type.


yjhmelody:
Oh, them are exported already. I closed.

Issue #740, opened 2022-12-10
improve message when generating from corrupted scale file


dzmitry-lahoda:
It is very unclear what happens here:
```
dz@dz-pc-11:~/github.com/AcalaNetwork/chopsticks$ RUST_BACKTRACE=full  subxt codegen --file=./result.scale
The application panicked (crashed).
Message:  proc-macro-error API cannot be used outside of `entry_point` invocation, perhaps you forgot to annotate your #[proc_macro] function with `#[proc_macro_error]
Location: /home/dz/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro-error-1.0.4/src/lib.rs:483

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 7 frames hidden ⋮                               
   8: proc_macro_error::diagnostic::Diagnostic::emit::h71ed7196a89e2746
      at <unknown source file>:<unknown line>
   9: proc_macro_error::diagnostic::Diagnostic::abort::h8f21112efec23e85
      at <unknown source file>:<unknown line>
  10: subxt_codegen::api::generate_runtime_api_from_bytes::h0ff1f2f3ac0feadd
      at <unknown source file>:<unknown line>
  11: subxt::commands::codegen::codegen::he8bf517fc01e777a
      at <unknown source file>:<unknown line>
  12: subxt::main::{{closure}}::h016da6de0ae33237
      at <unknown source file>:<unknown line>
  13: tokio::runtime::park::CachedParkThread::block_on::h2b6fa559db493273
      at <unknown source file>:<unknown line>
  14: tokio::runtime::scheduler::multi_thread::MultiThread::block_on::h6f350664f51835b8
      at <unknown source file>:<unknown line>
  15: tokio::runtime::runtime::Runtime::block_on::haa2dbc4ce2ac990f
      at <unknown source file>:<unknown line>
  16: subxt::main::hfbaa01add1608d1e
      at <unknown source file>:<unknown line>
  17: std::sys_common::backtrace::__rust_begin_short_backtrace::h3d6ba28366c1d5da
      at <unknown source file>:<unknown line>
  18: std::rt::lang_start::{{closure}}::h64023f3b354b6755
      at <unknown source file>:<unknown line>
[...]

PR #841, opened 2023-02-28
codegen: Add codegen error

lexnv:
This PR adds a specific `CodegenError` to ensure that codegen functions can be used outside of `proc_macro`s.

Previously, the codegen was using the `proc-macro-error` which panicked upon errors when used from subxt/cli.

Closes: https://github.com/paritytech/subxt/issues/740.


codegen/src/types/mod.rs
```diff
@@ -302,12 +303,13 @@ impl<'a> TypeGenerator<'a> {
...
+    pub fn type_derives(&self, ty: &Type<PortableForm>) -> Result<Derives, CodegenError> {
         let joined_path = ty.path().segments().join("::");
-        let ty_path: syn::TypePath = syn::parse_str(&joined_path).unwrap_or_else(|e| {
-            abort_call_site!("'{}' is an invalid type path: {:?}", joined_path, e,)
-        });
-        self.derives.resolve(&ty_path)
+        let ty_path: syn::TypePath = match syn::parse_str(&joined_path) {
+            Ok(ty_path) => ty_path,
+            Err(e) => return Err(CodegenError::InvalidTypePath(joined_path, e)),
+        };
```
jsdw, line 311: I wonder whether the `let else` stuff can be used (untested!):

```suggestion
        let Ok(ty_path) = syn::parse_str::<syn::TypePath>(&joined_path) else {
            return Err(CodegenError::InvalidTypePath(joined_path, e));
        };
```

codegen/src/api/storage.rs
```diff
@@ -41,22 +42,19 @@ pub fn generate_storage(
...
-        .map(|entry| {
-            generate_storage_entry_fns(metadata, type_gen, pallet, entry, crate_path)
-        })
-        .collect();
+    let mut storage_fns = Vec::new();
+    for entry in &storage.entries {
+        let storage_fn =
+            generate_storage_entry_fns(metadata, type_gen, pallet, entry, crate_path)?;
+        storage_fns.push(storage_fn);
+    }
```
jsdw, line 55: I guess some of this could have been done with a `.collect::<Result<Vec<_>,_>()` style thing instead of changing to loops, but I'm happy with either really!

jsdw (approved):
Looks great to me! I have a feeling that this will fix some other issue I ran into a while back with codegen stuff too, but off the top of my head I can't remember what :)

cli/src/commands/codegen.rs
```diff
@@ -107,6 +107,14 @@ fn codegen(
         type_substitutes,
         crate_path,
     );
-    println!("{runtime_api}");
+    match runtime_api {
+        Ok(runtime_api) => println!("{runtime_api}"),
+        Err(e) => {
+            // `Span` cannot be sent between threads.
```
niklasad1, line 113: I don't understand this comment, is that the reason why you can't propagate the error?

codegen/src/api/mod.rs
```diff
@@ -211,35 +271,36 @@ impl RuntimeGenerator {
 
         let metadata_hash = get_metadata_per_pallet_hash(&self.metadata, &pallet_names);
 
-        let modules = pallets_with_mod_names.iter().map(|(pallet, mod_name)| {
+        let mut modules = Vec::new();
```
niklasad1, line 274: I don't think you need to use a `Vec` here as as quote is implemented for iterators as well but it's fine either way

codegen/src/api/mod.rs
```diff
@@ -376,40 +438,41 @@ pub fn generate_structs_from_variants<F>(
...
-            error_message_type_name
-        )
+
+    let scale_info::TypeDef::Variant(variant) = ty.type_def() else {
+        return Err(CodegenError::InvalidType(error_message_type_name.into()))
+    };
+
+    let mut result = Vec::new();
+
+    for var in variant.variants() {
```
niklasad1, line 453: You could collect this this a `Result<Vec<_>, Error>` instead to avoid creating the vec and push.

Readable as it is right now, so optional :)

niklasad1 (approved):
nice, LGTM

codegen/src/api/calls.rs
```diff
@@ -40,12 +40,10 @@ pub fn generate_calls(
...
     types_mod_ident: &syn::Ident,
     crate_path: &CratePath,
-) -> TokenStream2 {
+) -> Result<TokenStream2, CodegenError> {
     // Early return if the pallet has no calls.
-    let call = if let Some(ref calls) = pallet.calls {
-        calls
-    } else {
-        return quote!()
+    let Some(call) = &pallet.calls else {
```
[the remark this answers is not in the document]
lexnv: Yep its a really cool syntax, tho I can see two small downsides:
- it cannot capture the error, that should be totally fine with Options tho
- rustfmt doesn't seem to work with this syntax, seems that anything in the else { } block is ignored for formatting

cli/src/commands/codegen.rs
lexnv: The `CodegenError` includes a `Span` in some variants that can be used to hint at better compile errors.
The `color_eyre` does impose `StdError + Send + Sync` on the returned error, to circumvent that I've just printed the error message in red. I'll update the comment :D

codegen/src/types/mod.rs
lexnv: `map_err` did the trick for this one, thanks!

codegen/src/api/storage.rs
lexnv: Have reverted to using `collect::<...>` everywhere, that sounds pretty cool!

One strange thing about it, the compiler seems to be quite strict in resolving `collect::<Result<Vec<_>, _>`.
If the Error is not explicitly typed in the closure, then the scope in which the compiler finds Error type is not increased. Such is the case where we use `?` and the error could be deduced from the function's return type.
This leads to some places where the `CodegenError` needs to be typed next to collect.

lexnv:
Have picked the `collect::<Result<..>>` as the way to go for loops.
While at it, have also improved the error messages we report back to the user.
This should give a small boost in understanding what the problem is (mostly regarding invalid metadata V14 formats). 
Thanks for the feedback!

codegen/src/api/mod.rs
```diff
@@ -47,6 +50,67 @@ use std::{
...
+    Io(String, std::io::Error),
+    /// Cannot decode the metadata bytes.
+    #[error("Could not decode metadata, only V14 metadata is supported: {0}")]
+    Decode(#[from] codec::Error),
+    /// Out of line modules are not supported.
+    #[error("Out-of-line subxt modules are not supported, make sure you are providing a body to your module: pub mod polkadot {{ ... }}")]
+    InvalidModule(Span),
+    /// Expected named or unnamed fields.
+    #[error("Fields should either be all named or all unnamed, make sure you are providing a valid metadata V14: {0}")]
+    InvalidFields(String),
```
jsdw, line 70: The error messages look good to me overall :)

jsdw (approved):

codegen/src/api/calls.rs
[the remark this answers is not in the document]
jsdw: Interesting on the latter point; I guess they will fix it eventually and we'll have to do some reformatting everywhere!

PR #842, opened 2023-02-28
Use scale-encode and scale-decode to encode and decode based on metadata

jsdw:
**Part of a series:**

- [scale-encode](https://github.com/paritytech/scale-encode): new crate to handle encoding using type info.
- [scale-decode](https://github.com/paritytech/scale-decode/pull/11): handle decoding using type info.
- [scale-value](https://github.com/paritytech/scale-value/pull/25): update Value type to use scale-encode/decode.
- !**[subxt](https://github.com/paritytech/subxt/pull/842): Start taking advantage of these in Subxt.**

This PR:

- Codegen to derive `EncodeAsType` and `DecodeAsType` on generated types (currently as well as `Encode` and `Decode`, but eventually `Encode`/`Decode` may become entirely optional as we use type information for all encoding and decoding). (derives stuff tweaked to allow attrs to also be added to types).
- Manually derive `EncodeAsType` and `DecodeAsType` (well, `IntoVisitor`/`Visitor`) on custom types that need it.
- Add `RootEvent` trait and codegen to assist in decoding (with type info) events into a Root event type.
- Move most things (building transactions, decoding events/storage/constants) to using `EncodeAsType` and `DecodeAsType` (ie to encode and decode using type info). **Note: Encoding StorageKeys still uses static encoding**.
- Consolidate the addresss/payload types into one type, now that we have one code path that can cater for static and dynamic values.
- Simplify storage codegen; no more hashers needed (we apply hashing based on runtime metadata for all codepaths now).
- Remove the generated top level Event type, instead aliasing to the `RuntimeEvent` that's generated already (in part I did this because I wanted to be able to get hold of the path to the `RuntimeCall` type, but for now I'm not trying to substitute it since there are other issues there.

Breakage:

- This PR didn't require much in terms of changes to tests/examples actually; unless you explicitly reference certain types, mostly things will just keep working I hope. A small change to the storage API was needed though; if you need to manually turn a storage address into bytes, the function to do that now lives on the `StorageClient` (ie `api.storage().address_root_bytes()`)

**What's the point of all this?**

Previously, static things encoded/decoded to/from bytes based solely on their structure and not based on the actual type information. Dynamic `Value` types could be used in many places though, and those values would encode and decode based on type information.

This PR takes steps towards unifying these approaches, and allows both static and dynamic values to be encoded in the same sort of way based on type information. This carries a couple of benefits:

1. Static encoding is now more tolerant w.r.t differences between static types and metadata (eg if we have a u8, and the type information requires a u32, we can now happily encode the u8 as a u32, or if we have a static type with fields that aren't needed, they can safely be ignored). There is no longer a need to use dyanmic `Value`s when that robustness is needed, making static types strictly better.
2. Static and dynamic encoding can start to share more common logic; this is seen as we collapse Static+Dynamic versions of various structs down into one struct that handles either.
3. For  things like calls to work across runtimes, they need to be encoded dynamically based on pallet/call index. But what if we want to make batch/multisig/nested calls? Currently we can hand craft them using the generated `Call` enum for the "inner" call, but this enum is hardcoded to the pallet/call indexes of the node it was generated from. Now, everything is encoded dynamically (based on variant name in this case) across the board, so nested calls will "just work" across runtimes (and we can provide a nicer API for this too now).
4. One code path for encoding static or dynamic things means hopefully more robusty code over two code paths where one (ie the dynamic) may not be so well battle tested)


subxt/src/events/events_type.rs
```diff
@@ -343,63 +347,98 @@ impl EventDetails {
...
 }
 
+/// This trait is implemented on the statically generated root event type, so that we're able
+/// to decode it properly via a pallet event that impls `DecodeAsMetadata`. This is necessary
+/// becasue the "root event" type is generated using pallet info but doesn't actually exist in the
+/// metadata types, so we have no easy way to decode things into it via type information and need a
+/// little help via codegen.
+#[doc(hidden)]
+pub trait RootEvent: Sized {
+    /// Given details of the pallet event we wawnt to decode, and the name of the pallet, try to hand
```
lexnv, line 432:
```suggestion
    /// Given details of the pallet event we want to decode, and the name of the pallet, try to hand
```

subxt/src/metadata/decode_encode_traits.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
```
lexnv, line 1: nit: Substrate has decided to remove the copyright years, I guess we could adopt the same style in subxt since we'll no longer care about incrementing years :D

subxt/src/utils/wrapper_opaque.rs
```diff
@@ -0,0 +1,263 @@
...
+            Kind,
+        };
+
+        let Some(ty) = types.resolve(type_id) else {
+            return Err(Error::new(ErrorKind::TypeNotFound(type_id)))
+        };
+
+        // Do a basic check that the target shape lines up.
+        let scale_info::TypeDef::Composite(_) = ty.type_def() else {
+            return Err(Error::new(ErrorKind::WrongShape { actual: Kind::Struct, expected: type_id }))
```
lexnv, line 103:
```suggestion
            return Err(Error::new(ErrorKind::WrongShape {
                actual: Kind::Struct,
                expected: type_id,
            }))
```

subxt/src/utils/wrapper_opaque.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        // Check that the name also lines up.
+        if ty.path().ident().as_deref() != Some("WrapperKeepOpaque") {
+            return Err(Error::new(ErrorKind::WrongShape {
+                actual: Kind::Struct,
+                expected: type_id,
+            }))
+        }
+
+        // Just blat the bytes out..
```
lexnv, line 114:
```suggestion
        // Just blat the bytes out.
```

subxt/src/utils/wrapper_opaque.rs
```diff
@@ -0,0 +1,263 @@
...
+            }))
+        }
+
+        // The field to decode is a compact len followed by bytes. Decode the length, then grab the bytes.
+        let Compact(len) = value
+            .decode_item(Compact::<u32>::into_visitor())
+            .expect("length checked")?;
+        let field = value.next().expect("length checked")?;
+
+        // Sanity check that the compact length we decoded lines up with the number of bytes encoded inthe next field.
```
lexnv, line 153:
```suggestion
        // Sanity check that the compact length we decoded lines up with the number of bytes encoded in the next field.
```

codegen/src/types/derives.rs
```diff
@@ -117,10 +149,21 @@ impl quote::ToTokens for Derives {
                     .to_string()
                     .cmp(&quote::quote!(#b).to_string())
```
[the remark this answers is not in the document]
jsdw: Yeah, I don't think the sorting matters at all for derives, and indeed cargo fmt will do the trick, so it seems better to spit things out in the order given :)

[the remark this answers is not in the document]
jsdw: Ah, actually since derives are in a HashMap, sorting them ensures a consistent order, which we want for the UI tests, so I added the sorting back :)

lexnv (approved):

subxt/src/storage/storage_address.rs
```diff
@@ -107,100 +127,24 @@ where
...
-        }
-        bytes
-    }
-
     /// Return bytes representing the root of this storage entry (ie a hash of
-    /// the pallet and entry name).
-    pub fn to_root_bytes(&self) -> Vec<u8> {
+    /// the pallet and entry name). Use [`crate::storage::StorageClient::address_bytes()`]
+    /// to obtain the bytes representing the entire address.
+    pub fn root_bytes(&self) -> Vec<u8> {
```
niklasad1, line 133: I like `to_root_bytes` better but no biggie

jsdw: I'm trying to remember why I changed it now!

jsdw: I changed it to to_root_bytes; you're right, it feels nicer!

codegen/src/types/substitutes.rs
```diff
@@ -108,48 +108,77 @@ impl TypeSubstitutes {
...
     }
 
+    /// Only insert the given substitution if a substitution at that path doesn't
+    /// already exist.
+    pub fn insert_if_not_exists(
+        &mut self,
+        source: syn::Path,
+        target: AbsolutePath,
+    ) -> Result<(), CodegenError> {
+        let (key, val) = TypeSubstitutes::parse_path_substitution(source, target.0)?;
```
niklasad1, line 118: Ok, this fails only if the parsing fails.

Can't we throw a warning or something if `insert` is ignored i.e, the given key already exist?

niklasad1 (approved):
awesome, looking forward to this PR

codegen/src/types/substitutes.rs
jsdw: I didn't bother only because this function is sortof dangling at the moment, but the point of it will be to override the RuntimeCall enum only if the user hasn't already overridden it via some type substitution of their own or whatever (which is necessary because we don't know where it lives at the point of setting up the default substitutions.. although maybe it could and I'll need to look into that).

Maybe it's best just to remove this function for now!

subxt/src/storage/storage_client.rs
```diff
@@ -57,6 +58,24 @@ where
...
+    /// to retrieve the entries at the root of the associated address.
+    pub fn address_root_bytes<Address: StorageAddress>(
+        &self,
+        address: &Address,
+    ) -> Vec<u8> {
+        utils::storage_address_root_bytes(address)
+    }
+
+    /// Convert some storage address into the raw bytes that would be submitted to the node in order
+    /// to retrieve an entry.
```
ascjones, line 72: Would be nice to document the error condition here and above for end users

testing/integration-tests/src/frame/contracts.rs
```diff
@@ -220,6 +220,8 @@ async fn tx_call() {
         .contracts()
         .contract_info_of(&contract);
 
+    let info_addr_bytes = cxt.client().storage().address_bytes(&info_addr).unwrap();
```
ascjones, line 223: Slightly annoying to have to do this extra step now to generate the bytes from `info_addr_bytes` from `cxt.client().storage()` (and handle the possible `Error`)

testing/test-runtime/build.rs
```diff
@@ -90,10 +90,6 @@ async fn run() {
         #[subxt::subxt(
             runtime_metadata_path = "{}",
             derive_for_all_types = "Eq, PartialEq",
-            substitute_type(
-                type = "sp_arithmetic::per_things::Perbill",
-                with = "::sp_runtime::Perbill"
```
ascjones, line 95: Yes exactly you can't do this anymore

ascjones (commented):
In general looks fantastic, just a couple of comments/questions.

subxt/src/metadata/decode_encode_traits.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl<T: scale_encode::EncodeAsType> EncodeWithMetadata for T {
+    /// SCALE encode this type to bytes, possibly with the help of metadata.
+    fn encode_with_metadata(
+        &self,
+        type_id: u32,
+        metadata: &Metadata,
+        bytes: &mut Vec<u8>,
+    ) -> Result<(), Error> {
+        self.encode_as_type_to(type_id, metadata.types(), bytes)?;
```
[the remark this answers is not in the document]
jsdw: I couldn't remember until I tried removing them, but one type in particular (`DecodedValueThunk`, which is just a thing that hasn't been decoded yet but can be turned into a dynamic value or decoded manually into some static type) can implement `DecodeWithMetadata` but not `DecodeAsType` (since it wants  to store the `Metadata`, but `DecodeAsType` provides the lower level `PortableRegistry`). I remember trying to work around this when I ran into it, but it just made most sense to have the higher level trait in the end.

codegen/src/types/derives.rs
```diff
@@ -86,26 +89,55 @@ impl Derives {
     /// to the set of default derives that reside in `subxt`.
     pub fn new(crate_path: &CratePath) -> Self {
         let mut derives = HashSet::new();
+        let mut attributes = HashSet::new();
+
+        derives.insert(syn::parse_quote!(#crate_path::ext::scale_encode::EncodeAsType));
```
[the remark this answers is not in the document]
jsdw: Ah yeah, that is a bummer for sure :(

I do wonder how much other Subxt users will be impacted by this too.

testing/test-runtime/build.rs
jsdw: Yeah; you'd have to copy the type over or substitute it with some wrapper like `Static<::sp_runtime::PerBill>`

testing/integration-tests/src/frame/contracts.rs
jsdw: (One thing I would like to do anyway is improve the storage APIs so that you don't ever feel the need to generate the address when using Subxt APIs. Eg here, using `fetch_keys` below just simplifies this test because you are happy to consider just 10 of the keys. Using `iter()` to iterate them all might be more correct but a little more verbose becasue you can't do a the nice chaining; I wonder whethrI should tweak this test to use the "iter" API; I think I almost did when I first saw this)

ascjones (approved):
It's a shame about losing the substituting of foreign types, but this inconvenience is massively outweighed by the overall improvement here: brilliant stuff :fist_oncoming:

jsdw:
> It's a shame about losing the substituting of foreign types, but this inconvenience is massively outweighed by the overall improvement here: brilliant stuff 👊

Thank you :) 

I'm just writing a small followup PR which adds a small `Static` util type; at least with this you can, I think, do something like:

```rust
#[subxt::subxt(
            runtime_metadata_path = "{}",
            derive_for_all_types = "Eq, PartialEq",
            substitute_type(
                type = "sp_arithmetic::per_things::Perbill",
                with = "Static<::sp_runtime::Perbill>"
)]
```

And it'll at least provide a workaorund today (but of course having the proper impls will always be more robust, and this will leave a Static type you'll need to dig into)

Issue #826, opened 2023-02-15
Broken ignored doctest in subxt macro output.


deepink-mas:
I have this code
```rust
#[subxt::subxt(runtime_metadata_path = "metadata.scale")]
pub mod node_runtime {}
```

which expands into almost 7k lines of code, among which is this doctest which does not even compile
```rust
		pub mod constants {
			use super::runtime_types;
			pub struct ConstantsApi;

			impl ConstantsApi {
				#[doc = " A fee mulitplier for `Operational` extrinsics to compute \"virtual tip\" to boost their"]
				#[doc = " `priority`"]
				#[doc = ""]
				#[doc = " This value is multipled by the `final_fee` to obtain a \"virtual tip\" that is later"]
				#[doc = " added to a tip component in regular `priority` calculations."]
				#[doc = " It means that a `Normal` transaction can front-run a similarly-sized `Operational`"]
				#[doc = " extrinsic (with no tip), by including a tip value greater than the virtual tip."]
				#[doc = ""]
				#[doc = " ```rust,ignore"]
				#[doc = " // For `Normal`"]
				#[doc = " let priority = priority_calc(tip);"]
				#[doc = ""]
				#[doc = " // For `Operational`"]
				#[doc = " let virtual_tip = (inclusion_fee + tip) * OperationalFeeMultiplier;"]
				#[doc = " let priority = priority_calc(tip + virtual_tip);"]
				#[doc = " ```"]
				#[doc = ""]
				#[doc = " Note that since we use `final_fee` the multiplier applies also to the regular `tip`"]
				#[doc = " sent with the transaction. So, not only does the transaction get a priority bump based"]
				#[doc = " on the `inclusion_fee`, but we also amplify the impact of tips applied to `Operational`"]
				#[doc = " transactions."]
				pub fn operational_fee_multiplier(
					&self,
				) -> ::subxt::constants::StaticConstantAddress<
					::subxt::metadata::DecodeStaticType<::core::primitive::u8>,
				> {
					::subxt::constants::StaticConstantAddress::new(
						"TransactionPayment",
						"OperationalFeeMultiplier",
						[
							141u8, 130u8, 11u8, 35u8, 226u8, 114u8, 92u8, 179u8, 168u8, 110u8,
							28u8, 91u8, 221u8, 64u8, 4u8, 148u8, 201u8, 193u8, 185u8, 66u8, 226u8,
[...]

PR #843, opened 2023-02-28
codegen: Opt out of documentation

lexnv:
This PR exposes the `--no-docs` flag by the CLI.

When the flag is active the generated subxt API will not include the documentation from the metadata.

While at it, have also modified the `codegen` tests to check that no documentation present in the metadata-v14 is extracted on the generated API.

Closes: https://github.com/paritytech/subxt/issues/826.

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -56,6 +56,7 @@ use syn::parse_quote;
 /// * `derives` - Provide custom derives for the generated types.
 /// * `type_substitutes` - Provide custom type substitutes.
 /// * `crate_path` - Path to the `subxt` crate.
+/// * `should_gen_docs` - True if the generated API contains the documentation from the metadata.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
 pub fn generate_runtime_api_from_path<P>(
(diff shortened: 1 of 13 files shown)


codegen/src/api/mod.rs
```diff
@@ -82,6 +82,7 @@ where
         derives,
         type_substitutes,
         crate_path,
+        true,
```
jsdw, line 85: I reckon we should make sure that the macro doesn't generate any docs, in this case (since they are sortof unnecessary there by default). We can always expose a flag in the macro args then to add docs if IDE's can ever learn to ebnefit from it :)

lexnv: That makes total sense! I've disabled by default the documentation generation for the `subxt` macro and added a flag to its attributes: `expose_documentation`, that when present the generated code will expose the metadata documentation

macro/src/lib.rs
```diff
@@ -79,9 +79,22 @@
...
 //!
 //! By default the path `::subxt` is used.
+//!
+//! ### Expose documentation
+//!
+//! In order to expose the documentation from the runtime metadata on the generated
+//! code, users must specify the `expose_documentation` flag:
+//!
+//! ```ignore
+//! #[subxt::subxt(expose_documentation)]
```
jsdw, line 93: Nit: I feel like something like `generate_docs` would make more sense, since that's really what's happening here

jsdw (approved):
Nice!

macro/src/lib.rs
niklasad1: I agree that generate_docs is a better choice:)

jsdw: I tweaked this quickly via the in-browser editor :)

PR #845, opened 2023-03-02
Generate only runtime types

pmikolajczyk41:
## Context

In some cases we are interested only in the `RuntimeCall` enum (or more generally, only in some runtime types). For example:
 - https://github.com/paritytech/ink/issues/1675
 - https://github.com/paritytech/try-runtime-cli/blob/3ce6500f5a5c0e9eab72683646a64b5d7b51db24/cli/main.rs#L23

Unfortunately, we cannot expose just a single enum, because it recursively depends on pallet-specific types as well as auxiliary Substrate types (like `sp_core::crypto::*`). Hence, we must generate a whole bunch of items.

## PR changes

This PR introduces a new flag to `subxt::subxt` macro: `runtime_types_only`. If it is set to `true`, only `runtime_types` module will be generated. Usually, it is 3x less code than the full version. We also enrich the `subxt codegen` CLI command with a corresponding flag.

## Tests / examples

An example has been added to showcase defining `Block` type outside Substrate node.

## Possible improvements:
 1. We could try investigating which types are really needed for `RuntimeCall` and drop unused ones. However, such code pruning should be done by the compiler itself, so it might be pointless to some extent.

### DIFF
--- a/examples/examples/runtime_call_enum.rs
+++ b/examples/examples/runtime_call_enum.rs
@@ -1,0 +1,60 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! In some cases we are interested only in the `RuntimeCall` enum (or more generally, only in some
+//! runtime types). We can ask `subxt` to generate only runtime types by passing a corresponding
+//! flag.
+//!
+//! Here we present how to correctly create `Block` type for Polkadot chain.
+
+use sp_core::H256;
+use sp_runtime::{
+    generic,
+    traits::{
+        BlakeTwo256,
+        Block as _,
+        Header as _,
+    },
+    Digest,
+};
+use subxt::PolkadotConfig;
+
+#[subxt::subxt(
+    runtime_metadata_path = "../artifacts/polkadot_metadata.scale",
+    derive_for_all_types = "Clone, PartialEq, Eq",
+    runtime_types_only
+)]
+pub mod polkadot {}
+
+type RuntimeCall = polkadot::runtime_types::polkadot_runtime::RuntimeCall;
+
+type UncheckedExtrinsic = generic::UncheckedExtrinsic<
+    <PolkadotConfig as subxt::Config>::Address,
+    RuntimeCall,
+    <PolkadotConfig as subxt::Config>::Signature,
+    // Usually we are not interested in `SignedExtra`.
+    (),
+>;
+
+type Header = generic::Header<u32, BlakeTwo256>;
+type Block = generic::Block<Header, UncheckedExtrinsic>;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt::init();
+
+    let polkadot_header = Header::new(
+        41,
+        H256::default(),
+        H256::default(),
+        H256::default(),
+        Digest::default(),
+    );
+
+    let polkadot_block = Block::new(polkadot_header, vec![]);
+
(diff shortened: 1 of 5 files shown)


jsdw:
> 1. We could try investigating which types are really needed for `RuntimeCall` and drop unused ones. However, such code pruning should be done by the compiler itself, so it might be pointless to some extent.

This applies for everything else that's generated too, so I wonder how useful it is to only generate the runtime call enum in general?

There's a risk that we add all sorts of things to only generate specific bits and pieces that people are interested in, so I'm a bit wary of heading down this route in general (and as you found, it's a bit tricky filtering down the actual types that we generate, so you end up with a bunch of cruft too that isn't really needed still).

pmikolajczyk41:
@jsdw  It is useful to keep generated module as part of a crate (so that e.g. your IDE can support it). When you need only `RuntimeCall`, we can easily reduce code 3 times at least.

jsdw:
Mmm, so this basically generates all of the runtime types but none of the interfaces for calls, storage etc. I think I could get behind that as a concept, but would probably be against much other cherry picking of what's generated for fear of making the code too fiddly to work with and a prolieration of options to opt out of various bits and pieces (at least, as the code currently stands).

pmikolajczyk41:
If I'm not mistaken this is the only reasonable subset thing to extract - all other parts potentially depend on runtime types and usually on each other. However, there are valid usecases for runtime types only. But I agree, further granulation would be bad

jsdw:
> If I'm not mistaken this is the only reasonable subset thing to extract - all other parts potentially depend on runtime types and usually on each other. However, there are valid usecases for runtime types only. But I agree, further granulation would be bad

Another area I have thought about before is being able to generate code only for specific pallets, with the idea that People might want to write apps that only require specific pallets and don't care about the rest (the contracts CLI tool is one example). For this to be useful, I'd also want to be able to avoid generating any types which weren't actually needed.

examples/examples/runtime_call_enum.rs
```diff
@@ -0,0 +1,60 @@
...
+        H256::default(),
+        Digest::default(),
+    );
+
+    let polkadot_block = Block::new(polkadot_header, vec![]);
+
+    println!("{polkadot_block:?}");
+
+    Ok(())
+}
```
jsdw, line 60: This part of the example doesn't reallyhave anything to do with the "runtime types only" flag or making use of the `RuntimeCall` generated type?

Maybe just some code that encodes some UncheckedExtrinsic which is using the generated RuntimeCall would make more sense as an example? And Perhaps I'd rename it to `runtime_types_only.rs` but I'm not so wedded to that!

jsdw:
@pmikolajczyk41 Thanks for the contribution! This looks good to me once the conflicts are resolved; just a comment on the example :)

examples/examples/runtime_call_enum.rs
pmikolajczyk41: Actually this example is my very motivation for the `runtime_types_only` flag :) I am using it in the alias `type RuntimeCall = polkadot::runtime_types::polkadot_runtime::RuntimeCall;`, which is required for specifying `UncheckedExtrinsic` type and then `Block` type. Usually, these types are created in runtime, where `RuntimeCall` enum is available from `construct_runtime!` macro. However, there are some usecases, where we are outside runtime, but still need to have `Block` type defined explicitly.

Nevertheless, if you prefer, I can change this example.

pmikolajczyk41: I've renamed the example and added a commented-out snippet explaining what we lose when using `runtime_types_only` flag

jsdw (approved):

lexnv (approved):
LGTM! Thanks for contributing!

Issue #844, opened 2023-03-02
PairSigner fails to be created from sr25519::Pair


lx38:
I'm trying to create a PairSigner to execute transaction, but unable to do so, here is the simple code. Can you help?

``` rust
use sp_core::Pair;
async fn main() {
    let pair = sp_core::sr25519::Pair::from_string_with_seed(
        "0x1470c36a50ea10822d6c7346ded0cecaccc6e40c9b520d457271082c08628771",
        None,
    )
    .unwrap()
    .0;

    let signer = subxt::tx::PairSigner::new(pair);
}
```
## ERROR

```
error[E0277]: the trait bound `sp_core::sr25519::Pair: sp_core::crypto::Pair` is not satisfied
  --> test-subxt/src/main.rs:13:34
   |
13 |     let signer = PairSigner::new(pair);
   |                  --------------- ^^^^ the trait `sp_core::crypto::Pair` is not implemented for `sp_core::sr25519::Pair`
   |                  |
   |                  required by a bound introduced by this call
   |
   = help: the following other types implement trait `sp_core::crypto::Pair`:
             sp_application_crypto::ecdsa::app::Pair
             sp_application_crypto::ed25519::app::Pair
             sp_application_crypto::sr25519::app::Pair
             sp_core::crypto::dummy::Dummy
             sp_core::ecdsa::Pair
             sp_core::ed25519::Pair
             sp_core::sr25519::Pair
note: required by a bound in `PairSigner::<T, Pair>::new`
  --> /Users/igorl/.cargo/registry/src/github.com-1ecc6299db9ec823/subxt-0.27.1/src/tx/signer.rs:56:15
   |
56 |         Pair: PairT,
   |               ^^^^^ required by this bound in `PairSigner::<T, Pair>::new`

```

To me it looks strange as it suggest  the trait `sp_core::crypto::Pair` is not implemented for `sp_core::sr25519::Pair`
but down the line say 
the following other types implement trait `sp_core::crypto::Pair`:
sp_core::sr25519::Pair

niklasad1:
My guess is that `sp_core` that you are using has a different version than the from subxt.

If you using the latest subxt version (v0.27) try with `sp-core v16.0.0` instead

lx38:
Thank you Niklas. I was trying to use versions 7, 17, 18, as I was thinking that might be the reason... but I did not go to 16 and it worked! I appreciate your time and help!

niklasad1:
[...]

PR #853, opened 2023-03-09
fix: re-export `sp-core` and `sp-runtime`

niklasad1:
When the `substrate-compat` is enabled let's re-export `sp-core` and `sp-runtime` again for compatibility reasons.

Further, it easy to pick the wrong version of `sp-core` and `sp-runtime` so let's make the usage a bit smoother.

Close https://github.com/paritytech/subxt/issues/844

### DIFF
--- a/subxt/src/lib.rs
+++ b/subxt/src/lib.rs
@@ -182,4 +182,8 @@ pub mod ext {
     pub use frame_metadata;
     pub use scale_bits;
     pub use scale_value;
+    #[cfg(feature = "substrate-compat")]
+    pub use sp_core;
+    #[cfg(feature = "substrate-compat")]
+    pub use sp_runtime;
 }


lexnv (approved):

niklasad1:
@deuszx 

Can you open another issue regarding that and explain why that is needed? 
I want to keep it out this PR :)

deuszx:
@niklasad1 But isn't it related? With `substrate-impl` we allow people to use `AccountId32` as defined in `sp_core` (an other types) but when they try to use `SubstrateConfig`, the `trait Config` will give them `Config::AccountId = subxt::utils::AccountId32` instead of `sp_core::crypto::AccountId32`. This will force devs to manually change the types.

niklasad1:
It's is related but not what this PR is trying to tackle, thus please open another issue.

Personally, I don't like the suggested `PartialEq/Eq` impls we already got`from/into impls` between subxt::AccoundId32/sp_runtime::AccountId32 so better to discuss it alone in another issue but I get your point roughly.

jsdw (approved):

PR #861, opened 2023-03-14
Expose signer payload to allow external signing

### LINKED ISSUE #836: How can we sign things via a browser?
At the moment, we can pass in a `Signer` implementation to eg `TxClient::create_signed_with_nonce` to sign the tx for us.

In order for that `Signer` implementation to call out to a browser and let it sign things, it would need to be `async`.

Alternately, perhaps we can tweak the interface such that one can do something like get hold of the pre-signed transaciton details (signer payload and tx or some struct representing this) and then once we have a signature, turn this into a `SubmittableExtrinsic` by providing it and finishing the construction of the tx.

Let's figure something out here, and in the process try to come up with a simple example of making use of it (possibly via a browser wasm thing) so that we know we can do this.

jsdw:
This PR adds a few methods to `client.tx()`; `create_partial_signed_with_nonce` and `create_partial_signed` (to mirror `create_signed_with_nonce` and `create_signed`). These return a `PartialExtrinsic`, which just needs a signature to be turned into a "proper" extrinsic, and exposes a `signer_payload()` method to get back the signer payload that needs to be signed.

The added test is currently the best example of this flow. I'll do a documentation pass quite soon to add a bunch of examples and document some common flows and such.

Closes #836

### DIFF
--- a/subxt/src/tx/tx_client.rs
+++ b/subxt/src/tx/tx_client.rs
@@ -157,7 +157,7 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
         Ok(PartialExtrinsic {
             client: self.client.clone(),
             call_data,
-            additional_and_extra_params
+            additional_and_extra_params,
         })
     }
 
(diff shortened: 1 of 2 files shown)


lexnv (approved):
LGTM! 👍

niklasad1 (approved):
clean :+1:

PR #875, opened 2023-03-21
Upgrade to `syn 2.0`

### LINKED ISSUE #874: Upgrade to Syn 2.0
Now it's a thing. Mainly to long-term avoid having two copies of Syn in our dependency graph, but perhaps also it makes one or two things a little nicer after 3.5 years!

ascjones:
Closes https://github.com/paritytech/subxt/issues/874.

Required waiting for update to Darling: https://github.com/TedDriggs/darling/pull/226. 

Now this is done, but is has meant one **breaking change**: we can no longer use a keyword (in our case `type`) as the name in meta name value attributes e.g. `(type = "...")`. It would be possible to implement our own parsing of meta name values to allow keywords in the name position. But for simplicity I have opted to rename `type` to `path`, allowing us to continue to use the build in `darling::ast::NestedMeta` parsing. Which I think makes sense in the context. Happy to change that name to something else.


jsdw (approved):
I'm happy with `path`; I pondered `from` and `to` for the substitute type stuff, but `path` aligns nicely with the `derive` stuff (ie works in both cases) so I can't think of anything better!

lexnv (approved):
Amazing! 👍 

We should probably take a note to update the documentation (if needed) -- the one from the subxt guide: https://github.com/paritytech/subxt/pull/890

PR #876, opened 2023-03-21
cargo fmt with stable

jsdw:
Just using the stable defaults, because I don't like having to remember (and remind others) to use `cargo +nightly fmt`. IMO in general I prefer this. Possibly vertical imports are nicer to slightly reduce the risk of merge conflicts in that area, but I'm not super fussed. What do you guys think? Any rules you want to see added?


lexnv (approved):
There are a few tradeoffs we'll make here in terms of readability, although nothing unreasonable. 

As you mentioned, the vertical imports were a bit clearer to read. Another aspect is that the stable tends to keep things closer to one-liner: 

```rust
// unstable
TypePathType::Primitive { def } => {
                syn::Type::Path(match def {
```
-> 
```rust
TypePathType::Primitive { def } => syn::Type::Path(match def {
```

I generally think that using stable is a good idea, at least to not worry about deprecated config flags. LGTM! 👍

jsdw:
I assume a bunch of the one-liner-isations are because it has 100 chars rather than 90 to play with now in terms of line length, though I didn't pay very close attention to the existing rules so perhaps some others also preferred splits!

I'm easy either way really; as long as it's stable I'm open to any rules and config

niklasad1 (approved):
I'm all good with "rustfmt stable" and I like the imports better but also keep in mind that I'm not that modern :)

Issue #629, opened 2022-08-22
CLI Tool: Strip metadata to only include certain pallets


jsdw:
A common use case of Subxt is to generate client logic to talk to specific pallets.

Currently the CLI tool downloads the entirety of the metadata from a node.

It'd be nice if we could instruct the client to strip metadata (either direct from a node or provided) to include only the information necessary for specific pallets to be called. This would allow for more compact codegen and smaller metadata files to be kept around among other things.

jsdw:
Ah, I hit a snag having a go at stripping metadata to only include certain pallets. Well, two things:

1. `PortableRegistry` is pretty opaque and I can't see a way offhand to modify it, ie remove a bunch of types from it. I could leave it in, but I think it accounts for a large chunk of the bytes associated with a pallet.
2. Since the ID of a type is its location in a `Vec`, and I'd be shuffling things around in the vec to remove types from it, many of the IDs will end up changing. So, I'd also need to then go around the rest of the metadata swapping type IDs out for their new ones. Doable but a bit annoying and means the stripped metadata would deviate from the original a bunch more than hoped.

My options:

1. Only strip pallets data; leave all the type info in there even if it's no longer used.
2. PR to scale-info to expose some way to get at/alter the PortableRegistry types and tedious re-numbering of IDs.
3. I could probably do something awful involving serializing to JSON, stripping down and deserializing back to PortableRegistry..
4. Admit defeat and don't try to strip type info from metadata at all, taking away valuable learnings from the endeavour.

I'm tempted to just leave it be with 4 for now, since 1 is easy but has marginal gains and the others are rather fiddly and is all work that will need duplicating when V15 comes along. Perhaps this is something that should exist in Substrate, anyway?

jsdw:
I like the Substrate approach personally. Another benefit of implementing stripping there, aside from the bandwidth savings, is that it only needs to be implemented once, rather than reimplemented on the periphery in various toold (like sidecar and subxt) that want to strip it.

lexnv:
I created a POC in subxt to gather some data regarding a single pallet metadata ([here](https://github.com/paritytech/subxt/tree/lexnv/pallet_metadata_poc)).

To achieve this, the code recursively examines all type IDs required to represent a pallet.

## Pallet Type IDs Only

The first number indicates the number of type IDs necessary to describe the pallet, relative to the total number of type IDs in the PortableRegistry.
However, this calculation does not consider the type IDs needed to represent extrinsics or the metadata's runtime type.

[...]

PR #879, opened 2023-03-23
Metadata: Retain a subset of metadata pallets

lexnv:
This PR strips the metadata to include only some provided pallets.

### Metadata size with single pallet

The `No Generics` means that the extrinsic types were stripped of the generic type parameters.
Those generics where not needed and include nested type IDs that are unnecessarily increasing
the final metadata size. (Relies on https://github.com/paritytech/scale-info/pull/176). 

| No Generics | With Generics | Pallet |
| - | - | - |
| 244K | 244K | Alliance.scale |
| 240K | 240K | AllianceMotion.scale |
| 8,0K | 240K | AssetTxPayment.scale |
|  32K | 244K | Assets.scale |
|  12K |  12K | Auctions.scale |
| 8,0K | 240K | AuthorityDiscovery.scale |
| 8,0K | 240K | Authorship.scale |
|  16K | 244K | Babe.scale |
|  16K | 244K | Balances.scale |
|  16K | 244K | Bounties.scale |
|  16K | 240K | ChildBounties.scale |
|  12K |  12K | Claims.scale |
|  16K |  16K | Configuration.scale |
|  28K | 252K | Contracts.scale |
|  16K | 240K | ConvictionVoting.scale |
|  16K | 240K | CoreFellowship.scale |
| 240K | 240K | Council.scale |
|  16K |  16K | Crowdloan.scale |
| 248K | 248K | Democracy.scale |
| 8,0K | 8,0K | Dmp.scale |
|  20K | 248K | ElectionProviderMultiPhase.scale |
|  16K | 244K | Elections.scale |
|  12K | 240K | FastUnstake.scale |
| 8,0K | 240K | Glutton.scale |
|  12K | 240K | Grandpa.scale |
| 8,0K | 240K | Historical.scale |
|  16K |  16K | Hrmp.scale |
|  20K | 244K | Identity.scale |
|  12K | 240K | ImOnline.scale |
|  12K | 240K | Indices.scale |
| 8,0K | 8,0K | Initializer.scale |
| 240K | 240K | Lottery.scale |
|  12K | 240K | MessageQueue.scale |
| 8,0K | 240K | Mmr.scale |
| 240K | 240K | Multisig.scale |
|  44K | 252K | Nfts.scale |
|  20K | 248K | Nis.scale |
|  28K | 248K | NominationPools.scale |
| 8,0K | 240K | Offences.scale |
|  12K |  12K | ParaInclusion.scale |
|  12K |  12K | ParaInherent.scale |
|  12K |  12K | ParaScheduler.scale |
| 8,0K | 8,0K | ParaSessionInfo.scale |
| 8,0K | 8,0K | ParachainsOrigin.scale |
|  20K |  20K | Paras.scale |
|  12K |  12K | ParasDisputes.scale |
| 8,0K | 8,0K | ParasShared.scale |
|  16K |  16K | PhragmenElection.scale |
| 8,0K | 240K | Pov.scale |
|  12K | 240K | Preimage.scale |
| 240K | 240K | Proxy.scale |
| 8,0K | 240K | RandomnessCollectiveFlip.scale |
|  12K | 240K | RankedCollective.scale |
| 248K | 248K | RankedPolls.scale |
| 244K | 244K | Recovery.scale |
| 248K | 248K | Referenda.scale |
|  12K |  12K | Registrar.scale |
| 8,0K | 240K | Remark.scale |
| 8,0K | 240K | RootTesting.scale |
|  12K | 240K | Salary.scale |
| 240K | 240K | Scheduler.scale |
|  12K | 240K | Session.scale |
|  12K |  12K | Slots.scale |
|  20K | 244K | Society.scale |
|  36K | 252K | Staking.scale |
|  12K | 240K | StateTrieMigration.scale |
| 240K | 240K | Sudo.scale |
| 312K | 312K | System.scale |
| 240K | 240K | TechnicalCommittee.scale |
|  12K | 240K | TechnicalMembership.scale |
| 8,0K | 240K | Timestamp.scale |
|  16K | 240K | Tips.scale |
| 8,0K | 240K | TransactionPayment.scale |
|  12K | 240K | TransactionStorage.scale |
|  12K | 240K | Treasury.scale |
|  12K |  12K | Ump.scale |
|  28K | 244K | Uniques.scale |
| 240K | 240K | Utility.scale |
|  12K | 240K | Vesting.scale |
|  12K | 244K | VoterList.scale |
| 240K | 240K | Whitelist.scale |
|  36K |  36K | XcmPallet.scale |
| 468K | 468K | polkadot_metadata.scale |

Closes https://github.com/paritytech/subxt/issues/629.

// CC @paritytech/subxt-team


metadata/src/lib.rs
```diff
@@ -8,6 +8,8 @@ use frame_metadata::{
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, Variant};
 use std::collections::HashSet;
 
+pub mod retain;
```
jsdw, line 11: let's just expose the single function we need from the module rather than the module itself :)

metadata/src/retain.rs
```diff
@@ -0,0 +1,209 @@
...
+    type_ids.insert(metadata.ty.id());
+
+    // Additionally, subxt depends on the `RuntimeError` and `DispatchError`.
+    for ty in metadata.types.types() {
+        let Some(ident) = ty.ty().path().ident() else {
+            continue
+        };
+        if ident == DISPATCH_ERROR || ident == RUNTIME_EVENT {
+            type_ids.insert(ty.id());
+        }
```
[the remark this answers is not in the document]
jsdw: This also makes me wonder; are there any other types we might realise we need that aren't "connected" to the metadata in any way? How are they generated I wonder; do they happen to be used in certain calls?

As an earlier PR I pointed to the existing RuntimeEvent rather than creating a new one, but if the RuntimeEvent was only there by chance because it happened to be used in some pallet thing, that's probably a bad idea..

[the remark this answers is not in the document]
jsdw: Ah ok so `RuntimeEvent` is generates because it's returned from `system::storage::events` normally.

I think we should go back to generating a `RuntimeEvent` rather than pointing to that one then in Subxt codegen; then when we strip pallets we can generate only the event variants we are keeping, and we don't need to keep hold of the "RuntimeEvent" type unless it's actually used elsewhere in metadata.

Re "DispatchError", we do need that type because we use it to help decode our errors. We find it in metadata with:

```
        let dispatch_error_ty = metadata
            .types
            .types()
            .iter()
            .find(|ty| ty.ty().path().segments() == ["sp_runtime", "DispatchError"])
            .map(|ty| ty.id());
```

So perhaps we should do the same search here to be accurate?

cli/src/commands/metadata.rs
```diff
@@ -24,25 +25,45 @@ pub struct Opts {
     /// The format of the metadata to display: `json`, `hex` or `bytes`.
     #[clap(long, short, default_value = "bytes")]
     format: String,
+    /// Generate a subset of the metadata that contains only the
+    /// types needed to represent the provided pallets.
+    #[clap(long, use_value_delimiter = true, value_parser)]
+    retain_pallets: Option<Vec<String>>,
```
jsdw, line 31: I think I'd just call it `pallets`, ie "which pallets do I want".

It'd be amazing if this command supported a url _or_ filepath, so that somebody could point and strip down a metadata file they have already downloaded.

The `codegen` command already does, and it would be great if we could be consistent across those. This PR tried this and had a Source type to do this: https://github.com/paritytech/subxt/pull/805/files#diff-9e1e89a7d233bed7d98edc4c5e165f98d125c8d3e91f5417d278950858811039R9. I'd suggest similar; perhaps tweaking that to something like a:

```rust
enum MetadataSource {
    FilePath(PathBuf),
    Uri(Uri)
}

impl MetadataSource {
    pub async fetch(&self) -> RuntimeMetadataV14 { ... }
}
```

metadata/src/lib.rs
```diff
@@ -8,6 +8,9 @@ use frame_metadata::{
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, Variant};
 use std::collections::HashSet;
 
+mod retain;
```
jsdw, line 11: Nit: put the `mod` at the top above the other `use`s :)

metadata/src/retain.rs
```diff
@@ -0,0 +1,174 @@
...
+    metadata.pallets.retain(|pallet| {
+        if filter(pallet) {
+            collect_pallet_types(pallet, &mut type_ids);
+            true
+        } else {
+            false
+        }
+    });
+
+    // Additionally, subxt depends on the `RuntimeError` and `DispatchError`.
```
jsdw, line 121:
```suggestion
    // Additionally, subxt depends on the `DispatchError`; we use the same
    // logic here that is used when building our `Metadata`:
```

testing/ui-tests/src/lib.rs
```diff
@@ -15,14 +15,15 @@ mod dispatch_errors;
...
-use crate::utils::MetadataTestRunner;
+use crate::utils::{MetadataTestRunner, PalletMetadataTestRunner};
 
 // Each of these tests leads to some rust code being compiled and
 // executed to test that compilation is successful (or errors in the
 // way that we'd expect).
 #[test]
 fn ui_tests() {
     let mut m = MetadataTestRunner::default();
+    let mut p = PalletMetadataTestRunner::new();
```
jsdw, line 26: Nit: maybe move this to be right above that while loop where it's used?

testing/ui-tests/src/utils/pallet_metadata_test_runner.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+use codec::{Decode, Encode};
+use frame_metadata::{RuntimeMetadataPrefixed, RuntimeMetadataV14};
+use std::io::Read;
+use subxt_metadata::retain_metadata_pallets;
+
+static TEST_DIR_PREFIX: &str = "subxt_generated_pallets_ui_tests_";
```
jsdw, line 10: I'm wondering; build.rs scripts are given an OUT_DIR env var to put things, which is nicely out of the way. I did the same as you with the `metadata_test_runner.rs` stuff, so potentially both could be improved. Def a nice to have though; not needed :)

metadata/src/retain.rs
```diff
@@ -0,0 +1,174 @@
...
+    let dispatch_error_ty = metadata
+        .types
+        .types()
+        .iter()
+        .find(|ty| ty.ty().path().segments() == ["sp_runtime", "DispatchError"])
+        .expect("Metadata must contain sp_runtime::DispatchError");
+
+    type_ids.insert(dispatch_error_ty.id());
```
jsdw, line 129: So just looking at the V14 top level type:

```rust
pub struct RuntimeMetadataV14 {
	/// Type registry containing all types used in the metadata.
	pub types: PortableRegistry,
	/// Metadata of all the pallets.
	pub pallets: Vec<PalletMetadata<PortableForm>>,
	/// Metadata of the extrinsic.
	pub extrinsic: ExtrinsicMetadata<PortableForm>,
	/// The type of the `Runtime`.
	pub ty: <PortableForm as Form>::Type,
}
```

I can see that we are retaining types from the pallets, but I can't see how we are keeping `extrinsic` and `ty` types above offhand (and yet I think they are contributing to the sizes being big so I feel like I'm missing something :))?

testing/ui-tests/src/lib.rs
lexnv: Yep, that was my intuition too, although it needs to be kept around until after the test suite is dropped. Otherwise, we'll delete the test files before the test suite gets to running them. Apparently, the UI tests are running once the test suite object is dropped

metadata/src/retain.rs
lexnv: My bad, I didn't push the revert on the master. I've retained the types from the `extrinsic` and `runtime_ty`

testing/ui-tests/src/lib.rs
jsdw: Ah interesting; yeah that rings a bell actually! very strange :)

codegen/src/api/mod.rs
```diff
@@ -461,14 +460,10 @@ impl RuntimeGenerator {
...
 
-                /// The statically generated runtime call type.
-                pub type Call = #runtime_path::RuntimeCall;
-
                 /// The error type returned when there is a runtime issue.
                 pub type DispatchError = #types_mod_ident::sp_runtime::DispatchError;
 
-                // Make the runtime event type easily accessible, and impl RootEvent to help decode into it.
-                pub type Event = #runtime_path::RuntimeEvent;
+                #outer_event
```
jsdw, line 466: Random thought while I remember; some pallets like sudo and multisig have calls that accept a `RuntimeCall` as a param. `RuntimeCall` is the outer enum for calls for every pallet, which means it prob points to a lot of types across pallets. I'd guess that this is what accounts for a bunch of extra size in those pallets.

Perhaps this will just improve when I substitute `RuntimeCall` for the Subxt call type (substituting a type is supposed to lead to that type not being generated any more in codegen, though I'm not 100% sure whether that works or not).

cli/src/commands/metadata.rs
lexnv: Added in the latest commit, thanks!

cli/src/commands/metadata.rs
```diff
@@ -7,37 +7,38 @@ use color_eyre::eyre;
...
     #[clap(long, use_value_delimiter = true, value_parser)]
-    retain_pallets: Option<Vec<String>>,
+    pallets: Option<Vec<String>>,
 }
 
 pub async fn run(opts: Opts) -> color_eyre::Result<()> {
-    let hex_data = fetch_metadata_hex(&opts.url).await?;
-
-    let bytes = hex::decode(hex_data.trim_start_matches("0x"))?;
+    let source = MetadataSource::new(opts.url, opts.file)?;
```
jsdw, line 37: Sorry; I think I wasn't very clear; this `MetadataSource` thing could be the value of one of the Opts directly (perhaps behind a `--source` flag), rather than being some glorified way to call a function to use either file or URL :)

An alternate approach that might be nice would be a struct like:

```rust
#[derive(Args)]
struct FileOrUrl {
    #[clap(...)]
    pub file: Option<PathBuf>
    #[clap(...)]
    pub url: Option<Uri>
}
```

And then using that with `#[command(flatten)]` in the opts, something like:

```rust
pub struct Opts {
    ...
    #[command(flatten)]
    file_or_url: FileOrUrl,
```

(with whatever annotation prevents both being used together if that's possible). Search for `flatten` in this for an example: https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html.

The `enum MetadataSource` is more "correct" in that it allows only one of a URL or file path, but the above allows for either `--file` or `--url` to be used. Maybe theres an even better way, but yeah I just thought we could standardise how we accept the metadata location in some way :)

codegen/src/api/mod.rs
jsdw: I'm leaning towards allowing somebody to transcode a call into some other type eg RuntimeCall now actually; the problem with substituting RuntimeCall for some `BoxedPayload` is that the things using `RuntimeCall` require Encode,Decode,EncodeAsType,DecodeAsType,Debug at minimum, and there's not a sensible way to impl `Encode` and `Decode` for our `BoxedPayload`.

Transcoding is more "expensive", but can be opted out of if performance is important (just write the RuntimeCall enum manually), is less intrusive to the generated types (we don't need another substitution), and means we can avoid implementing non-functional Encode/Decode impls on our BoxedPayload.

cli/src/commands/utils.rs
```diff
@@ -2,47 +2,42 @@
...
+            if self.url.is_some() {
                 eyre::bail!("specify one of `--url` or `--file` but not both")
             };
 
-            return Ok(Self::FilePath(file));
+            let mut file = fs::File::open(path)?;
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes)?;
+            return Ok(bytes);
         }
```
jsdw: This looks great :)

jsdw:
Awesome; looks like removing those generic types had a really notable impact! (allets like Sudo and MultiSig I guess depend on `RuntimeCall` for the inner stuff which depends on types from all other pallets, so it makes sense that they are still big :))

jsdw (approved):
Looks great; happy to see this merge :)

cli/src/commands/utils.rs
```diff
@@ -0,0 +1,43 @@
...
+                eyre::bail!("specify one of `--url` or `--file` but not both")
+            };
+
+            let mut file = fs::File::open(path)?;
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes)?;
+            return Ok(bytes);
+        }
+
+        let url = self.url.clone().unwrap_or_else(|| {
```
niklasad1, line 36: this clone isn't needed but no biggie

metadata/src/retain.rs
```diff
@@ -0,0 +1,231 @@
...
+    if let Some(error) = &pallet.error {
+        type_ids.insert(error.ty.id);
+    }
+}
+
+/// Remove the generic type parameters of the type ID from the `PortableRegistry`
+/// of the metadata.
+fn remove_generic_type_params(metadata: &mut RuntimeMetadataV14, id: u32) {
+    let portable_ty = metadata
+        .types
```
niklasad1, line 51: hmm, this is not super nice with `.types.types.get_mut(..)` but that's from frame-metadata

metadata/src/retain.rs
```diff
@@ -0,0 +1,231 @@
...
+            true
+        } else {
+            false
+        }
+    });
+
+    let mut extrinsic_types = HashSet::new();
+    collect_extrinsic_types(&metadata.extrinsic, &mut extrinsic_types);
+    for id in extrinsic_types {
+        remove_generic_type_params(metadata, id);
```
niklasad1, line 168: I find this these lines quite hard to read.

Not sure if it's just me that have troubles to understand that `remove_generics_type_params` just from the metadata itself.

Why do we need to insert that `id` in `type_ids`? That shouldn't exist anyway in the metadata?

metadata/src/retain.rs
```diff
@@ -0,0 +1,231 @@
...
+
+/// Update the given type using the new type ID from the portable registry.
+///
+/// # Panics
+///
+/// Panics if the [`scale_info::PortableRegistry`] did not retain all needed types.
+fn update_type(ty: &mut UntrackedSymbol<TypeId>, map_ids: &BTreeMap<u32, u32>) {
+    let old_id = ty.id;
+    let new_id = map_ids
+        .get(&old_id)
```
niklasad1, line 86: you could also call `get(&old_id).copied().unwrap_or_else(..)` then remove replace `(*new_id).into()` with `new_id.into()`

testing/ui-tests/src/utils/pallet_metadata_test_runner.rs
```diff
@@ -0,0 +1,101 @@
...
+            pallet_filter.name == pallet.name
+        });
+
+        let mut tmp_dir = std::env::temp_dir();
+        tmp_dir.push(format!("{TEST_DIR_PREFIX}{index}"));
+
+        let tmp_metadata_path = {
+            let mut t = tmp_dir.clone();
+            t.push("metadata.scale");
+            t.to_string_lossy().into_owned()
```
niklasad1, line 60: probably fine to unwrap here as well because I would regard `std::env::temp_dir()` to generate a path with non-unicode characters as a bug.

the dir wouldn't be found in those cases anyway but yeah it would be catched below `create::dir(&tmp_dir)`

niklasad1 (approved):

metadata/src/retain.rs
jsdw: I wasn't a fan of this either so much, so I removed it and took a slightly different approach :)

codegen/src/api/mod.rs
jsdw: I just strip the RuntimeCall now that we've exposed these types in `scale-info`, so smaller metadata across the board :)

cli/src/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+                let mut file = fs::File::open(path)?;
+                let mut bytes = Vec::new();
+                file.read_to_end(&mut bytes)?;
+                Ok(bytes)
+            }
+            // Fetch from --url
+            (None, Some(uri)) => Ok(subxt_codegen::utils::fetch_metadata_bytes(uri).await?),
+            // Default if neither is provided; fetch from local url
+            (None, None) => {
+                let uri = "http://localhost:9933"
```
niklasad1, line 40: You could also use `Uri::from_static
```suggestion
                let uri = Uri::from_static("http://localhost:9933");
```

cli/src/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+                    .parse()
```
niklasad1, line 41:
```suggestion
```

cli/src/utils.rs
```diff
@@ -0,0 +1,47 @@
...
+                    .expect("default url is valid");
```
niklasad1, line 42:
```suggestion
```

cli/src/utils.rs
jsdw: Ooh yeah that's nicer, thanks!

PR #886, opened 2023-03-27
Add 'Static' type and improve type substitution codegen to accept it 

### LINKED ISSUE #851: [feature request] Type substitution from CLI for `codegen` command
As in the title - it would be nice to be able to substitute types when generating API via `codegen`. 

`0.26.0` release changed, among many things, what `AccountId32` type is: previously it was `subxt::ext::sp_core::crypto::AccountId32` and now it's `subxt::utils::AccountId32`. Even when re-exporting `sp_core` and `sp_runtime` these types will remain the same - `subxt::utils::AccountId32`. It would be nice either opt-out from the default substitutions (from the CLI) or override it.

The UX could be:
```sh
subxt codegen --url ${NODE_URL} \
	--derive Debug --derive Clone ... \
	--substitute type::from::A:type::into::APrime \
    --substitute type::from::B:type::into::BPrime \
    | rustfmt ... > node_api.rs
```
**EDIT:** Alternatively, we could introduce an argument `--substrate-impl` (to follow the feature from base `subxt`) that would use `sp_core::*` types automatically.

jsdw:
This PR:
- Adds a `subxt::utils::Static<T>` type If the type `T` impls Encode/Decode (ie static encoding and decoding), then the `Static` type itself implements `EncodeAsType` and `DecodeAsType` (which it'll do by statically encoding and decoding).
- Update type substitution to support concrete types and nested generic substitution (essentially to allow users to wrap their substitute types in  `subxt::utils::Static` to make it possible to substitute in types which only impl `Encode` and `Decode` (albeit wrapped in this Static wrapper).
- Add a codegen CLI command to allow type substitution on the command line using the same logic (closes #851).
- Move errors into a separate `errors.rs` module to be more consistent, and extend them with TypeSubstitutionErrors.

The overall reason for this is to help provide a path forwards for users who are currently using type substitution to pull in types which don't impl `EncodeAsType` and `DecodeAsType` (eg types from sp_runtime/sp_core etc), now that those traits are required to be implemented on generated types.

So a user can now do things like:

```rust
#[subxt::subxt(
    runtime_metadata_path = "/path/to/metadata.scale",
    substitute_type(
        type = "sp_runtime::multiaddress::MultiAddress<A, B>",
        with = "::subxt::utils::Static<::sp_runtime::multiaddress::MultiAddress<A, B>>"
    )
)]
pub mod node_runtime {}
```

For instance, to use `sp_runtime`'s `MultiAddress` instead of our own Subxt implementation.

Note that, in general, it's better to rely on types which do impl `EncodeAsType` and `DecodeAsType` where at all possible (Subxt provides a `utils::MultiAddress` and `utils::AccountId32` for instance which _do_ impl these traits).


codegen/src/error.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum TypeSubstitutionError {
+    /// Substitute "to" type must be an absolute path.
+    #[error("The 'to' type must be a path prefixed with 'crate::' or '::'")]
+    ExpectedAbsolutePath(Span),
+    /// Substitute types must have a valid path.
+    #[error("Substitute types must have a valid path.")]
+    EmptySubstitutePath(Span),
+    #[error("Expected the from/to type generics to have the form 'Foo<A,B,C..>'.")]
+    ExpectedAngleBracketGenerics(Span),
```
lexnv, line 94: nit: missing docs here and on the `FetchMetadataError`.

jsdw: Hmm good point; we need a missing docs lint on this crate :)

codegen/src/types/substitutes.rs
```diff
@@ -265,14 +401,80 @@ fn is_absolute(path: &syn::Path) -> bool {
...
+                syn_path!(::some::path::Foo<::other::Path<A, B>>),
+                vec![
+                    (ident("A"), type_path!(::new::A)),
+                    (ident("B"), type_path!(::new::B)),
+                ],
+                syn_path!(::some::path::Foo<::other::Path<::new::A, ::new::B>>),
+            ),
+            (
+                syn_path!(
+                    ::some::path::Foo<::other::Path<A, ::more::path::to<::something::Argh<B>>>, C>
```
lexnv, line 460: nit: Would it also work for `  ::some::path::Foo<::other::Path<A, ::more::path::to<::something::Argh<B>>>, C, A>` where we have 2 `A` to replace with `::new::A`?

lexnv (approved):

codegen/src/types/substitutes.rs
jsdw: Good idea for a test; it should work like that; any same named ident will get replaced. Lemme add a test :)

lexnv:
The clippy CI failed step is unrelated to this PR, I've also encountered while #887 and found the [commit](https://github.com/paritytech/subxt/pull/887/commits/7db4f40435787f549e62528052888fc98de31873) helped resolve it.

codegen/src/error.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Type should be an variant/enum.
+    #[error(
+        "{0} type should be an variant/enum type. Make sure you are providing a valid metadata V14"
+    )]
+    InvalidType(String),
+}
+
+impl CodegenError {
+    /// Fetch the location for this error.
+    // Todo: Probably worth storing location outside of the variant,
```
niklasad1, line 57: can you file an issue this?

codegen/src/error.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: it is 2023? :)

```suggestion
// Copyright 2019-2023 Parity Technologies (UK) Ltd.
```

cli/src/commands/codegen.rs
```diff
@@ -102,13 +117,31 @@ fn codegen(
...
+                "Cannot parse substitute '{from_str}={to_str}': {}",
+                e.to_string()
+            )
+        })?;
+        type_substitutes
+            .insert(from, to)
+            .map_err(|e: TypeSubstitutionError| {
+                eyre::eyre!(
+                    "Cannot parse substitute '{from_str}={to_str}': {}",
+                    e.to_string()
```
niklasad1, line 141: `TypeSubstitionError` already implements `Display` right and the extra `to_string` can be removed?

subxt/src/utils/static_type.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright 2019-2022 Parity Technologies (UK) Ltd.
```
niklasad1, line 1: new file 2023?

testing/ui-tests/src/incorrect/substitute_path_not_absolute.stderr
```diff
@@ -1,4 +1,4 @@
-error: The substitute path must be a global absolute path; try prefixing with `::` or `crate`
+error: Type substitution error: The 'to' type must be a path prefixed with 'crate::' or '::'
```
niklasad1, line 1: This error could be better, `to` is not super friendly.

I would prefer:  "the replaced type must have a path prefix with `crate::` or `::`" but

codegen/src/error.rs
```diff
@@ -0,0 +1,123 @@
...
+    RequestError(#[from] jsonrpsee::core::Error),
+    #[error("'{0}' not supported, supported URI schemes are http, https, ws or wss.")]
+    InvalidScheme(String),
+}
+
+/// Error attempting to do type substitution.
+#[derive(Debug, thiserror::Error)]
+pub enum TypeSubstitutionError {
+    /// Substitute "to" type must be an absolute path.
+    #[error("The 'to' type must be a path prefixed with 'crate::' or '::'")]
```
niklasad1, line 90: ´
```suggestion
    #[error("The substituted type must be a path prefixed with 'crate::' or '::'")]
```

niklasad1 (approved):
this is great

codegen/src/error.rs
jsdw: I like!

PR #896, opened 2023-04-03
Update scale-info to 2.5.0

jsdw:
And fix the deprecated errors


niklasad1 (approved):
few deprecation warnings left, otherwise all good

PR #897, opened 2023-04-05
wait_for_finalized behavior if the tx dropped, usurped or invalid

### LINKED ISSUE #889: Does wait_for_finalized hang indefinitely if the tx dropped or usurped
https://github.com/paritytech/subxt/blob/42bcddeecb39db98e711075b44331d24a8fbeb4d/subxt/src/tx/tx_progress.rs#L102-L116

I have a question regarding a scenario where I send a transaction and invoke this method to wait for its finalization. Is there a risk that the program may hang indefinitely? This is because the status of the transaction may be usurped or dropped, causing the method to continue waiting without returning an error in such cases.

tadeohepperle:
Fixes #889 the following changes are made:
When a TxProgress gets any of the following events via RpcSubscription, the stream should end:
- Usurped
- Finalized
- FinalityTimeout
- Invalid
- Dropped
Before this change, Usurped, Invalid and Dropped would cause functions like `wait_for_finalized()` and `wait_for_in_block()` to run indefinitely. 

The OnlineClient<T> trait bound for TxProgress impl blocks that contain  `wait_for_finalized()` and `wait_for_in_block()` was not lifted, because too much depends on it. But the trait bound of the Stream impl block was downgraded to `Clone`.

### Tests
I added a few unit tests, that generate a predetermined stream substription, create a TxProgress from it and check if the stream implementation of TxProgress behaves as desired, especially in `wait_for_finalized()`.


subxt/src/rpc/rpc_client.rs
```diff
@@ -166,7 +166,8 @@ impl<Res> std::fmt::Debug for Subscription<Res> {
 }
 
 impl<Res> Subscription<Res> {
-    fn new(inner: RpcSubscription) -> Self {
+    /// Creates a new [`Subscription<Res>`].
```
niklasad1, line 169: this type param in `doc links` is needless and could be simplified?!

```suggestion
    /// Creates a new [`Subscription`].
```

subxt/src/tx/tx_progress.rs
```diff
@@ -69,10 +69,10 @@ where
...
     /// transaction progresses, use [`TxProgress::next_item()`] instead.
     ///
-    /// **Note:** transaction statuses like `Invalid` and `Usurped` are ignored, because while they
-    /// may well indicate with some probability that the transaction will not make it into a block,
-    /// there is no guarantee that this is true. Thus, we prefer to "play it safe" here. Use the lower
-    /// level [`TxProgress::next_item()`] API if you'd like to handle these statuses yourself.
+    /// **Note:** transaction statuses like `Invalid`/`Usurped`/`Dropped` indicate with some
+    /// probability that the transaction will not make it into a block but there is no guarantee
+    /// that this is true. In those cases the stream is closed however, so you currently have no
+    /// way to find out if they finally made it into a block or not.
```
niklasad1, line 75: ideally, we should have an integration test against substrate if this behavior gets changed but should be fine for now and the old RPC API will be replaced "soon" :)

niklasad1: to elaborate the docs in substrate on the transaction status says one thing and the RPC implementation does something else :P

subxt/src/tx/tx_progress.rs
```diff
@@ -95,10 +98,10 @@ where
...
```
niklasad1, line 104: it's possible to create another subscription and watch the extrinsic again if any of the transaction statuses wasn't finalized.

For example

```rust
let xt = create_xt()
submit(xt.clone).await.unwrap()
loop {
   match watch_until_in_finalized(xt).await? {
       TxStatus::Finalized(_) => break Ok(()),
       other => {
           log::warning!("Watch extrinsic {xt} failed: {other}, retrying again);
       }
   }
}
```

niklasad1: for sure complicated but possible

subxt/src/tx/tx_progress.rs
```diff
@@ -122,10 +128,10 @@ where
...
     /// use [`TxProgress::next_item()`] instead.
     ///
-    /// **Note:** transaction statuses like `Invalid` and `Usurped` are ignored, because while they
-    /// may well indicate with some probability that the transaction will not make it into a block,
-    /// there is no guarantee that this is true. Thus, we prefer to "play it safe" here. Use the lower
-    /// level [`TxProgress::next_item()`] API if you'd like to handle these statuses yourself.
+    /// **Note:** transaction statuses like `Invalid`/`Usurped`/`Dropped` indicate with some
+    /// probability that the transaction will not make it into a block but there is no guarantee
+    /// that this is true. In those cases the stream is closed however, so you currently have no
+    /// way to find out if they finally made it into a block or not.
```
niklasad1, line 134: > so you currently have no way to find out if they finally made it into a block or not.

so again it's possible but annoying to do

subxt/src/tx/tx_progress.rs
tadeohepperle: Ah okay, I did not know this, should I add your code example to the doc comment?

subxt/src/tx/tx_progress.rs
```diff
@@ -108,6 +111,9 @@ where
                 TxStatus::FinalityTimeout(_) => {
                     return Err(TransactionError::FinalitySubscriptionTimeout.into())
```
jsdw, line 112: Now we have these other errors, perhaps this variant should be renamed to `FinalityTimeout` to be consistent.

subxt/src/tx/tx_progress.rs
```diff
@@ -370,3 +403,113 @@ impl<T: Config, C: OnlineClientT<T>> TxInBlock<T, C> {
...
+    impl OnlineClientT<PolkadotConfig> for MockClient {
+        fn rpc(&self) -> &crate::rpc::Rpc<PolkadotConfig> {
+            panic!("just a mock impl to satisfy trait bounds")
+        }
+    }
+
+    #[tokio::test]
+    async fn wait_for_finalized_returns_err_when_usurped() {
+        let c = MockClient;
+        let stream_elements: Vec<SubstrateTxStatus<H256, H256>> = vec![
```
niklasad1, line 452: you could also make a few  type aliases to make this code more readable such as:

```rust
type Hash = PolkadotConfig::Hash;
type TxProgress = crate::tx::TxProgress<PolkadotConfig, MockClient>;
type SubstrateTxStatus = crate::rpc::types::SubstrateTxStatus<Hash, Hash>>;
```

subxt/src/tx/tx_progress.rs
niklasad1: That code doesn't really work for the `TxProgress` as it wraps subscription and extrinsic.
So you can just modify your comment `In those cases the stream is closed however, so you currently have no way to find out` to something like `In those cases you have to re-subscribe to the extrinsic/create a new TxProgress repeatedly until the extrinsic is finalized/finalizetimeout `

niklasad1: I'm not sure whether we should provide an API for this, what do you guys think?

jsdw: Ooh hold on; I didn't think that it was possible to "re subscribe" to a transaction again!

tadeohepperle: > I'm not sure whether we should provide an API for this, what do you guys think?

You mentioned above that "the old RPC API will be replaced soon", so does that mean that also our stream handling will change? If that is the case, will the logic of this API need to change? If not it is certainly a good Idea to have an API for resubscribing right now, something like `tx_progress.wait_until_finalized_until_timeout()` that does the resubscribing under the hood.

subxt/src/tx/tx_progress.rs
```diff
@@ -220,8 +243,16 @@ impl<T: Config, C: OnlineClientT<T>> Stream for TxProgress<T, C> {
...
+/// Even though these cases can happen, the server-side of the stream is closed, if one of the following is encountered:
+/// - Usurped
+/// - Finalized
+/// - FinalityTimeout
+/// - Invalid
+/// - Dropped
+/// In any of these cases the client side TxProgress stream is also closed.
+/// So there is currently no way for you to tell if an Dropped`/`Invalid`/`Usurped` transaction
+/// reappears in the pool again or not.
+/// You are free to unsubscribe from notifications at any point.
```
jsdw, line 255: I think I'd move this to the bottom fo the comment, in part because the rest of this comment is copied from the TxStatus struct in Substrate :)

subxt/src/tx/tx_progress.rs
```diff
@@ -100,16 +100,16 @@ where
     ///
     /// **Note:** transaction statuses like `Invalid`/`Usurped`/`Dropped` indicate with some
     /// probability that the transaction will not make it into a block but there is no guarantee
-    /// that this is true. In those cases the stream is closed however, so you currently have no
-    /// way to find out if they finally made it into a block or not.
+    /// that this is true. In those cases you have to re-subscribe to the extrinsic/create a new
+    /// TxProgress repeatedly until the extrinsic is finalized.
```
jsdw, line 104:
```suggestion
    /// that this is true. In those cases the stream is closed however, so you currently have no
    /// way to find out if they finally made it into a block or not.
```

Just to revert this since turns out you can't re-subscribe :)

subxt/src/tx/tx_progress.rs
```diff
@@ -71,16 +71,16 @@ where
...
```
niklasad1, line 75: sorry @tadeohepperle I was wrong please revert this :(

subxt/src/tx/tx_progress.rs
```diff
@@ -130,8 +130,8 @@ where
...
```
jsdw, line 134:
```suggestion
    /// that this is true. In those cases the stream is closed however, so you currently have no
    /// way to find out if they finally made it into a block or not.
```

Just to revert this since turns out you can't re-subscribe :)

subxt/src/tx/tx_progress.rs
niklasad1: There is not possible to resubscribe to an extrinsic https://github.com/paritytech/substrate/blob/master/client/rpc-api/src/author/mod.rs#L72-#L77

subxt/src/tx/tx_progress.rs
```diff
@@ -172,7 +172,7 @@ impl<T: Config, C: Clone> Stream for TxProgress<T, C> {
                 //
                 // Even though `Dropped`/`Invalid`/`Usurped` transactions might make it into a block eventually,
                 // the server considers them final and closes the connection, when they are encountered.
-                // curently there is no way of telling if that happens, because the server ends the stream before.
+                // In those cases you have to re-subscribe to the extrinsic/create a new TxProgress repeatedly until the extrinsic is finalized.
```
jsdw, line 175:
```suggestion
    /// In those cases the stream is closed however, so you currently have no way to find 
    /// out if they finally made it into a block or not.
```

Just to revert this since turns out you can't re-subscribe :)

subxt/src/tx/tx_progress.rs
```diff
@@ -249,14 +255,10 @@ impl<T: Config, C: Clone> Stream for TxProgress<T, C> {
...
+///
 /// In any of these cases the client side TxProgress stream is also closed.
-/// So there is currently no way for you to tell if an Dropped`/`Invalid`/`Usurped` transaction
-/// reappears in the pool again or not.
-/// You are free to unsubscribe from notifications at any point.
-/// The first one will be emitted when the block in which the transaction was included gets
-/// finalized. The `FinalityTimeout` event will be emitted when the block did not reach finality
-/// within 512 blocks. This either indicates that finality is not available for your chain,
-/// or that finality gadget is lagging behind.
+/// In those cases you have to re-subscribe to the extrinsic/create a new TxProgress repeatedly until the extrinsic is finalized.
```
jsdw, line 260:
```suggestion
    /// In those cases the stream is closed however, so you currently have no way to find 
    /// out if they finally made it into a block or not.
```

Just to revert this since turns out you can't re-subscribe :)

jsdw (approved):
Looks awesome; just need to revert a few doc comments :)

subxt/src/tx/tx_progress.rs
```diff
@@ -370,3 +405,116 @@ impl<T: Config, C: OnlineClientT<T>> TxInBlock<T, C> {
...
+            let r: Box<RawValue> = RawValue::from_string(s).unwrap();
+            Ok(r)
+        })));
+
+        let rpc_subscription: RpcSubscription = RpcSubscription {
+            stream: rpc_substription_stream,
+            id: None,
+        };
+
+        let sub: Subscription<SubstrateTxStatus<Hash, Hash>> = Subscription::new(rpc_subscription);
```
niklasad1, line 517:
```suggestion
        let sub: Subscription<MockSubstrateTxStatus> = Subscription::new(rpc_subscription);
```

niklasad1 (approved):
nice, LGTM

PR #900, opened 2023-04-07
added at_latest

### LINKED ISSUE #899: Add `at_latest()` to use instead of `.at(None)` in various places.
Currently we have APIs like:

```rust
let client = OnlineClient::<PolkadotConfig>::new();

// fetch latest block:
let block = client.blocks().at(None).await?;
// fetch latest events:
let events = client.events().at(None).await?;
// fetch latest storage entries:
let storage = client.storage().at(None).await?;
```

Fetching the latest block/event/storage seems like a common need, and moreso, we only need the `async/await` + `Result` for one of the paths (ie when `None` is provided). So how about we split into two code paths:

```rust
let client = OnlineClient::<PolkadotConfig>::new();

// fetch latest block:
let block = client.blocks().at_latest().await?;
// fetch block at `block_hash`:
let block = client.blocks().at(block_hash); // no more Option

// fetch latest events:
let events = client.events().at_latest().await?;
// fetch events at `block_hash`:
let events = client.events().at(block_hash); // no more Option

// fetch latest storage entries:
let storage = client.storage().at_latest().await?;
// fetch storage at block hash:
let storage = client.storage().at(block_hash); // no more Option
```

That way, the interface is a bit clearer, and we avoid unnecessary result unwrapping etc where it's not needed (ie somebody could get a block hash via `client.blocks().at_latest().await?.hash()` and then pass that to subsequent calls and avoid the awaiting/unwrapping.

tadeohepperle:
resolves #899 by providing a simplified API for `.at(Option<Hash>)` by specialization into `.at_latest()` and `.at(Hash)`  in various places:
- storage_client: the at() version does not require async + Result anymore
- blocks_client: both versions still require async + Result
- events_client: both versions still require async + Result

### DIFF
--- a/subxt/src/blocks/blocks_client.rs
+++ b/subxt/src/blocks/blocks_client.rs
@@ -49,6 +49,36 @@ where
     /// but may run into errors attempting to work with them.
     pub fn at(
         &self,
+        block_hash: T::Hash,
+    ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
+        self.at_optional_block_hash(Some(block_hash))
+    }
+
+    /// Obtain block details given the provided block hash, or the latest block if `None` is
+    /// provided.
+    ///
+    /// # Warning
+    ///
+    /// This call only supports blocks produced since the most recent
+    /// runtime upgrade. You can attempt to retrieve older blocks,
+    /// but may run into errors attempting to work with them.
+    pub fn at_latest(
+        &self,
+    ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
+        self.at_optional_block_hash(None)
+    }
+
+    /// Obtain block details given the provided block hash, or the latest block if `None` is
+    /// provided.
+    ///
+    /// # Warning
+    ///
+    /// This call only supports blocks produced since the most recent
+    /// runtime upgrade. You can attempt to retrieve older blocks,
+    /// but may run into errors attempting to work with them.
+    #[inline]
+    fn at_optional_block_hash(
+        &self,
         block_hash: Option<T::Hash>,
     ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
         let client = self.client.clone();
--- a/subxt/src/blocks/blocks_client.rs
+++ b/subxt/src/blocks/blocks_client.rs
@@ -39,8 +39,7 @@ where
     T: Config,
     Client: OnlineClientT<T>,
 {
-    /// Obtain block details given the provided block hash, or the latest block if `None` is
-    /// provided.
+    /// Obtain block details given the provided block hash.
     ///
     /// # Warning
     ///
@@ -54,14 +53,7 @@ where
         self.at_optional_block_hash(Some(block_hash))
     }
 
-    /// Obtain block details given the provided block hash, or the latest block if `None` is
-    /// provided.
-    ///
-    /// # Warning
-    ///
-    /// This call only supports blocks produced since the most recent
-    /// runtime upgrade. You can attempt to retrieve older blocks,
-    /// but may run into errors attempting to work with them.
+    /// Obtain block details of the latest block hash.
     pub fn at_latest(
         &self,
     ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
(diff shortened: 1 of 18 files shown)


subxt/src/blocks/blocks_client.rs
```diff
@@ -48,6 +48,36 @@ where
...
     /// but may run into errors attempting to work with them.
     pub fn at(
+        &self,
+        block_hash: T::Hash,
+    ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
+        self.at_optional_block_hash(Some(block_hash))
+    }
+
+    /// Obtain block details given the provided block hash, or the latest block if `None` is
+    /// provided.
```
lexnv, line 58:
```suggestion
    /// Obtain block details of the latest block hash.
```

subxt/src/blocks/blocks_client.rs
```diff
@@ -48,6 +48,36 @@ where
...
+    ///
+    /// # Warning
+    ///
+    /// This call only supports blocks produced since the most recent
```
lexnv, line 62: We can remove this warning from the documentation since we are always operating with the latest block.
And therefore, users have no way to utilize an older and incompatible block here.

subxt/src/blocks/blocks_client.rs
```diff
@@ -48,6 +48,36 @@ where
...
+    pub fn at_latest(
+        &self,
+    ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
+        self.at_optional_block_hash(None)
+    }
+
+    /// Obtain block details given the provided block hash, or the latest block if `None` is
+    /// provided.
+    ///
+    /// # Warning
```
lexnv, line 74: Nit: Since this function is not publically exposed by subxt, we could also remove the warning message.
I believe we placed that warning to explicitly state in the documentation that older blocks may not be supported, just to set up the user expectations.

subxt/src/blocks/blocks_client.rs
```diff
@@ -48,6 +48,36 @@ where
     /// runtime upgrade. You can attempt to retrieve older blocks,
     /// but may run into errors attempting to work with them.
     pub fn at(
+        &self,
```
lexnv, line 51: We could also change the documentation of this function to state `/// Obtain block details given the block hash.`

lexnv:
The code looks good, just a few nits regarding the documentation! 👍 

I wonder if we could also remove the need to have the `.at()` async for the `blocks_client` and `events_client`. 
Generally, all 3 clients (including `storage_client`) could have a similar interface, just to avoid user confusion of writing `storage_client.at(hash);` and `blocks_client.at(hash).await`.

For the `blocks_client` part, we call Substrate to fetch the `Block::Header`, which keeps us from removing the async of the `.at()`. One idea here would be to fetch `Block::Header` lazily since we already require an `async` to fetch events, body, and runtime API.

subxt/src/blocks/blocks_client.rs
```diff
@@ -39,15 +39,31 @@ where
...
+    /// Obtain block details of the latest block hash.
+    pub fn at_latest(
+        &self,
+    ) -> impl Future<Output = Result<Block<T, Client>, Error>> + Send + 'static {
+        self.at_optional_block_hash(None)
+    }
+
+    /// Obtain block details given the provided block hash, or the latest block if `None` is
+    /// provided.
+    #[inline]
```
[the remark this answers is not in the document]
tadeohepperle: @niklasad1 I wanted to keep consistency to how it was before. Before, the pub method was also not inlined. Now the inline will just generate two pub functions `at()` and `at_latest()`that are both not inlined (like before).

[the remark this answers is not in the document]
niklasad1: I see, makes sense but let's hear what @lexnv thinks :)

subxt/src/events/events_client.rs
```diff
@@ -37,6 +37,32 @@ where
...
+
+    /// Obtain events at some block hash.
+    ///
+    /// # Warning
+    ///
+    /// This call only supports blocks produced since the most recent
+    /// runtime upgrade. You can attempt to retrieve events from older blocks,
+    /// but may run into errors attempting to work with them.
+    #[inline]
+    fn at_optional_block_hash(
```
jsdw, line 65: As above, let's put this code directly into `at()` and `at_latest()` so that the former no longer needs to be `async`, and no need for `#[inline]` :)

jsdw: Ah I see; both have to be `async` anyway actually!

In this case, let's just remove the `#[inline]` and it's all good :)

subxt/src/blocks/blocks_client.rs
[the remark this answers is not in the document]
jsdw: Ah I see; it would have to be async anyway! In that case let';s just remove the `#[inline]` (if anything it's be best on `at()` and `at_latest()` I think) and this looks good to me!

PR #907, opened 2023-04-12
Remove need to import parity-scale-codec to use subxt macro

jsdw:
Tell the codec derive macro uses where to look for itself in the subxt exports.


ascjones (approved):
Nice

niklasad1 (approved):

PR #910, opened 2023-04-13
Partial fee estimates for SubmittableExtrinsic

### LINKED ISSUE #527: Return fee details for an Extrinsic
Is there a way to check the exact `final_fee` that the signer of an extrinsic was charged? The best I could think of was approximating using weights. Thank you!

tadeohepperle:
Partially fixes #527.

There are two rpc endpoints that can be used in principle to obtain an estimate of the partial_fee. That is the paid fee without tips. I added a function `partial_fee_estimate()` to `SubmittableExtrinsic`, that uses Method I, as it is simpler.
### Method I: TransactionPaymentApi_query_info

The struct `RuntimeDispatchInfo` returned by this endpoint contains a the partial_fee in question:
```
let encoded_with_len = self.encoded_with_len();
let RuntimeDispatchInfo{ partial_fee, ..} = self.client
 .rpc()
 .state_call_decoded::<RuntimeDispatchInfo>(
     "TransactionPaymentApi_query_info",
     Some(&encoded_with_len),
     None,
).await?;
```
Here the `partial_fee` is already the result we are looking f
### Method II: TransactionPaymentApi_query_fee_details
Make a state call to "TransactionPaymentApi_query_fee_details
```
let encoded_with_len = self.encoded_with_len();
let InclusionFee {
    base_fee,
    len_fee,
    adjusted_weight_fee,
} = self.client.rpc().state_call_decoded::<FeeDetails>(
    "TransactionPaymentApi_query_fee_details",
    Some(&encoded_with_len),
    None,
).await?.inclusion_fee.unwrap();
let partial_fee_2 = base_fee + len_fee + adjusted_weight_fee;
```

Referring to this [this stackexchange answer by jsdw](https://substrate.stackexchange.com/questions/2637/determining-the-final-fee-from-a-client/4224#4224):
```txt
partial_fee = base_fee + len_fee + (adjusted_weight_fee/estimated_weight*actual_weight)
```
We assume here, that _estimated_weight == actual_weight_.
So this should hold and give the same result as Method I:
```txt
partial_fee = base_fee + len_fee + adjusted_weight_fee
```
## Tests:
I added an integration test that verifies both methods work and return the same result.


subxt/src/rpc/types.rs
```diff
@@ -233,6 +233,70 @@ into_block_number!(u8 u16 u32 u64);
 /// Arbitrary properties defined in the chain spec as a JSON object.
 pub type SystemProperties = serde_json::Map<String, serde_json::Value>;
 
+/// taken from original type [`pallet_transaction_payment::types::FeeDetails<u128>`]
```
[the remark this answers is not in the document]
niklasad1: this won't work because the crate `pallet_transaction_payment` is not in your dependency tree

subxt/src/tx/tx_client.rs
```diff
@@ -425,6 +429,15 @@ where
     pub fn into_encoded(self) -> Vec<u8> {
         self.encoded.0
     }
+
+    /// Returns the SCALE encoded extrinsic bytes with the length attached to the end.
+    /// Useful for certain RPC-calls that expect this format.
+    ///
+    /// *Example*: let self.encoded.0 be [10, 20, 10, 20, 10], then encoded_with_len() will be [10, 20, 10, 20, 10, 5, 0, 0, 0]. So there are always 4 bytes appended representing the length of the extrinsic in bytes in little endian notation.
```
niklasad1, line 436: would be better with a `doc test` or `doc example` instead this long line very hard to read

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +478,67 @@ where
         let dry_run_bytes = self.client.rpc().dry_run(self.encoded(), at).await?;
         dry_run_bytes.into_dry_run_result(&self.client.metadata())
     }
+
+    /// returns an estimate for the partial fee. There are two ways of obtaining this, method I is used here in practice. Both methods should give the exact same result though.
+    /// ## Method I: TransactionPaymentApi_query_info
+    ///
+    #[cfg_attr(doctest, doc = " ````no_test")]
```
niklasad1, line 485:
```suggestion
    ```ǹo_run
```

subxt/src/rpc/types.rs
```diff
@@ -233,6 +233,70 @@ into_block_number!(u8 u16 u32 u64);
...
+    pub len_fee: u128,
+    ///
+    /// - `targeted_fee_adjustment`: This is a multiplier that can tune the final fee based on the
+    ///   congestion of the network.
+    /// - `weight_fee`: This amount is computed based on the weight of the transaction. Weight
+    /// accounts for the execution time of a transaction.
+    ///
+    /// adjusted_weight_fee = targeted_fee_adjustment * weight_fee
+    pub adjusted_weight_fee: u128,
+}
```
jsdw, line 263: As far as I can see, we don't need/use these types anywhere (except to demonstrate a second way to get fee details), so I'd just ditch them to keep the API surface smaller (and going forwards we'll eventually be generating these state_call types anyway, so we only want to hardcode the bare minimum :))

subxt/src/tx/tx_client.rs
```diff
@@ -425,6 +429,15 @@ where
...
     }
+
+    /// Returns the SCALE encoded extrinsic bytes with the length attached to the end.
+    /// Useful for certain RPC-calls that expect this format.
+    ///
+    /// *Example*: let self.encoded.0 be [10, 20, 10, 20, 10], then encoded_with_len() will be [10, 20, 10, 20, 10, 5, 0, 0, 0]. So there are always 4 bytes appended representing the length of the extrinsic in bytes in little endian notation.
+    pub fn encoded_with_len(&self) -> Vec<u8> {
+        let b: [u8; 4] = (self.encoded.0.len() as u32).to_le_bytes();
+        [&self.encoded()[..], &b[..]].concat()
+    }
```
jsdw, line 440: Offhand I don't think that this carries it's weight as a thing we want to expose as part of the public API; it's a fairly one-off need afaik so I'd just inline this logic into `partial_fee_estimate` myself :)

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +478,67 @@ where
...
+    /// ```txt
+    /// partial_fee = base_fee + len_fee + (adjusted_weight_fee/estimated_weight*actual_weight)
+    /// ```
+    /// We assume here, that _estimated_weight == actual_weight_.
+    ///
+    /// So this should hold and give the same result as Method I:
+    /// ```txt
+    /// partial_fee = base_fee + len_fee + adjusted_weight_fee
+    /// ```
+    pub async fn partial_fee_estimate(&self) -> Result<u128, Error> {
```
jsdw, line 528: I think I'd get rid of most of the comment above, which is more talking about some alternates and describing how the method works, and just leave something that focuses on what the output is eg:

```
/// This returns an estimate for what the extrinsic is expected to cost to execute, less any tips.
/// The actual amount paid can vary from block to block based on node traffic and other factors.
```

(I didn't think too carefully about this wording but yeah; we can be more concise :))

testing/integration-tests/src/client/mod.rs
```diff
@@ -582,3 +582,47 @@ async fn chainhead_unstable_unpin() {
...
+
+    // Method II: TransactionPaymentApi_query_fee_details + calculations
+    let encoded_with_len = signed_extrinsic.encoded_with_len();
+    let InclusionFee {
+        base_fee,
+        len_fee,
+        adjusted_weight_fee,
+    } = api
+        .rpc()
+        .state_call_decoded::<FeeDetails>(
```
jsdw, line 615: We could have those other types next to/in this test; I quite like this test as a general sanity check that we're understanding how this stuff works!

subxt/src/rpc/rpc.rs
```diff
@@ -382,6 +366,19 @@ impl<T: Config> Rpc<T> {
...
+        &self,
+        function: &str,
+        call_parameters: Option<&[u8]>,
+        at: Option<T::Hash>,
+    ) -> Result<Res, Error> {
+        let bytes = self.state_call(function, call_parameters, at).await?;
+        let cursor = &mut &bytes[..];
+        let res: Res = Decode::decode(cursor)?;
+        Ok(res)
+    }
```
jsdw, line 380: I'm on the fence about exposing `state_call` and `state_call_decoded` in the RPC interface (just to keep it as slim as possible; it's quite nice imo if each method corresponds to one actual RPC call; in the past this interface has been "muddier" and has had a mix of aggregated calls and "simple" ones, but then it becomes unclear what logic to put in here vs in the higher level interfaces :))

I'd probably just pick one to keep and ditch the other for that reason :)

tadeohepperle: I liked the interface of the `state_call_decoded` more, so I now renamed it to `state_call`. It saves us decoding in multiple places, while there is just one place (`call_raw()`) that requires raw data. I hope the interface change does not break much, I fixed all the tests it affects.

testing/integration-tests/src/client/mod.rs
tadeohepperle: Done, good idea.

subxt/src/rpc/types.rs
```diff
@@ -233,6 +233,70 @@ into_block_number!(u8 u16 u32 u64);
...
+/// taken from original type [`frame_support::dispatch::DispatchClass`]
+#[derive(Debug, Encode, Decode, Eq, PartialEq, Copy, Clone)]
+pub enum DispatchClass {
+    /// A normal dispatch.
+    Normal,
+    /// An operational dispatch.
+    Operational,
+    /// A mandatory dispatch.
+    Mandatory,
+}
```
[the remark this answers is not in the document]
tadeohepperle: I was also wondering this, the types I got this from are generic over a type called `Balance` which is the u128 here.

subxt/src/client/online_client.rs
```diff
@@ -136,10 +136,9 @@ impl<T: Config> OnlineClient<T> {
 
     /// Fetch the metadata from substrate using the runtime API.
     async fn fetch_metadata(rpc: &Rpc<T>) -> Result<Metadata, Error> {
-        let bytes = rpc.state_call("Metadata_metadata", None, None).await?;
-        let cursor = &mut &*bytes;
-        let _ = <Compact<u32>>::decode(cursor)?;
-        let meta: RuntimeMetadataPrefixed = Decode::decode(cursor)?;
+        let (_, meta) = rpc
+            .state_call::<(Compact<u32>, RuntimeMetadataPrefixed)>("Metadata_metadata", None, None)
+            .await?;
```
jsdw, line 141: Ooh much nicer!

tadeohepperle: Thanks!

subxt/src/rpc/rpc.rs
```diff
@@ -31,16 +31,19 @@
 //! # }
 //! ```
 
+use std::sync::Arc;
+
+use codec::{Decode, Encode};
+use frame_metadata::RuntimeMetadataPrefixed;
+use serde::Serialize;
+
+use crate::{error::Error, utils::PhantomDataSendSync, Config, Metadata};
```
jsdw, line 40: If possible we should add some `cargo fmt` rules to control how imports are grouped etc, just so it's standard everywhere

tadeohepperle: yes, that would be good

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +467,35 @@ where
...
+            weight_ref_time: u64,
+            #[codec(compact)]
+            weight_proof_size: u64,
+            /// Class of this dispatch; an enum variant index. Adapted from original type <https://paritytech.github.io/substrate/master/frame_support/dispatch/enum.DispatchClass.html>
+            class: u8,
+            /// The inclusion fee of this dispatch; what we want:
+            partial_fee: u128,
+        }
+
+        let len_bytes: [u8; 4] = (self.encoded.0.len() as u32).to_le_bytes();
```
jsdw, line 488: I wonder what this interface expects here in terms of length; often I would expect an encoded `Compact<u32>` to be expected. I'll see if I can have a dig around and find out!

subxt/src/tx/tx_client.rs
```diff
@@ -2,21 +2,23 @@
...
     utils::{Encoded, PhantomDataSendSync},
 };
-use codec::{Compact, Encode};
-use derivative::Derivative;
-use std::borrow::Cow;
-
 // This is returned from an API below, so expose it here.
 pub use crate::rpc::types::DryRunResult;
 
+use super::TxPayload;
```
niklasad1, line 20: I would prefer to move this `use` above the `pub use` as it was before to have clean structure of the use stuff.

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +467,35 @@ where
...
+
+    /// This returns an estimate for what the extrinsic is expected to cost to execute, less any tips.
+    /// The actual amount paid can vary from block to block based on node traffic and other factors.
+    pub async fn partial_fee_estimate(&self) -> Result<u128, Error> {
+        /// adapted from original type <https://paritytech.github.io/substrate/master/pallet_transaction_payment_rpc_runtime_api/struct.RuntimeDispatchInfo.html>
+        #[derive(Debug, Encode, Decode, Eq, PartialEq, Copy, Clone)]
+        pub struct RuntimeDispatchInfo {
+            /// Weight of this dispatch. Adapted from original type <https://paritytech.github.io/substrate/master/frame_support/weights/struct.Weight.html>
+            #[codec(compact)]
+            weight_ref_time: u64,
```
niklasad1, line 479: I'm pretty sure that substrate a has dedicated `Weight` type which has `proof_size` and `ref_time` but maybe that's why you tried to explain with the comment above?

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +467,35 @@ where
...
+            #[codec(compact)]
+            weight_proof_size: u64,
+            /// Class of this dispatch; an enum variant index. Adapted from original type <https://paritytech.github.io/substrate/master/frame_support/dispatch/enum.DispatchClass.html>
+            class: u8,
+            /// The inclusion fee of this dispatch; what we want:
+            partial_fee: u128,
```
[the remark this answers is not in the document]
tadeohepperle: I think the computing cost is so marginal, that I think the complexity increase and understandability descrease in doing so, might not be worth it.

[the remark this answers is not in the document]
niklasad1: Personally, I don't like these extra fields which makes it a bit harder to understand and reason about.

I did [something similar in the staking-miner](https://github.com/paritytech/staking-miner-v2/blob/main/src/epm.rs#L273-#L290) where `RuntimeDispatchInfo` only has the weight field but very much like this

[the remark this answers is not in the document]
lexnv: It would be good to see if the contract team uses the weight estimation as well. In that case, it might be worth adding it since we'd have more feedback and data points around it

subxt/src/tx/tx_client.rs
jsdw: Yeah, I just suggested to collapse it into this one since we just need to scale decode those bytes and move on; I have no strong opinion on whether a separate struct or not as a result :)

subxt/src/runtime_api/runtime_types.rs
```diff
@@ -42,11 +42,16 @@ where
...
         // which is a temporary thing we'll be throwing away quickly:
         async move {
-            let data = client
+            let call_parameters = call_parameters.unwrap_or_default();
+            let hex_encoded_call_params = format!("0x{}", hex::encode(call_parameters));
+            let bytes: types::Bytes = client
                 .rpc()
-                .state_call(function, call_parameters, Some(block_hash))
+                .request(
+                    "state_call",
```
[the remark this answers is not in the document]
jsdw: Yeah; let's just make the result of this `call_raw` be `Res: Decode` :)

subxt/src/tx/tx_client.rs
jsdw: Ok I finally dug around and found the answer, and I can see now that a u32 is the second argument for the runtime API so this is all ok.

In terms of code, a slight simplification would be:

```rust
// The first param is the encoded extrinsic:
let mut params = self.encoded().to_vec();
// The second param is the u32 length of said extrinsic:
self.encoded().len().encode_to(&mut params);

// .. pass `params` to the call
```

subxt/src/runtime_api/runtime_types.rs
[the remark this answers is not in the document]
tadeohepperle: @lexnv Good idea. Should I change something in the code now, or is that a thing for another PR? 
@jsdw I would like to make your suggested change. But the function name should stay `call_raw`? So the raw then only refers to the `call_parameters: Option<&'a [u8]>` and not to the return type anymore, right?

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +466,35 @@ where
         let dry_run_bytes = self.client.rpc().dry_run(self.encoded(), at).await?;
         dry_run_bytes.into_dry_run_result(&self.client.metadata())
     }
+
+    /// This returns an estimate for what the extrinsic is expected to cost to execute, less any tips.
+    /// The actual amount paid can vary from block to block based on node traffic and other factors.
+    pub async fn partial_fee_estimate(&self) -> Result<u128, Error> {
+        /// adapted from original type <https://paritytech.github.io/substrate/master/pallet_transaction_payment_rpc_runtime_api/struct.RuntimeDispatchInfo.html>
+        #[derive(Debug, Encode, Decode, Eq, PartialEq, Copy, Clone)]
+        pub struct RuntimeDispatchInfo {
```
niklasad1, line 475:
```suggestion
        struct RuntimeDispatchInfo {
```

subxt/src/tx/tx_client.rs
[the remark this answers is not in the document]
niklasad1: Ok, I see then.

Thus, we should mimic the type exactly as in substrate with same fields and so on or just mark the fields that we don't care about such as:

```rust    
        /// adapted from original type <https://paritytech.github.io/substrate/master/pallet_transaction_payment_rpc_runtime_api/struct.RuntimeDispatchInfo.html>   
        #[derive(Debug, Encode, Decode, Eq, PartialEq, Copy, Clone)]
        struct RuntimeDispatchInfo {
             /// This field is just used to make Compact encoding work.
             _: Weight,
              /// This field is just used to make Compact encoding work.
             _: DispatchClass,
             partial_fee: u128,
        }
```

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +466,35 @@ where
...
+            #[codec(compact)]
+            weight_proof_size: u64,
+            /// Class of this dispatch; an enum variant index. Adapted from original type <https://paritytech.github.io/substrate/master/frame_support/dispatch/enum.DispatchClass.html>
+            class: u8,
+            /// The inclusion fee of this dispatch; what we want:
+            partial_fee: u128,
+        }
+
+        let mut params = self.encoded().to_vec();
+        (self.encoded().len() as u32).encode_to(&mut params);
```
niklasad1, line 488: :+1:, nice much more readable now

subxt/src/runtime_api/runtime_types.rs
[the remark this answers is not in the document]
jsdw: Yeah, keep the `call_raw`; this is because in the future we'll be adding a much more ergonomic `call` method which will be the preferred way to use this interface :)

subxt/src/tx/tx_client.rs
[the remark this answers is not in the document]
jsdw: I'm easy either way, but leant towards removing the "noise" of 2 additional struct/enum definitions :)

When the Runtime API stuff is in metadata for real we can improve on all of this anyway

subxt/src/tx/tx_client.rs
tadeohepperle: I now just removed the struct completely here, as you and James recommended.

niklasad1 (approved):
LGTM, nice work

testing/integration-tests/src/client/mod.rs
```diff
@@ -582,3 +578,75 @@ async fn chainhead_unstable_unpin() {
...
+    let signed_extrinsic = api
+        .tx()
+        .create_signed(&tx, &alice, Default::default())
+        .await
+        .unwrap();
+
+    // Method I: TransactionPaymentApi_query_info
+    let partial_fee_1 = signed_extrinsic.partial_fee_estimate().await.unwrap();
+
+    // Method II: TransactionPaymentApi_query_fee_details + calculations
```
niklasad1, line 630: I wonder if it's sufficient to just rely on that the result from `Method I: TransactionPaymentApi_query_info` could be decoded as this should already be tested by substrate I reckon.

Then remove this extra code :point_down:, but this is a nice way to check the result so I have mixed feelings here your call

jsdw: I'm happy either way too; I don't think we need the test but it's interesting in documenting how the things work so I'm easy :)

subxt/src/tx/tx_client.rs
```diff
@@ -465,4 +466,23 @@ where
...
         dry_run_bytes.into_dry_run_result(&self.client.metadata())
     }
+
+    /// This returns an estimate for what the extrinsic is expected to cost to execute, less any tips.
+    /// The actual amount paid can vary from block to block based on node traffic and other factors.
+    pub async fn partial_fee_estimate(&self) -> Result<u128, Error> {
+        let mut params = self.encoded().to_vec();
+        (self.encoded().len() as u32).encode_to(&mut params);
+        // destructuring RuntimeDispatchInfo, see type information <https://paritytech.github.io/substrate/master/pallet_transaction_payment_rpc_runtime_api/struct.RuntimeDispatchInfo.html>
+        // data layout: {weight_ref_time: Compact<u64>, weight_proof_size: Compact<u64>, class: u8, partial_fee: u128}
```
jsdw, line 476: Nice; more compact code with nice comment to explain :)

jsdw (approved):
Looks good to me; nice job!

testing/integration-tests/src/client/mod.rs
tadeohepperle: If the logic changes in the future the test will break, then we can still change/remove it I think :)

PR #912, opened 2023-04-13
metadata: Use v15 internally

lexnv:
This PR upgrades the v14 metadata to the latest v15.

- Subxt can work with V14 and V15
- When loading a V14 metadata, the metadata is converted to V15 with additional added fields empty
- Pointing the CLI to a V14 metadata will return a V15

This lays the ground for the runtime metadata API work when a new substrate release includes https://github.com/paritytech/substrate/pull/13302, that will include:
- fetch V15 metadata using the new API
- generate runtime API for calling into the runtime of substrate


lexnv:
bot rebase

cli/src/commands/compatibility.rs
```diff
@@ -108,7 +110,8 @@ async fn fetch_runtime_metadata(url: &Uri) -> color_eyre::Result<RuntimeMetadata
     }
 
     match metadata.1 {
-        RuntimeMetadata::V14(v14) => Ok(v14),
+        RuntimeMetadata::V14(v14) => Ok(metadata_v14_to_latest(v14)),
```
jsdw, line 113: I feel like the CLI tool should just give back whatever metadata version is available and not try to transform it (that way people can just use it to download whatever the latest metadata version is from a node rather than specifically returning something that subxt wants. Then, the subxt macro/codegen can do this update function instead to be compatible with as many versions as possible (whether downloaded via our cli tool or other)

cli/src/commands/metadata.rs
```diff
@@ -39,9 +40,11 @@ pub async fn run(opts: Opts) -> color_eyre::Result<()> {
             }
         };
 
-        retain_metadata_pallets(metadata_v14, |pallet_name| {
+        retain_metadata_pallets(&mut metadata_v15, |pallet_name| {
```
jsdw, line 43: Ah so to retain specific pallets maybe it makes sense then to update the metadata to V15 so that we don't have to have logic to strip the metadata to certain pallets for different versions of it. But that's ok; we can make a note that if you use the `--pallets` thing then it'll update to V15 (currently unstable) as well.

cli/src/commands/compatibility.rs
lexnv: Yep, for internal functionality this PR uses the V15:
- retain only specific pallets
- validation hashing 
- codegen

Otherwise, we would need to handle both v14 and v15 internally, which would lead to extra code we'd have to maintain.

Transitioning from V14 to V15 shouldn't cause any breaking changes in the output, expect the output of `cli metadata --retain-pallets`.
And when users point to a V14 node and use plain `cli metadata` the outputted metadata would not be converted to V15.

cli/src/commands/metadata.rs
lexnv: Yep, will document around the `--pallets` option that the metadata will be updated to V15.

Another option for handling this would be to have bidirectional `From` traits.
We currently implement V14 -> V15 and we could implement V15 -> V14 to not affect the shape of the metadata while keeping the code footprint in `metadata/retain.rs` as small as possible.
However, that adds extra code complexity since upgrading and downgrading the metadata requires a fairly large method.

Weighing the options, I think that documenting this CLI option provides a better outcome for us

cli/src/commands/compatibility.rs
jsdw: Ah yup; I missed that we preserve V14 if you don't use `--pallets` so that looks good to me as-is :)

let's avoid implementing a way to downgrade to V14 again at the mo

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+        out[idx] = a ^ b;
+    }
+    out
+}
+
+/// Combine two hashes or hash-like sets of bytes together into a single hash.
+/// `xor` is OK for one-off combinations of bytes, but if we are merging
+/// potentially identical hashes, this is a safer way to ensure the result is
+/// unique.
+fn hash_hashes(a: [u8; 32], b: [u8; 32]) -> [u8; 32] {
```
niklasad1, line 48: maybe `concat_and_hash` is better name for this function

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+    let mut out = [0u8; 32 * 2];
+    for (idx, byte) in a.into_iter().chain(b).enumerate() {
```
niklasad1, line 50: you also use copy_from_slice which I think is slightly cleaner/more readable here:

```rust
 out[0..32].copy_from_slice(&a[..]);
 out[32..].copy_from_slice(&b[..]);
```

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+            bytes = xor(bytes, get_type_hash(registry, ty.id, visited_ids));
+        }
+        StorageEntryType::Map {
+            hashers,
+            key,
+            value,
+        } => {
+            for hasher in hashers {
+                // Cloning the hasher should essentially be a copy.
+                bytes = hash_hashes(bytes, [hasher.clone() as u8; 32]);
```
niklasad1, line 226: yeah, that type should really be copy

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+    Ok(hash)
+}
+
+/// Obtain the hash representation of a `frame_metadata::v15::PalletMetadata`.
+pub fn get_pallet_hash(
+    registry: &PortableRegistry,
+    pallet: &PalletMetadata<PortableForm>,
+) -> [u8; 32] {
+    // Begin with some arbitrary hash (we don't really care what it is).
+    let mut bytes = hash(&[19]);
```
niklasad1, line 320: huh, can you elaborate? :)

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+            bytes
+        }
+    }
+}
+
+/// Obtain the hash representation of a `scale_info::Type` identified by id.
+fn get_type_hash(registry: &PortableRegistry, id: u32, visited_ids: &mut HashSet<u32>) -> [u8; 32] {
+    // Guard against recursive types and return a fixed arbitrary hash
+    if !visited_ids.insert(id) {
+        return hash(&[123u8]);
```
niklasad1, line 167: does `hash(&[123u8])` correspond to some special value or is it arbitrary?

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+    }
+
+    let ty = registry.resolve(id).unwrap();
```
niklasad1, line 170: expect if this is infallible plz :)

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+        })
+        .collect();
+
+    // Sort by pallet name to create a deterministic representation of the underlying metadata.
+    pallets.sort_by_key(|&(name, _hash)| name);
+
+    // Note: pallet name is excluded from hashing.
+    // Each pallet has a hash of 32 bytes, and the vector is extended with
+    // extrinsic hash and metadata ty hash (2 * 32).
+    let mut bytes = Vec::with_capacity(pallets.len() * 32 + 64);
```
niklasad1, line 379: You could also add constants for these instead such `(pallet.len() * HASH_LEN + EXTRINSIC_HASH_LEN + METADATA_HASH_LEN)` for readability

The comment is good enough anyway

metadata/src/validation.rs
lexnv: Yep that makes sense, have renamed it! Thanks!

metadata/src/validation.rs
lexnv: I like it! Have changed it thanks!

metadata/src/validation.rs
lexnv: Have added the `MAGIC_PALLET_VALUE` constant and a comment when we use it.
The pallets could basically contain no data, for those cases, we still need to return a hashing value.
And we use this magic arbitrary value to signal that this is an "empty" pallet.

metadata/src/validation.rs
lexnv: Have introduced the `MAGIC_RECURSIVE_TYPE_VALUE`. This is a random / magic value that we return to signal recurrsion

metadata/src/validation.rs
lexnv: Yep that makes sense, we call this function internally and get the type ID from the metadata which should have the type registered in the registry. Have changed it to an `expect`

metadata/src/validation.rs
lexnv: That makes sense, have added a `HASH_LEN` which should be universal across this file since it is produced by our `hash` function, thanks!

metadata/src/validation.rs
```diff
@@ -0,0 +1,851 @@
...
+        })
+        .collect();
+
+    // Sort by pallet name to create a deterministic representation of the underlying metadata.
+    pallets_hashed.sort_by_key(|&(name, _hash)| name);
+
+    // Note: pallet name is excluded from hashing.
+    // Each pallet has a hash of 32 bytes, and the vector is extended with
+    // extrinsic hash and metadata ty hash (2 * 32).
+    let mut bytes = Vec::with_capacity(pallets_hashed.len() * 32);
```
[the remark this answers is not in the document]
lexnv: Good catch! This should only take into account the number of pallet hashes here (ie for the `get_metadata_per_pallet_hash` fn).

The comment was wrong and have adjusted it, while reusing the `HASH_LEN` constant introduced above. Thanks!

metadata/src/validation.rs
```diff
@@ -0,0 +1,864 @@
...
+    out
+}
+
+/// Combine two hashes or hash-like sets of bytes together into a single hash.
+/// `xor` is OK for one-off combinations of bytes, but if we are merging
+/// potentially identical hashes, this is a safer way to ensure the result is
+/// unique.
+fn concat_and_hash(a: [u8; 32], b: [u8; 32]) -> [u8; 32] {
+    let mut out = [0u8; 32 * 2];
+    out[0..32].copy_from_slice(&a[..]);
```
niklasad1, line 59: you could use `HASH_LEN` here as well :)

niklasad1 (approved):

metadata/src/lib.rs
```diff
@@ -3,844 +3,120 @@
...
+    get_pallet_hash, get_storage_hash, NotFound,
+};
 
-/// Obtain the hash representation of a `frame_metadata::RuntimeMetadataV14`
-/// hashing only the provided pallets.
+/// Latest runtime metadata version supported by subxt.
+pub type LatestRuntimeMetadata = RuntimeMetadataV15;
+
+/// Convert the metadata V14 to the latest metadata version.
+pub fn metadata_v14_to_latest(metadata: RuntimeMetadataV14) -> LatestRuntimeMetadata {
```
jsdw, line 22: I'm happy for this in this PR, but ultimately wonder whether we want to make the version of metadata we use in Subxt opaque, eg have a struct like

```
pub struct RuntimeMetadata(RuntimeMetadataV15);
```

and then having the functions in this crate accept this as an argument.

The idea is that then, changes to the metadata (esp as V15 is unstable and liable to change a little for instance) will not be a breaking change to Subxt, just an internal detail.

If we did this, there is a question about whether users may want access to the metadata that Subxt is using though. (Subxt wraps it and ends up with a `Metadata` struct, and currently users can call certain functions on that (the ones we use in Subxt internally). I don't really like that interface being exposed (I'd rather Metadata was entirely opaque) but a compromise is that we allow access to the underlying RuntimeMetadataV14 or whatever.

So ultimately I think we could:

- Have `RuntimeMetadata` as above to hide away the concrete version we are working with
- Make all of the functions on `Metadata` (in subxt lib crate) `pub(crate)`; that way we can have exactly the interfacce we need as Subxt evolves internally without worrying about breaking things for users (because they can't use that stuff
- Give users access to the underlying metadata by `impl RuntimeMetadata { pun fn into_runtime_metadata_v14(self) -> RuntimeMetadataV14 }` and eg `impl Metadata { pub fn runtime_metadata() }` (so that user can do `metadata.runtime_metadata().clone().into_runtime_metadata_v14()`; yes you'd need to clone but it'd be a one off need; maybe we can do better?)

Then, in the future we can add `into_runtime_metadata_v15()` etc and so on as a purely additive non breaking change and the actual metadata we use internally never matters to users (and can be eg unstable versions etc without breaking anything)

What do you reckon?

As I said at the top, this could be a followup PR if it makes sense to people! @ascjones I wonder what you reckon to this too?

jsdw: (this would mean adding a way to go from latest metadata to V14, V15 etc but going backwards should almost entirely mean just throwing away some information, and if it ever is harder than that then we can opt to break and remove an `into_runtime_metadata_vxx()` fn if we want)

metadata/src/lib.rs
```diff
@@ -3,844 +3,120 @@
...
-        assert_ne!(
-            to_hash(meta_type::<StructField>()),
-            to_hash(meta_type::<StructFieldSwap>())
-        );
+/// # Panics
+///
+/// Panics if the runtime metadata version is not supported.
+///
+/// Supported versions: v14 and v15.
+pub fn metadata_to_latest(meta: RuntimeMetadataPrefixed) -> LatestRuntimeMetadata {
```
jsdw, line 116: with my suggestion of `pub struct RuntimeMetadata(RuntimeMetadataV15)` above, this could be a fn like `impl From<RuntimeMetadataV14> for RuntimeMetadata { ... }`.

In general, I think we should avoid panicking in this public function, so until the above sort of thing happens I'd prefer that this didn't panic and either took a specific RuntimeMetadataV14 or whatever or returned an `Option<RuntimeMetadataV15>` so we can panic where it's called instead rather than potentially forget about the panicking at the call site

metadata/src/lib.rs
```diff
@@ -3,844 +3,120 @@
...
+pub use retain::retain_metadata_pallets;
+pub use validation::{
+    get_call_hash, get_constant_hash, get_metadata_hash, get_metadata_per_pallet_hash,
+    get_pallet_hash, get_storage_hash, NotFound,
+};
 
-/// Obtain the hash representation of a `frame_metadata::RuntimeMetadataV14`
-/// hashing only the provided pallets.
+/// Latest runtime metadata version supported by subxt.
+pub type LatestRuntimeMetadata = RuntimeMetadataV15;
```
jsdw, line 19: I wouldn't bother with this alias (and it's not used in a bunch of places where it could be anyway)

metadata/src/validation.rs
jsdw, line 1: I'm assuming all of this is just copy & paste and nothing has changed logic wise?

jsdw (approved):
Looking good to me; just a couple of small things, and a larger comment which could be a followup PR (I'd be interested in a bit of feedback on whether it's a sane idea anyway!)

metadata/src/lib.rs
jsdw: @cmichi I have this memory that you wanted some function or data added to the `Metadata` struct in Subxt a while ago (and also a memory of thinking I'd eventually forget that it was wanted since Subxt doesn't need it).

If we stopped exposing _all_ methods on `Metadata` but instead gave you a way to get back the `RuntimeMetadataV14` as described above, would that be sufficient? I'm quite eager to hide said APIs if I can, because they basically just exist for Subxt internal usage and that's crept out into the real world a bit too much, so I think the frame-metadata structs are a better thing to provide access to :)

metadata/src/validation.rs
lexnv: Initially, the logic was moved from `lib.rs` to `validation.rs`. However, Niklas suggested some valid improvement ideas to make the code look cleaner, without impacting the functionality: 
- moved a few hardcoded constants to a dedicated `const`
- renamed the function that concatenates two hashes
- a few places where we could copy from slices
- improved some bits of documentation

metadata/src/lib.rs
jsdw: Just to follow up; Alex pointed out on our call this morning that a bunch of traits rely on being able to get information out of metadata (and the functions exposed on Metadata, while they could be refined, exist as more optimal ways to get at what we need).

so, while we could have a RuntimeMetadata new type to hide the underlying metadata version we use, we'd still need to provide enough information in the Metadata interface to get everything anybody might need for decoding/encoding things.

So this needs more thought, basically!

metadata/src/lib.rs
lexnv: I've removed this after all since we were only using it in:
- codegen while extracting the V15 metadata
- metadata crate for unit tests

We could add it back later if need be when we have a better grasp of the metadata :D

PR #913, opened 2023-04-14
Factor substrate node runner into separate crate

### LINKED ISSUE #909: Test-runtime: Run Substrate node on port 0 for more reliable execution
We've ran into issues a couple of times now (eg #908) where a nightly build fails because it tried to connect to a Substrate node 6 times without success.

The way that we start said Substrate node up is by selecting a port to use like so:

```rust
fn next_open_port() -> Option<u16> {
    match TcpListener::bind(("127.0.0.1", 0)) {
        Ok(listener) => {
            if let Ok(address) = listener.local_addr() {
                Some(address.port())
            } else {
                None
            }
        }
        Err(_) => None,
    }
}
```

And then telling Substrate to start using that port.

I wonder whether the port isn't freed up quickly enough sometimes, and so Substrate picks another random port to start on or something like that, leading to an inability to connect to it?

In any case, in our tests we take a different approach; we start Substrate with a `--ws-port=0`, and then parse the `stdout` until we find the actual port that it picked. This may be a more reliable approach to take, since we'll always try connecting on the actual port that it used, and won't run the issue of providing a port that ends up being in use. So, let's update our build code to do the same thing.

jsdw:
So that we can re-use the logic for starting a test substrate node up in both the `test-runtime` crate (to hopefully avoid the retry errors we've been seeing a bit lately) and the `integration-tests` crate which also spawns a node.

I added support for [@niklasad1's upcoming PR](https://github.com/paritytech/substrate/pull/13384) in terms of capturing stdout but not in terms of handling the invalid flag; perhaps I should! (but until then, this should still be a bit of an improvement I hope)

I also removed the Subxt dep on the `build.rs` of `test-runtime` in the hopes that it would mean it is triggered much less often (ie not every time Subxt is changed), but let's see aye!

Closes #909


jsdw:
When https://github.com/paritytech/subxt/pull/907 merges I'll remove the `parity-scale-codec` dependency from `test-runtime` too

testing/substrate-runner/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Spawn the node, handing back an object which, when dropped, will stop it.
+    pub fn spawn(self) -> Result<SubstrateNode, Error> {
+        let mut cmd = Command::new(self.binary_path);
+
+        cmd.env("RUST_LOG", "info")
+            .arg("--dev")
+            .stdout(process::Stdio::piped())
+            .stderr(process::Stdio::piped())
+            .arg("--port=0")
```
niklasad1, line 63: why do we care about the lip2p port? :)

testing/substrate-runner/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+            .arg("--rpc-port=0")
```
niklasad1, line 64: You could omit the port for both `rpc--port, --ws-port` here because substrate would fallback to `0` anyway, this way it would be backward-compatible as well :)

testing/integration-tests/Cargo.toml
```diff
@@ -35,3 +35,4 @@ tracing = "0.1.34"
 tracing-subscriber = "0.3.11"
 wabt = "0.10.0"
 which = "4.4.0"
+substrate-runner = { path = "../substrate-runner" }
```
niklasad1, line 38:
```suggestion
substrate-runner = { path = "../substrate-runner" }

```

niklasad1 (approved):

tadeohepperle (approved):

testing/substrate-runner/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "substrate-runner"
+version = "0.28.0"
+edition = "2021"
+publish = false
+
+[dependencies]
```
lexnv, line 7:
```suggestion
[dependencies]

```

lexnv (approved):

testing/substrate-runner/src/lib.rs
jsdw: I think the reasoning is to prevent any overlap with the port used for other substrate instances we spawn; I guess Substrate will pick a random port for us already if the one we give isn't any good, but probably just getting the kernel to pick an OK port straight off is a touch better anyway?

testing/substrate-runner/src/lib.rs
jsdw: Hmm but ah true; backward compat is a good reason not to have them; I'll remove :)

testing/substrate-runner/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+impl Drop for SubstrateNode {
+    fn drop(&mut self) {
+        let _ = self.kill();
+    }
+}
+
+// Consume a stderr reader from a spawned substrate command and
+// locate the port number that is logged out to it.
+fn try_find_substrate_port_from_output(r: impl Read + Send + 'static) -> Option<u16> {
+    BufReader::new(r).lines().take(50).find_map(|line| {
```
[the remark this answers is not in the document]
jsdw: I guess I wanted it to fail reasonably quickly if for some reason it wasn't picking up the port; I started at 100 but I think I counted right now that it's something like 25 lines to get the port, so 50 was my "safe" number; maybe 100 to compromise a bit? It should never take that long for the port to be logged I'd hope :)

PR #918, opened 2023-04-18
Metadata V15: Generate Runtime APIs

lexnv:
This PR adds support for the runtime API using the metadata exposed by substrate in https://github.com/paritytech/substrate/pull/13302.

## API Design

- [runtime_apis.rs](codegen/src/api/runtime_apis.rs) generates a new runtime API interface that allows users to call directly into the runtime of substrate-based chains
- The `Metadata_metadata` runtime function call is accessible via `metadata().metadata()`. This preserves two indirections: one for the runtime API trait (identified by the first `metadata()`) and one for the concrete function to call from the runtime trait. This follows the design decisions in subxt to utilize empty struct objects instead of paths for improved ergonomics and formatting. 
- Arguments can be passed to the runtime functions that accept them. The provided arguments to the runtime function call are scale-encoded in the codegen directly.
 
- codegen API using the `subxt` macro
```rust
pub struct RuntimeApi;
impl RuntimeApi {
    pub fn metadata(&self) -> Metadata {
        Metadata
    }
    ...
    
 #[doc = " The `Metadata` api trait that returns metadata for the runtime."]
pub struct Metadata;
impl Metadata {
    #[doc = " Returns the metadata of a runtime."]
    pub fn metadata(
        &self,
    ) -> ::subxt::runtime_api::StaticRuntimeApiPayload<runtime_types::sp_core::OpaqueMetadata>
    {
        let mut result = Vec::new();
        ::subxt::runtime_api::StaticRuntimeApiPayload::new_static(
            "Metadata_metadata",
            result,
            [
                158u8, 223u8, 229u8, 177u8, 228u8, 49u8, 24u8, 219u8, 143u8, 82u8, 255u8,
                132u8, 245u8, 210u8, 160u8, 197u8, 218u8, 10u8, 63u8, 165u8, 115u8, 91u8,
                70u8, 151u8, 246u8, 51u8, 18u8, 235u8, 91u8, 143u8, 27u8, 127u8,
            ],
        )
    }
```

- API is similar to submitting extrinsics: users must first construct the API payload and then submit it to the chain.

```rust
    // Create a runtime API payload that calls into
    // `Core_version` function.
    let call = polkadot::apis().core().version();

    // Submit the runtime API call.
    let version = api.runtime_api().at_latest().await?.call(call).await;
    println!("Core_version: {:?}", version);
```

## Runtime API Payload

This PR introduces two types of payloads: `StaticRuntimeApiPayload` and `DynamicRuntimeApiPayload`.
The payload that these types are built around takes in two generic parameters:
- Argument data
- Return type

Each payload must implement the following trait:

```rust

pub trait RuntimeApiPayload {
    /// The target type of the return value.
    type ReturnType: DecodeWithMetadata;

    /// The runtime API function name.
    fn fn_name(&self) -> &str;

    /// Scale encode arguments data to the provided output.
    fn encode_args_to(&self, metadata: &Metadata, out: &mut Vec<u8>) -> Result<(), Error>;

    /// Returns the details needed to validate the runtime API call, which
    /// include a statically generated validation hash.
    fn validation_details(&self) -> Option<ValidationDetails>;
}
```

The associated `ReturnType` is used for decoding the result of the function back to a provided type (either by codegen or "dynamically" by the user).

### StaticRuntimeApiPayload

This payload leverages the API exposed by the codegen and is strongly typed. 

The arguments are provided by the codegen as scale-encoded bytes, which makes the generic argument data `Vec<u8>`. In this case  `encode_args_to` ignores the `Metadata` parameters and passes through the bytes
obtained from the codegen directly.

Validation is enabled by default for these payloads, users can opt-out however using the `unvalidated()` method.

### DynamicRuntimeApiPayload

This offers users more control when it comes to making dynamic calls. Similar to transactions, 
users can construct a dynamic runtime API payload from various objects and the subxt
will do its best to encode those objects to the shape of the runtime metadata.

Build on top of https://github.com/paritytech/subxt/pull/912.

Closes https://github.com/paritytech/subxt-team-milestones/issues/21.


codegen/src/utils/fetch_metadata.rs
```diff
@@ -33,25 +34,81 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+    let hex_data = format!("0x{}", hex::encode(bytes));
     Ok(hex_data)
 }
+enum ClientType {
+    WebSocket(jsonrpsee::async_client::Client),
+    Http(jsonrpsee::http_client::HttpClient),
+}
+
+impl ClientType {
+    fn request<'life0, 'life1, 'async_trait, R, Params>(
```
niklasad1, line 61: you could also remove the `ClientType` and use `async fn request (client: impl jsonrpsee::core::client::ClientT, method, params) -> Result<R, E>`

doesn't matter

subxt/src/client/online_client.rs
```diff
@@ -136,10 +134,8 @@ impl<T: Config> OnlineClient<T> {
 
     /// Fetch the metadata from substrate using the runtime API.
     async fn fetch_metadata(rpc: &Rpc<T>) -> Result<Metadata, Error> {
-        let (_, meta) = rpc
-            .state_call::<(Compact<u32>, RuntimeMetadataPrefixed)>("Metadata_metadata", None, None)
-            .await?;
-        Ok(meta.try_into()?)
+        const V15_METADATA_VERSION: u32 = u32::MAX;
```
niklasad1, line 137: because v15 is not stabilized yet?

examples/examples/runtime_calls.rs
```diff
@@ -0,0 +1,58 @@
...
+    // `Core_version` function.
+    let call = polkadot::apis().core().version();
+
+    // Submit the runtime API call.
+    let version = api.runtime_api().at_latest().await?.call(call).await;
+    println!("Core_version: {:?}", version);
+
+    // Show the supported metadata versions of the node.
+    // Calls into `Metadata_metadata_versions` runtime function.
+    let call = polkadot::apis().metadata().metadata_versions();
```
niklasad1, line 39: maybe `runtime_api_call` to be extra clear.

codegen/src/utils/fetch_metadata.rs
```diff
@@ -33,25 +34,81 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+    {
+        match self {
+            ClientType::WebSocket(client) => client.request(method, params),
+            ClientType::Http(client) => client.request(method, params),
+        }
+    }
+}
+
+/// Execute runtime API call and return the specified runtime metadata version.
+async fn fetch_latest_metadata(client: ClientType) -> Result<Vec<u8>, FetchMetadataError> {
```
niklasad1, line 87:
```suggestion
async fn fetch_latest_metadata(&client: impl jsonrpsee::core::client::Client) -> Result<Vec<u8>, FetchMetadataError> { ... }
```

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Each argument of the runtime function must be scale-encoded.
+pub trait RuntimeApiPayload {
+    /// The target type of the return value.
+    // Note: `DecodeWithMetadata` bound implied by the `encode_as_fields_to`.
+    type ReturnType: DecodeWithMetadata;
+
+    /// The runtime API function name.
+    fn fn_name(&self) -> &str;
+
+    /// Scale encode arguments data to the provided output.
```
niklasad1, line 42: I'm got confused by `output` in my head this is the input params to the runtime api call but I see it's just `out: Vec<u8> that you refer to.

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,185 @@
...
+            fn_name: Cow::Borrowed(fn_name),
+            args_data,
+            validation_hash: Some(hash),
+            _marker: std::marker::PhantomData,
+        }
+    }
+}
+
+/// Create a new [`DynamicRuntimeApiPayload`].
+pub fn dynamic(
```
niklasad1, line 174: it's not possible to call `DynamicRuntimeApiPaylod::new(...)` instead of this?

It looks like Payload has `fn new()`?

codegen/src/utils/fetch_metadata.rs
lexnv: Looks much better now thanks!

metadata/src/retain.rs
```diff
@@ -211,6 +211,10 @@ where
     }
     update_extrinsic_types(&mut metadata.extrinsic, &map_ids);
     update_type(&mut metadata.ty, &map_ids);
+
+    // The retain pallets method strips the runtime APIs information,
+    // as it only takes into account working with pallets.
+    metadata.apis = Default::default();
```
jsdw, line 217: Are runtime APIs distinct from pallets then? I had assumed that each pallet would bring in some of it's own runtime APIs?

codegen/src/utils/fetch_metadata.rs
```diff
@@ -33,25 +36,75 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+        let opaque: Option<frame_metadata::OpaqueMetadata> = Decode::decode(&mut &raw_bytes[..])?;
+        let bytes = opaque.ok_or(FetchMetadataError::Other(
+            "Metadata version not found".into(),
+        ))?;
+
+        return Ok(bytes.0);
+    }
+
+    // The `Metadata_metadata_at_version` RPC failed, fall back to the old `state_getMetadata`.
+    let raw: String = client.request("state_getMetadata", rpc_params![]).await?;
```
jsdw, line 87: Maybe would be better to fall back to the old runtime API call instead?

codegen/src/utils/fetch_metadata.rs
```diff
@@ -33,25 +36,75 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+
+    let raw_bytes = hex::decode(raw.trim_start_matches("0x"))?;
+    // Decode the metadata to inspect the RuntimeMetadata variant.
+    let meta: RuntimeMetadataPrefixed = Decode::decode(&mut &raw_bytes[..])?;
+    // Update available versions to V15.
+    let metadata = match meta.1 {
+        RuntimeMetadata::V14(v14) => metadata_v14_to_latest(v14),
```
jsdw, line 94: Doesn't this mean that we'll always be returning V15 metadata from a node in a bunch of our CLI commands?

As you know, I'm keen to make sure that whereveer possible, the CLI commands (eg `subxt metadata`) returns whatever the node has available rather than modifying anything (unless we need to strip pallets or whatever, which will cause an upgrade).

I guess I'd be keen for this `metadata_v14_to_latest` call to be explicit in whatever command requires the upgrade, rather than buried away deeper in the fetch logic, so that it's super clear where exactly we are upgrading it :)

examples/examples/runtime_calls.rs
jsdw, line 1: This is cool for now; I'll have to add docs for this stuff eventually and at that point this example would probably move or change a little :)

examples/examples/runtime_calls.rs
```diff
@@ -0,0 +1,73 @@
...
+    let runtime_api_call = polkadot::apis().metadata().metadata_versions();
+
+    // Submit the runtime API call.
+    let versions = api
+        .runtime_api()
+        .at_latest()
+        .await?
+        .call(runtime_api_call)
+        .await?;
+    println!("Metadata_metadata_versions: {:?}", versions);
```
jsdw, line 53: Awesome!

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,186 @@
...
+        let mut v = Vec::new();
+        self.encode_args_to(metadata, &mut v)?;
+        Ok(v)
+    }
+
+    /// Returns the details needed to validate the runtime API call, which
+    /// include a statically generated validation hash.
+    fn validation_details(&self) -> Option<ValidationDetails> {
+        None
+    }
```
jsdw, line 58: To align with constants/storage I'd just have `validation_hash(&self) -> Option<[u8; 32]>`

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// The payload already contains the scale-encoded argument bytes.
+///
+/// This is only expected to be used from codegen.
+pub type StaticRuntimeApiPayload<ReturnTy> = Payload<Vec<u8>, ReturnTy>;
+
+/// A dynamic runtime API payload.
+pub type DynamicRuntimeApiPayload = Payload<Composite<()>, DecodedValueThunk>;
+
+impl<ReturnTy: Decode, ArgsData> Payload<ArgsData, ReturnTy> {
```
jsdw, line 124: do we need `Decode` here? I don't think anything should need to rely on this?

And if this goes away, we can merge the below `impl` too so `new()` and `new_static()` can live next to eachother similarto eg `TxPayload` :)

subxt/src/runtime_api/runtime_payload.rs
jsdw: This is really just for consistency I think with other modules that expose a `dynamic` function which is then made available from the `dynamic` module

metadata/src/retain.rs
lexnv: I'm not sure if we can confidently identify which pallets are bringing their own APIs.

For example, the `Metadata` [trait](https://github.com/paritytech/substrate/blob/471b49b09fd0d2a3057090ab0720f4e39faed5c5/primitives/api/src/lib.rs#L721) lives as a separate entity and users define and implement these with `decl_runtime_apis` and `impl_runtime_apis` which I don't think will bind the API to any specific pallet.

Although it might be possible to "identify" the pallet by trait name, similar to the [system pallet](https://github.com/paritytech/substrate/blob/471b49b09fd0d2a3057090ab0720f4e39faed5c5/bin/node-template/runtime/src/lib.rs#L461-L465). However, that might not be deterministic, in the sense that users can name their APIs different from the pallet that defines them. 

For the retain metadata, I removed the runtime API for simplicity.

We could add a flag on the CLI to also retain specified runtime traits similar to
```bash
retain --pallets System --runtime-apis Metadata
```

That would imply we add two filters to those methods:
- one filter for identifying the pallets
- another for identifying the runtime APIs: identify just the `trait_name` (ie `Metadata` instead of `Metadata_metadata` and other fns)

I think that would give users the most control over what they want to retain, let me know what you think of it 🙏

codegen/src/utils/fetch_metadata.rs
lexnv: Yep, that makes sense! Will modify this to return:
- V15 as `RuntimeMetadataPrefixed` if the `Metadata_metadata_at_version(u32::MAX)` exists
- V14 (or latest stable) as `RuntimeMetadataPrefixed` via `Metadata_metadata` runtime call 

We can then upgrade it later in the codegen when we generate the interface

subxt/src/runtime_api/runtime_payload.rs
lexnv: Yep, have removed the bounds and merged the code :D

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,185 @@
...
+/// contains the `metadata_at_version` function. The corresponding runtime function
+/// is `Metadata_metadata_at_version`.
+///
+/// - encoded arguments
+///
+/// Each argument of the runtime function must be scale-encoded.
+pub trait RuntimeApiPayload {
+    /// The target type of the return value.
+    // Note: `DecodeWithMetadata` bound implied by the `encode_as_fields_to`.
+    type ReturnType: DecodeWithMetadata;
```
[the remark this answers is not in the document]
niklasad1: Ideally, this associated type would removed from the trait and the impls has to specify this bound such as.

```rust
impl<D: DecodeWithMetadata> RuntimeApiPayload for DynamicRuntimeApiPayload<D> {
    fn fn_name(&self) -> &str {
        &self.fn_name
    }

    fn encode_args_to(&self, metadata: &Metadata, out: &mut Vec<u8>) -> Result<(), Error> {
        let fn_metadata = metadata.runtime_fn(&self.fn_name)?;

        self.args_data
            .encode_as_fields_to(fn_metadata.fields(), metadata.types(), out)?;

        Ok(())
    }
}
```

Doesn't matter that much to me as this is an internal trait used by subxt and shouldn't be needed to impl by "subxt users"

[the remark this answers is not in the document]
niklasad1: Up to you but would be good to document `ReturnType` such as (I might be wrong here but just as example):

> The `ReturnType` is not needed by the trait itself but dynamic implementations might use `scale_value::encode_as_field_to` which requires this `DecodeWithMetadata trait bound`

niklasad1 (approved):
good stuff :heart:

subxt/src/runtime_api/runtime_payload.rs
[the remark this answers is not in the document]
jsdw: The `ReturnType` associated type is needed so that the Subxt API call fn can look like:

```rust
fn call<P: RuntimeApiPayload>(&self, payload: P) -> Result<P::ReturnType,..>
```

ie so that the subxt API knows what return type to decode the result into and hand back.

Statically, we know what the type should be (and codegen it). But We can also return some dynamic type by setting `ReturnType` to `DecodedValueThunk` or something :)

metadata/src/retain.rs
jsdw: Ah thanks for the info!

Let's add another issue to allow trimming of the runtime APIs in the CLI tool (we don't need to do it here; it's not super important but would be nice for people to be able to keep only the runtime APIs they care about, if any)

And indeed, trimming pallets and trimming Runtime APIs should be different filters then; triming pallets shouldn't also remove the runtime APIs and vice versa :)

codegen/src/api/runtime_apis.rs
```diff
@@ -0,0 +1,136 @@
...
+                    [#(#call_hash,)*],
+                )
+            }
+        ))
+    }).collect::<Result<_, _>>()?;
+
+    let trait_name = format_ident!("{}", trait_name);
+    // The structure that identifies the `TraitName` of the runtime call.
+    let runtime_api = quote!(
+        #( #[doc = #docs ] )*
```
jsdw, line 85: I guess we should hide docs like elsewhere if some `should_gen_docs` bool provided to this call is false

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    fn encode_args_to(&self, metadata: &Metadata, out: &mut Vec<u8>) -> Result<(), Error> {
+        let fn_metadata = metadata.runtime_fn(&self.fn_name)?;
+
+        self.args_data
+            .encode_as_fields_to(fn_metadata.fields(), metadata.types(), out)?;
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
lexnv: Yep, that totally makes sense! 

To achieve that have created the method input types (structs that contain the parameters of the method).
That felt a bit more involved initially, but turned out nice!

```rust
    // Struct generated for consistency with the payload trait impl.
    #[derive( :: subxt :: ext :: scale_encode :: EncodeAsType ...)
    pub struct AccountNonce {
        pub account: ::subxt::utils::AccountId32,
    }
    
    pub fn account_nonce(
      &self,
      account: ::subxt::utils::AccountId32,
  ) -> ::subxt::runtime_api::Payload<inputs::AccountNonce, ::core::primitive::u32>
  {
      ::subxt::runtime_api::Payload::new_static(
          "AccountNonceApi_account_nonce",
          
          // Pass the struct input directly here similar to TXs.
          inputs::AccountNonce { account },
          ...
  }
```

While at it, have added a test to double check we encode / decode properly

codegen/src/api/runtime_apis.rs
```diff
@@ -0,0 +1,169 @@
...
+            use super::#types_mod_ident;
+
+            #docs
+            pub struct #trait_name;
+
+            impl #trait_name {
+                #( #methods )*
+            }
+
+            pub mod inputs {
```
jsdw, line 107: I quite like this `inputs` mod; maybe we should do the same for calls (and storage? and whatever)?

Not a blocker or anythign though :)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -33,25 +34,62 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+    let bytes = fetch_metadata_bytes(url).await?;
+    let hex_data = format!("0x{}", hex::encode(bytes));
     Ok(hex_data)
 }
 
-async fn fetch_metadata_ws(url: &Uri) -> Result<String, FetchMetadataError> {
+/// Execute runtime API call and return the specified runtime metadata version.
+async fn fetch_latest_metadata(client: impl ClientT) -> Result<Vec<u8>, FetchMetadataError> {
+    const V15_METADATA_VERSION: u32 = u32::MAX;
+    let bytes = V15_METADATA_VERSION.encode();
```
jsdw, line 59: Given this this will be changing and won't necessarily be available on nodes or compatible with the `frame-metadata` that we'll try decoding against, I wonder whether we should only get the latest _stable_ metadata by default, and require users to call eg `--unstable-metadata` or something to ask for the unstable metadata?

I think I'd sleep better at night knowing that getting the unstable metadata is very much opt-in; it also means we won't be breaking the interface a bunch for "stable" users, while letting people experiment with runtime API call generation before it's stabilised.

(this also goes along with my desire to hide the unstable metadata we use internally I guess, though that is much more challenging to do in a nice way I feel!)

Can be done as a separate PR, but I think should be done :)

metadata/src/retain.rs
```diff
@@ -191,6 +221,9 @@ where
     // Keep the "runtime" type ID, since it's referenced in our metadata.
     type_ids.insert(metadata.ty.id);
 
+    // Keep the runtime APIs types.
+    collect_runtime_api_types(&metadata.apis, &mut type_ids);
```
jsdw, line 225: nice! So runtime types are preserved now when pallets are retained I assume :)

metadata/src/validation.rs
```diff
@@ -404,6 +495,19 @@ pub fn get_metadata_hash(metadata: &RuntimeMetadataV15) -> [u8; 32] {
         &mut visited_ids,
     ));
 
+    let mut apis: Vec<_> = metadata
```
jsdw, line 498: This is ok for now, but thoughts:

- we set the capacity of bytes above; can we now adjust that to take into account these extra things we're adding to them?
- more generally I wonder whether we can get rid of the allocations entirely; iirc if you XOR a set of `[u8;32]`'s together for instance, it doesnt matter what order they are `xor`'d in and they will output the same result. So maybe we don't need to be sorting at all.
- need to think carefully about `xor` use in other places though because of this; `concat_and_hash` is generally "safer" and all of our XOR-ing might lead to a bunch of stuff that'll collide more. (ie I think we should maybe be xor-ing when we don't care about order of inputs and concat_and_hash-ing when we do, eg if we end up with nameA,input1A,input2A,nameB,input1B,input2B then we should not just xor all of these together because it'll be identical to if we swapped the inputs around, but we could `xor(concat_and_hash(nameA,input1A,input2A), concat_and_hash(nameB,input1B,input2B))` for instance to say "we care about those internal valeus being hashed uniquely but dont care about the order of those named things

Let's open an issue to look into this all!

subxt/src/rpc/rpc.rs
```diff
@@ -362,11 +362,37 @@ impl<T: Config> Rpc<T> {
...
+            .state_call("Metadata_metadata_at_version", Some(&param), None)
+            .await?;
+
+        let bytes = opaque.ok_or(Error::Other("Metadata version not found".into()))?;
+
+        let meta: RuntimeMetadataPrefixed = Decode::decode(&mut &bytes.0[..])?;
+
+        let metadata: Metadata = meta.try_into()?;
+        Ok(metadata)
+    }
```
jsdw, line 394: The CLI tool has a fallback to use the older methods if the new ones aren't available.

I'm a bit worried that if we merge this, Subxt will break for everybody running on a slightly older node until the new metadata interface is available everywhere.

I think for now that we should fall back to fetching the old version if we can't fetch the new one, either here or in `fetch_metadata` (we could then have a call like this `metadata_at_version` that just fetches the old one, and then in `fetch_metadata` try one after the other, for instance. At least that would give us some wiggle room so that Subxt would continue to work with older versions of Substrate.

testing/integration-tests/src/runtime_api/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+    let runtime_api_call = node_runtime::apis()
+        .account_nonce_api()
+        .account_nonce(alice);
+    let nonce = api
+        .runtime_api()
+        .at_latest()
+        .await?
+        .call(runtime_api_call)
+        .await?;
+    assert_eq!(nonce, 1);
```
jsdw, line 34: Maybe could also do this before submitting the call to see what it's 0 initially (if possible)?

jsdw:
Looking good! To summarise my comments so far:

- Add `--version=14`/`--version=unstable` to the CLI tool so people don't fetch unstable metadata by default and have to opt in
- Fall back to using old metadata interface in Subxt for fetching metadata if the new one isn't there (so Subxt doesn't stop working on all not-very-new Substrate versions)
- Open issue to look at metadata hashing and use of `xor`/allocating; I think we can improve some things here.
- Consider adding `mod inputs` to other things to be consistent with your Runtime API codegen

codegen/src/api/runtime_apis.rs
lexnv: After looking at the code a bit, leaving here my thoughts on this:

- For Calls we do something similar, starting from the Call enum we generate the structures of the payloads.
From what I see those structs are used as parameters for the payload itself, as a result, not necessarily as an input: `subxt::tx::Payload<Send>`. 
We could rename the `inputs` mod to `types` to provide a more generic naming that could be used here as well.

```rust
pub struct AddMemo {
    pub index: runtime_types::polkadot_parachain::primitives::Id,
    pub memo: ::std::vec::Vec<::core::primitive::u8>,
}

pub struct TransactionApi;
impl TransactionApi {
    pub fn add_memo(
        &self,
        index: runtime_types::polkadot_parachain::primitives::Id,
        memo: ::std::vec::Vec<::core::primitive::u8>,
    ) -> ::subxt::tx::Payload<AddMemo> {
        ::subxt::tx::Payload::new_static(
            "Crowdloan",
            "add_memo",
            AddMemo { index, memo }, ...
```
We could place that `AddMemo` into `pub mod types` for example.

- We use a similar approach to Calls for generating Events, because we registered the event type in the metadata as an enum.
The enum module generated for a pallet only contains the generated structs; as opposed to Calls that also generate the `struct TransactionApi`. Since there's nothing that could collide, I would leave Events untouched.

- The return type of Constants is already generated by the `TypeGenerator` (because that specific type is already present / registered in the metadata; as opposed to the `Call` enum or vector of type IDs from which we build structs).

- For storage entries, we use the return type directly similar to Constants. For the function parameters however, we pass the parameters directly to the `subxt::storage::address::make_static_storage_map_key`. And since I believe we don't have enough type information to name the parameters, we use `_0`, `_1` etc. 

```rust
pub struct StorageApi;
impl StorageApi {
pub fn queries(
      &self,
      _0: impl ::std::borrow::Borrow<::core::primitive::u64>,
  ) -> ::subxt::storage::address::Address<...
  > {
      ::subxt::storage::address::Address::new_static(
          "XcmPallet", "Queries",
          vec![::subxt::storage::address::make_static_storage_map_key(
              _0.borrow(),
          )],
      )
  }
``` 

And the `make_static_storage_map_key` returns:
https://github.com/paritytech/subxt/blob/b597b5d331e0385ba7a2973baeaf505cd405e104/subxt/src/storage/storage_address.rs#L213-L215

I'm wondering if we could generate structs for those storage calls tho. 
I think we didn't do that because we'd need to hash and encode independently each field of the struct, and that might turn out to be a bit tricky.

TLDR; will change the input to types and do a similar change for calls. Let me know what you think of it 🙏

metadata/src/validation.rs
lexnv: Referencing the issue: https://github.com/paritytech/subxt/issues/937, thanks for opening it!

codegen/src/api/runtime_apis.rs
jsdw: The example for calls that you gave:

```rust
impl TransactionApi {
    pub fn add_memo(
        &self,
        index: runtime_types::polkadot_parachain::primitives::Id,
        memo: ::std::vec::Vec<::core::primitive::u8>,
    ) -> ::subxt::tx::Payload<AddMemo> {
        ::subxt::tx::Payload::new_static(
            "Crowdloan",
            "add_memo",
            AddMemo { index, memo }, ...
```

Isn't that identical to the runtime API usage of structs, ie `AddMemo` is an input type to the payload just like it's an input on the runtime apis?

That said, calling it `types` instead of `inputs` I'd have no issue at all with; it's a bit more generic in case it's a useful module name to use in other places :)

Not a big deal either way; happy with whatyou you think best here!

lexnv: > Isn't that identical to the runtime API usage of structs, ie AddMemo is an input type to the payload just like it's an input on the runtime apis?

Yep, they are identical, we generate them almost the same way. 

Thanks! Will go ahead and place them under the `types` module

cli/src/utils.rs
```diff
@@ -16,30 +16,98 @@ pub struct FileOrUrl {
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct MetadataVersion {
+    version: u32,
+}
+
+impl std::str::FromStr for MetadataVersion {
+    type Err = String;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
+        const SUPPORTED_VERSIONS: &[u32] = &[14];
```
jsdw, line 86: Personally I don't think I'd have this; I'd just try fetching whatever version they ask for (or checking the list of versions the node provides) and returning an error if the version isn't available on that node.

Subxt itself may only support certain versions, but the macro/codegen can be the thing that complains if the user asks for a version that's not supported by Subxt, and the CLI tool can just do it's best to download whatever is asked for from a node :)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -12,14 +13,30 @@ use jsonrpsee::{
...
 
+/// The metadata version that is fetched from the node.
+#[derive(Default)]
+pub enum MetadataVersion {
+    /// Version 14.
+    #[default]
+    V14,
+    /// Latest unstable version of the metadata.
+    Unstable,
+}
```
jsdw, line 24: As per my earlier comment, I guess I'd also "future proof" this and just allow it to represent any possible version that may exist going forwards eg:

```rust
pub enum MetadataVersion {
    Stable(u32),
    Unstable
}
```

And then error only at the point at which the metadata obtained is seen to not be suitable (ie in the codegen when trying to decode and use it)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -32,26 +49,80 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
 }
 
-async fn fetch_metadata_ws(url: &Uri) -> Result<String, FetchMetadataError> {
+/// Execute runtime API call and return the specified runtime metadata version.
+async fn fetch_metadata(
+    client: impl ClientT,
+    version: MetadataVersion,
+) -> Result<Vec<u8>, FetchMetadataError> {
+    const UNSTABLE_METADATA_VERSION: u32 = u32::MAX;
+    const LATEST_STABLE_VERSION: u32 = 14;
```
jsdw, line 84: Instead of a constant, it'd be very cool if we made a runtime call to get the list of supported versions from the node and used that to know what the latest stable version was, since maybe there will be a mix of nodes with stable V14 and stable V15 for a while, and of course new versions eventually too

subxt/src/rpc/rpc.rs
```diff
@@ -362,11 +362,54 @@ impl<T: Config> Rpc<T> {
...
+        at: Option<T::Hash>,
+    ) -> Result<Res, Error> {
+        let bytes = self.state_call_raw(function, call_parameters, at).await?;
         let cursor = &mut &bytes[..];
         let res: Res = Decode::decode(cursor)?;
         Ok(res)
     }
 
+    /// Execute runtime API call and return the specified runtime metadata version.
+    pub async fn metadata_at_version(&self, version: u32) -> Result<Metadata, Error> {
```
jsdw, line 382: Nit: the below is called `state_call_metadata` (ie prefixed with `state_call`) but this isn't. maybe `metadata_at_version` and `metadata_legacy` are good name choices to also emphasise the other one being the old one?

subxt/src/runtime_api/runtime_types.rs
```diff
@@ -50,4 +52,57 @@ where
             Ok(data)
         }
     }
+
+    /// Execute a runtime API call.
+    pub fn call<Call: RuntimeApiPayload>(
```
jsdw, line 57: Awesome!

codegen/src/utils/fetch_metadata.rs
```diff
@@ -32,26 +49,80 @@ fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
...
+
+    // Fetch the latest stable version with `Metadata_metadata` API.
+    if version_num == LATEST_STABLE_VERSION {
+        let raw: String = client
+            .request("state_call", rpc_params!["Metadata_metadata", "0x"])
+            .await?;
+        let raw_bytes = hex::decode(raw.trim_start_matches("0x"))?;
+        let bytes: frame_metadata::OpaqueMetadata = Decode::decode(&mut &raw_bytes[..])?;
+        return Ok(bytes.0);
+    }
```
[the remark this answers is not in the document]
jsdw: Given that we chatted and we can get the stable version, does my above make sense in the end? :)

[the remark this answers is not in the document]
lexnv: That makes sense, had to add a few extra checks to make sure that the returned list by Substrate is not empty and to consider the latest stable as the second largest element if the u32::MAX is present in the list.

This should work for all changes regarding Substrate, except for the case in which users are allowed to specify the Runtime Metadata via feature flags. If developers will gain control of the runtime metadata and are able to specify V14 as stable, although other stable versions are available I think we'll not be technically correct here, but everything will keep on working :D

In the end, I like how the version list turned out, although it required some changes, it should keep us future proof :D

cli/src/utils.rs
```diff
@@ -105,9 +99,9 @@ impl std::str::FromStr for MetadataVersion {
 impl From<MetadataVersion> for CodegenMetadataVersion {
     fn from(input: MetadataVersion) -> CodegenMetadataVersion {
         match input.version {
-            14 => CodegenMetadataVersion::V14,
             u32::MAX => CodegenMetadataVersion::Unstable,
-            _ => panic!("MetadataVersion and CodegenMetadataVersion are not in sync!"),
+            14 => CodegenMetadataVersion::Latest,
```
jsdw, line 103: Nit: maybe just manually have an `into_codegen_metadata_version(metadata_version: Option<MetadataVersion>) -> CodegenMetadataVersion` so we don't have to hardcode the 14 here?

jsdw: Or clap has `default_value = <str>` which we could use so that `MetadataVersion` could have a `Latest` variant and clap will default to using that. (maybe we can just use `CodegenMetadataVersion` directly in that case?)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -75,48 +77,122 @@ pub async fn fetch_metadata_hex(
...
+            rpc_params!["Metadata_metadata_versions", "0x"],
+        )
+        .await
+        .ok()
+        .map(|raw: String| {
+            let raw_bytes = hex::decode(raw.trim_start_matches("0x"))?;
+            let versions: Vec<u32> = Decode::decode(&mut &raw_bytes[..])?;
+            Ok::<Vec<u32>, FetchMetadataError>(versions)
+        })
+        .transpose()?;
```
jsdw, line 111: Does the `?` here lead to us exiting early if this call fails?

lexnv: The code here seems a bit complex after another look, I'll take a note to go back and do a plain-old match to make it easier to follow:

- we initially do a `.ok()` which throws away the `Result<String, Error>` from the client and wraps into a `Option<String>`
- we transpose an Option<Result<..>>, out of which the Result can only be part of the closure: hex decoding or coded::decod  failed

I believe we should be safe here, I'll be a bit more explicit with a simple match and such :D

cli/src/utils.rs
lexnv: That makes sense! Have ended up using just the `CodegenMetadataVersion` here :D Thanks!

jsdw (approved):
Looks awesome; good job on this!

PR #929, opened 2023-04-24
extrinsics: Decode extrinsics from blocks

lexnv:
This PR adds support to decode extrinsics from blocks.

The extrinsic is composed of:
- 1 byte to signal the version and if the payload is signed or not
- for a signed payload only:
   - Address
   - Signature
   - Extra
- Call data

The type IDs of the Address, Signature, Extra and Call data are dynamically extracted from the metadata.
This is used then during the raw decoding to skip over those fields.
Currently, the Signature and Extra fields are not exposed or tracked internally.

The interface exposes the ability to fetch bytes or decode on a custom type the
- `Address` signer of the extrinsic payload
- `Call` the raw function of the extrinsic

Output from the added example

```bash
 Block 0x3620a9782142e9d523f5aa8fa0a043e9824285130eb55df6e61a70544136eccf
  Extrinsic index 0
  Extrinsic call: Timestamp(set { now: 1682353806000 })
  Extrinsic index 1
  Extrinsic call: ParaInherent(enter { data: InherentData { bitfields: [UncheckedSigned { payload: AvailabilityBitfield(DecodedBits { .. } })
  Extrinsic index 2
  Extrinsic call: Balances(transfer { dest: Id(AccountId32([142, 175, 4, 21, 22, 135, 115, 99, 38, 201, 254, 161, 126, 37, 252, 82, 135, 97, 54, 147, 201, 18, 144, 156, 178, 38, 170, 71, 148, 242, 106, 72])), value: 1100000000 })
  Extrinsic address: Some(Ok(Id(AccountId32([212, 53, 147, 199, 21, 253, 211, 28, 97, 20, 26, 189, 4, 169, 159, 214, 130, 44, 133, 88, 133, 76, 205, 227, 154, 86, 132, 231, 165, 109, 162, 125]))))
```

### Next Steps
 - [x] Custom metadata wrapper for `ExtrinsicMetadata`
 - [x] Introduce a `StaticExtrinsic` trait to identify pallets and variant index
 - [x] Decode using `StaticExtrinsic` after changes in codegen


subxt/src/blocks/block_types.rs
```diff
@@ -82,6 +78,12 @@ where
         ))
     }
 
+    /// Return the extrinsics associated with the block.
+    pub async fn extrinsics(&self) -> Result<extrinsics::Extrinsics<T>, Error> {
+        let block_details = self.block_details().await?;
+        extrinsics::Extrinsics::new(self.client.metadata(), block_details.block)
+    }
+
```
jsdw, line 86: I was expecting all of this extrinsics stuff to end up coming from `BlockBody`, below, with the idea that:

- you get hold of a block, and then load the body with `block.bocy().await?` 
- and then from the body you can iterate thorugh extrinsics already with `body.extrinsics()`
- and then each extrinsic you get back can be decoded (and that `Extrinsic` thing can have an interface that is inspired by `EventDetails`.

I don't think it really makes sense to have a separate top level `extrinsics` folder, because we already have a `tx` folder for constructing transactions (extrinsics) anyway, and then this `blocks` folder is all about getting and working with blocks, whose bodies are simply a list of extrinsics :)

So yeah I guess I'd maybe move the `Extrinsic` type defined already below into a separate file in this folder if it'll be getting quite big and then adding to that type the logic for decoding them (and then no need for eg an `ExtrinsicsClient` :)

subxt/src/error/mod.rs
```diff
@@ -112,6 +115,25 @@ impl BlockError {
...
+    #[error("Expected more extrinsic bytes")]
+    InsufficientData,
+    /// Unsupported signature.
+    #[error("Unsupported extrinsic version, only version 4 is supported currently")]
+    /// The extrinsic has an unsupported version.
+    UnsupportedVersion(u8),
+    /// Decoding error.
+    #[error("Cannot decode extrinsic: {0}")]
+    DecodingError(codec::Error),
+}
```
jsdw, line 135: Given the above, these can then be a part of `BlockError` or something where needed

subxt/src/extrinsics/extrinsics_type.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// The pallet index.
+    pallet_index: u8,
+    /// The variant index.
+    variant_index: u8,
+    /// Extrinsic bytes.
+    bytes: Arc<[u8]>,
+}
+
+impl ExtrinsicDetails {
+    // Attempt to dynamically decode a single event from our events input.
```
jsdw, line 189: Comment needs updating :)

subxt/src/extrinsics/extrinsics_type.rs
```diff
@@ -0,0 +1,344 @@
...
+                scale_decode::visitor::IgnoreVisitor,
+            )
+            .map_err(scale_decode::Error::from)?;
+        }
+
+        let call_start_idx = extrinsic_bytes.len() - cursor.len();
+
+        // Ensure the provided bytes are sound.
+        scale_decode::visitor::decode_with_visitor(
+            cursor,
```
jsdw, line 255: Nit: instead of moving the cursor here and resetting it below, perhaps here use `&mut *cursor` to "copy" the cursor and then we don't need to re-declare it below?

subxt/src/extrinsics/extrinsics_type.rs
```diff
@@ -0,0 +1,344 @@
...
+            ids.call,
+            &metadata.runtime_metadata().types,
+            scale_decode::visitor::IgnoreVisitor,
+        )
+        .map_err(scale_decode::Error::from)?;
+
+        // Decode the pallet index, then the call variant.
+        let cursor = &mut &extrinsic_bytes[call_start_idx..];
+        let pallet_index: u8 = Decode::decode(cursor).map_err(Error::Codec)?;
+        let variant_index: u8 = Decode::decode(cursor).map_err(Error::Codec)?;
```
jsdw, line 265: If the decode above works, we could also just do:

```rust
let pallet_index = cursor[0];
let variant_index = cursor[1];
```

But I'm ok with having the extra safety :)

subxt/src/extrinsics/extrinsics_type.rs
```diff
@@ -0,0 +1,344 @@
...
+    pub fn as_address<T: Decode>(&self) -> Option<Result<T, Error>> {
+        self.address_bytes()
+            .map(|bytes| T::decode(&mut &bytes[..]).map_err(Error::Codec))
+    }
+
+    /// Attempt to statically decode the extrinsic call bytes into the provided type.
+    pub fn as_call<T: Decode>(&self) -> Result<T, Error> {
+        let bytes = &mut &self.call_bytes()[..];
+        T::decode(bytes).map_err(Error::Codec)
+    }
```
jsdw, line 338: Can we use `decode_as_fields` instead for this? Since we know the type ID we want to decode as and have metadata?

I'm keen to avoid statically decoding stuff with eg `Decode` wherever possible.

For events, we have a `StaticEvent` type which allows us to point at specific event fields, and know how to decode them then from the metadata. I think this is on your todo list already (but I wouldn't want to merge the `Decode` version basically :))

Also it's not super consistent at the mo because this `as_call` looks like it decodes the entire enum, whereas the event's `as_event` decodes just the event fields (and `as_root_event` decodes the entire enum), so yeah just have a check and make sure that these method names are consistent with the event names when done :)

subxt/src/blocks/block_types.rs
lexnv: Have moved the extrinsic to the `BlocksBody` to iterate over them as `extrinsics()`

One other thing is that we still have the `find` apis that we want to mirror from the events.
Have added them with `_extrinsic` prefix as in `find_extrisic`, for extra clarity.

That makes me think that we could add something of the following:
```
struct BlockBody {
    fn extrinsics() -> Extrinsics
}

struct Extrinsics {
   fn find() ...
   fn find_first() ...
   fn iter() -> ExtrinsicDetails 
}
```

However, the `Extrinsics` struct from here would contain almost the same information as the `BlockBody` in this example. 
Let me know what you think of it 🙏

subxt/src/extrinsics/extrinsics_type.rs
```diff
@@ -0,0 +1,344 @@
...
+use codec::Decode;
+use derivative::Derivative;
+use frame_metadata::v15::RuntimeMetadataV15;
+use std::{collections::HashMap, sync::Arc};
+
+/// A collection of extrinsics obtained from a block, bundled with the necessary
+/// information needed to decode and iterate over them.
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""), Clone(bound = ""))]
+pub struct Extrinsics<T: Config> {
```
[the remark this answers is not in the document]
lexnv: That makes sense!
The hash from the extrinsics is used to fetch the events in the implementation with `OnlineClient`

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -0,0 +1,888 @@
...
+            scale_decode::visitor::IgnoreVisitor,
+        )
+        .map_err(scale_decode::Error::from)?;
+
+        // Decode the pallet index, then the call variant.
+        let cursor = &mut &extrinsic_bytes[call_start_idx..];
+
+        if cursor.len() < 2 {
+            return Err(BlockError::InsufficientData.into());
+        }
```
jsdw, line 285: Aah yeah; in most places we just let `codec::Error` encapsulate this. Since we've decoded the whole thing above though, we must have more than 2 bytes and so I'd just remove this error check entirely and maybe a brief comment (that was we can remove that `InsufficientData` variant).

Your prev approach of decoding the bytes was probably safer; maybe I shouldn't have opened my mouth in the first place because that worked nicely even if a little more verbose :)

subxt/src/blocks/block_types.rs
```diff
@@ -92,13 +91,23 @@ where
...
     }
+
+    /// Fetch the block's body from the chain.
+    async fn block_details(&self) -> Result<ChainBlockResponse<T>, Error> {
+        let block_hash = self.header.hash();
+        match self.client.rpc().block(Some(block_hash)).await? {
+            Some(block) => Ok(block),
+            None => Err(BlockError::not_found(block_hash).into()),
+        }
+    }
```
jsdw, line 102: This is only used in one place again now so maybe best to put back into `fn body()`?

subxt/src/error/mod.rs
```diff
@@ -102,6 +102,19 @@ pub enum BlockError {
     /// An error containing the hash of the block that was not found.
     #[error("Could not find a block with hash {0} (perhaps it was on a non-finalized fork?)")]
     NotFound(String),
+    /// Extrinsic type ID cannot be resolved with the provided metadata.
+    #[error("Extrinsic type ID cannot be resolved with the provided metadata. Make sure this is a valid metadata")]
+    MissingType,
+    /// Expected more extrinsic bytes.
+    #[error("Expected more extrinsic bytes")]
+    InsufficientData,
```
jsdw, line 110: Hopefully we can remove this as per above comment :)

testing/integration-tests/src/blocks/mod.rs
```diff
@@ -118,3 +120,71 @@ async fn runtime_api_call() -> Result<(), subxt::Error> {
...
+    let extrinsics = block.body().await.unwrap().extrinsics();
+    assert_eq!(extrinsics.len(), 2);
+    assert_eq!(extrinsics.block_hash(), block_hash);
+
+    assert!(extrinsics
+        .has::<node_runtime::balances::calls::Transfer>()
+        .unwrap());
+
+    assert!(extrinsics
+        .find_first::<node_runtime::balances::calls::Transfer>()
```
jsdw, line 163: Awesome :)

testing/integration-tests/src/blocks/mod.rs
```diff
@@ -118,3 +120,71 @@ async fn runtime_api_call() -> Result<(), subxt::Error> {
...
+        .collect::<Vec<_>>();
+
+    assert_eq!(block_extrinsics.len(), 2);
+    let timestamp = block_extrinsics.get(0).unwrap();
+    timestamp.as_root_extrinsic::<node_runtime::Call>().unwrap();
+    timestamp
+        .as_pallet_extrinsic::<node_runtime::runtime_types::pallet_timestamp::pallet::Call>()
+        .unwrap();
+    timestamp
+        .as_extrinsic::<node_runtime::timestamp::calls::Set>()
```
jsdw, line 179: Aha nice!

jsdw:
This is awesome! Only one thing re extrinsic metadata that I think we need to make a `Result` because a panic could conceivably happen otherwise, but looking great!

subxt/src/blocks/extrinsic_types.rs
lexnv: Yep, that makes sense, we can utilize that codec::Error. Although I also like that we avoided some decodes which made the code a bit more efficient :D

subxt/src/blocks/block_types.rs
```diff
@@ -26,7 +27,7 @@ pub struct Block<T: Config, C> {
 
 // A cache for our events so we don't fetch them more than once when
 // iterating over events for extrinsics.
-type CachedEvents<T> = Arc<AsyncMutex<Option<events::Events<T>>>>;
+pub(crate) type CachedEvents<T> = Arc<AsyncMutex<Option<events::Events<T>>>>;
```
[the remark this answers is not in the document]
lexnv: We also use that in the `extrinsic_types.rs` which allows users to fetch events from an OnlineClient and some ExtrinsicDetails. I do like this functionality that allows the users to fetch the events of only a specific tx after doing some work with it :D

subxt/src/metadata/metadata_type.rs
```diff
@@ -485,11 +543,60 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
...
+        for variant in &type_def_variant.variants {
+            let pallet_name: Arc<str> = variant.name.to_string().into();
+            let pallet_index = variant.index;
+
+            // Pallet variants must contain one single call variant.
+            // In the following form:
+            //
+            // enum RuntimeCall {
+            //   Pallet(pallet_call)
+            // }
```
jsdw, line 571: I think it's optional that a pallet actually has any calls in it; do we _need_ to find them all here?

Ie if a pallet has no calls, then there's no need to try and generate a variant for it at all in the codegen?

subxt/src/metadata/metadata_type.rs
```diff
@@ -485,11 +543,60 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
             .find(|ty| ty.ty.path.segments == ["sp_runtime", "DispatchError"])
             .map(|ty| ty.id);
 
+        let extrinsic_ty = metadata
+            .types
+            .resolve(metadata.extrinsic.ty.id)
+            .ok_or(InvalidMetadataError::MissingType(metadata.extrinsic.ty.id))?;
+
+        let Some(call_id) = extrinsic_ty.type_params
```
jsdw, line 551: I geuss my comment should have been here; why is a missing call type an issue? :)

jsdw (approved):
Looks great to me :)

tadeohepperle (approved):

subxt/src/book/usage/extrinsics.rs
```diff
@@ -173,3 +175,52 @@
 //! Take a look at the API docs for [`crate::tx::TxProgress`], [`crate::tx::TxStatus`] and
 //! [`crate::tx::TxInBlock`] for more options.
 //!
+//! ## Decoding Extrinsics
+//!
+//! Subxt uses a block-centric API to allow users to decode extrinsics. This approach requires users to retrieve
+//! a block's body from the chain before being able to decode the extrinsics. By considering that the block's body
```
niklasad1, line 181:
```suggestion
//! a block's body from the chain before being able to decode the extrinsics. By considering that the block body
```

subxt/src/book/usage/extrinsics.rs
```diff
@@ -173,3 +175,52 @@
...
+//! consists of the extrinsics and an additional justification, this API aligns well with the operational nature of the chain
```
niklasad1, line 182: I think the polkadot docs is a slightly better

> The block body is made up of extrinsics representing the generalization of the concept of transactions. Extrinsics can contain any external data the underlying chain wishes to validate and track.

Mentioning justifications here is a bit hard to follow, it's specific finality proof which is not super important for subxt :)

niklasad1 (approved):
another really solid PR, good stuff

PR #932, opened 2023-04-25
Move all deps to workspace toml

jsdw:
Move all of our dependencies (and version) to the workspace toml, and reference them in crates where used.

Why?
- Prevents dupe versions of crates from being used across the workspace (or, makes it very much "opt-in" and explicit if you do need this)
- Makes bumping Subxt versions easy for releasing; just one file to change
- Makes bumping Substrate dep versions easier; all in one place.
- Makes it clear how many crates we import in total; might motivate us to cut down a bit ;)

**Note:** `package = "parity-scale-codec"` is needed in each crate, because the macro uses https://github.com/bkchr/proc-macro-crate, which relies on finding that key in the local TOML (see https://github.com/bkchr/proc-macro-crate/issues/34). The alternative is giving a manual crate path in all uses of the macro, but doesn't seem worth it.


tadeohepperle (approved):

testing/wasm-tests/target/CACHEDIR.TAG
niklasad1, line 1: remove?!

Cargo.toml
```diff
@@ -17,8 +21,69 @@ resolver = "2"
...
+sp-runtime = "23.0.0"
+sp-version = "21.0.0"
+
+# Subxt workspace crates:
+subxt = { version = "0.28.0", path = "subxt" }
+subxt-macro = { version = "0.28.0", path = "macro" }
+subxt-metadata = { version = "0.28.0", path = "metadata" }
+subxt-codegen = { version = "0.28.0", path = "codegen" }
+test-runtime = { path = "testing/test-runtime" }
+substrate-runner = { path = "testing/substrate-runner" }
```
niklasad1, line 89:
```suggestion
substrate-runner = { path = "testing/substrate-runner" }

```

Cargo.toml
```diff
@@ -7,6 +7,10 @@ members = [
     "testing/test-runtime",
     "testing/integration-tests",
     "testing/ui-tests",
+    ## This cannot be a workspace dependency, because it requires
```
niklasad1, line 10: can you move this comment above `exclude = ["testing/wasm-tests"]` instead? 

I found it a little bit hard to understand

niklasad1 (changes requested):
Great, just remove some unintended committed files and it should be ready to go

testing/wasm-tests/target/debug/.cargo-lock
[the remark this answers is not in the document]
jsdw: Argh these must have turned up when I ran `cargo hack` locally!

jsdw:
Comment moved and added files gone :)

testing/wasm-tests/Cargo.lock
niklasad1, line 1: do want this?

niklasad1 (approved):

testing/wasm-tests/Cargo.lock
jsdw: I debated that; I assume it's used to determine the deps to download for tests so maybe good to keep it to ensure complete consistency between us :)

PR #936, opened 2023-04-27
Add no_default_substitutions to the macro and cli

### LINKED ISSUE #927: Add `no_default_substitutions` to the macro/CLI tool
We have `no_default_derives` now to disable the default derives/attrs. Perhaps we should add `no_default_substitutions`, which is more widely useful (ie usable with code generated to work with Subxt) and would allow users complete control over exactly which generated types are substituted and which are not? 

# Cons:

1. This will have some caveats like `no_default_derives` does (eg certain defaults are added to work around things like eg `BTreeMap` imposing `Ord` constraints on generated types), and I wouldn't expect most users to want it, but it would be consistent with allowing people to opt out of default derives and users could conditionally derive what was needed and override things as they wish to work around issues themselves.
2. I'm wary of supporting more use cases that don't revolve around using the macro to generate code for the Subxt library; I don't want  us to end up with a bunch of complexity in the code to support more exotic uses that may be better off utilising their own codegen, since our macro stuff is geared towards generating a Subxt interface.

tadeohepperle:
fixes #927 

Similar to `no_default_derives` there is now a `no_default_substitutions` option for the macro and cli on how to perform type substitutions in the codegen.

Added a test to check that it really works.

### DIFF
--- a/cli/src/commands/codegen.rs
+++ b/cli/src/commands/codegen.rs
@@ -56,6 +56,11 @@ pub struct Opts {
     /// Defaults to `false` (default trait derivations are provided).
     #[clap(long)]
     no_default_derives: bool,
+    /// Do not provide default substitutions for the generated types.
+    ///
+    /// Defaults to `false` (default substitutions are provided).
+    #[clap(long)]
+    no_default_substitutions: bool,
 }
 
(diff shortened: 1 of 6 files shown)


jsdw (approved):
Nice; looks great!

lexnv (approved):
Nice job 👍

PR #945, opened 2023-05-03
Draft PR for checking the time it takes to execute CI jobs

tadeohepperle:


niklasad1:
@tadeohepperle I wonder if you could write a github action that get the time for each job in every PR and open issue if a job was slow such as over 5-8 mins or something like that?

tadeohepperle:
> a github action that get the time for each "test-job" in every PR and open issue if a "test-job" was slow such as over 5-8 mins or something like that?

That is a good idea, I will look into it. For now I will close this PR.

PR #947, opened 2023-05-04
Runtime APIs; don't ask for validation hash anywhere except new_static

jsdw:
It doesn't make any sense to expect people to provide it outside of `new_static` for our codegen; just a tiny tweak as I missed this in review :)


lexnv (approved):
Looks good to me! Thanks for adding this! 🙏

niklasad1 (approved):

PR #959, opened 2023-05-15
Have a pass over metadata validation

jsdw:
- Loosen struct and enum field ordering now that EncodeAsType and DecodeAsType render the order irrelevant
- Tighten up some other places where we used `xor` instead of `concat_and_hash`
- Get rid of sorting and such by relying on XOR to produce order-independent hashes
- merge the per pallet and global metadata hashing logic into one function (We may want to extend this for only specific runtime APIs too once we allow runtime APIs to be stripped from metadata too)

### DIFF
--- a/metadata/src/validation.rs
+++ b/metadata/src/validation.rs
@@ -486,6 +486,12 @@ pub struct MetadataHasher<'a> {
     specific_pallets: Option<Vec<&'a str>>,
 }
 
+impl<'a> Default for MetadataHasher<'a> {
+    fn default() -> Self {
+        MetadataHasher::new()
+    }
+}
+
 impl<'a> MetadataHasher<'a> {
     /// Create a new [`MetadataHasher`]
     pub fn new() -> Self {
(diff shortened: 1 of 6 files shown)


metadata/src/validation.rs
```diff
@@ -35,142 +32,154 @@ enum TypeBeingHashed {
...
     visited_ids: &mut HashSet<u32>,
-) -> [u8; 32] {
-    let mut bytes = get_type_hash(registry, field.ty.id, visited_ids);
-
-    // XOR name and field name with the type hash if they exist
-    if let Some(name) = &field.name {
-        bytes = xor(bytes, hash(name.as_bytes()));
-    }
+) -> [u8; HASH_LEN] {
+    let field_name_bytes = match &field.name {
```
lexnv, line 80: small nit: 
```rust
let field_name_bytes = field.name.as_ref().map(|name| hash(name.as_bytes())).unwrap_or_default()
```

lexnv (approved):

metadata/src/validation.rs
jsdw: I think the original is easier to read anyways :)

tadeohepperle (approved):
Looks good to me :)

PR #961, opened 2023-05-17
Retain specific runtime APIs

### LINKED ISSUE #933: CLI: Allow users to retain specific runtime APIs
Extracted from: https://github.com/paritytech/subxt/pull/918#discussion_r1176441186

tadeohepperle:
fixes #933 

I integrated the retaining of specific runtime APIs into the existing `retain_metadata_pallets` functionality, because it is more efficient to join the two operations in one place. As for the typing: I tried usind `Option<FnMut(&str) -> bool>` as a sensiple argument type for both `pallets_filter` and `runtime_apis_filter` but spent way too much time on lifetime issues and could not get it going well. So now I settled for just passing `|_| true` closures, when all elements should be retained. Not so nice, I know, but should work.

### DIFF
--- a/cli/src/commands/metadata.rs
+++ b/cli/src/commands/metadata.rs
@@ -49,24 +49,15 @@ pub async fn run(opts: Opts) -> color_eyre::Result<()> {
                 ));
             }
         };
-        match (opts.pallets.as_ref(), opts.runtime_apis.as_ref()) {
-            (Some(pallets), Some(runtime_apis)) => retain_metadata(
-                &mut metadata_v15,
-                |pallet_name| pallets.iter().any(|p| &**p == pallet_name),
-                |runtime_api_name| runtime_apis.iter().any(|p| &**p == runtime_api_name),
-            ),
-            (Some(pallets), None) => retain_metadata(
-                &mut metadata_v15,
-                |pallet_name| pallets.iter().any(|p| &**p == pallet_name),
-                |_| true,
-            ),
-            (None, Some(runtime_apis)) => retain_metadata(
-                &mut metadata_v15,
-                |_| true,
-                |runtime_api_name| runtime_apis.iter().any(|p| &**p == runtime_api_name),
-            ),
-            (None, None) => {}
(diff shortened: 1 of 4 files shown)


cli/src/commands/metadata.rs
```diff
@@ -25,27 +25,48 @@ pub struct Opts {
...
+                |_| true,
+            ),
+            (None, Some(runtime_apis)) => retain_metadata(
+                &mut metadata_v15,
+                |_| true,
+                |runtime_api_name| runtime_apis.iter().any(|p| &**p == runtime_api_name),
+            ),
+            (None, None) => {}
+        }
         metadata = metadata_v15.into();
```
jsdw, line 70: The way to avoid the repetition would be by boxing the closures like so:

```rust
let retain_pallets_fn: Box<dyn Fn(&str) -> bool> = match opts.pallets.as_ref() {
    Some(pallets) => Box::new(|name| pallets.iter().any(|p| &**p == name)),
    None => Box::new(|_| true)
};
let retain_runtime_apis_fn: Box<dyn Fn(&str) -> bool> = match opts.runtime_apis.as_ref() {
    Some(apis) => Box::new(|name| apis.iter().any(|p| &**p == name)),
    None => Box::new(|_| true)
};
retain_metadata(&mut metadata_v15, retain_pallets_fn, retain_runtime_apis_fn);
metadata = metadata_v15.into();
```

jsdw (approved):
Nice work!

cli/src/commands/metadata.rs
tadeohepperle: Ah, nice, yeah that is definitely a solution. I still want to keep a check that at least one of the arguments is set, because if not we can avoid calling the `retain_metadata` function alltogether.

jsdw: Yeah I wouldn't reaplce the initial check, just the highlighted lines :)

lexnv (approved):
Amazing PR! 👍

Issue #964, opened 2023-05-18
[light-client] Wasm compatible light-client


lexnv:
Expose a light client that is wasm compatible to be utilized with subxt.

Smoldot provides an example that interacts with javascript world.
The example relies on the javascript to expose a few core functions (see [here](https://github.com/smol-dot/smoldot/blob/43b33066bf0f2058c0c39c1cb735d8e8745c8f7f/wasm-node/rust/src/bindings.rs#L18-L69)).

There are a few ways that this could be approached:

- smoldot: implement required functions within rust wasm compatible (probably using `gloo`)
- smoldot: wasm-compatible bindgens to import js part of the project in rust
- investigate if using `substrate-connect` would be a viable / faster option

Issue #68, opened 2020-02-05
Unsigned extrinsics


leoyvens:
At a glance this library seems to assume extrinsics are signed, it would be nice to support unsigned extrinsics.

PR #965, opened 2023-05-18
light-client: Add experimental light-client support

### LINKED ISSUE #962: [light-client] Experimental support with smoldot
Add experimental support for the light-client integration in subxt.

The end goal is for subxt to expose a `LightClient` that utilises smoldot under the scenes.

- expose an `experimental-light-client` feature flag
- implement `RpcClientT` for the exposed light client
- adjust/modify/remove results of requests/subscriptions coming from Smoldot to align with Substrate produces responses

lexnv:
This PR exposes the `LightClient` that uses smoldot to connect to chains.

Smoldot exposes a single non-clonable object that exposes all responses of all methods and all subscriptions via
a `next()` method call. Because of this, a background task is spawned to multiplex the responses back
to the appropriate submitters.

Smoldot generates some extra events while submitting an extrinsic, this is related to the new RPC Spec V2.
As a result of that, spurious events emitted by `transaction_unstable_watchEvent` are ignored.

The `SubstrateTxStatus` needed to be changed as well to support both lower camelCase and upper CamelCase.

When using the smoldot from crates-io, the implementation panics at:

```bash
thread 'main' panicked at 'attempt to subtract with overflow', /Users/lexnv/.cargo/registry/src/index.crates.io-6f17d22bba15001f/smoldot-0.6.0/src/chain_spec.rs:444:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
However, using the latest from the smoldot repo does not reproduce the issue.

The PR exposes 2 ways to construct the Light Client:
- from a spec str 
- from a URL from which the chain spec is fetched

The former version is working as expected, while the latter seems to cause an overflow issue in smoldot

```bash
thread 'main' panicked at 'attempt to subtract with overflow', /Users/lexnv/.cargo/git/checkouts/smoldot-2dcbf637e11a77d4/43b3306/lib/src/chain_spec.rs:445:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Testing Done

Initially using smoldot to target a local node caused the client to hang with `GrandPa warp sync` warnings.
Using the latest branch, I was able to start a node as normal:

1. `./polkadot-6dc9e84dde2 --dev --alice --validator`
2. `cargo run --example tx_basic_light_client`

### Available functionality
The following functionality is tested both on a local node, as well as on the polkadot live chain:

- Submit and watch a transaction
- Subscribes to all finalized blocks using the old RPC method
- Subscribes to the head of the chain using the new `chainHead` RPC method
- Dynamically query constants
- Dynamically decode the events of the latest block
- Various RPC calls to ensure proper shape of the response.

### Known limitations

- Iterating over keys is not expected to work: the light client tries its best to provide some answers, although I've encountered the same 3 responses submitted over again
- chainHead_unstable_body response contains a `value: ["0xhex", "0xhex"]` shape, where subxt expects from substrate chains `result: "0xhex"`
- chainHead_unstable_storage response contains a `value: "0xhex"` shape, where subxt expects from substrate chains `result: "0xhex"` (this is handled internally)
- submitting a transaction using the old API will generate spurious events from the new API (chainHead_unstable_transaction -- handled internally)

Closes: https://github.com/paritytech/subxt/issues/962


ascjones:
Been dreaming about this for a long time :rocket: 

IIRC @rphmeier also expressed an interest in this.

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,275 @@
...
+                // Drain the subscription cache, that holds messages that
+                // were not propagated to this subscription yet (because the
+                // RPC server produced a notification before the user registered
+                // to receive notifications).
+                if let Some(cached_responses) = self.subscriptions_cache.remove(&id) {
+                    tracing::debug!(target: LOG_TARGET, "Some messages were cached before susbcribing");
+
+                    for response in cached_responses {
+                        if sender.send(response).await.is_err() {
+                            tracing::warn!(target: LOG_TARGET, "Cannot send notification to susbcription {:?}", id);
```
niklasad1, line 80:
```suggestion
                            tracing::warn!(target: LOG_TARGET, "Cannot send notification to subscription {:?}", id);
```

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,275 @@
...
+            BackendMessage::Request { id, sender } => {
+                self.requests.insert(id, sender);
+            }
+            BackendMessage::Subscription { id, sender } => {
+                // Drain the subscription cache, that holds messages that
+                // were not propagated to this subscription yet (because the
+                // RPC server produced a notification before the user registered
+                // to receive notifications).
+                if let Some(cached_responses) = self.subscriptions_cache.remove(&id) {
+                    tracing::debug!(target: LOG_TARGET, "Some messages were cached before susbcribing");
```
niklasad1, line 76:
```suggestion
                    tracing::debug!(target: LOG_TARGET, "Some messages were cached before subscribing");
```

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Handle the registration messages received from the user.
+    async fn handle_register(&mut self, message: BackendMessage) {
+        match message {
+            BackendMessage::Request { id, sender } => {
+                self.requests.insert(id, sender);
+            }
+            BackendMessage::Subscription { id, sender } => {
+                // Drain the subscription cache, that holds messages that
```
niklasad1, line 71: I wonder if this is needed anymore, IIRC this was a bug in the old jsonrpc server

fine let's keep this for now

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,275 @@
...
+                    }
+                    return;
+                }
+
+                // Subscription ID not registered yet, cache the response.
+                // Note: Compiler complains about moving the `result` for `.entry.and_modify(..).or_insert(..)`,
+                // because it sees it as used on both closures and apparently cannot determine that only one
+                // closure can be executed.
+                match self.subscriptions_cache.entry(id) {
+                    std::collections::hash_map::Entry::Occupied(mut entry) => {
```
niklasad1, line 125: nit, import `std::collections::hash_map::Entry` to avoid repeating here and a few lines below?

Cargo.toml
```diff
@@ -73,6 +73,11 @@ wabt = "0.10.0"
 wasm-bindgen-test = "0.3.24"
 which = "4.4.0"
 
+# Light client support:
+smoldot-light = { git = "https://github.com/smol-dot/smoldot.git", default-features = false }
```
niklasad1, line 77: do we want to `pin a commit rev here`? I reckon it smoldot can have a breaking changes etc, I guess will also prevent us from publishing subxt on crates.io?

Perhaps the idea is to make a wrapper crate with smoldot to begin with?

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        tracing::trace!(target: LOG_TARGET, "Task closed");
+    }
+}
+
+/// The RPC response from the light-client.
+/// This can either be a response of a method, or a notification from a subscription.
+#[derive(Debug, Clone)]
+enum RpcResponse {
```
niklasad1, line 209: you could also use the types from jsonrpsee but I guess jsonrpsee isn't used here anyway..

subxt/src/rpc/lightclient/client.rs
```diff
@@ -0,0 +1,376 @@
...
+    ///
+    /// # Note
+    ///
+    /// The notification messages are buffered internally to ensure that users will receive all
+    /// messages in the following case:
+    ///
+    /// * T0. [`Self::register_request()`].
+    /// * T1. submit a plain RPC method request.
+    /// * T2. the subscription produces a notification. (T2 happens before the user calls this method)
+    /// * T3. user parses the subscription ID from (T1) and calls [`Self::register_subscription`].
```
niklasad1, line 115: ok, I see why the cache is needed then. A little weird I would expect to register_subscription to occur before then reading the subscription notifications.

niklasad1 (commented):
Looks good,

How to deal with the `git dependency` is a blocker from my side otherwise really cool to have support for this in subxt

rphmeier:
Nice, looking forward to trying this out. Is it also intended to work in-browser with Wasm compilation as well as native executables?

lexnv:
Yep, we plan to add support for Wasm workloads as well in the context of https://github.com/paritytech/subxt/issues/964.

At present, this PR addresses the non-wasm light-clients (https://github.com/paritytech/subxt/issues/962).

Cargo.toml
jsdw: Hmm yeah, we need to reference something on crates.io really.

subxt/src/rpc/lightclient/background.rs
jsdw: Since jsonrpsee is behind a feature flag it'd be best to avoid using it outside of where necessary anyways :)

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,301 @@
...
+/// than the substrate full node.
+const BUFFER_NUM_NOTIFICATIONS: usize = 16;
+
+const LOG_TARGET: &str = "light-client-background";
+
+/// Message protocol between the front-end client that submits the RPC requests
+/// and the backend handler that produces responses from the chain.
+///
+/// The light client uses a single object [`smoldot_light::JsonRpcResponses`] to
+/// handle all requests and subscriptions from a chain. A background task is spawned
```
tomaka, line 28: That's correct, however "a chain" is actually reference-counted in smoldot's internals. If you add two chains with the same chain spec, it will only connect once and sync once. Rather than multiplex the requests, I would suggest you create multiple "chains", one per user/client/whatever. I can also add a `clone_chain` function to make this easier.

tomaka:
> while the latter seems to cause an overflow issue in smoldot

This indicates an "invalid" checkpoint.
I've put invalid in quotes, because the format of checkpoints is a clusterfuck that was never tackled properly: https://github.com/paritytech/substrate/issues/11184
Basically, Substrate just dumps a bunch of internal structs, and then smoldot tries to magically parse it. This panic indicates a parsing error.

Of course it is still a bug in smoldot because smoldot is never supposed to panic, but the bugfix that I will do is to return an error instead. I don't know how to fix the underlying issue.

tomaka:
> the implementation panics at:

After looking into it, it happens when the checkpoint in the chain spec was generated before the first block was authored.
I'll fix it in the next couple days.

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,340 @@
...
+                            " Cannot send method response to id {:?}",
+                            id
+                        );
+                    }
+                }
+            }
+            Ok(RpcResponse::Subscription { method, id, result }) => {
+                // Subxt calls into `author_submitAndWatchExtrinsic`, however the smoldot produces
+                // `{"event":"broadcasted","numPeers":1}` which is part of the RPC V2 API. Ignore
+                // this spurious event.
```
tomaka, line 130: Smoldot is also supposed to produce a `validated` event according to the new API. It doesn't because I haven't finished implementing it, but you might want to ignore it as well.

subxt/src/rpc/types.rs
```diff
@@ -239,30 +239,41 @@ pub type SystemProperties = serde_json::Map<String, serde_json::Value>;
 ///
 /// This is copied from `sp-transaction-pool` to avoid a dependency on that crate. Therefore it
 /// must be kept compatible with that type from the target substrate version.
+///
+/// Substrate produces `camelCase` events, while smoldot produces `CamelCase` events.
```
tomaka, line 243: Smoldot is not supposed to produce `CamelCase` events

tomaka: Opened https://github.com/smol-dot/smoldot/issues/610

tomaka: I didn't realize when I read this comment, but it's smoldot that generates `transaction_unstable_watchEvent` events when it shouldn't.

I've noticed this independently, and this has now been fixed in https://github.com/smol-dot/smoldot/pull/677

lexnv: Thanks a lot for fixing this! It will make the code simpler from our side!

subxt/Cargo.toml
```diff
@@ -90,3 +106,15 @@ sp-runtime = { workspace = true }
 sp-keyring = { workspace = true }
 sp-version = { workspace = true }
 assert_matches = { workspace = true }
+tracing-subscriber = { workspace = true }
```
lexnv, line 109: Note: this is really useful for making sure the light-client is working (ie GrandPa warp sync)

subxt/src/book/usage/light_client.rs
```diff
@@ -0,0 +1,87 @@
...
+//!     .build()
+//!     .await?;
+//! ```
+//!
+//! The following example connects to the live Polkadot chain and performs several queries using the Light Client.
+//!
+//! You can run the example using the following command:
+//!
+//! ```bash
+//! cargo run --example unstable_light_client_tx_basic --features="unstable-light-client"
```
lexnv, line 53: Note: this feels a bit too explicit in how to run the examples. Even after adding the `[[example]]` with required-features in Cargo.toml the examples will not start with appropriate feature flag, it seems related to: https://github.com/rust-lang/cargo/issues/4663.

subxt/Cargo.toml
```diff
@@ -40,6 +40,16 @@ jsonrpsee-web = ["jsonrpsee/async-wasm-client", "jsonrpsee/client-web-transport"
 # latest features exposed by the metadata.
 unstable-metadata = []
 
+# Activate this to expose the Light Client functionality.
+# Note that this feature is experimental and things may break or not work as expected.
+unstable-light-client = [
```
niklasad1, line 45: I'm think exposing a feature flag `unstable` which has all unstable stuff i.e, both metadata, light client and other stuff would be better.

subxt/Cargo.toml
niklasad1: you mean by getting the logs? 

Is there a specific smoldot log target/level that is useful for this? 
Could be useful with a comment in the example code I guess..

artifacts/polkadot_spec.json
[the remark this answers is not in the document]
niklasad1: aight, I see this is only for the example...

subxt/examples/unstable_light_client_tx_basic.rs
```diff
@@ -0,0 +1,47 @@
...
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt::init();
+
+    // Create a light client by fetching the chain spec of a local running node.
+    // In this case, because we start one single node, the bootnodes must be overwritten
+    // for the light client to connect to the local node.
+    let light_client = LightClientBuilder::new()
+        .trusted_url("ws://127.0.0.1:9944")
+        .bootnodes(
```
niklasad1, line 18: nice

subxt/Cargo.toml
jsdw: I think keeping them separate prob makes sense here, because `unstable-metadata` doesn't just add something but also changes how Subxt fetches metadata from a node, so one may plausibly want to try light clients without also having subxt try to fetch V15 metadata :)

artifacts/polkadot_spec.json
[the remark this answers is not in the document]
lexnv: Yep, we only use it for the live-chain example :D

subxt/Cargo.toml
lexnv: The logs look similar to:

```bash
2023-06-16T11:33:36.099647Z  INFO smoldot: Chain initialization complete for dev. Name: "Development". Genesis hash: 0x4e1e…05c2. State root hash: 0x50e69d13db00a26a2834fa5ed6d6f2d99c5b965a6170da09b682c2e817f105ae. Network identity: 12D3KooWFmJ4hSvYaYTVDnAmPvDgH775XG67XkwxfbAAALKGw8CY. Chain specification or database starting at: 0x4e1e…05c2 (#0)

2023-06-16T11:33:36.100055Z  WARN json-rpc-dev: The JSON-RPC client has just called a JSON-RPC function from the legacy JSON-RPC API (chain_getBlockHash). Legacy JSON-RPC functions have loose semantics and cannot be properly implemented on a light client. You are encouraged to use the new JSON-RPC API <https://github.com/paritytech/json-rpc-interface-spec/> instead. The legacy JSON-RPC API functions will be deprecated and removed in the distant future.

2023-06-16T11:33:37.439069Z  INFO sync-service-dev: GrandPa warp sync finished to #68 (0x4f1b…3651)
2023-06-16T11:33:37.439437Z  INFO runtime-dev: Finalized block runtime ready. Spec version: 9390. Size of `:code`: 1.4 MiB.

2023-06-16T11:33:38.681854Z  INFO tx-service-dev: Successfully validated transaction 0x8f19…4c46
```

The `INFO` works fine for use, we care about the `INFO sync-service-dev: GrandPa warp sync finished to #68 (0x4f1b…3651)` and `INFO runtime-dev: Finalized block runtime ready. Spec version: 9390. Size of `:code`: 1.4 MiB.` with which I had some problems in the past :D

subxt/src/book/usage/mod.rs
```diff
@@ -10,12 +10,14 @@
...
 //! - [Blocks](blocks)
 //! - [Runtime APIs](runtime_apis)
+//! - [Unstable Light Client](light_client)
 //!
 //! Alternately, [go back](super).
 
 pub mod blocks;
 pub mod constants;
 pub mod events;
+pub mod light_client;
```
jsdw, line 20: Probably we need to think about where is suitablke to put this in the book; I think it doesn't make sense here, but I can have a look at this properly though after this PR is merged so this is all good for now :)

subxt/src/book/usage/light_client.rs
jsdw: I think this is fine, though I'm actually surprised that doing a `cargo check --examples` in the `subxt` crate doesn't lead to some compilation error, since the feature isn't enabled by default and so the examples would not work... I'd be curious to know why/how this works?!

subxt/examples/unstable_light_client.rs
```diff
@@ -0,0 +1,150 @@
...
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt::init();
+
+    // Create a light client from the provided chain spec.
+    // Note: this connects to the live polkadot chain.
+    let light_client = LightClientBuilder::new()
+        .chain_spec(include_str!("../artifacts/polkadot_spec.json"))
+        .build()
+        .await?;
+    let api = OnlineClient::<PolkadotConfig>::from_rpc_client(Arc::new(light_client)).await?;
```
jsdw, line 36: I'm wondering about how this is exposed.

Right now the lightclient stuff is an implementation of `RpcClientT`, but I wonder if it's useful to expose it at that level, or whether it'd be better if we exposed `LightClient<T>` right next to `OnlineClient<T>` and `OfflineClient<T>` in the `subxt::client` module (and when it's stable we could expose it right at the top level too). 

Ie we could have an itnerface like:

```rust
// subxt/client/light_client.rs:

pub struct LightClientBuilder {...}

pub struct LightClient<T>(OnlineClient<T>);

impl OnlineClient<T> for LightClient<T> {...}
impl OfflineClient<T> for LightClient<T> {...}

// usage:

use subxt::client::{ LightClient, LightClientBuilder };

let api: LightClient<PolkadotConfig> = LightClientBuilder::new()
    .chain_spec(include_str!("../artifacts/polkadot_spec.json"))
    .build::<PolkadotConfig>()
    .await?;
```

Then the fact that it's internally just implementing `RpcClientT` is just an implementation detail and doesn't really matter. Also lets us change what sort of things we expose on `LightClient<T>` in the future if we want etc.

subxt/examples/unstable_light_client.rs
jsdw, line 1: This is awesome to check that it's all actually working as expected, though as an example for users to follow, the only really important bit is how to construct a LightClient, and then everything else is basically identical :)

I'd probably suggest that ultimately we only need the "basic" example here, but when we figure out how to test this better we take this sort of logic and have tests instead to check that the lightclient is working as expected.

I wonder whether with your recent relvations and findings to get this all working, it might be a bit easier now to know how to programmatically spin up a substrate node with the relevant properties, access the spec details from it and have a few tests under the "unstable-lightclient" feature flag which can run just to check that it can do what we hope in this way? :) 

(this can all be a follow up PR though; for now I'm fine with this example sticking around while we figure out docs and tests for this :))

subxt/src/rpc/lightclient/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn chain_spec(mut self, chain_spec: &'a str) -> Self {
+        self.chain_spec = Some(chain_spec);
+        self
+    }
+
+    /// Overwrite the bootnodes of the chain specification.
+    ///
+    /// Can be used to provide trusted entities to the chian spec, or for
+    /// testing environments.
+    pub fn bootnodes(mut self, bootnodes: impl Iterator<Item = &'a str>) -> Self {
```
jsdw, line 113: Nit; `impl IntoIterator` would make this a touch more ergonomic :)

subxt/src/rpc/lightclient/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// The chain spec must be obtained from a trusted entity.
+    ///
+    /// It can be fetched from a trused node with the following command:
+    /// ```bash
+    /// curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "sync_state_genSyncSpec", "params":[true]}' http://localhost:9944/ | jq .result > res.spec
+    /// ```
+    ///
+    /// # Note
+    ///
+    /// For testing environments, please populate the "bootNodes" if the not already provided.
```
jsdw, line 97: Or use `bootnodes()` on this builder thing to configure some? (which might be easier than editing the spec, so maybe is a better way to push people?)

subxt/src/rpc/lightclient/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+        .await
+        .map_err(|_| LightClientError::Handshake)?;
+
+    let client = ClientBuilder::default()
+        .request_timeout(core::time::Duration::from_secs(180))
+        .max_notifs_per_subscription(4096)
+        .build_with_tokio(sender, receiver);
+
+    client
+        .request("sync_state_genSyncSpec", rpc_params![true])
```
jsdw, line 238: I wonder how we'll get hold of this with Pierre's new API; is there some Runtime API to get hold of it for instance? If an RPC API we should use that to help future us, and if not then we should ask Pierre maybe?

subxt/examples/unstable_light_client.rs
jsdw: (and just to add; one nice thing about exposing a very minimum interface like this, rather than exposing any of the underlying RpcClientT impl etc, is that when the WASM version lands, we can just swap this `LightClient<T>` struct out with the wasm impl of it (if need by) and the rest of the interface hopefully remains identical :))

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+
+    /// Perform the main background task:
+    /// - receiving user registration for RPC method / subscriptions
+    /// - providing the results from the light client back to users.
+    pub async fn start_task(
+        &mut self,
+        backend: mpsc::Receiver<BackendMessage>,
+        rpc_responses: smoldot_light::JsonRpcResponses,
```
jsdw, line 271: Nit: The word `backend` for "messages coming from subxt" confuses me a little, because I see these messages as coming from the "frontend" or user if anything; for me, something like `from_subxt` and `from_node` might be clearer :)

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        let mut backend_event_fut = backend_event.next();
+        let mut rpc_responses_fut = rpc_responses_event.next();
+
+        loop {
+            match future::select(backend_event_fut, rpc_responses_fut).await {
+                // Message received from the backend: user registered.
+                Either::Left((backend_value, previous_fut)) => {
+                    let Some(message) = backend_value else {
+                        tracing::trace!(target: LOG_TARGET, "Frontend channel closed");
```
jsdw, line 292: This is exactly one of the confusing things; the thing is called `backend_value` but if it's not there, the "frontend" is closed :D

subxt/src/rpc/lightclient/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn max_subscriptions(mut self, max_subscriptions: u32) -> Self {
+        self.max_subscriptions = max_subscriptions;
+        self
+    }
+
+    /// After a chain has been added, it is possible to extract a "database" (in the form of a
+    /// simple string). This database can later be passed back the next time the same chain is
+    /// added again.
+    /// A database with an invalid format is simply ignored by the client.
+    pub fn database_content(mut self, database_content: &'a str) -> Self {
```
jsdw, line 143: Is this like a saved data file or something which can allow smoldot to save some work, I wonder?

There's no way to get back the database content, so I guess it may or may not be useful to expose this :)

subxt/src/book/usage/light_client.rs
```diff
@@ -0,0 +1,87 @@
...
+//! (i.e., it is the only process started), you need to manually add the following to the spec:
+//!
+//! ```bash
+//! {
+//!    "bootNodes": [
+//!      "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp"
+//!    ],
+//!    "genesis": { ... },
+//! }
+//! ```
```
jsdw, line 33: Is this true, or can one just use `.bootnodes()` on the builder to add the relevant boot node and not need to modify the downloaded spec? If so, I'd push people that way :)

subxt/examples/unstable_light_client.rs
```diff
@@ -0,0 +1,150 @@
...
+pub mod polkadot {}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt::init();
+
+    // Create a light client from the provided chain spec.
+    // Note: this connects to the live polkadot chain.
+    let light_client = LightClientBuilder::new()
+        .chain_spec(include_str!("../artifacts/polkadot_spec.json"))
```
jsdw, line 33: Is it a good idea to keep the `artifacts/polkadot_spec.json` around?

For documentation, we can just show an example that points to some made up spec path (since we `ignore` the docs pointing to this anyway). And then in the "full" example we can build from url anyway :))

Maybe it's worth keeping around still but I'm not sure if it'll end up falling out of date or something (or maybe it never would)? It's quite large (5.5MB) so I'm hesitant :)

subxt/src/rpc/lightclient/background.rs
```diff
@@ -0,0 +1,425 @@
...
+                    if sender.send(Ok(result)).is_err() {
+                        tracing::warn!(
+                            target: LOG_TARGET,
+                            " Cannot send method response to id {:?}",
+                            id
+                        );
+                    }
+                } else if let Some((sub_id_sender, sender)) = self.id_to_subscription.remove(&id) {
+                    let mut sub_id = result.to_string();
+                    sub_id.retain(|ch| ch.is_ascii_digit());
```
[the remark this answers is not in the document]
lexnv: The `Box<RawValue>` conversion `to_string()` will have extra quotes around our subscription ID.
This is because we parse the ` subscription_id: "123" ` directly to a `Box<RawValue>` I believe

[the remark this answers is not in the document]
jsdw: I wonder whether to `trim` the `"`'s then perhaps and then convert `to_string()`? that way if something goes wrong it'll be less likely to be something weird :D

.github/workflows/rust.yml
```diff
@@ -166,6 +166,37 @@ jobs:
...
+                  override: true
+
+            - name: Rust Cache
+              uses: Swatinem/rust-cache@988c164c3d0e93c4dbab36aaf5bbeb77425b2894 # v2.4.0
+
+            - name: Run tests
+              uses: actions-rs/cargo@v1.0.3
+              with:
+                  command: test
+                  args: --package integration-tests --features unstable-light-client -- --test-threads=2
```
niklasad1, line 198: hrm? can you explain this? :)

lexnv: This was testing an assumption about failures encountered in the CI, but have found the reason to be that the light client needs at least a few blocks to be produced from the Substrate node before connecting to it.

Will remove `--test-threads=2` soon :D

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+    client: smoldot_light::Client<Platform>,
+    /// The ID of the chain used to identify the chain protocol (ie. substrate).
+    ///
+    /// Note: A single chain is supported for a client. This aligns with the subxt's
+    /// vision of the Client.
+    chain_id: ChainId,
+    /// Unique ID for RPC calls.
+    request_id: usize,
+    /// Map the request ID of a RPC method to the frontend `Sender`.
+    requests: HashMap<String, oneshot::Sender<MethodResponse>>,
```
niklasad1, line 62: why are request IDs stored as Strings instead of usize? 

I prefer usize here and avoid a needless allocation :)

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+                method,
+                params,
+                sender,
+            } => {
+                let id = self.next_id();
+                let request = format!(
+                    r#"{{"jsonrpc":"2.0","id":"{}", "method":"{}","params":{}}}"#,
+                    id, method, params
+                );
+                let id = id.to_string();
```
niklasad1, line 108: same here would be nice to avoid this allocation

subxt/src/client/lightclient/background.rs
lexnv: It's something I wasn't entirely sure too. I was basically afraid that the subscription ID might be handed back in hex format since it's actually a number inside a string (ie result: "123").
But I think for now we could go away without this allocation and revisit things if subscription ID breaks compatibility over time

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+                        || result.to_string().contains("validated"))
+                {
+                    tracing::debug!(target: LOG_TARGET, "Ignoring notification {:?}", result);
+                    return;
+                }
+
+                if let Some(sender) = self.subscriptions.get_mut(&id) {
+                    // Send the current notification response.
+                    if sender.send(result).await.is_err() {
+                        tracing::warn!(
```
niklasad1, line 255: you could even downgrade these to `debug` as this really can only occur if the receiver is dropped but probably good to have verbose logging for now..

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// Perform the main background task:
+    /// - receiving requests from subxt RPC method / subscriptions
+    /// - provides the results from the light client back to users.
+    pub async fn start_task(
+        &mut self,
+        from_subxt: mpsc::Receiver<FromSubxt>,
+        from_node: smoldot_light::JsonRpcResponses,
+    ) {
+        let from_subxt_event = tokio_stream::wrappers::ReceiverStream::new(from_subxt);
+        let from_node_event = futures_util::stream::unfold(from_node, |mut from_node| async {
```
niklasad1, line 281: is `stream::unfold` used here to make this Unpin? Or why not use https://docs.rs/smoldot-light/latest/smoldot_light/struct.JsonRpcResponses.html#method.next directly?

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+                            id
+                        );
+                    }
+                }
+            }
+        };
+    }
+
+    /// Parse the response received from the light client and sent it to the appropriate user.
+    async fn handle_rpc_response(&mut self, response: String) {
```
niklasad1, line 183: looks like this function doesn't have to be async

subxt/src/client/lightclient/builder.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl LightClientBuilder {
+    /// Create a new [`LightClientBuilder`].
+    pub fn new() -> LightClientBuilder {
+        LightClientBuilder::default()
+    }
+
+    /// Overwrite the bootnodes of the chain specification.
+    ///
+    /// Can be used to provide trusted entities to the chian spec, or for
```
niklasad1, line 48:
```suggestion
    /// Can be used to provide trusted entities to the chain spec, or for
```

subxt/src/client/lightclient/builder.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// testing environments.
+    pub fn bootnodes<'a>(mut self, bootnodes: impl IntoIterator<Item = &'a str>) -> Self {
```
niklasad1, line 50: could be a good idea to make this fallible and decode as json values here directly instead of detecting that when the chainspec has been downloaded....

subxt/src/rpc/types.rs
tomaka: This is not really my problem, but my two is I'd suggest you update smoldot and revert these changes. There's no need to leave code that bypasses smoldot bugs when these bugs have been fixed.

subxt/src/client/lightclient/builder.rs
niklasad1: should probably be infallible anyway so shouldn't matter I guess..

subxt/src/client/lightclient/background.rs
lexnv: The sender channel for the subscription returns a future

subxt/src/client/lightclient/rpc.rs
```diff
@@ -0,0 +1,215 @@
...
+        ));
+
+        let smoldot_light::AddChainSuccess {
+            chain_id,
+            json_rpc_responses,
+        } = client
+            .add_chain(config)
+            .map_err(|err| LightClientError::AddChainError(err.to_string()))?;
+
+        let (to_backend, backend) = mpsc::channel(128);
```
niklasad1, line 58: could be good make this configurable at some point, just a warning that backpressure deadlocks are super hard to detect/debug.

Initially I would make these unbounded but your call :)

subxt/src/client/lightclient/rpc.rs
```diff
@@ -0,0 +1,215 @@
...
+        params: String,
+    ) -> Result<
+        (
+            oneshot::Receiver<MethodResponse>,
+            mpsc::Receiver<Box<RawValue>>,
+        ),
+        SendError<FromSubxt>,
+    > {
+        let (sub_id, sub_id_rx) = oneshot::channel();
+        let (sender, receiver) = mpsc::channel(128);
```
niklasad1, line 109: could be good make this configurable at some point, just a warning that backpressure deadlocks are super hard to detect/debug.

Initially I would make these unbounded but your call :)

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,428 @@
...
+                // Subxt calls into `author_submitAndWatchExtrinsic`, however the smoldot produces
+                // `{"event":"broadcasted","numPeers":1}` and `{"event":"validated"}` which are part
+                // of the RPC V2 API. Ignore those events.
+                if method == "transaction_unstable_watchEvent"
+                    && (result.to_string().contains("broadcasted")
+                        || result.to_string().contains("validated"))
+                {
+                    tracing::debug!(target: LOG_TARGET, "Ignoring notification {:?}", result);
+                    return;
+                }
```
tomaka, line 268: Same remark here. This is fixed.

testing/integration-tests/src/lib.rs
```diff
@@ -4,24 +4,27 @@
...
 #![deny(unused_crate_dependencies)]
 
-#[cfg(test)]
+#[cfg(all(test, not(feature = "unstable-light-client")))]
 mod codegen;
 #[cfg(test)]
 mod utils;
 
-#[cfg(test)]
+#[cfg(all(test, not(feature = "unstable-light-client")))]
```
niklasad1, line 12: you could add another module called `full_client` or something to reduce the feature compilation noise

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -0,0 +1,197 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! # Light Client Initialization and Testing
```
niklasad1, line 5: good explanation

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,446 @@
...
+
+                self.requests.insert(id, sender);
+
+                let result = self.client.json_rpc_request(request, self.chain_id);
+                if let Err(err) = result {
+                    tracing::warn!(
+                        target: LOG_TARGET,
+                        "Cannot send RPC request to lightclient {:?}",
+                        err.to_string()
+                    );
```
tomaka, line 120: I just want to point out that this can realistically happen. If a `TooBusy` error is returned, you're supposed to back-pressure the requests sender.

I understand that from your point of view it's not very convenient, so I've opened https://github.com/smol-dot/smoldot/issues/804 to clarify this thing.

niklasad1 (approved):
great PR, looks good to me

subxt/src/client/lightclient/background.rs
tomaka: To clarify: there will always be an error here. But maybe I can provide some configuration option that guarantees a minimum buffer size, so that you have a runtime guarantee that it won't happen.

subxt/src/client/lightclient/builder.rs
lexnv: Oh yeap, this simplifies the code quite a bit!
Have stored them as `Vec<serde_json::Value>` in the builder directly, thanks!

subxt/src/client/lightclient/rpc.rs
lexnv: Yep, I trust your intuition, have made these channels unbounded :D

subxt/src/rpc/types.rs
tomaka, line 243: Posting a new comment for visibility, because GitHub's UI isn't great.

Since this bug is fixed in smoldot, I'd suggest revert the changes you've made here.

subxt/src/client/lightclient/background.rs
lexnv: Yep exactly, without the `unfold` stream the `full_node.next()` would not implement `Unpin` and we'll get an error similar

```
within `impl futures::Future<Output = std::option::Option<std::string::String>>`, the trait `Unpin` is not implemented for `[async fn body@JsonRpcResponses::next::{closure#0}]`
```

subxt/src/client/lightclient/background.rs
lexnv: Yep, it totally works now! Thanks a lot for fixing this and for the review! 🙏

subxt/src/client/lightclient/background.rs
lexnv: Yep, I think we'll eventually push for a stable `light-client` feature flag and then we'll come back to this :D

testing/integration-tests/src/lib.rs
lexnv: Yep that makes a lot of sense, thanks!

subxt/src/client/lightclient/background.rs
tomaka: Sorry I was a bit confused this morning since I was working on something else in parallel.

What you should do is pass `max_pending_requests: u32::max_value()` and `max_subscriptions: u32::max_value()` when adding the chain. There's no need to make these parameters configurable.

These parameters exist to make sure that the memory usage of smoldot is bounded in situations where the JSON-RPC client is potentially malicious and can spam you with requests. This isn't relevant here, since you are the JSON-RPC client.

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,439 @@
...
+                sub_id,
+                sender,
+            } => {
+                // For subscriptions we need to make a plain RPC request to the subscription method.
+                // The server will return as a result the subscription ID.
+                let id = self.next_id();
+                let request = format!(
+                    r#"{{"jsonrpc":"2.0","id":"{}", "method":"{}","params":{}}}"#,
+                    id, method, params
+                );
```
tadeohepperle, line 155: I think constructing the `request` variable and incrementing the id could be done before the match arms to avoid code duplication:
```
let id = self.next_id();
let (FromSubxt::Request { method, params, .. } | FromSubxt::Subscription { method, params, .. } )= &message;
let request = format!(
    r#"{{"jsonrpc":"2.0","id":"{}", "method":"{}","params":{}}}"#,
    id, method, params
);
```

subxt/src/client/lightclient/background.rs
```diff
@@ -0,0 +1,439 @@
...
+use super::LightClientError;
+use smoldot_light::{platform::default::DefaultPlatform as Platform, ChainId};
+
+const LOG_TARGET: &str = "light-client-background";
+
+/// The response of an RPC method.
+pub type MethodResponse = Result<Box<RawValue>, LightClientError>;
+
+/// The type of a subscription ID.
+type SubscriptionId = usize;
```
tadeohepperle, line 21: nit: I would probably do the same for `RequestId`, that are also just represented as a usize or remove the type alias.

tadeohepperle (approved):
Very good work, I like the multiplexing over the background task a lot.

testing/integration-tests/Cargo.toml
```diff
@@ -15,6 +15,9 @@ description = "Subxt integration tests that rely on the Substrate binary"
 [features]
 default = ["subxt/integration-tests"]
 
+# Enable to run the tests with Light Client support.
+unstable-light-client = ["subxt/integration-tests", "subxt/unstable-light-client"]
```
jsdw, line 19: I did a grep and offhand can't see any reference to `integration-tests` nowadays, so I think this feature can just be removed from here and Subxt? (would be ideal; logic that is required specifically for testing is never desirable if it can be avoided :))

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+async fn various_rpc_calls(api: &Client) -> Result<(), subxt::Error> {
+    let _system_chain = api.rpc().system_chain().await?;
+    let _system_name = api.rpc().system_name().await?;
+    let _finalized_hash = api.rpc().finalized_head().await?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn light_client_testing() -> Result<(), subxt::Error> {
```
jsdw, line 182: This is exactly what I'd have done in the end too; nice testing!

testing/integration-tests/src/utils/wait_for_blocks.rs
```diff
@@ -11,4 +11,11 @@ pub async fn wait_for_blocks<C: Config>(api: &impl OnlineClientT<C>) {
     let mut sub = api.rpc().subscribe_all_block_headers().await.unwrap();
     sub.next().await;
     sub.next().await;
+
+    #[cfg(feature = "unstable-light-client")]
```
jsdw, line 15: Nit; maybe tweak the comment to note that if unstable-light-client is enabled it'll wait for 3 more blocks, because presumably that helps it work faster? :)

testing/substrate-runner/src/lib.rs
```diff
@@ -56,7 +56,7 @@ impl SubstrateNodeBuilder {
     pub fn spawn(self) -> Result<SubstrateNode, Error> {
         let mut cmd = Command::new(self.binary_path);
 
-        cmd.env("RUST_LOG", "info")
+        cmd.env("RUST_LOG", "info,libp2p_tcp=debug")
```
jsdw, line 59: TIL I didn't know about this syntax!

testing/integration-tests/src/lib.rs
```diff
@@ -5,27 +5,32 @@
...
-mod metadata;
-#[cfg(test)]
-mod runtime_api;
-#[cfg(test)]
-mod storage;
+// We expose the utils crate's contents as public to avoid
+// a few unused warnings when running the tests under the
+// "unstable-light-client" flag.
+#[cfg(all(test, feature = "unstable-light-client"))]
+pub use utils::*;
```
jsdw, line 14: I'd probably either:

1. Just always expose as pub; the less switching based on feature flags the better IMO (every feature flag usage makes our lives a little more complicated), or
2. Don't change the visibility (it's a bit of a hack) and just add the relevant `#[cfg(test, allow(unused)]` type thing (whatever it is) if that works; it more directly addresses the actual issue so feels clearer to me :)

subxt/src/client/lightclient/builder.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// # Note
+    ///
+    /// For testing environments, please populate the "bootNodes" if the not already provided.
+    /// See [`Self::bootnodes`] for more details.
+    ///
+    /// ```json
+    ///   "bootNodes": [
+    ///       "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp"
+    ///    ],
+    /// ```
```
jsdw, line 130: I think we should just point at `Self::bootnodes` here, and then if you want, the docs for that function can say something like "setting bootnodes here is equivalent to adding them to the chain spec like so"

subxt/examples/unstable_light_client_tx_basic.rs
```diff
@@ -0,0 +1,52 @@
...
+    // In this case, because we start one single node, the bootnodes must be overwritten
+    // for the light client to connect to the local node.
+    //
+    // The `12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp` is the P2P address
+    // from a local polkadot node starting with
+    // `--node-key 0000000000000000000000000000000000000000000000000000000000000001`
+    let api: LightClient<PolkadotConfig> = LightClientBuilder::new()
+        .bootnodes(
+            ["/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp"]
+                .into_iter(),
```
jsdw, line 26: No need for `into_Iter()` here now I think :)

jsdw (approved):
A few remaining nits and small things, but in general this is amazing stuff! Clean, well documented and tested. Good job!!

PR #967, opened 2023-05-19
Ensure unique types in codegen

### LINKED ISSUE #966: Types overwritten in codegen when parameters differ but paths are the same.
On westmint, the `RuntimeCall` enum has these variants:

```json
[
    // ...
    {
        "name": "Assets",
        "fields": [
        {
            "type": 279,
            "typeName": "self::sp_api_hidden_includes_construct_runtime::hidden_include::dispatch\n::CallableCallFor<Assets, Runtime>"
        }
        ],
        "index": 50
    },
    // ...
    {
        "name": "ForeignAssets",
        "fields": [
        {
            "type": 308,
            "typeName": "self::sp_api_hidden_includes_construct_runtime::hidden_include::dispatch\n::CallableCallFor<ForeignAssets, Runtime>"
        }
        ],
        "index": 53
    }
    ]
```

Note that we have two versions of the same "Assets" pallet under different names, and with slightly different types used.

The first pallet points to this inner enum:

```json
{
  "id": 279,
  "type": {
    "path": [
      "pallet_assets",
      "pallet",
      "Call"
    ],
    "params": [
      {
        "name": "T",
        "type": null
      },
      {
        "name": "I",
        "type": null
      }
    ],
    "def": {
      "variant": {
        "variants": [
          {
            "name": "create",
            "fields": [
              {
                "name": "id",
                "type": 48,
                "typeName": "T::AssetIdParameter"
              },
              // ....
```

The second to this enum:

```json
{
  "id": 308,
  "type": {
    "path": [
      "pallet_assets",
      "pallet",
      "Call"
    ],
    "params": [
      {
        "name": "T",
        "type": null
      },
      {
        "name": "I",
        "type": null
      }
    ],
    "def": {
      "variant": {
        "variants": [
          {
            "name": "create",
            "fields": [
              {
                "name": "id",
                "type": 45,
                "typeName": "T::AssetIdParameter"
              },
              // ....
```

Notes: 
- The paths to these types are identical.
- The "id" type is different (48 vs 45), ie `T::AssetIdParameter` is different.
- The `"params"` entries are both `null` in either case, so not very helpful, possibly because we rely on associated types.

Now, let's look at the codegen:

```rust
pub enum RuntimeCall {
    // ...
    #[codec(index = 50)]
    Assets(runtime_types::pallet_assets::pallet::Call),
    // ...
    #[codec(index = 53)]
    ForeignAssets(runtime_types::pallet_assets::pallet::Call),
}
```

Both versions of the assets pallet point to the exact same `Call` type, which looks like:

```rust
pub enum Call {
    // ...
    create {
        id: runtime_types::xcm::v3::multilocation::MultiLocation,
        admin: ::subxt::utils::MultiAddress<::subxt::utils::AccountId32, ()>,
        min_balance: ::core::primitive::u128,
    },
    // ...
```

So it seems as if the later `ForeignAssets` version of the call is what we're generating, and the other version is lost.

I think that the fix here is to modify the type generation such that if two types in the metadata have different IDs but the same path, we append eg a number to the end of the path for one of those IDs, so that we'd end up with a `Call` and a `Call2` being generated in the above case. 

I've attached the metadata that has this issue; one can use this to codegen and explore the JSON and ultiamtely confirm that the fix has worked.

[METADATA.scale.zip](https://github.com/paritytech/subxt/files/11515777/METADATA.scale.zip)

jsdw:
This feels like a bit of a hack, but basically the issue is:

- Multiple types with the same path can appear in the metadata.
- When those types have known type params, Subxt de-duplicates them and only generates one type.
- But we've run into a case where the type has no type params and the same path (owing to having two instances of a similar pallet, and perhaps something about generic types where only associated types are used not being captured properly in the metadata or something).
- Anyway, this leads to one of the types being overridden by the other. This PR aims to fix that by finding any dupe types that have no params and share a path, and de-duping the paths/names.

Closes #966


lexnv (approved):
LGTM! 👍

codegen/src/api/mod.rs
```diff
@@ -175,15 +175,56 @@ impl RuntimeGenerator {
...
+
         RuntimeGenerator { metadata }
     }
 
+    /// Ensure that every unique type we'll be generating or referring to also has a
+    /// unique path, so that types with matching paths don't end up overwriting each other
+    /// in the codegen. We ignore any types with generics; Subxt actually endeavours to
+    /// de-duplicate those into single types with a generic.
+    fn ensure_unique_type_paths(metadata: &mut RuntimeMetadataV15) {
+        let mut visited_path_counts = std::collections::HashMap::<Vec<String>, usize>::new();
```
niklasad1, line 194: hmm, are these type hints really needed?

testing/integration-tests/src/codegen/codegen_tests.rs
```diff
@@ -0,0 +1,97 @@
...
+            apis: vec![],
+        };
+
+        RuntimeMetadataPrefixed::from(metadata)
+    }
+
+    let metadata = generate_metadata_with_duplicate_types();
+    let interface = generate_runtime_interface_from_metadata(metadata);
+
+    assert!(interface.contains("FirstDupeTypeVariant"));
```
niklasad1, line 95: see my comment above, I don't understand if the the variants are merged to a single type or something else?

codegen/src/api/mod.rs
```diff
@@ -175,15 +175,56 @@ impl RuntimeGenerator {
...
+            }
+
+            let has_valid_type_params = ty.ty.type_params.iter().any(|tp| tp.ty.is_some());
+
+            // Ignore types which have generic params that the type generation will use.
+            // Ordinarily we'd expect that any two types with identical paths must be parameterized
+            // in order to share the path. However scale-info doesn't understand all forms of generics
+            // properly I think (eg generics that have associated types that can differ), and so in
+            // those cases we need to fix the paths for Subxt to generate correct code.
+            if has_valid_type_params {
```
niklasad1, line 208: Can't we just skip this check and modify duplicate the path the regardless to be on the safe side?

Maybe I don't understand the comment properly but I regard it as if it's generic is safe to have 
duplicate types because when those are instantiated it isn't an issue?

Would be cool to to have test for it :)

testing/integration-tests/src/codegen/codegen_tests.rs
```diff
@@ -0,0 +1,97 @@
...
+        registry.register_type(&meta_type::<RuntimeEvent>());
+        registry.register_type(&meta_type::<DispatchError>());
+
+        // Now we duplicate some types with same type info:
+        enum Foo {}
+        impl TypeInfo for Foo {
+            type Identity = Self;
+            fn type_info() -> scale_info::Type {
+                scale_info::Type::builder()
+                    .path(scale_info::Path::new("DuplicateType", "dupe_mod"))
```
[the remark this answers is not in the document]
jsdw: We end up with two different types with the same path and name here, ie the first case.

I think scale-info uses `std::any::TypeId`, so if I tried adding the same type to the registry twice it would not add the second one, hence the need for two separate types with the same info :)

testing/integration-tests/src/codegen/codegen_tests.rs
jsdw: We end up with 2 variant types in the registry that both have the same path and name, and then when doing the codegen we spot that the paths are identical and add a "2" to one of the names to make sure that they don't override eachother. (ie if you remove the `ensure_unique_types` line and run this test again it'll fail)

codegen/src/api/mod.rs
jsdw: If the path is same and the type has generics, `scale-info` already handles that case ok.

For example, in the metadata you might see types like `Vec<u8>`, `Vec<bool>` and `Vec<Foo>` mentioned. When we generate the code, we see that they all have the same path but don't care, so we just end up with a single `Vec<Something>` in the generated output. (I didn't realise this at first and indeed didn't have this check to begin with, but then the codegen ended up with `Vec1<T>`, `Vec2<T>`, `Vec3<T>` and so on for loads of types).

The problem we ran into here was that there were two types mentioned in the metadata, `pallet::Call` and `pallet::Call`. They actually had some different types in their variants, but didn't have any generics that the code gen could use, so appeared the same. Now, we spot that case and change one to `pallet::Call2` so that both get generated in the output and one doesn't end up overriding the other.

Yeah I can probably add a test to show that generics end up being output as just one type :)

jsdw:
@niklasad1 I added another test to show that types with generics + same path are de-duplicated; hopefully with my other replies that addresses things :)

codegen/src/api/mod.rs
jsdw: I needed the `usize` bit because nothing else determines that number type, and so I figure we may as well have both for clarity :) (I did import Hashmap though to shorten this line)

niklasad1 (approved):
cool, easier to understand now :)

PR #974, opened 2023-05-22
Introduce Metadata type

jsdw:
To now, we've been internally using `frame_metadata::RuntimeMetadataV15` (or V14 or whatever in the past). We have various internal functionality that uses this, and then in Subxt we wrap it with some caching and other bits and pieces to expose things in a way Subxt needs.

This PR instead transforms the wire format `frame_metadata::RuntimeMetadataV14`/`frame_metadata::RuntimeMetadataV15` into a `Metadata` struct. Why?

1. To take control of the metadata API we expose; validation/retain stuff is all now provided on this `Metadata` struct, and we expose an interface that gives access to all aspects of the metadata but allows it to internally change over time.
2. Changes to `frame_metadata` going forwards (eg new metadata versions) do not force breaking changes in Subxt; we can expose functionality to decode metadata into our `Metadata` type and continue to operate and expose this same type as it is extended with new functionality etc.
3. Rely more on scale-info types; rather than wrapping the `RuntimeMetadata` and duplicating `PalletMetadata`, `ConstantMetadata` etc, we just provide direct access to the variants already in the type registry. Practically, this means that we expose everything that we know about more directly, and so don't need to replicate APIs to do the same.
4. As a result of the above, the `subxt_metadata` crate becomes much more of a complete crate on its own, rather than a collection of functions that operate on some version of frame_metadata.

For reviewing, I'd suggest that you focus most of your efforts on the `metadata` folder and `subxt_metadata` crate (and then see how it's made available also in `subxt/src/metadata/metadata_type.rs`. Mostly everything else is just fallout from these changes to shift everything to using this new metadata type. I'll add comments for other changes that might be worth more attention.


codegen/src/api/runtime_apis.rs
```diff
@@ -63,21 +60,21 @@ fn generate_runtime_api(
...
+            ))
+        };
 
         let method = quote!(
             #docs
             pub fn #method_name(&self, #( #params, )* ) -> #crate_path::runtime_api::Payload<types::#struct_name, #output> {
                 #crate_path::runtime_api::Payload::new_static(
-                    #runtime_fn_name,
+                    #trait_name_str,
+                    #method_name_str,
```
jsdw, line 77: I moved Runtime APIs to be split like calls etc, so that you provide separately the trait name and then method name. One reason for doing this was simply allow consistency in the metadata between how Runtime APIs and Pallet based things are accessed (ie you ask for trait name and then can ask for a specific method or just get docs for the trait ot whatever.

codegen/src/api/mod.rs
```diff
@@ -632,9 +621,9 @@ impl RuntimeGenerator {
 
                 /// check whether the Client you are using is aligned with the statically generated codegen.
                 pub fn validate_codegen<T: #crate_path::Config, C: #crate_path::client::OfflineClientT<T>>(client: &C) -> Result<(), #crate_path::error::MetadataError> {
-                    let runtime_metadata_hash = client.metadata().metadata_hash(&PALLETS);
+                    let runtime_metadata_hash = client.metadata().hasher().only_these_pallets(&PALLETS).hash();
```
jsdw, line 624: Something I bumped into here; we validate codegen based on only the generated pallets, but we don't take into account if runtime APIs have also been stripped. We should open an issue to ensure that this validation properly takes into account stripped runtime APIs too.

subxt/src/metadata/metadata_type.rs
jsdw, line 1: The metadata exposed from `subxt` is now a simple wrapper around `subxt_metadata::Metadata` so that it can be cloned cheaply.

testing/ui-tests/src/utils/pallet_metadata_test_runner.rs
```diff
@@ -106,18 +91,10 @@ impl PalletMetadataTestRunner {
...
-                }
-
-                let mut tmp_dir = std::env::temp_dir();
-                tmp_dir.push(format!("{TEST_DIR_PREFIX}{i}"));
-                std::fs::remove_dir_all(tmp_dir).expect("cannot cleanup temp files");
-            };
+        for idx in 0..self.index {
+            let mut tmp_dir = std::env::temp_dir();
+            tmp_dir.push(format!("{TEST_DIR_PREFIX}{idx}"));
+            let _ = std::fs::remove_dir_all(tmp_dir);
```
jsdw, line 97: Simplified because we only care about index and not whether we are iterating all or specific pallets.

codegen/src/api/runtime_apis.rs
jsdw: Issue opened for this: #976

subxt/src/client/online_client.rs
```diff
@@ -19,8 +19,7 @@ use crate::{
 };
 use derivative::Derivative;
 use futures::future;
-use parking_lot::RwLock;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
```
lexnv, line 22: Nice!

testing/integration-tests/src/codegen/codegen_tests.rs
```diff
@@ -11,7 +11,7 @@ use subxt_codegen::{CratePath, DerivesRegistry, RuntimeGenerator, TypeSubstitute
 
 fn generate_runtime_interface_from_metadata(metadata: RuntimeMetadataPrefixed) -> String {
     // Generate a runtime interface from the provided metadata.
-    let generator = RuntimeGenerator::new(metadata);
+    let generator = RuntimeGenerator::new(metadata.try_into().unwrap());
```
lexnv, line 14: nit: Maybe we can provide an expect message to align with the below `generator`?

lexnv (approved):
LGTM! The new metadata API looks solid! 👍

metadata/src/utils/ordered_map.rs
```diff
@@ -0,0 +1,137 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+use std::collections::HashMap;
+
+/// A minimal ordered map to let one search for
+/// things by key or get the values in insert order.
+#[derive(Debug, Clone)]
+pub struct OrderedMap<K, V> {
```
niklasad1, line 10: I'm not super happy about this because it's quite hard to understand such as `push_insert` which inserts one item at the end of the `Vec` and one item at `key` in the Map.

but fine :P

subxt/src/constants/constants_client.rs
```diff
@@ -39,13 +39,14 @@ impl<T: Config, Client: OfflineClientT<T>> ConstantsClient<T, Client> {
             let expected_hash = self
                 .client
                 .metadata()
-                .constant_hash(address.pallet_name(), address.constant_name())?;
+                .pallet_by_name(address.pallet_name())
```
niklasad1, line 42: dq: is this to ensure that the pallet name was stripped?

subxt/src/events/events_type.rs
```diff
@@ -292,19 +300,27 @@ impl EventDetails {
...
-    pub fn event_metadata(&self) -> &EventMetadata {
-        self.metadata
-            .event(self.pallet_index(), self.variant_index())
-            .expect("this must exist in order to have produced the EventDetails")
+    /// Fetch details from the metadata for this event.
+    pub fn event_metadata(&self) -> EventMetadataDetails {
+        let pallet = self
+            .metadata
+            .pallet_by_index(self.pallet_index())
+            .ok_or_else(|| MetadataError::PalletIndexNotFound(self.pallet_index()))
```
niklasad1, line 316: I don't get why the ok_or_else is needed because you expect the line after?!

```suggestion
```

metadata/src/utils/ordered_map.rs
jsdw: I'm happy to hear any suggestions here! My goal is just to preserve the order of items as seen in metadata, but also allow access by some other key (name, mostly). It just exists so that there's no need to manually keep a vec and hashmap in sync for doing this, but perhaps there's a better approach or maybe some api changes that would make it easier to understand? :)

subxt/src/events/events_type.rs
jsdw: Ah whoops, you're right, they aren't needed! I just copied the code from somewhere where they were needed and forgot to remove them :D

testing/integration-tests/src/codegen/codegen_tests.rs
jsdw: MAkes sense; I'll add one :)

subxt/src/constants/constants_client.rs
jsdw: I'm not sure I follow; perhaps you could elaborate?:)

niklasad1: I was referring to changed API:

previous version:

```rust
 client.metadata().constant_hash(address.pallet_name(), address.constant_name())?;
```

new code:

```rust
   client
        .metadata()
        .pallet_by_name(address.pallet_name())
        .ok_or_else(|| MetadataError::PalletNameNotFound(address.pallet_name().to_owned()))?
        .constant_hash(address.constant_name())
        .ok_or_else(|| MetadataError::ConstantNameNotFound(address.constant_name().to_owned()))?;
```

metadata/src/lib.rs
```diff
@@ -2,101 +2,619 @@
...
+            .and_then(|i| self.pallets.get_by_index(*i))?;
+
+        Some(PalletMetadata {
+            inner,
+            types: self.types(),
+        })
+    }
+
+    /// Access a pallet given its name.
+    pub fn pallet_by_name(&self, pallet_name: &str) -> Option<PalletMetadata<'_>> {
```
niklasad1, line 102: As my previous comment, it would nice to return Result<PalletMetadata<'_>>, PalletNameNotFound> here to avoid doing it when it's called.

niklasad1 (approved):
good stuff, I like the new metadata API

metadata/src/utils/ordered_map.rs
niklasad1: ok, I didn't look at the code where it's used and yes I see why you added it

subxt/src/constants/constants_client.rs
jsdw: Aah, I had a couple of reasons for breaking the API into two here:

1. By getting the pallet first, we then have the information needed to find the relevant constant/storage/etc rather than needing hashmaps with both pallet and constant/sotrage/whatever keys in.
2. Instead of needing to provide `Result`s back to say whether it was the pallet or constant that couldn't be found, the metadata can just return `Option`s in a lot of cases, and leave it up to eg Subxt how to present the error, which felt cleaner design wise.
3. There may be other reasons to want to access a pallet by its name or whatever, and in some cases (eg in this PR) it's useful to have both the pallet and the specific event/whatever details to hand back anyway :)

metadata/src/lib.rs
jsdw: I think `Option`'s are cleaner here, and then the thing using the metadata can decide how it should map those into errors (if it wishes), but yeah that bit we can make nicer :)

subxt/src/runtime_api/runtime_payload.rs
```diff
@@ -74,16 +79,42 @@ impl<ArgsData: EncodeAsFields, ReturnTy: DecodeWithMetadata> RuntimeApiPayload
...
     fn encode_args_to(&self, metadata: &Metadata, out: &mut Vec<u8>) -> Result<(), Error> {
-        let fn_metadata = metadata.runtime_fn(&self.fn_name)?;
+        let api_trait = metadata
+            .runtime_api_trait_by_name(&self.trait_name)
+            .ok_or_else(|| MetadataError::RuntimeTraitNotFound((*self.trait_name).to_owned()))?;
+        let api_method = api_trait
+            .method_by_name(&self.method_name)
+            .ok_or_else(|| MetadataError::RuntimeMethodNotFound((*self.method_name).to_owned()))?;
+
+        // TODO [jsdw]: This isn't good. The options are:
```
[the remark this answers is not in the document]
jsdw: Good reminder, thank you :)

PR #977, opened 2023-05-25
Cli tool tests

### LINKED ISSUE #946: CLI: Check the CLI produces the expected output
Ensure that the CLI is tested in terms of:
- expected flags
- non-compatible flag combination
- downloading of metadata for different versions
- code-generation for old / new / unstable versions

tadeohepperle:
partly fixes #946

I added a Writer (`&mut impl std::io::Write`) to each cli command such that we can write to it instead of `stdout`.
This makes it easy to test the expected output of each command. 

Added tests for some commands.


jsdw (approved):
Looks good; I'll always suggest preferring unit tests over e2e style ones whenever possible, but good to have the option to test the output too for when it's harder to do that (or there's something we want to ensure we are outputting) :)

cli/src/commands/explore/mod.rs
```diff
@@ -140,3 +145,63 @@ fn print_available_pallets(metadata: &Metadata) -> String {
...
+        args.append(&mut split);
+        let opts: Opts = clap::Parser::try_parse_from(args)?;
+        let mut output: Vec<u8> = Vec::new();
+        run(opts, &mut output)
+            .await
+            .map(|_| String::from_utf8(output).unwrap())
+    }
+
+    #[tokio::test]
+    async fn test_commands() {
```
niklasad1, line 168: I would prefer to split this test into smaller ones because it's much harder to debug if something breaks.

cli/src/utils/type_description.rs
```diff
@@ -271,3 +265,39 @@ fn format_type_description(input: &str) -> String {
...
+        let portable_registry: scale_info::PortableRegistry = types.into();
+        (id.id, portable_registry)
+    }
+
+    #[test]
+    fn test_type_description() {
+        let (foo_type_id, foo_registry) = make_type::<Foo>();
+        let description = print_type_description(&foo_type_id, &foo_registry).unwrap();
+        let mut output = String::new();
+        writeln!(output, "struct Foo {{").unwrap();
```
niklasad1, line 297: may this works as well:

```rust
let s = r#"struct Foo {
        hello: String,
        num: i32
}"#;
```

niklasad1 (approved):
LGTM, thanks for taking care of this :muscle:

cli/src/commands/explore/mod.rs
tadeohepperle: Sure, I will split it up in another PR today.

PR #993, opened 2023-06-01
move examples to main repo so we can include_str them for publishing

jsdw:
It turns out that `cargo publish` failed for Subxt because it could no longer find the `include_str`'s in the documentation, which linked to examples in the "external" examples crate.

The easiest fix for this is to move them back into the Subxt crate, so this is what I've done here.

I actually think this tidies things a little anyway, since then the external `examples` folder can be used solely for our larger examples.

Anyway, hopefully this passes CI.


niklasad1 (approved):

lexnv (approved):

jsdw:
I've tweaked the CI to not try building examples with different feature combinations, which will go wrong.

niklasad1:
The downside with this is that `cargo run --examples` won't work anymore?

jsdw:
`cargo run --example foo` will work as it did before :)

PR #1000, opened 2023-06-05
Fix codegen validation when Runtime APIs are stripped

### LINKED ISSUE #976: Fix codegen validation when Runtime APIs are stripped
The CLI tool can now strip out specific runtime APIs as well as specific pallets.

We have a method in the codegen called `validate_codegen` which compares some node metadata against the generated code. It takes into account only those pallets that we explicitly generated, but does not do the same for Runtime APIs (so when runtime APIs are stripped the metadata will never compare equal to node metadata). 

Let's fix that (preferably after #974 merges since it changes things in this area).

tadeohepperle:
fixes #976


codegen/src/api/mod.rs
```diff
@@ -321,11 +321,14 @@ impl RuntimeGenerator {
             .collect();
         let pallet_names_len = pallet_names.len();
 
-        let metadata_hash = self
+        let runtime_api_names: Vec<_> = self
             .metadata
-            .hasher()
-            .only_these_pallets(&pallet_names)
```
tadeohepperle, line 327: I removed the `only_these_pallets(&pallet_names)` here because I feld it is unnecessary. It includes all pallet_names from the metadata anyway.

jsdw: I think that makes sense now; we should make sure that the hashes actually do validate properly but I think this is all good :)

codegen/src/api/mod.rs
```diff
@@ -621,7 +624,12 @@ impl RuntimeGenerator {
 
                 /// check whether the Client you are using is aligned with the statically generated codegen.
                 pub fn validate_codegen<T: #crate_path::Config, C: #crate_path::client::OfflineClientT<T>>(client: &C) -> Result<(), #crate_path::error::MetadataError> {
-                    let runtime_metadata_hash = client.metadata().hasher().only_these_pallets(&PALLETS).hash();
+                    static RUNTIME_APIS: [&str; #runtime_api_names_len] = [ #(#runtime_api_names,)* ];
```
jsdw, line 627: For consistency I'd probably move this up next to the PALLETS static on line 544 (you could also move the PALLETS static down here; I wonder whether it's useful to publically expose these or not).

codegen/src/api/mod.rs
```diff
@@ -623,7 +629,11 @@ impl RuntimeGenerator {
 
                 /// check whether the Client you are using is aligned with the statically generated codegen.
                 pub fn validate_codegen<T: #crate_path::Config, C: #crate_path::client::OfflineClientT<T>>(client: &C) -> Result<(), #crate_path::error::MetadataError> {
-                    let runtime_metadata_hash = client.metadata().hasher().only_these_pallets(&PALLETS).hash();
+                    let runtime_metadata_hash = client
+                                                    .metadata()
+                                                    .hasher()
+                                                    .only_these_pallets(&PALLETS)
+                                                    .only_these_runtime_apis(&RUNTIME_APIS).hash();
```
jsdw, line 636: Is it just GitHub or are these lines super indented? :D

jsdw:
Looks perfect! Do we have any tests to check that the hashes line up when we strip pallets and runtime APIs now from metadata? Let's add one or two just to make sure that the hashes all line up :)

metadata/src/utils/validation.rs
```diff
@@ -424,30 +426,54 @@ impl<'a> MetadataHasher<'a> {
         self
     }
 
+    /// only hash the provided runtime APIs instead of hashing every runtime API
```
lexnv, line 429:
```suggestion
    /// Only hash the provided runtime APIs instead of hashing every runtime API.
```

lexnv (approved):

jsdw:
I had a pass over this to add a few tests for the validation. I also ended up:

- renaming `validate_codegen(..) -> Result<(), Error>` to `is_codegen_valid_for(metadata) -> bool` (having it return a bool made more sense to me, and then I wanted to tweak the name to match).
- getting rid of the `PalletMetadataTestRunner`, and upgrading `MetadataTestRunner` to do validation, too.

So probably somebody that isn't me should review it :)

lexnv (approved):

testing/ui-tests/src/utils/metadata_test_runner.rs
```diff
@@ -2,72 +2,179 @@
...
+
+    /// Set the test name.
+    pub fn name(mut self, name: impl AsRef<str>) -> Self {
+        self.name = name.as_ref().to_owned();
+        self
+    }
+
+    /// Set metadata to be validated against the generated code.
+    /// By default, we'll validate the same metadata used to generate the code.
+    pub fn validation_metadata(mut self, md: impl Into<Metadata>) -> Self {
```
niklasad1, line 79: I think `set_metadata` is better name for this method...

niklasad1 (approved):

testing/ui-tests/src/utils/metadata_test_runner.rs
jsdw: The `build()` call is where you pass the actual metadata that the macro will use; this `validation_metadata` is just the optional metadata you can use to test the `is_codegen_valid_for` function against :)

PR #1002, opened 2023-06-07
Get account nonce via state_call; remove RPC method

jsdw:
Some chains don't have the RPC method that we were using, and using state_call will be more future proof going forwards anyway.

Also make the function public because it's come up a couple of times recently where people have needed to fetch this.


niklasad1 (approved):
LGTM, I assume the trait bounds are changed because `state_call` just takes `Bytes`

jsdw:
> LGTM, I assume the trait bounds are changed because `state_call` just takes `Bytes`

Yeah exactly; instead of deserializing form a JSON response we're decoding from a runtime API response :)

lexnv (approved):

PR #1014, opened 2023-06-14
chore(subxt): make payload details public

clearloop:
the payload details are really useful for logging, for our situation, we have arch with dynamic API with subxt which manages the calls with the built `payload`, so making these two fields public again is the simpliest way we can get them for logging

https://github.com/gear-tech/gear/blob/640136c0b83f9fc16b978c1ec1aacfc7f1a9f4e4/gsdk/src/signer/calls.rs#L426


tadeohepperle (approved):
Looks good to me.

niklasad1 (approved):

lexnv (approved):

jsdw (approved):

PR #1015, opened 2023-06-14
Metadata difference command

tadeohepperle:
Explore the differences between two nodes

## Example
```
subxt diff ./artifacts/polkadot_metadata_small.scale ./artifacts/polkadot_metadata_tiny.scale
subxt diff ./artifacts/polkadot_metadata_small.scale wss://rpc.polkadot.io:443
 ```

Example output (colored in cli):
```
Pallets:
    ~ Balances
        Calls:
            - transfer_allow_death
            - force_set_balance
            + set_balance
            - upgrade_accounts
            - set_balance_deprecated
        Constants:
            - MaxHolds
            - MaxFreezes
        Storage Entries:
            - Freezes
            ~ Account (Changed: value type, default value)
            - Holds
    + Scheduler
    + Session
    + Democracy
    + Claims
    + Crowdloan
    + Registrar
    + FastUnstake
    + Initializer
    + Offences
    + Bounties
    + Treasury
    ~ System
        Storage Entries:
            ~ Events (Changed: value type)
            ~ Account (Changed: value type, default value)
    + ChildBounties
    + Babe
    + ElectionProviderMultiPhase
    + Configuration
    + ParaInherent
    + VoterList
    + Hrmp
    + Ump
    + ParasShared
    ~ Multisig
        Calls:
            ~ as_multi
            ~ as_multi_threshold_1
    + PhragmenElection
    + Utility
    + Proxy
    + ParachainsOrigin
    + Paras
    + NominationPools
    + Auctions
    + Timestamp
    + Indices
    + Grandpa
    + AuthorityDiscovery
    + Slots
    + TransactionPayment
    + Historical
    + TechnicalMembership
    + ImOnline
    + ParaSessionInfo
    + Authorship
    + Dmp
    + ParasDisputes
    + Preimage
    + ParaScheduler
    + TechnicalCommittee
    + Vesting
    + Identity
    + XcmPallet
    + Council
    + Tips
    + ParaInclusion
Runtime APIs:
    - TransactionPaymentCallApi
    - Core
    - MmrApi
    - BeefyApi
    - GrandpaApi
    - AccountNonceApi
    - BlockBuilder
    - TaggedTransactionQueue
    - StakingApi
    - AuthorityDiscoveryApi
    - SessionKeys
    - BabeApi
    - Metadata
    - NominationPoolsApi
    - ParachainHost
    - OffchainWorkerApi
    - TransactionPaymentApi
```

### DIFF
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -4,8 +4,6 @@ use codec::Decode;
 use frame_metadata::RuntimeMetadataPrefixed;
 use std::collections::HashMap;
 
-
-
 use crate::utils::FileOrUrl;
 use color_eyre::owo_colors::OwoColorize;
 
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -305,7 +305,6 @@ impl TryFrom<(&str, Option<MetadataVersion>)> for FileOrUrl {
     type Error = ();
 
     fn try_from(value: (&str, Option<MetadataVersion>)) -> Result<Self, Self::Error> {
-        println!("value {}", &value.0);
         let path = std::path::Path::new(&value.0);
         if path.exists() {
             Ok(FileOrUrl {
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -1,17 +1,17 @@
 use clap::Parser as ClapParser;
 use codec::Decode;
-use color_eyre::eyre::eyre;
+
 use frame_metadata::RuntimeMetadataPrefixed;
 use std::collections::HashMap;
-use std::path::PathBuf;
-use std::str::FromStr;
+
+
 
 use crate::utils::FileOrUrl;
 use color_eyre::owo_colors::OwoColorize;
-use jsonrpsee::client_transport::ws::Uri;
+
 use scale_info::form::PortableForm;
 use scale_info::Variant;
-use subxt_codegen::utils::MetadataVersion;
+
 use subxt_metadata::{
     ConstantMetadata, Metadata, PalletMetadata, RuntimeApiMetadata, StorageEntryMetadata,
     StorageEntryType,
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -7,9 +7,13 @@ use color_eyre::eyre::eyre;
 use frame_metadata::{RuntimeMetadata, RuntimeMetadataPrefixed};
 use frame_metadata::v15::RuntimeMetadataV15;
 use jsonrpsee::client_transport::ws::Uri;
+use scale_info::form::PortableForm;
+use scale_info::Variant;
 use subxt_codegen::utils::{MetadataVersion};
-use subxt_metadata::{ConstantMetadata, Metadata, PalletMetadata, RuntimeApiMetadata};
-use crate::utils::FileOrUrl;
+use subxt_metadata::{ConstantMetadata, Metadata, PalletMetadata, RuntimeApiMetadata, StorageEntryMetadata};
+use crate::utils::{FileOrUrl};
+use std::io::Write;
+use color_eyre::owo_colors::OwoColorize;
 
 /// todo: add docs
 #[derive(Debug, ClapParser)]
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: cli/src/commands/diff.rs [hunk 1]
```diff
@@ -0,0 +1,526 @@
...
+/// Explore the differences between two nodes
+///
+/// # Example
+/// ```
+/// subxt diff ./artifacts/polkadot_metadata_small.scale ./artifacts/polkadot_metadata_tiny.scale
+/// subxt diff ./artifacts/polkadot_metadata_small.scale wss://rpc.polkadot.io:443
+/// ```
+#[derive(Debug, ClapParser)]
+pub struct Opts {
+    node1: String,
```
niklasad1, line 29: I would prefer to change this to `entry1: FileOrUrl` and `entry2: FileOrUrl` then clap will just use the `TryFrom` or `FromStr` impl and you don't have to deal with that yourself.

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,526 @@
...
+                pallets.remove(pallet_metadata.name());
+                continue;
+            }
+        }
+        *e2 = Some(pallet_metadata);
+    }
+
+    pallets
+        .into_values()
+        .map(|tuple| Diff::try_from(tuple).unwrap())
```
niklasad1, line 504: `Err()` should be unreachable based on the code above but would be good with an `expect` and a small explanation.

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,526 @@
...
+/// ```
+/// subxt diff ./artifacts/polkadot_metadata_small.scale ./artifacts/polkadot_metadata_tiny.scale
+/// subxt diff ./artifacts/polkadot_metadata_small.scale wss://rpc.polkadot.io:443
+/// ```
+#[derive(Debug, ClapParser)]
+pub struct Opts {
+    node1: String,
+    node2: String,
+    #[clap(long, short)]
+    version: Option<MetadataVersion>,
```
niklasad1, line 32: I don't get why `version` is a CLI parameter, how would one insert `--version` here and why? 

Better to read that from the metadata itself?

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,526 @@
...
+    #[clap(long, short)]
+    pallet: Option<String>,
```
niklasad1, line 34: This isn't used AFAIU, what's intention behind it? Just compare a single pallet in the metadata?

If we want this, it would be better with Vec<String> to compare a one or more pallets

cli/src/commands/diff.rs
jsdw: The version is which version of metadata is being requested when a url is given (eg are we downloading V14 metadatas or unstable-v15 metadatas or "whatever is latest stable"; something like that)

xlc:
doesn't subwasm already have something like this?

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,496 @@
...
+            if e1_inner.hash() == runtime_api.hash() {
+                runtime_apis.remove(runtime_api.name());
+                continue;
+            }
+        }
+        *e2 = Some(runtime_api);
+    }
+    runtime_apis
+        .into_values()
+        .map(|tuple| Diff::try_from(tuple).unwrap())
```
niklasad1, line 442:
```suggestion
        .map(|tuple| Diff::try_from(tuple).expect("At least one value is inserted when the key exist; qed"))
```

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,496 @@
...
+            if e1_hash == e2_hash {
+                constants.remove(constant.name());
+                continue;
+            }
+        }
+        *e2 = Some(constant);
+    }
+    constants
+        .into_values()
+        .map(|tuple| Diff::try_from(tuple).unwrap())
```
niklasad1, line 408:
```suggestion
        .map(|tuple| Diff::try_from(tuple).expect("At least one value is inserted when the key exist; qed")))
```

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,496 @@
...
+                    calls.remove(call_variant.name.as_str());
+                    continue;
+                }
+            }
+            *e2 = Some(call_variant);
+        }
+    }
+    calls
+        .into_values()
+        .map(|tuple| Diff::try_from(tuple).unwrap())
```
niklasad1, line 375:
```suggestion
        .map(|tuple| .map(|tuple| Diff::try_from(tuple).expect("At least one value is inserted when the key exist; qed"))))
```

niklasad1 (approved):

niklasad1:
> doesn't subwasm already have something like this?

Yes, subwasm has something similar but it compares `wasm files` and this/subxt operates on metadata either by passing an URL or a path to the metadata.

Both are useful but if you are already a user of subxt the `diff` command is more ergonomic to use, most likely one has already embedded a static metadata file in the binary.

jsdw:
Some small feedback bits re the output:

-  wonder whether `<ENTRY1>` and `<ENTRY2>` could be followed by help like "metadata file or node URL" in the following?:

```text
error: the following required arguments were not provided:
  <ENTRY1>
  <ENTRY2>

Usage: subxt diff <ENTRY1> <ENTRY2>

For more information, try '--help'.
```

- It appears (and please correct me if I'm wrong) that `subxt diff A B` will show changes relative to "A", ie which pallets were added or removed in A. I'd normally expect a `diff` commant to show changes with respect to "B". Could this be swapped around?

Otherwise, the output looks fantastic and I love the use of colours :)

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,496 @@
...
+                    .call_hash(&call_variant.name)
+                    .expect("call should be present");
+                if e1_hash == e2_hash {
+                    calls.remove(call_variant.name.as_str());
+                    continue;
+                }
+            }
+            *e2 = Some(call_variant);
+        }
+    }
```
jsdw, line 372: The general pattern of "give me a list of the differences between two things" feels like something that could be abstracted here since it's repeated several times; the only logic that changes each time is "how do I turn a thing into something comparable". 

I'd propose it'd be a fun challenge to have a go at doing this abstracting; I could imagine ultimately ending with a signature something like this:

```rust
fn diff<T, C: PartialEq>(
    items_a: impl IntoIterator<Item = T>,
    comparable_a: impl Fn(&T) -> C
    items_b: impl IntoIterator<Item = T>,
    comparable_b: impl Fn(&T) -> C
) -> Vec<Diff<T>>
```

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,496 @@
...
+///
+/// # Example
+/// ```
+/// subxt diff ./artifacts/polkadot_metadata_small.scale ./artifacts/polkadot_metadata_tiny.scale
+/// subxt diff ./artifacts/polkadot_metadata_small.scale wss://rpc.polkadot.io:443
+/// ```
+#[derive(Debug, ClapParser)]
+pub struct Opts {
+    entry1: FileOrUrl,
+    entry2: FileOrUrl,
```
jsdw, line 28: Perhaps a slightly more descriptive name for these like `metadata_or_url1` and `metadata_or_url2` would make the output a little clearer.

I wonder whether adding comments above these would lead to nicer help being shown in the output too :)

jsdw:
Looks great; just a few suggestions :)

tadeohepperle:
> it appears (and please correct me if I'm wrong) that `subxt diff A B` will show changes relative to "A", ie which pallets were added or removed to get to A.

@jsdw I think maybe you understood it wrong. It shows things with a `+` that are present in B but not in A, so things that were added to get from A to B. And vice versa. For example consider this:
`subxt diff ./artifacts/polkadot_metadata_small.scale ./artifacts/polkadot_metadata_tiny.scale`
gives as output 
```
Pallets:
    - Staking
    - Balances
    - Multisig
    - System
Runtime APIs:
    ~ TransactionPaymentCallApi
```
indicating that some pallets are removed going from `small` to `tiny`.
Or maybe I am understanding it wrong?

jsdw:
Oh interesting; hopefully I just got it wrong then, lemme check! I compared kuama to polkadot iirc and it seemed as if it was the otherway round but I'll try again to confirm :)

metadata/src/lib.rs
```diff
@@ -303,8 +313,8 @@ impl StorageMetadata {
     }
 
     /// An iterator over the storage entries.
-    pub fn entries(&self) -> impl ExactSizeIterator<Item = &StorageEntryMetadata> {
```
jsdw, line 306: Not really a big deal, but I was wodnering whether there was a reason for this change?

IIRC the reason I tended to iterate over other metadata structs exposed here rather than give back slices was just to give a little more wiggle room in case we wanted to change the internals a bit in the future. There are a couple of other places in this API that return the ExactSizeIterator even though they could return slices too for this reason :)

cli/src/commands/diff.rs
```diff
@@ -0,0 +1,437 @@
...
+}
+
+#[derive(Debug, Clone)]
+enum Diff<T> {
+    Added(T),
+    Changed { from: T, to: T },
+    Removed(T),
+}
+
+fn diff<T, C: PartialEq, I: Hash + PartialEq + Eq>(
```
jsdw, line 399: Awesome! Much prefer this :)

jsdw: (and with this logic now separated, you could also have a quick unit test or two that diffing things using this function works as expected :))

tadeohepperle: I added a unit test. This also made me realize that the output order was random because of the random nature of hash maps. So I added alphabetical ordering as well.

jsdw (approved):

jsdw:
Awesome stuff; good job @tadeohepperle!

lexnv (approved):
Lovely PR! Amazing job here 🙏

PR #1016, opened 2023-06-14
Add `subxt_signer` crate for native & WASM compatible signing

jsdw:
The `subxt-signer` crate replaces the need to use `sp_core::sr25519::Pair` for sr25519 signing of transactions (which is something we allow via the `substrate-compat` feature flag.

Reasons to use `subxt_signer` over `sp_core::sr25519::Pair`
- It can be compiled to WASM, so can be used to sign things in Subxt WASM programs.
- Fewer dependencies; we only bring in what we need for singing and not a bunch of other `sp_core` deps.
- A relatively small and simple API for creating and using keypairs.

This crate can be used independently of `subxt` by disabling the `subxt` feature flag, in case you'd like to use it to sign things independently of Subxt.

I also rejigged the feature flags a bit with this in mind. Subxt now has these main feature flags:

- "native" and "web": pick exactly one of these to write native or web compatible subxt programs. (if you're writing a lib, best to re-expose these to allow upstream users to pick one).
- "jsonrpsee": use to pull in `jsonrpsee`. Works with either of  "native" or "web"
- "substrate-compat": use to pull in key sp_* crates, sign things via them, and add compat impls to make using them a little nicer. Doesn't work with "web".


altaua (approved):

subxt/src/book/usage/transactions.rs
```diff
@@ -60,28 +60,54 @@
...
+//! for chains which require sr25519 signatures (requires the `subxt` feature to be enabled).
+//! 2. Alternately, Subxt can use instances of Substrate's [`sp_core::Pair`] to sign things by wrapping
+//! them in a [`crate::tx::PairSigner`] (requires the `substrate-compat` feature to be enabled).
+//!
+//! Going for 1 leads to fewer dependencies being imported and WASM compatibility out of the box via
+//! the `web` feature flag. Going for 2 is useful if you're already using the Substrate dependencies or
+//! need additional signing algorithms that `subxt_signer` doesn't support, and don't care about WASM
+//! compatibility.
+//!
+//! Let's see how to use each of these approaches:
```
tadeohepperle, line 77: Good explanation, avoids confusion for users!

subxt/src/lib.rs
```diff
@@ -33,23 +33,31 @@
 )]
 #![allow(clippy::type_complexity)]
 
+#[cfg(any(
+    all(feature = "web", feature = "native"),
+    not(any(feature = "web", feature = "native"))
+))]
+compile_error!("subxt: exactly one of the 'web' and 'native' features should be used.");
```
tadeohepperle, line 40: This is a smart way to handle mutually exclusive features! Haven't seen that yet.

tadeohepperle (approved):
Looks good to me, good work! :)

signer/src/sr25519.rs
```diff
@@ -0,0 +1,403 @@
...
+    Hex(#[from] hex::FromHexError),
+}
+
+/// Dev accounts, helpful for testing but not to be used in production,
+/// since the secret keys are known.
+pub mod dev {
+    use super::*;
+    use std::str::FromStr;
+
+    once_static_cloned! {
```
lexnv, line 209: Nice!

signer/src/crypto/derive_junction.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// Create a new hard (hardened) DeriveJunction from a given, encodable, value.
+    ///
+    /// If you need a soft junction, use `soft()`.
+    pub fn hard<T: Encode>(index: T) -> Self {
+        Self::soft(index).harden()
+    }
+
+    /// Consume self to return the chain code.
+    pub fn unwrap_inner(self) -> [u8; JUNCTION_ID_LEN] {
```
niklasad1, line 56: I would prefer `into_inner` instead of `unwrap_inner` because I regard `unwrap` as something fallible but maybe just me :P

lexnv (approved):
LGTM! Amazing PR 👍

subxt/src/lib.rs
```diff
@@ -33,23 +33,31 @@
 )]
 #![allow(clippy::type_complexity)]
 
+#[cfg(any(
+    all(feature = "web", feature = "native"),
```
niklasad1, line 37: > subxt: exactly one of the 'web' and 'native' features should be used.

should this be `not(all(feature = "web", feature = "native"))` then?

signer/src/crypto/secret_uri.rs
```diff
@@ -0,0 +1,153 @@
...
+/// ```
+/// # use subxt_signer::{SecretUri, DeriveJunction, DEV_PHRASE, ExposeSecret};
+/// # use std::str::FromStr;
+/// let suri = SecretUri::from_str("0xe5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a//Alice").expect("Parse SURI");
+///
+/// assert_eq!(vec![DeriveJunction::from("Alice").harden()], suri.junctions);
+/// assert_eq!("0xe5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a", suri.phrase.expose_secret());
+/// assert!(suri.password.is_none());
+/// ```
+pub struct SecretUri {
```
niklasad1, line 77: you could maybe state that this code is extracted from substrate or something

I'm a little scared of it to be honest but looks identical to `SecretUri` in substrate...

subxt/src/lib.rs
niklasad1: the smart way is not to have mutually exclusive features :)

subxt/Cargo.toml
```diff
@@ -15,31 +15,49 @@ description = "Submit extrinsics (transactions) to a substrate node via RPC"
...
+
+# Enable this to use jsonrpsee (allowing for example `OnlineClient::from_url`).
+jsonrpsee = [
+    "dep:jsonrpsee"
+]
+
+# Enable this to pull in extra Substrate dependencies which make it possible to
+# use the `sp_core::crypto::Pair` Signer implementation, as well as adding some
+# `From` impls for types like `AccountId32`. Cannot be used with "web".
 substrate-compat = [
```
niklasad1, line 45: so, `features = ["jsonrpsee", "native"]` should make it work but not bring in sp-core and all these deps, correct? 

This will be great

subxt/src/lib.rs
niklasad1: Doesn't really have to be addressed in this PR but would be great if we could have APIs similar [to](https://github.com/paritytech/jsonrpsee/blob/master/client/ws-client/src/lib.rs#L127-#L131) to avoid having to deal with "mutually exclusive features`

but ofc it won't work for if something doesn't work for WASM.

niklasad1 (commented):
very excited about this one, but I have a few questions :)

subxt/src/lib.rs
jsdw: > the smart way is not to have mutually exclusive features :)

Yeah, ideally the features would be purely additive in general!

Probably I need to keep thinking about feature flags here, and also which flags are enabled by default (it's annoying re doc examples etc if you don't have feature flags enabled that you want to use but might be surmountable).

When I did this I figured that a non additive "native or web" feature and then everything else being additive (in terms of APIs exposed at least) felt like the nicest way to handle it (otherwise you end up with the non-additiveness being pushed to other features anyway like jsonrpsee-web and jsonrpsee-native, or lightclient-web and lightclient-native not playing well together).

Def open to more thoughts on all of this though!

signer/src/crypto/derive_junction.rs
jsdw: Agreed! This is basically copied verbatim from sp_core so I hadn't really paid it much attention :)

subxt/Cargo.toml
jsdw: Yup exactly :)

The only reason I left `substrate_compat` in the default deps iirc was so that doc examples (well, one doc example iirc, and maybe a couple of tests or something) would work. Probably need one more go-over on these features to sort out niggles like that because would be nicer not to have it by default at all!

signer/src/crypto/secret_uri.rs
jsdw: Yeah; in the `mod.rs` of the crypto folder I say that everything in the crypto folder comes from sp_core::crypto :) (though with tiny tweaks or doc comment updates).

I didn't like bringing it in particularly either but it is just a straight copy of the logic! Probably I could test this in isolation better too to prove it aligns.

subxt/src/lib.rs
jsdw: Hmmm, I think:

`not(all(feature = "web", feature = "native"))` specifically matches when there are not exactly both features enabled. Ie if 1 feature is enabled, it would be true.

`not(any(feature = "web", feature = "native"))` matches when there are not any of the features enabled, ie none of them are activated.

niklasad1: I mean that we would need something similar to:

```rust
#[cfg(all(
    // ensure that only of one `web` and `native` is used
    not(all(feature = "web", feature = "native")),
    // ensure that at least one of the feature `web` or `native` is enabled.
    any(feature = "web", feature = "native")
))]
```

PR #1018, opened 2023-06-16
cli: Write output to file

### LINKED ISSUE #969: Documentation for subxt-cli and windows.
I had followed the instructions for using subxt-cli to export the metadata but when I imported the metadata into my project it always showed errors saying it could not understand it. I tried this even with the substrate-node-template with the same results. I eventually tried running subxt-cli in WSL and that seems to have fixed it. So it seems there is some incompatibility with windows. At least temporarily it might be wise to add this to the documentation.

lexnv:
This PR adds the ability to output the result of the subxt/cli metadata command to a dedicated file.

This is in response to https://github.com/paritytech/subxt/issues/969, where the output from the Windows process to the file results in metadata byte differences.

With this patch applied:

```bash
$> Get-FileHash .\win.scale

Algorithm       Hash                                                                   Path
---------       ----                                                                   ----
SHA256          637D22238919A0293847BB0DA3F4E3E3FCF0711AF07DE7BF3E8527437962E8BE       D:\subxt\cli\win.scale


$> shasum -a 256 mac.scale                                                     
637d22238919a0293847bb0da3f4e3e3fcf0711af07de7bf3e8527437962e8be  mac.scale
```

Closes https://github.com/paritytech/subxt/issues/969

### DIFF
--- a/cli/src/commands/metadata.rs
+++ b/cli/src/commands/metadata.rs
@@ -7,7 +7,7 @@ use clap::Parser as ClapParser;
 use codec::{Decode, Encode};
 use color_eyre::eyre::{self, bail};
 use frame_metadata::{v15::RuntimeMetadataV15, RuntimeMetadata, RuntimeMetadataPrefixed};
-use std::io::Write;
+use std::{io::Write, path::PathBuf};
 use subxt_metadata::Metadata;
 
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

jsdw (approved):
Lovely!

PR #1024, opened 2023-06-21
chain_getBlock extrinsics encoding

### LINKED ISSUE #1010: chain_getBlock extrinsics encoding
When I query the chain_getBlock extrinsics, subxt deserialize the extrinsics bytes with scale

https://github.com/paritytech/subxt/blob/1ad9a20ce2a156fae7c3b3d0fb4d408c5d07c374/subxt/src/rpc/types.rs#L113-L126

As a result if I compare the result of substrate export-blocks command or if I make a raw query to the rpc, I get different values for these extrinsics. If the subxt client is just supposed to return the result of the API call it should not try to interpret the payload.

Instead we should probably use something like this 
```rust
pub struct ChainBlockExtrinsic(#[serde(with = "impl_serde::serialize")] pub Vec<u8>);
```

tadeohepperle:
fixes #1010.

The scale encoded extrinsic bytes are now only decoded when the `ExtrinsicDetails` are constructed from the raw bytes (so a bit more down the line and not directly in the RPC call).

### DIFF
--- a/subxt/src/blocks/extrinsic_types.rs
+++ b/subxt/src/blocks/extrinsic_types.rs
@@ -17,6 +17,7 @@ use codec::Decode;
 use derivative::Derivative;
 use scale_decode::DecodeAsFields;
 use std::{collections::HashMap, sync::Arc};
+use crate::utils::strip_compact_prefix;
 
 /// Trait to uniquely identify the extrinsic's identity from the runtime metadata.
 ///
@@ -216,13 +217,13 @@ where
         let metadata = client.metadata();
 
         // removing the compact encoded prefix:
-        let extrinsic_bytes: Vec<u8> = Decode::decode(&mut &extrinsic_bytes[..])?;
+        let bytes: Arc<[u8]> =  strip_compact_prefix(extrinsic_bytes)?.1.into();
 
         // Extrinsic are encoded in memory in the following way:
         //   - first byte: abbbbbbb (a = 0 for unsigned, 1 for signed, b = version)
         //   - signature: [unknown TBD with metadata].
         //   - extrinsic data
-        let first_byte: u8 = Decode::decode(&mut &extrinsic_bytes[..])?;
+        let first_byte: u8 = Decode::decode(&mut &bytes[..])?;
 
         let version = first_byte & VERSION_MASK;
         if version != LATEST_EXTRINSIC_VERSION {
@@ -232,13 +233,13 @@ where
         let is_signed = first_byte & SIGNATURE_MASK != 0;
 
         // Skip over the first byte which denotes the version and signing.
-        let cursor = &mut &extrinsic_bytes[1..];
+        let cursor = &mut &bytes[1..];
 
         let mut address_start_idx = 0;
         let mut address_end_idx = 0;
 
         if is_signed {
-            address_start_idx = extrinsic_bytes.len() - cursor.len();
+            address_start_idx = bytes.len() - cursor.len();
 
             // Skip over the address, signature and extra fields.
             scale_decode::visitor::decode_with_visitor(
@@ -248,7 +249,7 @@ where
                 scale_decode::visitor::IgnoreVisitor,
             )
             .map_err(scale_decode::Error::from)?;
-            address_end_idx = extrinsic_bytes.len() - cursor.len();
+            address_end_idx = bytes.len() - cursor.len();
 
             scale_decode::visitor::decode_with_visitor(
                 cursor,
@@ -267,17 +268,17 @@ where
             .map_err(scale_decode::Error::from)?;
         }
 
-        let call_start_idx = extrinsic_bytes.len() - cursor.len();
+        let call_start_idx = bytes.len() - cursor.len();
 
         // Decode the pallet index, then the call variant.
-        let cursor = &mut &extrinsic_bytes[call_start_idx..];
+        let cursor = &mut &bytes[call_start_idx..];
 
         let pallet_index: u8 = Decode::decode(cursor)?;
         let variant_index: u8 = Decode::decode(cursor)?;
 
         Ok(ExtrinsicDetails {
             index,
-            bytes: extrinsic_bytes.into(),
+            bytes,
             is_signed,
             address_start_idx,
             address_end_idx,
--- a/subxt/src/blocks/extrinsic_types.rs
+++ b/subxt/src/blocks/extrinsic_types.rs
@@ -786,14 +786,8 @@ mod tests {
         let ids = ExtrinsicPartTypeIds::new(&metadata).unwrap();
 
         // Decode with empty bytes.
-        let result = ExtrinsicDetails::decode_from(
-            1,
-            &vec![],
-            client,
-            H256::random(),
-            Default::default(),
-            ids,
-        );
+        let result =
+            ExtrinsicDetails::decode_from(1, &[], client, H256::random(), Default::default(), ids);
         assert_matches!(result.err(), Some(crate::Error::Codec(_)));
     }
 
@@ -806,7 +800,7 @@ mod tests {
         // Decode with invalid version.
         let result = ExtrinsicDetails::decode_from(
             1,
-            &3u8.encode(),
+            &vec![3u8].encode(),
             client,
             H256::random(),
             Default::default(),
@@ -845,7 +839,7 @@ mod tests {
         // The length is handled deserializing `ChainBlockExtrinsic`, therefore the first byte is not needed.
         let extrinsic = ExtrinsicDetails::decode_from(
             1,
-            tx_encoded.encoded()[1..].into(),
+            tx_encoded.encoded(),
             client,
             H256::random(),
             Default::default(),
--- a/subxt/src/blocks/extrinsic_types.rs
+++ b/subxt/src/blocks/extrinsic_types.rs
@@ -119,7 +119,7 @@ where
             } else {
                 match ExtrinsicDetails::decode_from(
                     index as u32,
-                    extrinsics[index].0.clone().into(),
+                    &extrinsics[index].0,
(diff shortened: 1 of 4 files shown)


subxt/src/blocks/extrinsic_types.rs
```diff
@@ -264,17 +266,19 @@ where
...
 
         // Decode the pallet index, then the call variant.
-        let cursor = &mut &extrinsic_bytes[call_start_idx..];
+        let cursor = &mut &decoded_extrinsic_bytes[call_start_idx..];
 
         let pallet_index: u8 = Decode::decode(cursor)?;
         let variant_index: u8 = Decode::decode(cursor)?;
 
+        let decoded_extrinsic_bytes: Arc<[u8]> =
+            Arc::from(decoded_extrinsic_bytes.into_boxed_slice());
```
niklasad1, line 278: I don't why this boxing is needed here, isn't better to use the `From<&[T]> for Arc<[T]> impl` here?

But I guess the underlying implementation still has to copy to the bytes but should be better anyway is my hunch :P

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -841,7 +846,7 @@ mod tests {
         // The length is handled deserializing `ChainBlockExtrinsic`, therefore the first byte is not needed.
```
niklasad1, line 846: remove this comment?

subxt/src/tx/tx_client.rs
```diff
@@ -445,8 +445,7 @@ where
     /// and obtain details about it, once it has made it into a block.
     pub async fn submit_and_watch(&self) -> Result<TxProgress<T, C>, Error> {
         // Get a hash of the extrinsic (we'll need this later).
-        let ext_hash = T::Hasher::hash_of(&self.encoded);
-
+        let ext_hash = T::Hasher::hash_of(&self.encoded.encode());
```
niklasad1, line 448: can you explain why this is encoded twice?

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -215,11 +215,13 @@ where
 
         let metadata = client.metadata();
 
+        let decoded_extrinsic_bytes: Vec<u8> = Decode::decode(&mut &encoded_extrinsic_bytes[..])?;
```
niklasad1, line 218: doesn't this mess up the indexing after decoding it here?

niklasad1 (commented):
I'm afraid I don't know this code good enough but I can't really follow this fix.

I guess because you decode the bytes and store them in `ExtrinsicDetails` might make 
other code/names inconsistent and I can't follow it properly...

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: I think doing it the way I did it is quite efficient. It does not involve any heap allocations / memcpy of bytes on the heap, because only the pointers on the stack are given control from Vec -> Box -> Arc and that is it. I think the From<&[T]> will create a copy of the bytes on the heap. What do you think?

subxt/src/tx/tx_client.rs
tadeohepperle: I'll add a comment. I think it is like this: The `self.encoded` refers to the raw bytes of the extrinsic. Calling the `.encode()` on these bytes basically just prefixes it with the compact encoded length, IIRC.

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: This is basically just removing the prefix of the bytes. The idea in this PR is that we keep the prefix for as long as possible in the `ChainBlockExtrinsic` and only remove it (e.g. calling `Decode::decode` when we actually need to get details out of the extrinsic bytes).

subxt/src/tx/tx_client.rs
niklasad1: I would rather add two separate fields `raw_bytes` and `scale_bytes` in both `SubmittableExtrinsic` and `ExtrinsicDetails` to avoid this `encode()`

These bytes are already allocated and can be reused later and it makes the code easier to understand.

Regarding https://github.com/paritytech/subxt/blob/master/subxt/src/rpc/rpc.rs#L401, that is encoding it to `sp_core::Bytes` so it may work just to pass the `raw_bytes` into there...

tadeohepperle: I also thought about that, but it will not save us allocations, because inside of the `Rpc::watch_extrinsic()` function, we need to construct an `RpcParams` struct that needs to take ownership of the scale encoded bytes. So we need to clone them, because we cannot move them out of the SubmittableExtrinsic. What do you think about having the  `Rpc::watch_extrinsic()` function return the hash of the bytes of the `RpcParams`? This would avoid any unnecessary allocations. I think I need to talk with James about this issue on Monday, I also don't know if the fix I did is desired in this way.

niklasad1: I don't care that much of the allocations to be honest but it would make the code much more readable.

The `rpc_params!` doesn't need to take ownership of the bytes FWIW, it just serializes them. 
So you could do `rpc_params![&data]`

subxt/src/tx/tx_client.rs
```diff
@@ -417,12 +420,13 @@ where
     pub fn from_bytes(client: C, tx_bytes: Vec<u8>) -> Self {
         Self {
             client,
+            encoded_prefixed: tx_bytes.encode(),
```
jsdw, line 423: The bytes provided from eg line 394 are already prefixed with length, so we shouldn't need to do this again I think?

subxt/src/tx/tx_client.rs
```diff
@@ -444,12 +448,12 @@ where
     /// Returns a [`TxProgress`], which can be used to track the status of the transaction
     /// and obtain details about it, once it has made it into a block.
     pub async fn submit_and_watch(&self) -> Result<TxProgress<T, C>, Error> {
-        // Get a hash of the extrinsic (we'll need this later).
-        let ext_hash = T::Hasher::hash_of(&self.encoded);
-
+        // Get a hash of the extrinsic (we'll need this later)
+        // We use the `encoded_prefixed` to calculate the hash, because it has the same hash
```
jsdw, line 452: Here, I think the hash should be as it was before; ie the hash of the exact bytes we're sending to the `watch_extrinsic` rpc call and not a hash of those bytes with another length prefixed.

To check, the hash returned here should be identical to the hash returned from making the `submit` call below with the same extrinsic; worth trying that to make sure one way or the other :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -215,6 +215,9 @@ where
 
         let metadata = client.metadata();
 
+        // removing the compact encoded prefix:
+        let extrinsic_bytes: Vec<u8> = Decode::decode(&mut &extrinsic_bytes[..])?;
```
jsdw, line 219: Here and in the `.position(..)` thing below, we could avoid the allocating by instead decoding the `Compact<u64>` off the front of the bytes and then seeing how many bytes were consumed in doing this.
(I say "avoid allocating" but this was copying the bytes into an `Arc` before, so either way you may allocate, but at least have the chocie of exactly where)

This "compact length" value (or in other words an "extrinsic_start_index" could be saved if useful in `ExtrinsicDetails` so that it's easy to get the rest of the bytes from it when needed.

I'd have to look at the code in more detail to see what the `Arc` was used for before and whether it's then good to keep it, or whether perhaps the `Vec<ChainBlockExtrinsic>` in `Extrinsics` could be instead something like `Arc<[ChainBlockExtrinsic]>` so that we don't allocate at all when we `iter()` over the extrinsics :)

tadeohepperle: We could also give the `ExtrinsicDetails` an explicit lifetime that is bound by the lifetime of `Extrinsics`. This way we can degrade the bytes field of `ExtrinsicDetails` to `bytes: &[u8]` instead of `bytes: Arc<[u8]>` where the slice points to the `i..` bytes of the Vec<u8> of the respective ChainBlockExtrinsic, where i is the compact encoded length. This also avoids allocation. I am not sure though if the Arc was necessary somewhere.

subxt/src/tx/tx_progress.rs
```diff
@@ -389,7 +389,10 @@ impl<T: Config, C: OnlineClientT<T>> TxInBlock<T, C> {
             .iter()
             .position(|ext| {
                 use crate::config::Hasher;
-                let hash = T::Hasher::hash_of(&ext.0);
+                let Ok((_,stripped)) = strip_compact_prefix(&ext.0)else {
```
jsdw, line 392:
```suggestion
                let Ok((_,stripped)) = strip_compact_prefix(&ext.0) else {
```

subxt/src/utils/mod.rs
```diff
@@ -35,6 +35,12 @@ impl codec::Encode for Encoded {
     }
 }
 
+pub(crate) fn strip_compact_prefix(bytes: &[u8]) -> Result<(u64, &[u8]), codec::Error> {
```
jsdw, line 38: Nit: maybe worth a quick doc commanet like

```rust
/// Decodes a compact encoded value from the beginning of the provided bytes,
/// returning the value and any remaining bytes.
```

jsdw (approved):
LGTM, just a couple of tiny nits!

I guess you decided against using lifetimes in ExtrinsicDetails etc to avoid the cloining; I think that's something we could explore as a future PR though anyway, and I think now that this PR doesnt add any additional allocating over the original (and probably just removes a little as it's no longer also needed when decoding now), so yup it's all good!

lexnv (approved):
LGTM!

PR #1025, opened 2023-06-23
Adjust jsonrpsee web features

lexnv:
There were a couple of places that still use the `jsonrpsee-web` feature.

Subxt has transitioned to using the `web` feature to signal the wasm builds.

While at it, bump a few dependencies from Cargo.lock.


jsdw (approved):
Ooh good spots!

niklasad1 (approved):

PR #1026, opened 2023-06-23
Add light client platform WASM compatible 

### LINKED ISSUE #964: [light-client] Wasm compatible light-client

lexnv:
This PR exposes a pure-rust code base for light-client interaction in WASM environments.

- The low-level light-client implementation is moved to a dedicated crate part of the subxt workspace. This is done for modularity and better feature flag support while abstracting the environment interface.
- The low-level `smoldot_light::PlatformRef` is implemented for both Native and WASM. Although the Native could use the `smoldot_light::DefaultPlatform`, this is reused with minimal code for better light-client code coverage. The main advantage of this approach is pure-rust code for WASM environments without gluing JavaScript code. A downside is that the common code of the platform needs maintenance.
- A minimal dependency low-level `WebSocket` is implemented using `wasm-bindgen`
  - This takes inspiration from gloo-net, a crate that we also use in jsonrpsee for WASM environments
  - Most of the code is boilerplate for registering JavaScript closures/callbacks for open, message, error and close event
  - Besides minimal dependencies, the in-house `WebSocket` allows us to efficiently implement `AsyncRead` and `AsyncWrite` by levering a `VecDeque`
 
To enable the light-client users need to enable the "unstable-light-client" feature flag.
For wasm support, the "web" feature flag is also expected, which keeps consistency across subxt flags.
 
 ### Testing Done
 
 Similarly to the native light-client, a WASM compatible client syncs with the `Polkadot` live chain and subscribes to a few finalized chain blocks. 

Builds upon: https://github.com/paritytech/subxt/pull/965
Closes https://github.com/paritytech/subxt/issues/964.


subxt/src/client/lightclient/platform/native.rs
```diff
@@ -0,0 +1,161 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
```
tomaka, line 1: I've already asked this question to the `cumulus` team, and I'm repeating it here: why rewrite this implementation of `Platform`?

Smoldot [provides an implementation](https://github.com/smol-dot/smoldot/blob/main/light-base/src/platform/default.rs), and https://github.com/paritytech/subxt/pull/965 uses it. Why decide that copy-pasting its code here (without caring for the licensing, but let's forget that) when there's already a fully working and maintained version provided by smoldot itself?

To me you (and Cumulus) are really shooting yourself in the foot by doing this, and I just don't understand it.

We all know that this code will rot over time. You have the choice between the simple option that will work forever, and the complicated option that will likely require effort to maintain, and you go for the complicated one.

tomaka:
While it's not my problem, I just want to warn you about `wasm-bindgen`.
I tried for around six months to use `wasm-bindgen`, and after completely pulling my hair off out of frustration I have decided to get rid of it and write JavaScript code manually on top.

Smoldot notably is capable of running itself in a worker thread (which is important when you have CPU-heavy operations), compressing the .wasm file, and runs perfectly well on all bundlers, which are things that are complicated (to say the least) with wasm-bindgen. Maybe the situation has improved nowadays, but I highly doubt it.

I also want to point out that the code in this PR doesn't support WebRTC. As soon as WebRTC lands in Substrate (which, granted, is more than 6 months late already) and is working reasonably well, we want to deprecate the hell out of WebSockets.

subxt/src/client/lightclient/platform/native.rs
lexnv: Hey, I've implemented the low level `Platform` here to expose a pure-rust based implementation for wasm. While at it, I've also ported the platform for native (using tokio sockets) to allow us to reuse the same code for both platforms for better coverage. The differences are mainly in the `connect` function which uses a custom websocket with minimal dependencies that also implements `AsyncRead` / `AsyncWrite`.


Yep, I also felt the frustration with the `wasm-bindgen`, fortunately, we are using it for the custom socket. The socket is an oversimplification of [gloo-net](https://github.com/rustwasm/gloo/blob/aa0dbcdf3e27fdf359efe1259e5021e6e20fd3fb/crates/net/src/websocket/futures.rs#L48) (this is what we use by default with jsonrpsee for wasm environments).
I'm not quite sure if we could have a pure-rust based implementation without some types of bindings for the websocket 🤔 I would be up for transitioning to something different

> I also want to point out that the code in this PR doesn't support WebRTC.

Yep, we'd need to maintain the platform in the meanwhile if we pick this road in the end and look for alternatives of webrtc

jsdw: @lexnv what are the major blockers to using @tomaka's `Platform` implementation? I agree that it would be nicer to rely on an implementation that was supported and kept uptodate for us, so would be good to understand exactly the issues with doing so :)

lexnv: Pierre's platform works perfectly for native environments and there's also an alternative for WASM.
To make it fully compatible with our WASM needs the functions from [lightclient/src/platform/wasm.rs](https://github.com/paritytech/subxt/pull/1026/files#diff-1aa5737258d86a40987b2ab081bf272a5960e7051d042ce41e954ab933df57a6) are added to this PR.
For WASM we need: `spawn`, `now_from_unix_epoch`, `now`, `sleep`, and most importantly `connect` together with a few types, while for native we (including Cumulus) need just the `connect` to use a different socket.

Maybe we could leverage 90% of the smoldot Platform by having an extension trait.
The parts of the `Platform` that handles buffers are common between environments (`lightclient/src/platform/default.rs`), and the trickiest to implement right. Then we would need to implement just the connect for native and avoid duplicating the code here and in Cumulus.

```rust
// Extension trait to reuse the common parts of the Platform that handles buffers internally
trait PlatformRefBuffered {
 type Instant;
 type Delay;
 
 fn spawn() ..
 fn now() ..
 }

impl<T: PlatformRefBuffered> Platform for T {
...
}
```

### Smoldot Wasm Node

There's a crate in smoldot that implements this.
The crate requires the javascript realm to expose these [bindings](https://github.com/smol-dot/smoldot/blob/c300c082ec34a749108e71f12912e1355e168c67/wasm-node/rust/src/bindings.rs#L63-L64).

```rust
#[link(wasm_import_module = "smoldot")]
extern "C" {
  pub fn buffer_copy(buffer_index: u32, target_pointer: u32);
  ...
```

These are expected to be placed under the linking module "smoldot".
Before manually implementing the `Platform` here I also tried to create these bindings in rust, then reuse the wasm platform from this crate.

However, I was unable to expose any external functions under the `smoldot` module.
The WAT binary would expect the bindings as:

```bash
(import "smoldot" "buffer_copy" (func $_ZN18smoldot_light_wasm8bindings11buffer_copy17hf21d3e4eee7463c1E (type 6)))
(import "smoldot" "advance_execution_ready" (func $_ZN18smoldot_light_wasm8bindings23advance_execution_ready17h100f38f424c0220dE (type 0)))
(import "smoldot" "connection_new" (func $_ZN18smoldot_light_wasm8bindings14connection_new17hab2362b6b1ccdf39E (type 28)))
```
While regardless of where I tried to implement the functions they would be exported from rust as

```bash
(export "buffer_copy" (func $buffer_copy))
(export "buffer_size" (func $buffer_size))
(export "json_rpc_responses_non_empty" (func $json_rpc_responses_non_empty))
```

We could overcome this eventually by modifying smoldot and removing `#[link(wasm_import_module = "smoldot")]`.
However, all the binding functions are designed to work within the JavaScript world.
For example, the `buffer_copy` function takes as parameter `buffer_index: u32` which would mean we'd also need to implement in Rust a way to map a buffer's memory region to an index.
There's an example in [timers.rs](https://github.com/smol-dot/smoldot/blob/c300c082ec34a749108e71f12912e1355e168c67/wasm-node/rust/src/timers.rs) from the wasm-node, which uses some static variables to handle the mappings.

Just for reference, [here](https://github.com/smol-dot/smoldot/tree/main/wasm-node/javascript) is the JavaScript counterpart of the wasm platform.
An alternative to this would be writing the JavaScript code then binding to it in rust. Or trying to bind directly to the JavaScript code from wasm-platform.

That felt a bit more involved than having our slim socket bindings and using that common parts of the platform and I would be up for adjusting/modifying this to make things better for us 🙏

tomaka: > The parts of the Platform that handles buffers are common between environments (lightclient/src/platform/default.rs), and the trickiest to implement right

What I'm suggesting is to simply de-duplicate the code.

Sure the code is the same as in smoldot, but the choice here is between:

- Copy-pasting part of smoldot's `Platform` implementation into subxt and use it for both native and wasm sockets.
- Copy-pasting part of smoldot's `Platform` implementation into subxt and use only for wasm sockets.

I don't really see how solution 1 (which you're going for) is in any way better than solution 2.

> Before manually implementing the Platform here I also tried to create these bindings in rust,

That's not possible and I don't think it's a good approach nonetheless.

My personal intuition about the way forward is that I should try improve the `Platform` trait to be easier to implement, and/or maybe allow easily reusing smoldot's JavaScript.
But I can't really guarantee in any way that I'll succeed here.

lexnv: That totally makes sense! 🙏 

Indeed removing the native Platform from here and keeping only the WASM part will simplify the code for us.
The main advantage here would be that most users will be using a fully tested Platform provided by smoldot that we can easily update.
I was considering the native Platform from this PR to lay the foundation for testing and give us a bit more confidence, but I think we could overcome that if we get manage to get a bit more WASM testing.
And since we'll release this under an experimental/unstable feature flag in the beginning, I think we are fine for this PR.

> My personal intuition about the way forward is that I should try improve the Platform trait to be easier to implement, and/or maybe allow easily reusing smoldot's JavaScript.

Thanks! That would help us indeed! Just a crazy idea, maybe initially having a small platform extension for native only would also simplify the work from Cumulus side. I think Cumulus just needs a way to change the socket from `fn connect`

tomaka: > if we get manage to get a bit more WASM testing

IMO you need WASM-specific testing anyway, because browser APIs have many weird corners. I don't think it's a good idea to assume that things "just work" if you just plug APIs together.

> Thanks! That would help us indeed! Just a crazy idea, maybe initially having a small platform extension for native only would also simplify the work from Cumulus side. I think Cumulus just needs a way to change the socket from fn connect

I'm planning to rework the `connect` function quite a lot, for example to pass an already-parsed multiaddr, and to let smoldot do the WebSocket implementation on top of TCP (in other words the native implementation would just need to support TCP). But again these are just ideas and there are many more things on my to-do list that have priority.

Cargo.toml
```diff
@@ -77,10 +78,21 @@ wasm-bindgen-test = "0.3.24"
 which = "4.4.0"
 
 # Light client support:
+smoldot = { version = "0.8.0", default-features = false }
 smoldot-light = { version = "0.6.0", default-features = false }
 tokio-stream = "0.1.14"
 futures-util = "0.3.28"
 
+# Light client wasm:
+web-sys = { version = "0.3.61", features = [ "BinaryType", "CloseEvent", "MessageEvent", "WebSocket"] }
```
niklasad1, line 87:
```suggestion
web-sys = { version = "0.3.61", features = ["BinaryType", "CloseEvent", "MessageEvent", "WebSocket"] }
```

codegen/src/utils/fetch_metadata.rs
```diff
@@ -62,7 +62,7 @@ pub fn fetch_metadata_hex_blocking(
 
 // Block on some tokio runtime for sync contexts
 fn tokio_block_on<T, Fut: std::future::Future<Output = T>>(fut: Fut) -> T {
-    tokio::runtime::Builder::new_multi_thread()
+    tokio::runtime::Builder::new_current_thread()
```
niklasad1, line 65: can you explain this change? :)

lexnv: Oh yea, that shouldn't change! Initially, tokio was included with features in the workspace that did not play nice with wasm environment. Since then I've changed the tokio dependency to not include features and forgot about this! Thanks for spotting it!

altaua (approved):

lightclient/src/platform/mod.rs
jsdw, line 1: I find some of the names/organisation a touch confusing here; my understanding is that the code in this `platform` module is mostly actually for the WASM platform, and then the native platform is just using the smoldot default one.

So currently the `default` mod is the WASM platform definition (which confusingly for me uses `mod wasm as platform_impl`), the `wasm` mod is just helper code for this definition and `wasm_socket` a websocket impl.

I'd prolly go with something like:

```rust
/// WASM platform; exposes a `Platform`. (formally `default.rs`)
mod wasm_platform;
/// WASM helper definitions: (formally `wasm.rs`)
mod wasm_helpers;
/// WASM web socket impl (left as is
mod wasm_socket;
```

And then just reference `wasm_helpers::X` where needed rather than renaming to `platform_impl` so that it's clearer where things are coming from :)

lightclient/Cargo.toml
jsdw, line 1: Good idea to have this as a separate crate I think!

lightclient/src/platform/default.rs
```diff
@@ -0,0 +1,299 @@
...
+    type Delay = platform_impl::Delay;
+    type Yield = future::Ready<()>;
+    type Instant = platform_impl::Instant;
+    type Connection = std::convert::Infallible;
+    type Stream = platform_impl::Stream;
+    type ConnectFuture = future::BoxFuture<
+        'static,
+        Result<PlatformConnection<Self::Stream, Self::Connection>, ConnectError>,
+    >;
+    type StreamUpdateFuture<'a> = future::BoxFuture<'a, ()>;
```
jsdw, line 43: Oooh GATs!

lightclient/src/platform/default.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    fn next_substream<'a>(&self, c: &'a mut Self::Connection) -> Self::NextSubstreamFuture<'a> {
+        // This function can only be called with so-called "multi-stream" connections. We never
+        // open such connection.
+        match *c {}
+    }
+
+    fn update_stream<'a>(&self, stream: &'a mut Self::Stream) -> Self::StreamUpdateFuture<'a> {
+        Box::pin(future::poll_fn(|cx| {
+            let Some((read_buffer, write_buffer)) = stream.buffers.as_mut() else { return Poll::Pending };
```
jsdw, line 117: `Poll::Pending` here would get stuck forever (nothing would ever wake it up again except by chance); does it make more sense to return an error of some kind?

lightclient/src/platform/default.rs
```diff
@@ -0,0 +1,299 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if update_stream_future_ready {
+                Poll::Ready(())
+            } else {
+                Poll::Pending
```
jsdw, line 226: Whenever `Poll::Pending` is returned, the `Context` (cx) _must_ at some point be used to wake the future up again, so that it continues to make progress. in at least one case, it looks like we return `Pending` here when nothing would be planning to ever call `wake()` on the context to wake it up again. Def something to ponder I think, so that things don't get stuck waiting forever :)

lightclient/src/platform/wasm_socket.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+impl AsyncRead for WasmSocket {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        let mut inner = self.inner.lock().expect("Mutex is poised; qed");
```
jsdw, line 167: Another `Pending`/`Waker` thing:

Because this `WasmSocket` is behind an `Arc<Mutex<..>>`, a question I have is; what happens if we make, say, 3 copies of this, and then for each copy, in some different task, we `.read().await`?

Each copy would call `poll_read`, and then the waker given to each `poll_read` would be saved into `inner` to be woken up. But the wakers would overwrite eachother; 2 of the wakers would therefore never be used, and 2 of the three tasks would never make progress.

I had to consider this in some personal thing where I implemented a small MPSC style channel (https://github.com/jsdw/asyncified/blob/main/src/channel.rs), and there I had each clone of the sender allocate itself a new ID, and then I stored all wakers in a `HashMap<ID, waker>`) so that I would remove them at needed on cleanup, and wake them all up on progress.

That all said, `pub struct WasmSocket` is not itself clonable, so, does it need the `Arc` inside it at all? (or the mutex for that matter)? Because if you could clone the socket and read from it, different clones would be getting different parts of the output anyway which probably isn't what we want.

So maybe we can just remove the `Arc<Mutex>`? Or at least the `Arc`? And then the type system enforces that it can only be used in one place, with one waker, at a time.

testing/wasm-tests/tests/wasm.rs
```diff
@@ -22,3 +38,29 @@ async fn wasm_ws_transport_works() {
     let chain = client.rpc().system_chain().await.unwrap();
     assert_eq!(&chain, "Development");
 }
+
+#[wasm_bindgen_test]
+async fn light_client_works() {
```
jsdw, line 43: I guess that by enabling the "unstable-light-client" feature on this crate, we no longer are testing the "wasm but not via light client" route; maybe we should have two crates? `wasm-rpc-test` and `wasm-lightclient-test`? What do you think?

jsdw:
Generally looking great; just some `Poll::Pending` things to think about :)

lightclient/Cargo.toml
```diff
@@ -0,0 +1,86 @@
...
+smoldot-light = { workspace = true, optional = true }
+either = { workspace = true, optional = true }
+tokio = { workspace = true, optional = true }
+tokio-stream = { workspace = true, optional = true }
+futures-util = { workspace = true, optional = true }
+js-sys = { workspace = true, optional = true }
+send_wrapper = { workspace = true, optional = true }
+web-sys = { workspace = true, optional = true }
+wasm-bindgen = { workspace = true, optional = true }
+wasm-bindgen-futures = { workspace = true, optional = true }
```
jsdw, line 71: Just as a curiosity; since we don't use wasm_bindgen for generating any glue code by the looks of it (we rely on web_sys etc to interact with JS), are we importing these crates just to use certain types from them? And could we import those types directly from some smaller crate or whatever and not have any wasm-bindgen at all?

jsdw (commented):
Generally looking good; just a few questions and thoughts mainly around `Poll::Pending` uses which could be problematic :)

lightclient/src/platform/default.rs
lexnv: Indeed, the `Poll::Pending` here will never wake up, since we don't store the context anywhere. 

The stream's buffers can only be empty if the connection is in the `Reset` state. The `fn connect(&self, multiaddr: &str) -> Self::ConnectFuture ` method is called first to establish the connection that initializes the buffers here.

Looking a bit at smoldot, my understanding is:
- A task is spawned to handle incoming connection attempts ([here](https://github.com/smol-dot/smoldot/blob/5bb664978d0576fe4d19b9ed6f7537798d238e43/light-base/src/network_service.rs#L1058-L1068))
- The task checks in a loop that progress can be made on the connect; and updates some internal state machine
- The loop is ended when it cannot send a message back to the background task; or when the socket is closed 
- The last step of that loop waits for one of the following to happen
  - stream update (this future) to complete
  - receive a message from the coordinator (some internal detail of smoldot)
  - a timeout to happen
  
I believe that we drop this future after some period of time when we return `Poll::Pending` here.

The smoldot's `type StreamUpdateFuture<'a>: Future<Output = ()> + Unpin + Send + 'a;` forces us to return `()` and considering the above we might introduce a busy wait loop if we could return `Poll::Ready(Err()))`

// cc @tomaka would love to hear your thoughts on this 🙏

tomaka: You don't need to read smoldot's source code, you just have to read the documentation of the function you're implementing: https://docs.rs/smoldot-light/latest/smoldot_light/platform/trait.PlatformRef.html#tymethod.update_stream

It says:

> Returns a future that becomes ready when “something” in the state has changed.

Once a connection is in the reset state, it state never changes, and thus returning an infinite future is correct.

tomaka: The reason why the connections-related functions in the `Platform` trait are so different from `AsyncRead` and `AsyncWrite` is that `AsyncRead` and `AsyncWrite` are full of weird corner cases.

What errors can actually be returned and what do they represent? For example, can you suddely get a `PermissionDenied` error?
Can the reading side return errors and the writing side still work, or vice versa?
Are some errors recoverable? Can you continue reading/writing again despite an error being returned?
If you call `poll_close` and it returns `Pending`, can you start calling `poll_write` again after?

So to solve all of this I created my own abstraction which follows what I think are straight forward rules. There are no errors, and just 5 possibles states (open, read open write closed, read closed write open, both closed, and reset).
It's a bit annoying to plug into existing APIs, but I'm going to provide helpers in the future.

testing/wasm-tests/tests/wasm.rs
lexnv: Yep, that makes a lot of sense, have split up the testing into separate crates! Thanks for pointing out the light-client RPC!

lightclient/src/platform/mod.rs
lexnv: Indeed that looks a lot cleaner! Have adjusted the modules, thanks!
Initially, I was thinking of the `default.rs` as a common ground between a `native` and a `wasm`, but since have removed the native support in favor of the light-client's default platform and this makes more sense now!

lightclient/src/platform/wasm_socket.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// Websocket for WASM environments.
+///
+/// This is a rust-based wrapper around browser's WebSocket API.
+pub struct WasmSocket {
+    /// Inner data shared between `poll` and web_sys callbacks.
+    ///
+    /// This is safe in wasm environments.
+    inner: SendWrapper<Rc<RefCell<InnerWasmSocket>>>,
```
jsdw, line 31: Ah, so the reason for the `Arc<Mutex<..>>` was because this thing needed sharing across some different closures below; I missed that bit! And when compiled to WASM the mutex will disappear anyway since it's single threaded.

I think given that the outer struct is not cloneable, I'd remove the `SendWrapper` thing (and crate) and just put the `Arc<Mutex<..>>` back (it feels cleaner than this `SendWrapper<Rc<RefCell<..>>>` thing). Just maybe a dev comment on `WasmSocket` that the thing cannot impl `Clone` would do I think :)

lightclient/src/platform/wasm_socket.rs
jsdw: Ah, so the reason for the `Arc<Mutex<..>>` was because this thing needed sharing across some different closures below; I missed that bit! And when compiled to WASM the mutex will disappear anyway since it's single threaded.

I think given that the outer struct is not cloneable, I'd remove the `SendWrapper` thing (and crate) and just put the `Arc<Mutex<..>>` back (it feels cleaner than this `SendWrapper<Rc<RefCell<..>>>` thing). Just maybe a dev comment on `WasmSocket` that the thing cannot impl `Clone` would do I think :)

lightclient/src/platform/default.rs
jsdw: Thankyou for adding the comments; as a reviewer it's scary to see the `Pending`'s without context wakers being used/stored/whatever, but I think I understand how it's working a bit better now!

jsdw (approved):
Excellent work @lexnv; thanks for looking into those `Pending` things and checking that they are ok (ie smoldot is dropping the future or whatever anyway in those cases).

I'm happy to see this merge now and we can iterate on it from there if we run into things :)

PR #1041, opened 2023-06-29
Utilize Metadata V15

lexnv:
This PR makes use of metadata V15, marking the stabilization of the metadata in Subxt.

- V15 is the preferred metadata version in subxt (after unstable)
- codegen utilizes the RuntimeError, RuntimeEvent and RuntimeCall directly from the metadata
- hashing and stripping of metadata are modified to reflect the outer enum types
- converting from V14 to V15 and vice-versa builds and appends the needed PortableTypes to the metadata registry
- CI link for the substrate binary is updated -- prior we were using a 9-week old release
- tests are adjusted for the latest substrate binary
  - contracts: all contracts need to import the memory from env `(import "env" "memory" (memory 1 1))` https://github.com/paritytech/substrate/pull/14449
  - staking: controller account soon to be deprecated; all controlling functions are moved in the stash account https://github.com/paritytech/substrate/pull/14039 (merged about 8 weeks ago)

### Testing

```bash
// Main outer enum types
"outer_enums": {
"call_enum_ty": 140,
"event_enum_ty": 20,
"error_enum_ty": 877
},

// RuntimeCall type
  "id": 140,
"type": {"path": [ "kitchensink_runtime", "RuntimeCall" ],
    "def": { "variant": { "variants":  { "name": "System",..

// System variant of the RuntimeCall. 
id": 141,
"type": { "path": [ "frame_system", "pallet", "Call",
    "def": { "variant": { "variants": [ { "name": "remark", ...

// RuntimeEvent type
"id": 20,
"type": {"path": [kitchensink_runtime", "RuntimeEvent"],
    "def": {"variant": { "variants": [ { "name": "System", ..
          
// System variant of the RuntimeEvent. 
  "id": 21,
    "type": {  "path": [ "frame_system", "pallet","Event" ],
        "def": { "variant": { "variants": [ { "name": "ExtrinsicSuccess", ..

// RuntimeError type
"id": 877,
"type": { "path": [ "kitchensink_runtime", "RuntimeError" ],
    "def": { "variant": { "variants": [ { "name": "System", ...

// System variant of the RuntimeError. 
"id": 450,
    "type": { "path": [ "frame_system","pallet", "Error"],
        "def": {"variant": { "variants": [ { "name": "InvalidSpecName", ..

### CODE REVIEW

File: metadata/src/from_into/v14.rs [hunk 1]
```diff
@@ -178,17 +226,190 @@ fn v14_to_v15(metadata: v14::RuntimeMetadataV14) -> v15::RuntimeMetadataV15 {
...
+
+fn generate_outer_enums(
+    metadata: &mut v14::RuntimeMetadataV14,
+) -> v15::OuterEnums<scale_info::form::PortableForm> {
+    let call_enum = metadata
+        .types
+        .types
+        .iter()
+        .find(|ty| {
+            let Some(ident) = ty.ty.path.ident() else { return false };
```
niklasad1, line 316: I don't follow the `return false` thingy here, if `an Ident` is missing what does that mean?

metadata/src/from_into/v14.rs
```diff
@@ -178,17 +226,190 @@ fn v14_to_v15(metadata: v14::RuntimeMetadataV14) -> v15::RuntimeMetadataV15 {
...
+                index: pallet.index,
+                docs: vec![],
+            })
+        })
+        .collect();
+
+    let error_type = scale_info::Type {
+        path: scale_info::Path {
+            segments: vec![
+                "kitchensink_runtime".to_string(),
```
niklasad1, line 398: why is this injected? 

to work with node-template/node?

metadata/src/utils/validation.rs
```diff
@@ -195,11 +212,38 @@ pub fn get_type_hash(
 fn get_extrinsic_hash(
     registry: &PortableRegistry,
     extrinsic: &ExtrinsicMetadata,
+    specific_pallets: &Option<Vec<String>>,
```
niklasad1, line 215: I reckon that it would be cleaner to change this to (if possible):
```suggestion
    specific_pallets: Option<&[String]>,
```

niklasad1 (approved):
Looks good to me, added some minor questions/suggestions

Happy to get the latest substrate releases running in the CI again

metadata/src/from_into/v14.rs
lexnv: Subxt uses the latest stable metadata V15 under the hood, just to make sure we expose the latest features (runtime API etc).

Prior to this PR, codegen generated the RuntimeError type manually. For V15 this is already exposed.
We still want to want subxt compatible with V14. When we are targeting a V14 node, we are generating the type here to give subxt a working V15 :D

altaua (approved):

testing/test-runtime/build.rs
```diff
@@ -37,8 +37,8 @@ async fn run() {
 
     // Download metadata from binary. Avoid Subxt dep on `subxt::rpc::types::Bytes`and just impl here.
     // This may at least prevent this script from running so often (ie whenever we change Subxt).
-    const V15_METADATA_VERSION: u32 = u32::MAX;
-    let bytes = V15_METADATA_VERSION.encode();
+    const LATEST_STABLE_METADATA_VERSION: u32 = 15;
+    let bytes = LATEST_STABLE_METADATA_VERSION.encode();
```
jsdw, line 41: Nit: I prefer `V15_METADATA_VERSION`, because "LATEST" can change at any time and render this inaccurate :)

metadata/src/from_into/v14.rs
jsdw: Do we currently test that things work fine if V14 metadata is provided (and needs to be upgraded to V15) rather than just when V15 metadata is provided directly? It'd be good to have confidence that everything works as expected in both cases, ie the shims we insert when upgrading to V15 all work as expected (maybe this is caught in the act of upgrading metadata into the `Metadata` struct?)

metadata/src/utils/validation.rs
```diff
@@ -195,11 +212,38 @@ pub fn get_type_hash(
...
+            .types
+            .get(extrinsic.call_ty as usize)
+            .expect("Metadata should contain call enum type in registry");
+
+        // If the `RuntimeCall` is a variant, then filter out the pallets of interest.
+        // Note: this can be a different type def for our testing.
+        if let TypeDef::Variant(variant) = &ty.ty.type_def {
+            get_type_def_variant_hash(registry, variant, specific_pallets, &mut visited_ids)
+        } else {
+            get_type_hash(registry, extrinsic.signature_ty, &mut visited_ids)
```
jsdw, line 235: Why `extrinsic.signature_ty` here; is this supposed to just be some arbitrary hash that's returned to make testing a little easier?

If so, maybe just return some arbitrary thing, orr call `get_type_hash` on the entire `ty` if it's not a variant for "correctness"?

metadata/src/utils/validation.rs
```diff
@@ -104,6 +104,30 @@ fn get_variant_hash(
     concat_and_hash2(&variant_name_bytes, &variant_field_bytes)
 }
 
+fn get_type_def_variant_hash(
+    registry: &PortableRegistry,
+    variant: &TypeDefVariant<PortableForm>,
+    specific_pallets: &Option<Vec<&str>>,
```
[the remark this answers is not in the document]
jsdw: But, a follow-up; why is this actually needed at all? Since we expect to call `retain_pallets_in_runtime_outer_types`, which strips out unwanted variants already, then do we need to have any "specific pallets" type checking in `get_extrinsic_hash`?

metadata/src/utils/retain.rs
```diff
@@ -136,38 +157,36 @@ fn update_type(ty: &mut u32, map_ids: &BTreeMap<u32, u32>) {
     *ty = new_id;
 }
 
-/// Strip any pallets out of the RuntimeCall type that aren't the ones we want to keep.
-/// The RuntimeCall type is referenced in a bunch of places, so doing this prevents us from
-/// holding on to stuff in pallets we've asked not to keep.
-fn retain_pallets_in_runtime_call_type<F>(metadata: &mut Metadata, mut filter: F)
+/// Retain the enum type identified by ID and keep only the variants that
+/// match the provided filter.
+fn retain_enum_type<F>(metadata: &mut Metadata, id: u32, mut filter: F)
```
jsdw, line 162: Nit; maybe a name like `retain_variants_in_enum_type` would be more descriptive of what it's doing? :)

subxt/src/client/online_client.rs
```diff
@@ -131,14 +134,34 @@ impl<T: Config> OnlineClient<T> {
...
+                Err(_) => OnlineClient::fetch_latest_stable_metadata(rpc).await,
             }
         }
 
         #[cfg(not(feature = "unstable-metadata"))]
+        OnlineClient::fetch_latest_stable_metadata(rpc).await
+    }
+
+    /// Fetch the latest stable metadata from the node.
+    async fn fetch_latest_stable_metadata(rpc: &Rpc<T>) -> Result<Metadata, Error> {
```
jsdw, line 148: I think, since Subxt supports V14 and V15 metadata explicitly and doesn't support others (atm), I'd be explicit about trying to get V15 metadata from the call (incase some node decides to return some other version that we don't support decoding from or whatever).

I pondered removing the `unstable-metadata` feature flag but I think I'm happy keeping it as an "opt-in that will fall back to latest supported stable if not exists", just because why not. :)

So let's explicitly ask for V15 here, and if that fails, call `rpc.metadata().await` to get V14 which we also know we can decode (little easier anyway since we don't need to bother querying versions even if we don't want to). Then we'll keep working regardless of any future/different metadata versions, and can upgrade subxt to support them whenever we like.

metadata/src/utils/validation.rs
lexnv: Thanks for spotting this! I mean to have `call_ty` here instead of `signature_ty`.
If the `RuntimeCall` is an enum (as should be defined by substrate), then we know which variants of the enum to include in the hash. Otherwise, we hash everything from that type (since we may add a struct or a tuple in the metadata for our testing).

metadata/src/utils/validation.rs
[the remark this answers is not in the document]
lexnv: Yep, that makes sense!

Users may still want to compare some partially stripped metadata to the full metadata of a node. 
For those cases, the stripped metadata will contain only some specific variants for the `RuntimeCall`, while the full metadata will differ containing all pallets as variants.

The metadata also includes the `RuntimeError` and `RuntimeEvent` enums that are similar to the `RuntimeCall`.
The first 2 enums aren't hashed specifically in the validation phase, since the values of those variants are hashed indirectly while passing through the pallet's metadata.

The main reason for hashing the `RuntimeCall` explicitly here is because it is a component of the extrinsic's metadata. We could potentially generate the same extrinsic hash for different extrinsic metadatas (depending on the Runtime variants).

I believe we have two options for this:
- don't hash the `RuntimeCall` as part of the extrinsic metadata and rely on the pallets's parts to hash indirectly the variant's value; and have a slightly inconsistent extrinsic metadata hash
- hash the `RuntimeCall`; and have some extra steps to filter the variants

I'd love to get your thoughts on this! I'm not super sure if we may need to maintain hashing compatibility at the `ExtrinsicMetadata`'s level in the future

metadata/src/utils/validation.rs
jsdw: Makes sense to me :)

metadata/src/lib.rs
```diff
@@ -440,18 +447,37 @@ impl ConstantMetadata {
 /// Metadata for the extrinsic type.
 #[derive(Debug, Clone)]
 pub struct ExtrinsicMetadata {
-    /// The type of the extrinsic.
-    ty: u32,
+    /// The type of the address that signes the extrinsic
```
jsdw, line 450: signs :)

metadata/src/utils/validation.rs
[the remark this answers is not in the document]
jsdw: Summary because that was longer than planned:

I think we should hash (with `only_these_pallets` if given) the `outer_enum` types, and if we do that I don't think we need to hash the `extrinsic_metadata.call_ty` one, since it's the same I think?

codegen/src/api/mod.rs
```diff
@@ -551,7 +496,8 @@ impl RuntimeGenerator {
                 /// The error type returned when there is a runtime issue.
                 pub type DispatchError = #types_mod_ident::sp_runtime::DispatchError;
 
-                #outer_event
+                /// The outer event enum.
+                pub type Event = #event_path;
 
                 impl #crate_path::events::RootEvent for Event {
```
jsdw, line 502: Just to add another simplification; `RootEvent`, `RootExtrinsic` and `RootError` traits can all be removed now, since we now have type IDs and thus are able to decode the call/error/tx into whatever type implementing `DecodeAsType` makes sense. This should simplify the code a bunch more :)

metadata/src/utils/validation.rs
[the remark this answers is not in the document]
lexnv: Yep, that makes a lot of sense, sounds like a plan!

metadata/src/from_into/v14.rs
lexnv: Ops, missed this thread! Have added some extra tests to check we propagate correctly the newly constructed types (extrinsic / outerenums). While going over the testing, have simplified a bit how we generate the `RuntimeError` variants

.github/workflows/rust.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
 env:
     CARGO_TERM_COLOR: always
     # TODO: Currently pointing at latest substrate; is there a suitable binary we can pin to here?
-    SUBSTRATE_URL: https://releases.parity.io/substrate/x86_64-debian:stretch/latest/substrate/substrate
+    SUBSTRATE_URL: https://releases.parity.io/substrate/x86_64-debian:bullseye/latest/substrate/substrate
```
jsdw, line 22: Next time it'd prolly be good to get the CI fixes in a separate PR too so we can merge them more easily, but I think this will be able to merge shortly anyway hopefully now :)

codegen/src/api/mod.rs
```diff
@@ -382,118 +381,6 @@ impl RuntimeGenerator {
             })
             .collect::<Result<Vec<_>, CodegenError>>()?;
 
-        let outer_event_variants = self.metadata.pallets().filter_map(|p| {
```
jsdw, line 385: Wooo, even more red than last time :D

metadata/src/from_into/v14.rs
```diff
@@ -178,17 +226,375 @@ fn v14_to_v15(metadata: v14::RuntimeMetadataV14) -> v15::RuntimeMetadataV15 {
...
+            let error_ty = metadata
+                .types
+                .resolve(pallet_error.ty.id)
+                .expect("Error type is present in the metadata; qed");
+
+            // Note: the type name is the name of the type as it appears in substrate, we cannot
+            // accurately reproduce it when it contains more than 1 generic.
+            // The first generic is always the `Runtime`, the others represent instances.
+            let instances = ",Instance".repeat(error_ty.type_params.len() - 1);
+            let path = format!("{}<Runtime{}>", error_ty.path, instances);
```
jsdw, line 371: This looks super hacky, and I think it's just to populate the "type_name" thing? My question is; do we actually need this? Would anything break if we set `type_name` to be eg `{PalletName}Error`?

metadata/src/from_into/v14.rs
```diff
@@ -178,17 +226,375 @@ fn v14_to_v15(metadata: v14::RuntimeMetadataV14) -> v15::RuntimeMetadataV15 {
...
+    metadata.types.types.push(scale_info::PortableType {
+        id: error_type_id,
+        ty: error_type,
+    });
+
+    error_type_id
+}
+
+#[cfg(test)]
+mod tests {
```
jsdw, line 407: Like the tests; converting to v14 and back to v15 again and checking that things line up still is a good way to check :)

metadata/src/utils/validation.rs
```diff
@@ -198,8 +215,16 @@ fn get_extrinsic_hash(
 ) -> [u8; HASH_LEN] {
     let mut visited_ids = HashSet::<u32>::new();
 
-    let mut bytes = concat_and_hash2(
-        &get_type_hash(registry, extrinsic.ty, &mut visited_ids),
+    // Get the hashes of the extrinsic type.
+    let address_hash = get_type_hash(registry, extrinsic.address_ty, &mut visited_ids);
+    // The `RuntimeCall` type is intentionally emitted and hashed by the outer enums instead.
```
jsdw, line 220: Nit: omitted :)

metadata/src/utils/validation.rs
```diff
@@ -215,6 +240,57 @@ fn get_extrinsic_hash(
...
+        only_these_variants: Option<&[&str]>,
+        visited_ids: &mut HashSet<u32>,
+    ) -> [u8; HASH_LEN] {
+        let ty = registry
+            .types
+            .get(id as usize)
+            .expect("Metadata should contain enum type in registry");
+
+        if let TypeDef::Variant(variant) = &ty.ty.type_def {
+            get_type_def_variant_hash(registry, variant, only_these_variants, visited_ids)
```
jsdw, line 264: I'm pondering the use of `visited_ids` a bit now; eg is it right to pass the same `visited_ids` set into the call for each enum? What's the effect of doing so versus not doing so.

Doing so means that hashing things that share types might be faster; any type that's been seen will hash to some fixed "recursive" hash. The downside though is that the individual hashes become "weaker"; a bunch of more primitive types all are assigned the "recursive" hash when hashing the second and third enum. This makes them all effectively interchangeable; we'll no longer catch changes to them.

The point of `visited_ids`  is only to prevent infinite loops, but given the above, let's pass an empty `visited_ids` set into each new top level call; so here `get_enum_hash` could create an empty Set each time for instance so that the types are entirely independently hashed.

metadata/src/utils/validation.rs
```diff
@@ -481,8 +557,19 @@ impl<'a> MetadataHasher<'a> {
 
         let extrinsic_hash = get_extrinsic_hash(&metadata.types, &metadata.extrinsic);
         let runtime_hash = get_type_hash(&metadata.types, metadata.runtime_ty(), &mut visited_ids);
```
jsdw, line 559: The `visited_ids` passed in here is only used in this one place anyway, which is good (see prev comment). Let's just pass in an `&mut HashSet::new()` to this function to ensure that we don't use it in some other place in the future and be consistent on "pass new visited_ids into each new top level hashing call".

jsdw: (looking around this file more I think we should re-think the `visited_ids` thing a bit or at least check that we're not overly using the same one; something I've probably done in other places too. I think ultimately we will want `get_type_hash` to _not_ take in `visited_ids` and instead construct it internally and then an internal version of `get_type_hash` can accept it for the recursive calls to it. Basically, each unique type that we hash should have start from a fresh set of visited IDs. I'll open an issue to do this more thoroughly)

jsdw: Opened https://github.com/paritytech/subxt/issues/1066 to generally fix/prevent misuse of `visited_type_ids`

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -416,22 +399,17 @@ where
         }
     }
 
-    /// Attempt to decode these [`ExtrinsicDetails`] into a root extrinsic type (which includes
+    /// Attempt to decode these [`ExtrinsicDetails`] into a outer call enum type (which includes
```
jsdw, line 402:
```suggestion
    /// Attempt to decode these [`ExtrinsicDetails`] into an outer call enum type (which includes
```

subxt/src/error/dispatch_error.rs
```diff
@@ -191,17 +191,24 @@ pub struct ModuleErrorDetails<'a> {
...
-    pub pallet_index: u8,
-    /// Raw error bytes.
-    pub error: [u8; 4],
+    /// The raw bytes of the module error.
+    ///
+    /// Bytes representation:
+    ///  - `bytes[0]`:   pallet index
+    ///  - `bytes[1]`:   error index
+    ///  - `bytes[2..]`: 3 bytes specific for the module error
+    pub bytes: [u8; 5],
```
jsdw, line 200: Let's not expose `bytes` like this, because the `RawModuleError` structure is copied from Substrate (with a `pallet_index` and `error` field as seen before), soooo how about:

- Remove `RawModuleError` entirely now that it doesn't align with the Substate type anymore anyway (and its easiewr for us just to keep all of the bytes for easy decoding etc)
- Add `bytes()`, `pallet_index()` and `error_index()` to `ModuleError` so that one can still obtain the relevant parts and doesn't need to care about any internal details.

jsdw:
This looks awesome, nice work!

Two minor comments left:
- The hacky `type_name` thing; do we need it? (maybe? but maybe not?)
- Get rid of `RawModuleError` since it doesn't align with Substrate anyway now.

I wouldn't worry too much about the `visited_ids` stuff; we can handle that in #1066 :)

metadata/src/from_into/v14.rs
lexnv: I believe we could use the `PalletError` name here, have removed it :D

subxt/src/error/dispatch_error.rs
lexnv: That makes sense! Have removed the `RawModuleError` thanks!

metadata/src/from_into/v14.rs
jsdw: Ok awesome! Yeah, I had a look and `type_name` isn't really relevant; it's just for "disgnostic" info: https://github.com/paritytech/scale-info/blob/cc97f4c6496acf0f92d9caa45856331f094a7f38/src/ty/fields.rs#L55

The `name` of the field is the enum variant and the thing that decoding/encoding cares about :)

jsdw (approved):
Awesome stuff!

PR #1043, opened 2023-06-30
Example: How to connect to parachain

tadeohepperle:
The example shows how someone might construct the _subxt::Config_ and the _metadata module_ for a certain parachain if they develop an application with subxt.

The example is using the Statemint (Polkadot Asset Hub) parachain as an example.


examples/parachain-example/src/bin/submit_transaction.rs
```diff
@@ -0,0 +1,146 @@
...
+//! The parachain is only registered after the first epoch. So we need to wait 2 minutes, until the parachain becomes interactive and produces blocks.
+//!
+//!
+//! # Running the example
+//!
+//! After you have the network running, you should see something like [https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:41955#/explorer]
+//! as the polkadot.js link to the parachain (e.g. "collator01") in the zombienet output. Here `41955` refers to the port,
+//! which we will need shortly.
+//!
+//! In this example we use the _uniques_ pallet of the polakdot asset hub parachain.
```
niklasad1, line 61:
```suggestion
//! In this example we use the _uniques_ pallet of the polkadot asset hub parachain.
```

examples/parachain-example/src/bin/submit_transaction.rs
```diff
@@ -0,0 +1,146 @@
...
+//!
+//! You might have notices, that we use `chain = "rococo-local"` in the `asset-hub-zombienet.toml` file for the relay chain.
+//! This is just to make the epoch time shorter and should have no effect on your interactions with the parachain.
+//! Polkadot / Kusama / Rococo have different epoch times of `24h` / `2h` / `2min` respectively.
+//! The parachain is only registered after the first epoch. So we need to wait 2 minutes, until the parachain becomes interactive and produces blocks.
+//!
+//!
+//! # Running the example
+//!
+//! After you have the network running, you should see something like [https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:41955#/explorer]
```
niklasad1, line 57: can't we instruct zoombienet to use a specific port to avoid more configurations to run this example?

examples/parachain-example/asset-hub-zombienet.toml
```diff
@@ -0,0 +1,25 @@
+[relaychain]
+default_image = "docker.io/parity/polkadot:latest"
+default_command = "polkadot"
+default_args = [ "-lparachain=debug" ]
```
niklasad1, line 4:
```suggestion
default_args = ["-lparachain=debug"]
```

examples/parachain-example/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn encode_additional_to(&self, v: &mut Vec<u8>) {
+        self.additional_params.encode_to(v);
+    }
+}
+
+////////////////////////////////////////////////////////////
+// Third Config (using the Substrate and Polkadot Config)
+////////////////////////////////////////////////////////////
+
+pub enum StatemintConfig3 {}
```
niklasad1, line 356: I don't see why you need to duplicate this config, it would be better to just to call this `PolkadotConfig` via type alias or something.

This example contains already plenty of code :)

examples/parachain-example/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn encode_additional_to(&self, v: &mut Vec<u8>) {
+        self.additional_params.encode_to(v);
+    }
+}
+
+////////////////////////////////////////////////////////////
+// Second Config (just using subxt types)
+////////////////////////////////////////////////////////////
+
+pub enum StatemintConfig2 {}
```
niklasad1, line 276: maybe you could rename this to `StatemintConfigAssetTip`

examples/parachain-example/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    #[codec(compact)]
+    tip: u128,
+    asset_id: Option<u32>,
+}
+
+////////////////////////////////////////////////////////////
+// First Config (verbose and detailed)
+////////////////////////////////////////////////////////////
+
+pub enum StatemintConfig {}
```
niklasad1, line 196: maybe you could rename this to `StatemintConfigChargeAssetTxPayment`

examples/parachain-example/src/lib.rs
tadeohepperle: Just because the ExtrinsicParams are from Substrate instead of from Polkadot here. So it would be inaccurate to take just the PolkadotConfig.

.gitignore
```diff
@@ -4,3 +4,5 @@
 cargo-timing*
 /examples/wasm-example/dist
 /examples/wasm-example/target
+/examples/parachain-example/target
+/examples/parachain-example/metadata/target
```
lexnv, line 8: nit: new line here?

examples/parachain-example/examples/create_nft.rs
```diff
@@ -0,0 +1,141 @@
...
+//! The parachain is only registered after the first epoch. So we need to wait 2 minutes, until the parachain becomes interactive and produces blocks.
+//!
+//! # Running the example
+//!
+//! After you have the network up and running, you should see something like [https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:42069#/explorer]
+//! as the polkadot.js link to the parachain (e.g. "collator01") in the zombienet output.
+//!
+//! In this example we use the _uniques_ pallet of the polkadot asset hub parachain.
+//! The dev account _alice_ creates an NFT collection, then mints an NFT that is part of the collection.
+//! Beware that this example can take up to a minute to run. Specify the port like so:
```
lexnv, line 61: nit: should a configurable `--port 10101` be provided in the example?

subxt/src/book/mod.rs
```diff
@@ -71,8 +71,9 @@
 //! recent releases). Typically, to use Subxt to talk to some custom Substrate node (for example a
 //! parachain node), you'll want to:
 //!
-//! 1. [Generate an interface](setup::codegen).
-//! 2. [Configure and instantiate the client](setup::client).
+//! 1. [Generate an interface](setup::codegen)
+//! 2. [Create a config](setup::config)
+//! 2. [Use the config to instantiate the client](setup::client)
```
lexnv, line 76: nit: 3?

examples/parachain-example/examples/create_nft.rs
tadeohepperle: Ah, I forgot to remove this line. The port is fixed in the example (hard coded in the `asset-hub-zombienet.toml` and the `create_nft.rs` files).

examples/parachain-example/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+//! # Tutorial: Subxt setup to interact with custom chains
```
jsdw, line 1: I think that both this and the WASM example should have a README.md so that it's immediately obvious where to look to find instructions on running them, so I'd move these docs into a README file instead I reckon.

Having great Rust docs for this (and the other example) crate isn't super important anyway since we don't publish them or expect anybody to consume the "apis" of them or whatever :)

examples/parachain-example/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+//!
+//! There are many parachains on Polkadot and Kusama. Using Subxt to connect to a parachain
+//! (or any other custom, substrate based chain) requires 2 steps:
```
jsdw, line 4:
```suggestion
//! (or any other custom, substrate based chain) typically involves 2 steps:
```

(just because you can use the dynamic interface and not do step 1 at all if you don't want to, so it's not a requirement)

tadeohepperle (commented):
The restructuring looks good to me, I like how you broken down the type replacements into different sections and the example is now more slim.

jsdw (approved):

lexnv (approved):

jsdw:
Ah one mo; hit a compilation error with the example; need to add CI or something and fix it

.github/workflows/rust.yml
```diff
@@ -109,7 +109,7 @@ jobs:
             # Check WASM examples, which aren't a part of the workspace and so are otherwise missed:
             - name: Cargo check WASM examples
               run: |
-                cargo check --manifest-path examples/wasm-example/Cargo.toml
+                cargo check --manifest-path examples/wasm-example/Cargo.toml --target wasm32-unknown-unknown
```
jsdw, line 112: Nit: I think this should already be the case since `target: wasm32-unknown-unknown` means that it's the only target that will be installed, but nothing wrong with having some extra reassurance (I've done similar!)

jsdw (approved):

PR #1065, opened 2023-07-13
git: Increase light client test timeout to 25 mins

lexnv:
This has been filtered as flaky by our CI, increase the timeout to ensure the light client has proper time to sync with the chain and perform the tests.


jsdw (approved):

jsdw:
We'll have to keep an eye on this and see if there is something we can do to speed it up or whatever; isn't it wayyy faster on macs for some reason that we don't really understand?

tadeohepperle (approved):

PR #1067, opened 2023-07-13
Browser extension signing example

tadeohepperle:
I split the Yew WASM App into routes for the 2 examples.
This example constructs a `System->remark` call with a remark parsed from user input. 
We then let the user choose an account in the Talisman or PolkadotJS browser extension and sign the extrinsic in the extension. (Currently tested with the Dev version of the [polkadotjs extension](https://github.com/polkadot-js/extension))
The full `SubmittableExtrinsic` is then sent to a local node where we wait for it to be successful.

image
image
image
image
image
image

**What is missing?**
Right now `era`, `tip` and `mortality_checkpoint` of BaseExtrinsicParams are not used for the extrinsic and we use hardcoded values instead (no tip and immortal)


examples/wasm-example/src/main.rs
```diff
@@ -1,140 +1,59 @@
...
+mod routes;
 mod services;
 
-fn main() {
-    yew::Renderer::<SubxtExamplesComponent>::new().render();
+#[derive(Routable, PartialEq, Eq, Clone, Debug)]
+pub enum Route {
+    #[at("/fetching")]
+    Fetching,
+    #[at("/signing")]
```
jsdw, line 13: Yew is quite nice really; I like the built in browser history/routing stuff, and actually I generally find it quite easy to follow overall!

I wonder how the other frameworks differ; I should really try some of them at some point!

examples/wasm-example/Cargo.toml
```diff
@@ -11,3 +11,10 @@ subxt = { path = "../../subxt", default-features = false, features = ["jsonrpsee
 yew = {version = "0.20.0", features = ["csr"] }
 web-sys = "0.3.63"
 hex = "0.4.3"
+yew-router = "0.17.0"
+js-sys = "0.3.63"
+wasm-bindgen = "0.2.86"
+wasm-bindgen-futures = "0.4.36"
+anyhow = "1.0.71"
+serde = "1.0.163"
+serde_json = "1.0.96"
```
lexnv, line 20: nit: we could add a new line here

examples/wasm-example/index.js
```diff
@@ -0,0 +1,92 @@
+/// The `@polkadot/extension-dapp` package can be dynamically imported.
+/// Usually it is wise to use a package manager like npm or yarn to install it as a dependency.
+/// The `getPolkadotJsExtensionMod`
```
lexnv, line 3: I'm not super familiar with js docs, but is this line complete here? Or should we remove it?

subxt/src/config/extrinsic_params.rs
```diff
@@ -57,13 +57,20 @@ pub trait ExtrinsicParams<Hash>: Debug + 'static {
...
 #[derivative(Debug(bound = "Tip: Debug"))]
 pub struct BaseExtrinsicParams<T: Config, Tip: Debug> {
-    era: Era,
-    nonce: u64,
-    tip: Tip,
-    spec_version: u32,
-    transaction_version: u32,
-    genesis_hash: T::Hash,
-    mortality_checkpoint: T::Hash,
+    /// era
```
lexnv, line 60:
```suggestion
    /// Era defines how long the transaction will be valid for.
```

subxt/src/config/extrinsic_params.rs
```diff
@@ -57,13 +57,20 @@ pub trait ExtrinsicParams<Hash>: Debug + 'static {
...
+    pub nonce: u64,
+    /// tip
+    pub tip: Tip,
+    /// spec version
+    pub spec_version: u32,
+    /// transaction version
+    pub transaction_version: u32,
+    /// genesis hash of the chain
+    pub genesis_hash: T::Hash,
+    /// mortality checkpoint
```
lexnv, line 72:
```suggestion
    /// The block after which the transaction becomes valid.
```

subxt/src/config/extrinsic_params.rs
```diff
@@ -57,13 +57,20 @@ pub trait ExtrinsicParams<Hash>: Debug + 'static {
...
-    tip: Tip,
-    spec_version: u32,
-    transaction_version: u32,
-    genesis_hash: T::Hash,
-    mortality_checkpoint: T::Hash,
+    /// era
+    pub era: Era,
+    /// nonce (account nonce sent along an extrinsic, such that no extrinsic is submitted twice)
+    pub nonce: u64,
+    /// tip
```
lexnv, line 64:
```suggestion
    /// The tip you would like to give to the block author for this transaction.
    /// Note: Can be zero.
```
```suggestion
    /// tip
```

subxt/src/tx/tx_client.rs
```diff
@@ -318,7 +318,8 @@ where
 pub struct PartialExtrinsic<T: Config, C> {
     client: C,
     call_data: Vec<u8>,
-    additional_and_extra_params: T::ExtrinsicParams,
+    /// Extrinsic params (Additional and Extra)
+    pub additional_and_extra_params: T::ExtrinsicParams,
```
lexnv, line 322: Nit: Similarly to `call_data` we could create a getter for these params

```rust
/// Returns the additional and extra extrinsic params.
pub fn extra_params(&self) -> &T::ExtrinsicParams { .. }
```

examples/wasm-example/src/services.rs
```diff
@@ -71,3 +79,100 @@ pub(crate) async fn subscribe_to_finalized_blocks(
...
+}
+
+/// DTO to communicate with JavaScript
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Account {
+    /// account name
+    pub name: String,
+    /// name of the browser extension
+    pub source: String,
+    pub ty: String,
```
lexnv, line 98: nit: missing docs here

examples/wasm-example/src/services.rs
```diff
@@ -71,3 +79,100 @@ pub(crate) async fn subscribe_to_finalized_blocks(
...
+    format!("0x{}", hex::encode(bytes.as_ref()))
+}
+
+fn encode_to_hex<E: Encode>(input: &E) -> String {
+    format!("0x{}", hex::encode(input.encode()))
+}
+
+/// this is used because numeric types (e.g. u32) are encoded as little-endian via scale (e.g. 9430 -> d6240000)
+/// while we need a big-endian representation for the json (e.g. 9430 -> 000024d6).
+fn encode_to_hex_reverse<E: Encode>(input: &E) -> String {
```
lexnv, line 124: nit: `encoode_to_hex_big_endian`

examples/wasm-example/src/routes/signing.rs
```diff
@@ -0,0 +1,392 @@
...
+                        .system()
+                        .remark(self.message.as_bytes().to_vec());
+
+                    let api = self.online_client.as_ref().unwrap().clone();
+
+                    ctx.link()
+                        .send_future(
+                            async move {
+                                let partial_extrinsic =
+                                    match api.tx().create_partial_signed(&remark_call, &account_id, Default::default()).await {
```
jsdw, line 144: I'm a little wary of exposing `extrinsic_params` from `PartialExtrinsic` (and generally of exposing the `BaseExtrinsicParams` props). In part because I don't want us to be "committed" to those details in a public API (I like the idea that we can provide an interface to build extrinsic params etc, but don't think people should have any access to the internals so that we are free to change it, as we will be doing soon). 

I also don't think that it's necessary to expose them, because they have all been provided beforehand by the user or are available on the client anyway. So I'd do something more explicit, and gather each of the properties that you need, like:

```rust
let spec_version = api.runtime_version().spec_version;
let transaction_version = api.runtime_version().transaction_version;
let genesis_hash = api.genesis_hash();
let mortality_checkpoint = api.genesis_hash();
let era = subxt::config::Era::Immortal;
let method = api.tx().call_data(&remark_call)?;
let nonce = api.tx().account_nonce(&account_id).await?;
// ...
```

(Maybe eg `extension_signature_for_partial_extrinsic` takes in the client and call data and fetches the rest, or something)

subxt/src/config/extrinsic_params.rs
```diff
@@ -57,13 +57,20 @@ pub trait ExtrinsicParams<Hash>: Debug + 'static {
...
+    pub tip: Tip,
+    /// spec version
+    pub spec_version: u32,
+    /// transaction version
```
lexnv, line 68: Nit: could use a bit more details for these docs since we'd expose them to the end-user

jsdw:
Looks great to me, and it works well! 

I'd prefer that we don't expose any of those extra props in Subxt, and instead gather the information up from elsewhere; see my above comment :)

examples/wasm-example/src/routes/signing.rs
```diff
@@ -0,0 +1,392 @@
...
+    message: String,
+    remark_call_bytes: Vec<u8>,
+    online_client: Option<OnlineClient<PolkadotConfig>>,
+    stage: SigningStage,
+}
+
+impl SigningExamplesComponent {
+    /// # Panics
+    /// panics if self.online_client is None.
+    fn set_message(&mut self, message: String) {
```
lexnv, line 25: Could it be possible that `set_message` is called before the `OnlineClient` is created here?

I'm trying to think of the scenario where the light-client might take a bit to initialize while the user is eager to use the example.

If so, I wonder if we could warn the user somehow to try again later? Maybe a (red) message somewhere that the client is not initialized yet and to try again later?

lexnv: That makes sense, thanks for the info!

examples/wasm-example/src/services.rs
```diff
@@ -71,3 +80,105 @@ pub(crate) async fn subscribe_to_finalized_blocks(
...
+    let mut bytes = input.encode();
+    bytes.reverse();
+    format!("0x{}", hex::encode(bytes))
+}
```
jsdw, line 130: Yeah, it's really because we aren't scale encoding some of these numbers at all; we're just providing the hex representation of them. So really we should just `hex::encode` without the `input.encode()` bit :)

jsdw: Maybe the functions should be:

```rust
fn scale_encoded_hex<E: Encode>(input: &E) -> String { ... }
fn hex_encoded_bytes(input: &[u8]) -> String { ... }
```

And then `.to_be_bytes` or whatever on the numbers we pass to the latter?

jsdw (approved):
A couple of minor nits but assuming it still works when tested again, it looks great to me!

lexnv (approved):

PR #1074, opened 2023-07-17
Remove `Index` type from Config trait

### LINKED ISSUE #1061: Config: Remove Index param
One of the treickiest parts of getting started with Subxt is setting up the Config.

As a first step to simplifying it, can we remove the `Index` param? I believe it only exists so we know how to decode the value we get back from the runtime call, but I think we can be smarter and intelligently decode this value based on how many bytes it is. When we encode nonces for tx's it looks like we always encode to u64 anyway for that signed extension.

tadeohepperle:
fixes #1061 

Whereever we previously required a `Config::Index`, now any type that implements `Into<u64>` can be used.


subxt/src/config/extrinsic_params.rs
```diff
@@ -27,7 +27,7 @@ pub trait ExtrinsicParams<Index, Hash>: Debug + 'static {
     fn new(
         spec_version: u32,
         tx_version: u32,
-        nonce: Index,
+        nonce: impl Into<u64>,
```
jsdw, line 30: I'd keep this simple and just have it take a concrete `u64`; it's only called internally anyway, and so let's avoid extra generics unless they make the API more ergonomic for users :)

jsdw:
Looks good to me! I'd just remove the `impl Into<u64>` generics (it's easy for a user to provide a u64 off the bat; I don't see the point of making it generic really :)

jsdw:
Scrap that; @tadeohepperle pointed out that the nonce type is compact encoded in the signed extension, and so it's irrelevant what numeric type it actually is, so we're still good :)

lexnv (approved):

subxt/src/book/usage/transactions.rs
```diff
@@ -138,7 +138,7 @@
 //! // here, or can use `create_partial_signed` to fetch the correct nonce.
 //! let partial_tx = client.tx().create_partial_signed_with_nonce(
 //!     &payload,
-//!     0,
+//!     0u64,
```
jsdw, line 141: Nit: no more `u64` type needed now :)

jsdw (approved):
Looks great!

PR #1075, opened 2023-07-17
Prevent bug when reusing type ids in hashing

### LINKED ISSUE #1066: Metadata Validation: prevent reuse of `visited_type_ids`
We pass a `visited_type_ids` HashSet into a bunch of functions during validation of metadata. The point of this is to spot when we are recursing into the same type ID we've already seen, and insetad of getting stuck in an infintie loop, returning some "magic" hash.

The problem is that we reuse this `visited_type_ids` when hashing multiple types in some cases, so eg if we hash the types:

```rust
enum Foo {
    A(Bar, Wibble),
}

enum Bar { ... }

enum Wibble { ... }
```

We end up with a hashing approach something like:

```rust
hash("Foo") + 
    xor(
        hash("A") + hash(Bar) + hash(Wibble)
    )

hash("Bar") == RECURSIVE_HASH

hash("Wibble") == RECURSIVE_HASH
```

Ie, we hash `Foo` without any issues, but then when we try to hash `Bar` and `Wibble` reusing `visited_type_ids`, we have seen both types already and so both get replaced with some constant `RECURSIVE_HASH` to avoid infinite loops.

The solution is to ensure that `visited_ids` is not reused across type hashings at all.

One way to help ensure this is to make it so that `get_type_hash` does _not_ take `visited_ids` as an argument and instead creates it internally. It then internally will call `get_type_def_hash` and whatever by passing this new `visited_ids` in, and anything needing to recusviely call `get_type_hash` again can call some inner `get_type_hash_recursive` call or something which _does_ accept `visited_ids` and is not exposed/available for anything else to call.

With that, every attempt to hash a type will atuomatically use a new internal "recursive ids set" and we'll prevent accidental re-use.

See https://github.com/paritytech/subxt/pull/1041#discussion_r1262610682 for a bit of context

tadeohepperle:
fixes #1066.

I first thought a simple restructuring approach like @jsdw mentioned in #1066 would work, but correct me if I'm wrong, I think it cannot solve the recursive type problem, because creating a new HashSet in `get_type_id()` will eventually run into an infinite loop. 

Instead of using a `HashSet` for keeping track of the visited ids, we can use a `HashMap` that saves some info about if the type has been visited yet and also if a complete hash has been calculated yet.

I added a test to make sure it works (Without my code changes, it fails).


jsdw:
> because creating a new HashSet in get_type_id() will eventually run into an infinite loop.

What I meant was that `get_type_id` could be exposed to the rest of the module to get type IDs, and this would internally create a new `HashSet` or whatever.

And then "internally" (and not exposed to the rest of the crate) one could use a `get_type_id_with_cache` or whatever that looks like the current one does.

Maybe some stuff would end up in a different module or something to support that.

And the point of doing this would just be to guarantee that `get_type_id` always starts from a fresh state and doesn't re-use anything from other invocations which could change the result or make it worse :)

But anyways, maybe that's irrelevant now and I'll have a gander at the approach you took!

metadata/src/utils/validation.rs
```diff
@@ -195,31 +190,53 @@ fn get_type_def_hash(
...
+    // Values in the cache can be thought of as a mapping like this:
+    // type id -> not contained           if not seen yet
+    //         -> 0                       if within get_type_hash() calculation for the type
+    //         -> (pseudo-)random u32     if get_type_hash() calculation has completed for the type
+
+    if let Some(short_hash) = cache.get(&id) {
+        return hash(&short_hash.to_be_bytes());
     }
-
+    cache.insert(id, 0);
```
jsdw, line 212: I like the idea of caching the hashes like this in general!

There's maybe a small chance that the first 4 bytes of `type_hash` are actually `[0,0,0,0]`, which would make that type look like the special recursive type that leads to an early return.

I also wonder; is it better to store the full `[u8; HASH_LEN]` type hash in the cache? We use more memory with this, but then don't need to `hash` it, and it means that the cache is truly caching the computed hashes for various types rather than caching something that's not actually the original hash we computed for a type.

Combining the two thoughts above, I wonder whether the cache should look like:

```rust
type Cache = HashMap<u32, CachedValue>;

enum CachedValue {
    Recursive,
    Hash([u8; HASH_LEN])
}
```

Where we insert `CachedValue::Recursive` initially and update that to a final hash when done. (if `get_type_hash` looks in the cache and sees a `CachedValue::Recursive` it can return with eg `hash(MAGIC_RECURSIVE_HASH_VALUE)`, which is functionally the same as right now where it would return with `hash(0)`).

I'm suggesting a new enum above over using `Option` just because it would be a little more self-documenting, but either works really!

tadeohepperle: We could definitely do that, thought about something like that as well, it is more secure at the cost of a little bit more memory. I will change it! :)

jsdw: That'd be great, and then perhaps also we could add a test which uses the cache across a couple of invocations of hashing the same type to comfirm that indeed you get the same hash back both times (first time caches the hash, second time just returns it straight off). This should even work if the type has a recursive definition in which is cool :)

metadata/src/utils/validation.rs
```diff
@@ -743,6 +768,36 @@ mod tests {
         assert_eq!(hash, hash_swap);
     }
 
+    #[test]
```
jsdw, line 771: It'd also be good now to add a test wherein you basically hash the same type ID twice (or a few times), reusing the same cache for both. The hashes should be identical each time, showing that the cache is truly caching the type hashes and not something else :)

jsdw (approved):
This looks great! Just one tiny suggestion to add a test showing that the cache is cahcing actual hashes and thus can safely be reused across multiple calls to hash the same types or whatever, but otherwise looks good to me!

metadata/src/utils/validation.rs
```diff
@@ -798,6 +798,25 @@ mod tests {
...
+        let mut cache = HashMap::new();
+
+        let a_hash = get_type_hash(&registry, a_type_id, &mut cache);
+        let b_hash = get_type_hash(&registry, b_type_id, &mut cache);
+
+        let CachedHash::Hash(a_cache_hash) = cache[&a_type_id] else { panic!() };
+        let CachedHash::Hash(b_cache_hash) = cache[&b_type_id] else { panic!() };
+
+        assert_eq!(a_hash, a_cache_hash);
+        assert_eq!(b_hash, b_cache_hash);
```
jsdw, line 817: Can we also check that `a_hash == b_hash` :)

tadeohepperle: I don't think we want the hashes to be the same, because `A` and `B` are different structs, with different fields:
```
struct A { pub b: Box<B> }
struct B { pub a: Box<A> }
```
Maybe I understood you wrong, what did you mean?

jsdw: Ah, sorry yeah, I mean, it'd be nice to have a test like:

```rust
let a_type_id = registry.register_type(&meta_type::<A>()).id;

let a_hash1 = get_type_hash(&registry, a_type_id, &mut cache);
let a_hash2 = get_type_hash(&registry, a_type_id, &mut cache);

assert_eq!(a_hash1, a_hash2);
```

ie, we want to check that sharing the cache across different uses doesn't change the hashes that are produced for some type.

jsdw (approved):

lexnv (approved):

PR #1076, opened 2023-07-19
runtime API: Substitute `UncheckedExtrinsic` with custom encoding

### LINKED ISSUE #14584: Bad metadata for TransactionPaymentApi.query_info call in Metadata v15.
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Description of bug

While working on the Swift SDK I found that the new RuntimeApi metadata has inconsistency for the `TransactionPaymentApi.query_info` call (and `query_fee_details` call too).

Right now the call is described as:
```
TransactionPaymentApi_query_info {
    uxt: sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic(Vec<u8>),
    len: u32
}
```
`UncheckedExtrinsic` has the `sequence` type inside, which should be encoded with a size prefix. But in reality, call expects `uxt` to be encoded without prefix, as fixed data. This inconsistency prevents the dynamic encoding of this runtime call.

### Steps to reproduce

_No response_

lexnv:
The `substrate::UncheckedExtrinsic` type is substituted in codgen with a local type that implements custom encoding.

The old `UncheckedExtrinsic(pub Vec<u8>)` would re-encode the inner bytes of the extrinsic leading to a length prefix byte. This payload is then rejected by substrate WASM code, since the `TransactionPaymentApi` expects the bytes without the length prefix.

The following type was generated by the codegen:

```rust
pub struct UncheckedExtrinsic<_0, _1, _2, _3>(
    pub ::std::vec::Vec<::core::primitive::u8>,
    #[codec(skip)] pub ::core::marker::PhantomData<(_1, _0, _2, _3)>,
);
```

The type could not be passed to the runtime API because the encoding would expect only a sequence of data.
In contrast, this type is deduced to be a tuple. 

```bash
Error: Encode(Error { context: Context { path: [Location { inner: Field("uxt") }] }, kind: WrongShape { actual: Tuple, expected: 13 } })
```

- Add a new type for `UncheckedExtrinsic` that is capable of passing through the provided bytes without any further encoding
- Substitute `substrate::UncheckedExtrinsic` with our custom `UncheckedExtrinsic`
- Validate a raw runtime API call with the statically generated API from codegen

Closes: https://github.com/paritytech/substrate/issues/14584

### DIFF
--- a/subxt/src/utils/mod.rs
+++ b/subxt/src/utils/mod.rs
@@ -9,6 +9,7 @@ pub mod bits;
 mod multi_address;
 mod multi_signature;
 mod static_type;
+mod unchecked_extrinsic;
 mod wrapper_opaque;
 
 use codec::{Compact, Decode, Encode};
@@ -18,6 +19,7 @@ pub use account_id::AccountId32;
 pub use multi_address::MultiAddress;
 pub use multi_signature::MultiSignature;
 pub use static_type::Static;
+pub use unchecked_extrinsic::UncheckedExtrinsic;
 pub use wrapper_opaque::WrapperKeepOpaque;
 
 // Used in codegen
--- a/subxt/src/utils/mod.rs
+++ b/subxt/src/utils/mod.rs
@@ -28,7 +28,7 @@ pub use primitive_types::{H160, H256, H512};
 
 /// Wraps an already encoded byte vector, prevents being encoded as a raw byte vector as part of
 /// the transaction payload
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Decode)]
 pub struct Encoded(pub Vec<u8>);
 
 impl codec::Encode for Encoded {
--- a/subxt/src/utils/unchecked_extrinsic.rs
+++ b/subxt/src/utils/unchecked_extrinsic.rs
@@ -114,7 +114,8 @@ pub mod tests {
 
     #[test]
     fn unchecked_extrinsic_encoding() {
-        let tx_bytes = vec![1, 2, 3];
+        // A tx is basically some bytes with a compact length prefix; ie an encoded vec:
+        let tx_bytes = vec![1u8, 2, 3].encode();
 
         let unchecked_extrinsic = UncheckedExtrinsic::<(), (), (), ()>::new(tx_bytes.clone());
         let encoded_tx_bytes = unchecked_extrinsic.encode();
@@ -124,12 +125,12 @@ pub mod tests {
 
         // However, for decoding we expect to be able to read the extrinsic from the wire
         // which would be length prefixed.
-        let wire_bytes = tx_bytes.encode();
-        let decoded_tx =
-            UncheckedExtrinsic::<(), (), (), ()>::decode(&mut &wire_bytes[..]).unwrap();
+        let decoded_tx = UncheckedExtrinsic::<(), (), (), ()>::decode(&mut &tx_bytes[..]).unwrap();
+        let decoded_tx_bytes = decoded_tx.bytes();
         let encoded_tx_bytes = decoded_tx.encode();
 
-        assert_eq!(decoded_tx.bytes(), encoded_tx_bytes);
-        assert_eq!(tx_bytes, encoded_tx_bytes);
+        assert_eq!(decoded_tx_bytes, encoded_tx_bytes);
+        // Ensure we can decode the tx and fetch only the tx bytes.
+        assert_eq!(vec![1, 2, 3], encoded_tx_bytes);
     }
 }
--- a/subxt/src/utils/unchecked_extrinsic.rs
+++ b/subxt/src/utils/unchecked_extrinsic.rs
@@ -1,0 +1,91 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! The "default" Substrate/Polkadot UncheckedExtrinsic.
+//! This is used in codegen for runtime API calls.
+//!
+//! The inner bytes represent the encoded extrinsic expected by the
+//! runtime APIs. Deriving `EncodeAsType` would lead to the inner
+//! bytes to be re-encoded (length prefixed).
+
+use std::marker::PhantomData;
+
+use codec::Decode;
+use scale_decode::{visitor::DecodeAsTypeResult, IntoVisitor, Visitor};
+
+/// The unchecked extrinsic from substrate.
+#[derive(Clone, Debug, Eq, PartialEq, Decode)]
+pub struct UncheckedExtrinsic<Address, Call, Signature, Extra>(
+    pub Vec<u8>,
+    #[codec(skip)] pub PhantomData<(Address, Call, Signature, Extra)>,
+);
+
+impl<Address, Call, Signature, Extra> codec::Encode
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn encode(&self) -> Vec<u8> {
+        self.0.to_owned()
+    }
+}
+
+impl<Address, Call, Signature, Extra> scale_encode::EncodeAsType
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn encode_as_type_to(
+        &self,
+        _type_id: u32,
+        _types: &scale_info::PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), scale_encode::Error> {
+        out.extend_from_slice(&self.0);
+        Ok(())
+    }
+}
+
+impl<Address, Call, Signature, Extra> scale_encode::EncodeAsFields
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn encode_as_fields_to(
+        &self,
+        _fields: &mut dyn scale_encode::FieldIter<'_>,
+        _types: &scale_info::PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), scale_encode::Error> {
+        out.extend_from_slice(&self.0);
+        Ok(())
+    }
+}
+
+pub struct UncheckedExtrinsicDecodeAsTypeVisitor<Address, Call, Signature, Extra>(
+    PhantomData<(Address, Call, Signature, Extra)>,
+);
+
+impl<Address, Call, Signature, Extra> Visitor
+    for UncheckedExtrinsicDecodeAsTypeVisitor<Address, Call, Signature, Extra>
+{
+    type Value<'scale, 'info> = UncheckedExtrinsic<Address, Call, Signature, Extra>;
+    type Error = scale_decode::Error;
+
+    fn unchecked_decode_as_type<'scale, 'info>(
+        self,
+        input: &mut &'scale [u8],
+        _type_id: scale_decode::visitor::TypeId,
+        _types: &'info scale_info::PortableRegistry,
+    ) -> DecodeAsTypeResult<Self, Result<Self::Value<'scale, 'info>, Self::Error>> {
+        use scale_decode::{visitor::DecodeError, Error};
+        let decoded = UncheckedExtrinsic::decode(input)
+            .map_err(|e| Error::new(DecodeError::CodecError(e).into()));
+        DecodeAsTypeResult::Decoded(decoded)
+    }
+}
+
+impl<Address, Call, Signature, Extra> IntoVisitor
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    type Visitor = UncheckedExtrinsicDecodeAsTypeVisitor<Address, Call, Signature, Extra>;
+
+    fn into_visitor() -> Self::Visitor {
+        UncheckedExtrinsicDecodeAsTypeVisitor(PhantomData)
+    }
+}
--- a/subxt/src/utils/unchecked_extrinsic.rs
+++ b/subxt/src/utils/unchecked_extrinsic.rs
@@ -40,12 +40,11 @@ impl<Address, Call, Signature, Extra> scale_encode::EncodeAsType
 {
     fn encode_as_type_to(
         &self,
-        _type_id: u32,
-        _types: &scale_info::PortableRegistry,
+        type_id: u32,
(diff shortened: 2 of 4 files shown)


subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub Vec<u8>,
+    #[codec(skip)] pub PhantomData<(Address, Call, Signature, Extra)>,
+);
+
+impl<Address, Call, Signature, Extra> codec::Encode
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn encode(&self) -> Vec<u8> {
+        self.0.to_owned()
+    }
```
jsdw, line 29: Nit: Better to impl `encode_to` I think, which is given a buffer to write to; this will save cloning the vec in the event that the caller already has a buffer :)

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn encode_as_fields_to(
+        &self,
+        _fields: &mut dyn scale_encode::FieldIter<'_>,
+        _types: &scale_info::PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), scale_encode::Error> {
+        out.extend_from_slice(&self.0);
+        Ok(())
+    }
+}
```
jsdw, line 58: I'm not sure it makes sense to have this impl, since we can't sensible encode the thing to any set of fields (and the target type is an "UncheckedExtrinsic" type anyway and not some fields.

Does everything work ok without it? If so let's remove it :)

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use codec::Decode;
+use scale_decode::{visitor::DecodeAsTypeResult, IntoVisitor, Visitor};
+
+/// The unchecked extrinsic from substrate.
+#[derive(Clone, Debug, Eq, PartialEq, Decode)]
+pub struct UncheckedExtrinsic<Address, Call, Signature, Extra>(
+    pub Vec<u8>,
+    #[codec(skip)] pub PhantomData<(Address, Call, Signature, Extra)>,
+);
```
jsdw, line 22: Potentially we could reuse some other code more by having this be:

```rust
pub struct UncheckedExtrinsic<Address, Call, Signature, Extra>{
    encoded: Static<Encoded>,
    #[codec(skip)] pub PhantomData<(Address, Call, Signature, Extra)>,
};
```

Then, all of the impls could delegate to the `Static<Encoded>` type, which I think already does what we want it to.

In any case though, let's give this a nicer interface, something like:

```rust
// make a new one:
UncheckedExtrinsic::new(bytes)
// can convert bytes into this type easily:
impl From<Vec<u8>> for UncheckedExtrinsic
```

Hopefully type inference will sort out the generic params when it's used, ie

```rust
let runtime_api_call = node_runtime::apis()
        .transaction_payment_api()
        .query_fee_details(tx_bytes.into(), len);
```

Will hopefully Just Work, and infer all of the types properly!

subxt/src/utils/unchecked_extrinsic.rs
jsdw: (I note that `Encoded` impls `encode` and not `encode_to` so that could be improved)

testing/integration-tests/src/full_client/runtime_api/mod.rs
```diff
@@ -47,3 +49,57 @@ async fn account_nonce() -> Result<(), subxt::Error> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn unchecked_extrinsic_encoding() -> Result<(), subxt::Error> {
```
jsdw, line 54: I like the test; great to check that this works :)

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -38,7 +68,7 @@ impl<Address, Call, Signature, Extra> scale_encode::EncodeAsType
         _types: &scale_info::PortableRegistry,
         out: &mut Vec<u8>,
     ) -> Result<(), scale_encode::Error> {
-        out.extend_from_slice(&self.0);
+        out.extend_from_slice(self.bytes());
```
jsdw: Or, `self.0.encode_to(type_id, types, out)` to fully defer to the internal impl :)

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -14,18 +14,48 @@ use std::marker::PhantomData;
...
+        bytes.0 .0 .0
+    }
+}
+
 impl<Address, Call, Signature, Extra> codec::Encode
     for UncheckedExtrinsic<Address, Call, Signature, Extra>
 {
     fn encode_to<T: codec::Output + ?Sized>(&self, dest: &mut T) {
-        dest.write(&self.0)
+        dest.write(self.bytes())
```
jsdw: Or `self.0.encode_to(dest)` to fully defer to the internal impl (no space saved but one less place to check impl for in the future)

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,85 @@
...
+        input: &mut &'scale [u8],
+        _type_id: scale_decode::visitor::TypeId,
+        _types: &'info scale_info::PortableRegistry,
+    ) -> DecodeAsTypeResult<Self, Result<Self::Value<'scale, 'info>, Self::Error>> {
+        use scale_decode::{visitor::DecodeError, Error};
+        let decoded = UncheckedExtrinsic::decode(input)
+            .map_err(|e| Error::new(DecodeError::CodecError(e).into()));
+        DecodeAsTypeResult::Decoded(decoded)
+    }
+}
```
jsdw, line 75: Aah, I was hoping that you could get rid of more of this by using `Static<Encoded>` instead!

I wonder whether something like this would work to lean on the inner impl better (while not making any assumptions about it):

```rust
impl<Address, Call, Signature, Extra> Visitor
    for UncheckedExtrinsicDecodeAsTypeVisitor<Address, Call, Signature, Extra>
{
    type Value<'scale, 'info> = UncheckedExtrinsic<Address, Call, Signature, Extra>;
    type Error = scale_decode::Error;

    fn unchecked_decode_as_type<'scale, 'info>(
        self,
        input: &mut &'scale [u8],
        type_id: scale_decode::visitor::TypeId,
        types: &'info scale_info::PortableRegistry,
    ) -> DecodeAsTypeResult<Self, Result<Self::Value<'scale, 'info>, Self::Error>> {
        DecodeAsTypeResult::Decoded(self.0.decode_as_type(input, type_id.0, types))
    }
}
```

Maybe this is a sign that we can do something better with the `DecodeAsType` macro to allow things to be decoded based on some inner type or something.

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use std::marker::PhantomData;
+
+use codec::{Decode, Encode};
+use scale_decode::{visitor::DecodeAsTypeResult, IntoVisitor, Visitor};
+
+use super::{Encoded, Static};
+
+/// The unchecked extrinsic from substrate.
+#[derive(Clone, Debug, Eq, PartialEq, Encode, Decode, scale_encode::EncodeAsType)]
```
jsdw, line 20: Ah alas we need to decalre it ourselves because the derive `EncodeAsType` impl will try to be smarter and make sure it's encoding this whole thing as a tuple, where we just want to defer to the `Static` impl  and ignore this surrounding type entirely (this is why the test is failing)

lexnv: Yep exactly, we could at least remove the manual `Encode` part. I do wonder, could we make the `EncodeAsType` somehow ignore any fields that have `#[codec(skip)]` on them (similar to the codec)? 🤔

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,96 @@
...
+impl<Address, Call, Signature, Extra> scale_encode::EncodeAsType
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn encode_as_type_to(
+        &self,
+        _type_id: u32,
+        _types: &scale_info::PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), scale_encode::Error> {
+        self.0.encode_to(out);
```
jsdw, line 47: I still think that delegating to the inner impl is better overall, because then we never have to care/worry about what this impl does :)

eg

```rust
self.0.encode_as_type_to(type_id, types, out)
```

subxt/src/utils/unchecked_extrinsic.rs
jsdw: And good idea on using `Encode` and `Decode` derives; that helps!

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl<Address, Call, Signature, Extra> UncheckedExtrinsic<Address, Call, Signature, Extra> {
+    /// Construct a new [`UncheckedExtrinsic`].
+    pub fn new(bytes: Vec<u8>) -> Self {
+        Self(Static(Encoded(bytes)), PhantomData)
+    }
+
+    /// Get the bytes of the encoded extrinsic.
+    pub fn bytes(&self) -> &[u8] {
+        self.0 .0 .0.as_slice()
```
tadeohepperle, line 34: This formatting looks so weird, never seen that before :)

tadeohepperle (approved):
Looks good to me!

subxt/src/utils/unchecked_extrinsic.rs
lexnv: Same, I used to write it without spaces but I've changed my editor to run `rustfmt` for each file save which changed it like so :D

subxt/src/utils/mod.rs
```diff
@@ -26,7 +28,7 @@ pub use primitive_types::{H160, H256, H512};
 
 /// Wraps an already encoded byte vector, prevents being encoded as a raw byte vector as part of
 /// the transaction payload
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Decode)]
```
jsdw, line 31: Ah hmmm, I'm wondering about this automatic `Decode`, because:

- decoding like this will first decode a Compact length for the vec length and then put the relevant number of butes into the vec.
- encoding is not the opposite of this; it will just encode the bytes in the vec to the target; no compact length or whatever at the beginning.

Having a look, an extrinsic is actually just a compact encoded length and then that number of bytes.

So probably you were right in a previous iteration, and `UncheckedExtrinsic` should have a custom `Decode` impl that will decode the compact length and then the relevant number of bytes following it all into the `Static<Encoded>` struct, which will then encode (from the impl below) precisely those bytes back.

jsdw:
All looks great; I think we just need to put back a manual impl of `Decode` on `UncheckedExtrinsic` which decodes the compact length and then following bytes and puts all of those bytes in a vec.

Conceptually something like:

```rust
fn decode(bytes) {
    // the exttrinsic is basically identical to the encoded form of this:
    let xt_vec = <Vec<u8>>:decode(&mut bytes)?;
    UncheckedExtrinsic(Static(Encoded(xt_vec.encode())))
}
``` 

We could save the extra vec allocation by manually decoding the compact length etc, but maybe it's easier just to do the above?

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -35,6 +35,15 @@ impl<Address, Call, Signature, Extra> UncheckedExtrinsic<Address, Call, Signatur
     }
 }
 
+impl<Address, Call, Signature, Extra> Decode
+    for UncheckedExtrinsic<Address, Call, Signature, Extra>
+{
+    fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+        let xt_vec: Vec<u8> = Decode::decode(input)?;
```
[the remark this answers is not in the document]
jsdw: (is it worth adding a quick test just to check that encoding->decoding->encoding an UncheckedExtrinsic gives the same bytes, to confirm this impl is doing what we want?)

jsdw (approved):
Nice work!

subxt/src/utils/unchecked_extrinsic.rs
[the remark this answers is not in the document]
lexnv: Have added a test for this to check that:
- encoded format is the same as provided bytes of the extrinsic
- for decoding we want to be able to construct this from over the wire data, we decode both compact<length> and tx_bytes, while keeping just the tx_bytes 

Thanks for the suggestion! 🙏

subxt/src/utils/unchecked_extrinsic.rs
```diff
@@ -107,3 +107,29 @@ impl<Address, Call, Signature, Extra> IntoVisitor
...
+
+        // However, for decoding we expect to be able to read the extrinsic from the wire
+        // which would be length prefixed.
+        let wire_bytes = tx_bytes.encode();
+        let decoded_tx =
+            UncheckedExtrinsic::<(), (), (), ()>::decode(&mut &wire_bytes[..]).unwrap();
+        let encoded_tx_bytes = decoded_tx.encode();
+
+        assert_eq!(decoded_tx.bytes(), encoded_tx_bytes);
+        assert_eq!(tx_bytes, encoded_tx_bytes);
```
jsdw, line 133: Nit: 

I think this does what I'm hoping but it's a touch hard for me to follow (this stuff is in general because we are talking about encoding vec-like bytes into another vec :D)

The test could start with:

```rust
// A tx is basically some bytes with a compact length prefix; ie an encoded vec:
let tx_bytes =  vec![1,2,3].encode();
```

And then we have a "valid" TX format that can be used for all the tests, rather than needing a `let wire_bytes = tx_bytes.encode();` to make it valid for the second half :)

jsdw (approved):

subxt/src/utils/unchecked_extrinsic.rs
jsdw: I think rustfmt is just being a bit naff here, but ah well :)

PR #1078, opened 2023-07-20
Remove `substrate-compat` default feature flag

### LINKED ISSUE #1077: Remove "substrate-compat" default feature flag
The flags we have enabled by default are ["jsonrpsee", "native", "substrate-compat"].

I'd very much like to remove "substrate-compat" from this default list; it leads to various issues when substrate crates are out of date (common atm since they aren't being published again yet), and pulls in a load of dependencies that aren't needed so much now we have `subxt-signer`.

So, can we remove it, and feature flag the code/examples that depend on it (and make sure we still do test those in CI)?

I recently found a handy trick for doc examples; you can add eg this around an example to make compiling it conditional on a feature flag:

```rust
//! ```rust
//! # // don't compile when feature flag not present:
//! # #[cfg(not(feature = "foo"))]
//! # fn main() {}
//! #
//! # #[cfg(feature = "foo")]
//! # fn main() {
//! 
//! // example requiring feature goes here
//!
//! # }
```

Not super pretty but I think there's only one/two doc examples using "substrate-compat" :)

tadeohepperle:
fixes #1077 

I added `target_arch = "wasm32"` to the `jsonrpsee_helpers` mod because otherwise `jsonrpsee` does not expose the `client_transport::web` module.


.github/workflows/rust.yml
```diff
@@ -271,8 +275,8 @@ jobs:
                   # `listen-addr` is used to configure p2p to accept websocket connections instead of TCP.
                   # `node-key` provides a deterministic p2p address.
                   substrate --dev --node-key 0000000000000000000000000000000000000000000000000000000000000001 --listen-addr /ip4/0.0.0.0/tcp/30333/ws > /dev/null 2>&1 &
-                  wasm-pack test --headless --firefox
-                  wasm-pack test --headless --chrome
+                  wasm-pack test --no-default-features --features jsonrpsee,web --headless  --firefox 
+                  wasm-pack test --no-default-features --features jsonrpsee,web --headless --chrome
```
jsdw, line 279: These should be unnecessary shouldn't they? The `wasm-rpc-tests` crate doesn't expose any features, and so this should have no effect.

subxt/src/book/usage/transactions.rs
```diff
@@ -77,40 +77,43 @@
...
-//! let keypair = PairSigner::<PolkadotConfig,_>::new(keypair);
-//! #
-//! # // Test that these all impl Signer trait while we're here:
-//! #
-//! # fn is_subxt_signer(_signer: impl subxt::tx::Signer<PolkadotConfig>) {}
-//! # is_subxt_signer(subxt_signer::sr25519::dev::alice());
-//! # is_subxt_signer(subxt_signer::ecdsa::dev::alice());
-//! # is_subxt_signer(PairSigner::<PolkadotConfig,_>::new(sp_keyring::AccountKeyring::Alice.pair()));
+//! #[cfg(feature = "substrate-compat")]
+//! {
```
jsdw, line 81: Ooh, putting the feature on a block is much cleaner than my suggesiton; good idea!

I'd suggest hiding these lines though (with a `#` prefix) and not indenting the rest, since (a) most of it doesn't require that feature and (b) we talk about needing it above anyway, so keeping the example code neater feels better :)

subxt/Cargo.toml
```diff
@@ -17,7 +17,7 @@ keywords = ["parity", "substrate", "blockchain"]
 [features]
 # For dev and documentation reasons we enable more features than are often desired.
 # it's recommended to use `--no-default-features` and then select what you need.
-default = ["jsonrpsee", "native", "substrate-compat"]
+default = ["jsonrpsee", "native"]
```
jsdw, line 20: 🎉

.github/workflows/rust.yml
```diff
@@ -60,6 +60,10 @@ jobs:
             - name: Cargo check all targets.
               run: cargo check --all-targets
 
+            # A basic check for WASM target specifically
+            - name: Cargo check WASM target
+              run: cargo check --no-default-features --features jsonrpsee,web --target wasm32-unknown-unknown
```
jsdw, line 65: This will need to be in a separate wasm job so that the `wasm32-unknown-unknown` target can be installed (I checked recently and you can't specify multi ple targets in the toolchain action above at the mo :(

eg, I added this in the parachain-example PR so you could copy and tweak this:

```
    wasm_check:
        name: Cargo check (WASM)
        runs-on: ubuntu-latest
        steps:
            - name: Checkout sources
              uses: actions/checkout@v3

            - name: Install Rust stable toolchain
              uses: actions-rs/toolchain@v1
              with:
                  profile: minimal
                  toolchain: stable
                  target: wasm32-unknown-unknown
                  override: true

            - name: Rust Cache
              uses: Swatinem/rust-cache@dd05243424bd5c0e585e4b55eb2d7615cdd32f1f # v2.5.1

            # Check WASM examples, which aren't a part of the workspace and so are otherwise missed:
            - name: Cargo check WASM examples
              run: |
                cargo check --manifest-path examples/wasm-example/Cargo.toml
```

subxt/src/book/usage/transactions.rs
jsdw: @tadeohepperle this is the only thing left that I'd like to see done :)

subxt/src/book/usage/transactions.rs
```diff
@@ -77,43 +77,46 @@
...
+//!         .expect("valid URI");
+//!     let keypair = PairSigner::<PolkadotConfig,_>::new(keypair);
+//!     #
+//!     # // Test that these all impl Signer trait while we're here:
+//!     #
+//!     # fn is_subxt_signer(_signer: impl subxt::tx::Signer<PolkadotConfig>) {}
+//!     # is_subxt_signer(subxt_signer::sr25519::dev::alice());
+//!     # is_subxt_signer(subxt_signer::ecdsa::dev::alice());
+//!     # is_subxt_signer(PairSigner::<PolkadotConfig,_>::new(sp_keyring::AccountKeyring::Alice.pair()));
+//! }
```
jsdw, line 116:
```suggestion
//! # #[cfg(feature = "substrate-compat")]
//! # {
//! use subxt::config::PolkadotConfig;
//! use std::str::FromStr;
//!
//! //// 1. Use a `subxt_signer` impl:
//! use subxt_signer::{ SecretUri, sr25519 };
//!
//! // Get hold of a `Signer` for a test account:
//! let alice = sr25519::dev::alice();
//!
//! // Or generate a keypair, here from an SURI:
//! let uri = SecretUri::from_str("vessel ladder alter error federal sibling chat ability sun glass valve picture/0/1///Password")
//!     .expect("valid URI");
//! let keypair = sr25519::Keypair::from_uri(&uri)
//!     .expect("valid keypair");
//!
//! //// 2. Use the corresponding `sp_core::Pair` impl:
//! use subxt::tx::PairSigner;
//! use sp_core::Pair;
//!
//! // Get hold of a `Signer` for a test account:
//! let alice = sp_keyring::AccountKeyring::Alice.pair();
//! let alice = PairSigner::<PolkadotConfig,_>::new(alice);
//!
//! // Or generate a keypair, here from an SURI:
//! let keypair = sp_core::sr25519::Pair::from_string("vessel ladder alter error federal sibling chat ability sun glass valve picture/0/1///Password", None)
//!     .expect("valid URI");
//! let keypair = PairSigner::<PolkadotConfig,_>::new(keypair);
//! #
//! # // Test that these all impl Signer trait while we're here:
//! #
//! # fn is_subxt_signer(_signer: impl subxt::tx::Signer<PolkadotConfig>) {}
//! # is_subxt_signer(subxt_signer::sr25519::dev::alice());
//! # is_subxt_signer(subxt_signer::ecdsa::dev::alice());
//! # is_subxt_signer(PairSigner::<PolkadotConfig,_>::new(sp_keyring::AccountKeyring::Alice.pair()));
//! # }
```

jsdw (approved):
Looks good to me!

lexnv (approved):
LGTM!

PR #1079, opened 2023-07-21
Storage: Support iterating over NMaps with partial keys

### LINKED ISSUE #1044: Storage: Support iterating over NMaps with partial keys
Currently, when accessing storage maps (ie storage entries that require some key to be provided) we have two options:

```rust
// 1. provide _all_ of the keys to get a single value back:
runtime::storage().pallet().entry(key1, key2, ...);

// 2. provided no keys. This allows iterating over _everything_:
runtime::storage().pallet().entry_root();
```

When more than one key is needed, it'd be nice if people could provide any subset of the required keys to be able to iterate over all of the remaining values. In other words it'd be nice if the API looked something like:

```rust
// 1. provide _all_ of the keys to get a single value back:
runtime::storage().pallet().entry(key1, key2, key3);

// 2. provide some subset of keys. All of these allow iterating over results beneath them:
runtime::storage().pallet().entry_iter0();
runtime::storage().pallet().entry_iter1(key1);
runtime::storage().pallet().entry_iter2(key1, key2);
```

Here, `entry` would be `Fetchable` but not `Iterable`, and all of the `_iter` functions would be Iterable but not `Fetchable`.

Right now, one can call `.to_root_key()` on some entry to be able to iterate over the root things. Additionally, something like `runtime::storage().pallet().entry(key1, key2, key3)` is `Iterable` (iteration calls `.to_root_key()`). I'd propose scrapping both of these, ie:

- Iterating over a key should never call `.to_root_key()`; it'll just iterate whatever was given.
- If you provide every key eg `runtime::storage().pallet().entry(key1, key2, key3);`, you cannot iterate over this. Instead you'll have to construct a new entry with as many partial keys as you want to iterate, so it's explicit what you're actually iterating over.

See #1038 for the original inspiration.

tadeohepperle:
fixes #1044 

Changes to the Codegen to support iterating over partial keys.

Instead of 
```rs
runtime::storage().pallet().entry(key1, key2, key3); // fetchable and iterable (redirects to entry_root())
runtime::storage().pallet().entry_root(); // iterable
```
We now have:
```rs
runtime::storage().pallet().entry(key1, key2, key3); // only fetchable
runtime::storage().pallet().entry_iter(); // only iterable
runtime::storage().pallet().entry_iter1(key1); // only iterable
runtime::storage().pallet().entry_iter2(key1, key2); // only iterable
```


subxt/examples/storage_iterating_partial.rs
jsdw, line 1: Just a reminder that any new examples should be linked from the book somewhere (this ensures that they are "valuable" enough to warrant having in the first place, but also means they will be discovered when people read the relevant sections).

Probably the book should be tweaked anyway; it says:

```rust
//! As well as accessing specific entries, some storage locations can also be iterated over (such as
//! the map of account information). To do this, suffix `_root` onto the query constructor 
```

And  possibly at that point we should mention that you can iterate with `_iter`, and for maps with multiple keys, we can `_iter2`, `_iter3` etc to iterate with any number of the map keys. Then, probably we can just have an inline code example showing this (if there are any good examples of storage entries) or just leave the user to discover it if not.

tadeohepperle: I could not find a good example to show, so I removed that file, but I added some explanations to the book, how to use the `_iterX` constructors.

codegen/src/api/storage.rs
```diff
@@ -61,147 +62,105 @@ fn generate_storage_entry_fns(
...
-                    };
-
-                    (fields, key_impl)
-                }
+                TypeDef::Tuple(tuple) => tuple
+                    .fields
+                    .iter()
+                    .enumerate()
+                    .map(|(i, f)| {
+                        let ident: Ident = format_ident!("_{}", syn::Index::from(i));
```
jsdw, line 75: Nit; is `Ident` type explicitly needed?

subxt/src/book/usage/storage.rs
```diff
@@ -39,19 +39,26 @@
...
-//! let storage_query = subxt::dynamic::storage_root("System", "Account");
+//! let storage_query = subxt::dynamic::storage_iter("System", "Account");
 //! ```
 //!
+//! Some storage entries require multiple keys to be accessed, for example
+//! `polkadot::storage().preimage().preimage_for(_, _)`, which takes a `H256` and a `u32` as keys.
+//! For these multi-key storage entries, the generated static interface also offers _partial iteration_.
+//! This means iterating over a subset of storage entries, where just a subset of the keys match.
+//! These generated constructors have a suffix `_iter1` / `_iter2` / ... added, to iterate over all
+//! storage entries given the first 1 / 2 / ... keys.
```
jsdw, line 60: Perhaps an example will make this point clearer, eg:

```suggestion
//! Some storage entries are maps with multiple keys. As an example, we might end up with
//! an API like `runtime::storage().foo().bar(u8, bool, u16, String)` to fetch some entry "bar".
//! When this is the case, the codegen will generate multiple iterator query functions alongside
//! the function to fetch an individual value:
//!
//! - `runtime::storage().foo().bar(u8, bool, u16, String)`: fetch a single entry from the "bar" map.
//! - `runtime::storage().foo().bar_iter()`: iterate over all of the entries in the "bar" map.
//! - `runtime::storage().foo().bar_iter1(u8)`: iterate over all of the entries in the "bar" map under 
//!   a given `u8`.
//! - `runtime::storage().foo().bar_iter2(u8, bool)`: iterate over all of the entries in the "bar" map under 
//!   a given `u8` and `bool` value.
//! - `runtime::storage().foo().bar_iter2(u8, bool, u16)`: iterate over all of the entries in the "bar" map under 
//!   a given `u8`, `bool` and `u16` value.
```

subxt/src/storage/storage_address.rs
```diff
@@ -229,7 +229,7 @@ pub fn make_static_storage_map_key<T: codec::Encode>(t: T) -> StaticStorageMapKe
 }
 
 /// Construct a new dynamic storage lookup to the root of some entry.
-pub fn dynamic_root(
+pub fn dynamic_iter(
```
jsdw, line 232: Ah, but now the `dynamic_iter` doesn't actually provide the ability to iterate over subsets of keys like the static one does.

Maybe we should just remove this function entirely, since people can just use `dynamic` below with an empty vec to achieve the same, or with a subset of keys to hopefully be able to iterate whatever they like.

jsdw: (and if we remove this, remember to update the guide :)

jsdw:
Looks good overall! Just a few small comments ( a load of spaces in CHANGELOG to revert, comment update, remove `dynamic_root` now I think is best).

Also, I had al ook and found these storage entries with multiple values that might be easier to "set up" for testing:

```
multisig.multisigs(AccountId32, [u8; 32])
staking.erasStakers(u32, AccountId32)
xcmPallet.remoteLockedFungibles(u32, AccountId32, XcmVersionedAssetId)
```

Multisigs can be created and so maybe that's not too hard to write a test for? I mainly included the bottom one as the only example I could see of more than 2 keys :)

It would be great to have a test if possible to check that this stuff works (and with the multisig one we could also test the `dynamic` version to check it can do the same.

codegen/src/api/storage.rs
tadeohepperle: Ah, yes it is needed because of the `quote!( #field_name: impl ::std::borrow::Borrow<#field_ty> )` in Line 136, where `field_name` is the `ident` from above.

codegen/src/api/storage.rs
```diff
@@ -61,147 +62,231 @@ fn generate_storage_entry_fns(
...
+        )*
     })
 }
+
+fn primitive_type_alias(type_path: &TypePath) -> TokenStream {
+    // Vec<T> is cast to [T]
+    if let Some(ty) = type_path.vec_type_param() {
+        return quote!([#ty]);
+    }
+    // String is cast to str
```
lexnv, line 166: nice!

codegen/src/api/storage.rs
```diff
@@ -61,147 +62,231 @@ fn generate_storage_entry_fns(
...
+        } else {
+            let fn_name = if n_keys == 0 {
+                format_ident!("{snake_case_name}_iter")
+            } else {
+                format_ident!("{snake_case_name}_iter{}", n_keys)
+            };
+            (fn_name, false, true)
+        };
+        let is_fetchable_type = is_fetchable.then_some(quote!(#crate_path::storage::address::Yes)).unwrap_or(quote!(()));
+        let is_iterable_type = is_iterable.then_some(quote!(#crate_path::storage::address::Yes)).unwrap_or(quote!(()));
```
lexnv, line 125: nit: I found this a bit hard to follow, maybe a match would state the intent a bit better.
And since the `is_fetchable` and `is_iterable` are mutually exclusive, we could probably do with just one of them, up to you

jsdw (approved):
Looks great to me :)

subxt/src/book/usage/storage.rs
```diff
@@ -39,19 +39,33 @@
...
+//! When this is the case, the codegen will generate multiple iterator query functions alongside
+//! the function to fetch an individual value:
+//!
+//! - `runtime::storage().foo().bar(u8, bool, u16, String)`: fetch a single entry from the "bar" map.
+//! - `runtime::storage().foo().bar_iter()`: iterate over all of the entries in the "bar" map.
+//! - `runtime::storage().foo().bar_iter1(u8)`: iterate over all of the entries in the "bar" map under
+//!   a given `u8`.
+//! - `runtime::storage().foo().bar_iter2(u8, bool)`: iterate over all of the entries in the "bar" map under
+//!   a given `u8` and `bool` value.
+//! - `runtime::storage().foo().bar_iter2(u8, bool, u16)`: iterate over all of the entries in the "bar" map under
```
lexnv, line 66:
```suggestion
//! - `runtime::storage().foo().bar_iter3(u8, bool, u16)`: iterate over all of the entries in the "bar" map under
```

subxt/examples/storage_iterating_partial.rs
lexnv, line 1: I wonder if this partial example should (also) be part of the integration-tests

lexnv (approved):
👍

codegen/src/api/storage.rs
jsdw: I had a go and came up with something like this:

```rust
let is_iterator = n_keys != keys.len();

let fn_name = match (is_iterator, keys.len()) {
    (false, _) => format_ident!("{snake_case_name}"),
    (true, 0) => format_ident!("{snake_case_name}_iter"),
    (true, n) => format_ident!("{snake_case_name}_iter{}", n)
};

let yes_ty = || quote!(#crate_path::storage::address::Yes);
let no_ty = || quote!(());

let is_fetchable_type = is_iterator.then(no_ty).unwrap_or_else(yes_ty);
let is_iterable_type = is_iterator.then(yes_ty).unwrap_or_else(no_ty);
```

subxt/examples/storage_iterating_partial.rs
tadeohepperle: I tried that but had some issues. Let me see again, it is a good idea.

PR #1092, opened 2023-07-24
codegen: Fetch and decode metadata version then fallback

### LINKED ISSUE #1091: How to declare metadata version
I am getting this error on upgrading from v0.29.0 to v0.30.0. I assume that it defaults to V15 metadata, but the runtimes it is calling do not return this. Can I declare V14 metadata, or does the metadata returned have a way to tell subxt which version it is?
```
error: Could not decode metadata, only V14 and V15 metadata are supported: Could not decode `RuntimeMetadataPrefixed.1`:
           Could not decode `RuntimeMetadata::V15.0`:
               Could not decode `RuntimeMetadataV15::extrinsic`:
                   Could not decode `ExtrinsicMetadata::extra_ty`:
                       Could not decode `UntrackedSymbol::id`:
                           out of range decoding Compact<u32>
 --> src/types.rs:3:1
  |
3 | / #[subxt::subxt(
4 | |     runtime_metadata_url = "wss://kusama-rpc.polkadot.io:443",
5 | |     derive_for_all_types = "PartialEq, Clone"
6 | | )]
  | |__^
  |
  = note: this error originates in the attribute macro `subxt::subxt` (in Nightly builds, run with -Z macro-backtrace for more info)
```

lexnv:
The https://github.com/paritytech/subxt/issues/1091 exposed an error on out-of-date chains in terms of metadata with the latest substrate.

The code could fetch the latest unstable metadata, but that metadata version would fail to decode.
Failing to decode the metadata is a terminal error in our codegen.

To mitigate this behavior, the codegen would fetch and decode the metadata, the fallback on fetching older versions if that fails.

The order of fetching operations:
1. Unstable if the `unstable-metadata` attribute is provided
3. V15
4. V14

This is a bit overkill but hopefully would continue to work fine with fallbacks always on stable v15 and v14.

Closes: https://github.com/paritytech/subxt/issues/1091

### DIFF
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -83,6 +83,7 @@ where
 /// * `crate_path` - Path to the `subxt` crate.
 /// * `should_gen_docs` - True if the generated API contains the documentation from the metadata.
 /// * `runtime_types_only` - Whether to limit code generation to only runtime types.
+/// * `unstable_metadata` - Whether to fetch the unstable metadata first.
 ///
 /// **Note:** This is a wrapper over [RuntimeGenerator] for static metadata use-cases.
 pub fn generate_runtime_api_from_url(
@@ -93,21 +94,24 @@ pub fn generate_runtime_api_from_url(
     crate_path: CratePath,
     should_gen_docs: bool,
     runtime_types_only: bool,
+    unstable_metadata: bool,
 ) -> Result<TokenStream2, CodegenError> {
     fn fetch_metadata(url: &Uri, version: MetadataVersion) -> Result<Metadata, CodegenError> {
         let bytes = fetch_metadata_bytes_blocking(url, version)?;
         Ok(Metadata::decode(&mut &bytes[..])?)
     }
 
-    let metadata = match fetch_metadata(url, MetadataVersion::Unstable) {
-        Ok(metadata) => metadata,
-        Err(_) => match fetch_metadata(url, MetadataVersion::Latest) {
+    let metadata = unstable_metadata
+        .then(|| fetch_metadata(url, MetadataVersion::Unstable).ok())
+        .flatten();
+
+    let metadata = if let Some(unstable) = metadata {
+        unstable
+    } else {
+        match fetch_metadata(url, MetadataVersion::Version(15)) {
             Ok(metadata) => metadata,
-            Err(_) => match fetch_metadata(url, MetadataVersion::Version(15)) {
-                Ok(metadata) => metadata,
-                Err(_) => fetch_metadata(url, MetadataVersion::Version(14))?,
-            },
-        },
+            Err(_) => fetch_metadata(url, MetadataVersion::Version(14))?,
+        }
     };
 
     generate_runtime_api_with_metadata(
--- a/codegen/src/api/mod.rs
+++ b/codegen/src/api/mod.rs
@@ -94,15 +94,25 @@ pub fn generate_runtime_api_from_url(
     should_gen_docs: bool,
     runtime_types_only: bool,
 ) -> Result<TokenStream2, CodegenError> {
-    // Fetch latest unstable version, if that fails fall back to the latest stable.
-    let bytes = match fetch_metadata_bytes_blocking(url, MetadataVersion::Unstable) {
-        Ok(bytes) => bytes,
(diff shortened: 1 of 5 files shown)


codegen/src/api/mod.rs
```diff
@@ -94,15 +94,25 @@ pub fn generate_runtime_api_from_url(
...
-    // Fetch latest unstable version, if that fails fall back to the latest stable.
-    let bytes = match fetch_metadata_bytes_blocking(url, MetadataVersion::Unstable) {
-        Ok(bytes) => bytes,
-        Err(_) => fetch_metadata_bytes_blocking(url, MetadataVersion::Latest)?,
+    fn fetch_metadata(url: &Uri, version: MetadataVersion) -> Result<Metadata, CodegenError> {
+        let bytes = fetch_metadata_bytes_blocking(url, version)?;
+        Ok(Metadata::decode(&mut &bytes[..])?)
+    }
+
+    let metadata = match fetch_metadata(url, MetadataVersion::Unstable) {
```
jsdw, line 102: We shouldnt try to fetch unstable metadata now unless subxt has the unstable-metadata feature activated

one way to do this might be to expose from subxt eg a const boolean called UNSTABLE_METADATA that is true if the feature is enabled and false if it's not (via feature flag) and is doc(hidden).

this crate could then check that to decide whether to try and fetch unstable first.

jsdw: Actually that's no good because the crate doesn't depend on subxt. Let's just have it look for V15 and then V14, and optionally we can also add an "unstable_metadata" attr to enable it fetching unstable if we want :)

subxt/src/lib.rs
```diff
@@ -294,6 +295,21 @@ pub mod ext {
...
+/// #[subxt::subxt(
+///     runtime_metadata_url = "wss://rpc.polkadot.io:443"
+///     unstable_metadata,
+/// )]
+/// mod polkadot {}
+/// ```
+///
 /// **Note**: At the moment, you must derive at least one of `codec::Encode` or `codec::Decode` or `scale_encode::EncodeAsType` or
 /// `scale_decode::DecodeAsType` (because we add `#[codec(..)]` attributes on some fields/types during codegen), and you must use this
 /// feature in conjunction with `runtime_types_only` (or manually specify a bunch of defaults to make codegen work properly when
```
jsdw, line 315: These lines are meant to be a part of the `no_default_derives` section :)

subxt/src/lib.rs
```diff
@@ -294,6 +295,21 @@ pub mod ext {
 /// mod polkadot {}
 /// ```
 ///
+/// ## `unstable_metadata`
+///
+/// This attribute works only in combination with `runtime_metadata_url`. By default, the macro will fetch the latest stable
+/// version of the metadata from the target node. This attribute makes the codegen fetch the unstable version of the metadata.
+/// This can be useful in CI, but is **not recommended** in production code, since it runs at compile time and will cause
+/// compilation to fail if the node at the given address is unavailable or unresponsive.
```
jsdw, line 303:
```suggestion
/// This attribute works only in combination with `runtime_metadata_url`. By default, the macro will fetch the latest stable
/// version of the metadata from the target node. This attribute makes the codegen attempt to fetch the unstable version of 
/// the metadata first. This is **not recommended** in production code, since the unstable metadata a node is providing is likely 
/// to be incompatible with Subxt.
```

macro/src/lib.rs
```diff
@@ -131,8 +133,15 @@ pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
     }
 
     let should_gen_docs = args.generate_docs.is_present();
+    let unstable_metadata = args.unstable_metadata.is_present();
     match (args.runtime_metadata_path, args.runtime_metadata_url) {
         (Some(rest_of_path), None) => {
+            if unstable_metadata {
+                abort_call_site!(
+                    "The 'unstable_metadata' is expected only for `runtime_metadata_url`"
```
jsdw, line 141:
```suggestion
                    "The 'unstable_metadata' attribute requires `runtime_metadata_url`"
```

jsdw (approved):
Looks good to me; the builder pattern cleans things up a bit too :)

tadeohepperle (approved):
👍🏻

PR #1097, opened 2023-07-27
Set minimum supported `rust-version` to `1.70`

ascjones:
`OnceLock`, utilized in `subxt_signer` was only [stabilized](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#oncecell-and-oncelock) in `1.70`.

From https://github.com/paritytech/cargo-contract/issues/1240


jsdw (approved):

jsdw:
Ooh thanks; that number has obviously been neglected a bit; probably worth a note in the RELEASING.md guide!

ascjones:
> Ooh thanks; that number has obviously been neglected a bit; probably worth a note in the RELEASING.md guide!

I guess it is not easy to do without trying to build with different versions. Never used it but maybe an automated tool such as https://github.com/foresterre/cargo-msrv would help, and can even be used in CI (though I imagine it takes a while and you don't want to run it all the time).

jsdw:
Thanks, that's a good idea! I'll open an issue so we can think about this in any case! I knew that OnceCell or whatever was new to 1.70 but forgot that we even mentioned an msrv anywhere to be honest!

lexnv (approved):

PR #1102, opened 2023-07-31
Tests: support 'substrate-node' too and allow multiple binary paths

### LINKED ISSUE #1095: `SUBSTRATE_NODE_PATH` should default or accept `substrate-node`
Following the instructions here: https://github.com/paritytech/subxt/tree/master#integration-testing, substrate is actually installed as `substrate-node` rather than `substrate`, so running the examples would just fail. Recommendation would be to accept both `substrate` and `substrate-node` as possible binaries paths or add instructions to rename the binary.

jsdw:
Support passing multiple binary paths, and default to trying `substrate-node` and then `substrate`.

Closes #1095

### DIFF
--- a/testing/integration-tests/src/utils/node_proc.rs
+++ b/testing/integration-tests/src/utils/node_proc.rs
@@ -76,7 +76,7 @@ impl TestNodeProcessBuilder {
     {
         let mut node_builder = SubstrateNode::builder();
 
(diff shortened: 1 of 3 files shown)


lexnv (approved):
LGTM!

tadeohepperle (approved):
LGTM!

PR #1105, opened 2023-08-02
Support more types in Storage entry constructors 

### LINKED ISSUE #610: Support more types in Storage entry constructors
Basically, for any types needed in storage entries, we allow `impl Borrow<#ty>` for it. We have some custom logic so that if the type is `Vec<T>` we allow `impl Borrow<[T]>` for more flexibility. Can we do the same for types like `String` (prefer `str`) and `Box<T>` (prefer `T`) and so on?

See https://github.com/paritytech/subxt/pull/593#discussion_r938598968

tadeohepperle:
Fixes #610

This PR adds support for accessing storage items where the keys are of type `String` through `impl Borrow<str>` to allow more flexibility (before it would have been `impl Borrow<String>` ). In Polkadot no storage item is currently accessed via String though.
I added a function that makes it easy to add other rules and a test that validates that the code generation is working correctly. Currently it checks these types:
|original type|borrowed as|
|--|--|
|`Vec<u8>`         |`u8`
|`Box<u16>`     |`u16`
|`String`            |`str`
|`&'static str`    |`str`
|`Cow<'_, str>` |`str`


CHANGELOG.md
jsdw, line 1: Screewed up CHANGELOG here too :)

jsdw: Aah this is building off the other storage PR. For now maybe best to base the branch off the other PR so we only see the relevant changes here :)

codegen/src/api/storage.rs
```diff
@@ -61,147 +62,231 @@ fn generate_storage_entry_fns(
...
+    // Vec<T> is cast to [T]
+    if let Some(ty) = type_path.vec_type_param() {
+        return quote!([#ty]);
+    }
+    // String is cast to str
+    if type_path.is_string() {
+        return quote!(::core::primitive::str);
+    }
+    quote!(#type_path)
+}
```
jsdw, line 171: Looks sensible to me :)

PR #1106, opened 2023-08-03
V15 Metadata: Support accessing custom types

### LINKED ISSUE #1042: V15 Metadata: Support accessing custom types
V15 metadata introduces a new custom type map, something like `BTreeMap<String, { type_id: u32, value: Vec<u8> }>`.

Once this is stable, let's expose them; maybe with an API something like:

```rust
let custom_types = api.custom_types();

// Get some custom type "Foo", `None` if not found.
let ty = custom_types.get("Foo")?;

// Attempt to decode into a concrete type (Foo impls DecodeAsType):
// can use this to decode into `Value` too. `as_` because we have `as_event`,
// `as_extrinsic` or whatever. `decode_as_type` might be nicer.
let something: Foo = ty.as_type()?;

// Return the raw bytes:
let scale_bytes = ty.encoded();

// Return the raw type ID:
let ty_id = ty.type_id();
```

tadeohepperle:
fixes #1042.

Custom types are now available from `OfflineClientT` like this:
```rs
let api: OfflineClient<Config>;
let ty = api.custom_types().get("Foo")?;
let foo: Foo = ty.as_type()?;
```


subxt/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! Subxt is a library for interacting with Substrate based nodes. Using it looks something like this:
 //!
 //! ```rust,ignore
-#![doc = include_str!("../examples/tx_basic.rs")]
+#![doc = include_str ! ("../examples/tx_basic.rs")]
```
lexnv, line 8: nit: Is this intended?

lexnv (commented):
Lovely PR and useful to expose those custom fields to users!
Do you think it would be worth exposing the custom types in the codegen as well?

jsdw:
> Do you think it would be worth exposing the custom types in the codegen as well?

That's an interesting thought. Being quite arbitrary, some of the types and values may not line up (and so codegen should be lenient and just ignore any errors), and the string names are also arbitrary and need not be valid idents or anything, but one _could_ expose codegen for the cases where things appear to line up. It'd be cool to be able to do `let query = runtime::custom().foo(); api.custom().at(&query)` to access things in custom, decoded to some concrete type (if possible).

I'll raise an issue so we can consider this as a follow-up

metadata/src/lib.rs
```diff
@@ -132,6 +138,11 @@ impl Metadata {
         })
     }
 
+    /// Returns custom user defined types
+    pub fn custom_metadata(&self) -> &CustomMetadata<PortableForm> {
+        &self.custom
+    }
```
jsdw, line 144: I think we should hide the internals (frame-metadata stuff) a little better here and have an api more like:

```rust
let custom = metadata.custom();

let Some(value) = custom.get("Foo") else { return };

let type_id: u32 = value.ty();
let data: &[u8] = value.bytes();

// so maybe thats implemented something like:

pub struct CustomMetadata {
    // We can leave the frame_metadata map as is:
    values: BTreeMap<String, frame_metadata::CustomValueMetadata<T>>
}

pub struct CustomMetadataValue<'a> {
    type_id: u32,
    data: &'a Vec<u8>
}

impl CustomMetadata {
    pub fn get(&'a self, &str) -> Option<CustomMetadataValue<'a>>
}

impl CustomMetadataValue {
    pub fn ty(&self) -> u32;
    pub fn bytes(&Self) -> &[u8]
}
```

subxt/src/client/online_client.rs
```diff
@@ -292,6 +295,11 @@ impl<T: Config> OnlineClient<T> {
         <Self as OfflineClientT<T>>::constants(self)
     }
 
+    /// Access custom types.
+    pub fn custom_types(&self) -> CustomTypes {
```
jsdw, line 299: It's really the values that we are accessing (and the values happen to come with type info) so maybe `custom_values -> CustomValues` makes more sense?

subxt/src/custom_types/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+//! Types associated with accessing custom types
+
+use crate::error::MetadataError;
+use crate::Metadata;
+use scale_decode::DecodeAsType;
+
+/// gives you access to all `CustomValueMetadata` entries that are available as custom types on the metadata.
+pub struct CustomTypes {
+    metadata: Metadata,
+}
```
jsdw, line 14: Let's mirror constants etc and call this something like `CustomValuesClient` and have it store `Client` in it.

That way, we are consistent with the other "top level" APIs, but also give ourselves room to add eg codegen or whatever to aceess and decode custom types (which I hadn't thought about to be honest, but Alex bought it up and it seems like a reasonable sort of thing to want here, so that people can easily get hold of concrete decoded values).

Now the more contentious bit:

For the same reason (having concrete types/codegen), I wonder whether we should make the API look like the others also with respect to passing in some payload to access the value (ie copy the approach taken in the `constants` APIs)? That way we can have dynamic or static "queries" to access the value much like other APIS.

Pros of doing this:
- Consistent with other APIs in subxt
- Can decode to static values, or `DecodedValueThunk`'s just like constants

Cons:
- More overkill in terms of code to write
- `DecodedValueThunk` copies the bytes when it decodes which is less performant.

I think on balance it's probably worth it to be consistent and give room for dynamic or static decoding. What do you think?

tadeohepperle: I think we should make a CustomValuesClient that can only query values dynamically. I do not see a point in giving static queries to it, because we could just as well expose the static values in the codegen. 
For example if there is a custom type `Foo` with the structure `struct Foo {a: u8, b: bool}` that has an associated value `vec![3,0]`, we could probably generate:
```
struct Foo {a: u8, b: bool}
impl Default for Foo{
    fn default() -> Foo{
        const data: [u8; 2] = [3,0];
        Foo::decode(&mut &data[..]).expect("should work, otherwise Foo value is broken.")
    } 
}
```
This will make it much less confusing for users to use. Sadly we cannot make the Foo default value constant directly, because it might contains vecs and other types with heap allocations.
What do you think about that? Maybe we can also go the same path for constants.

metadata/src/from_into/v15.rs
```diff
@@ -17,6 +17,7 @@ use std::collections::HashMap;
 // Converting from V15 metadata into our Subxt repr.
 mod from_v15 {
     use super::*;
+    use crate::CustomMetadata;
```
jsdw, line 20: We don't need this for all of the other types?

metadata/src/lib.rs
```diff
@@ -631,6 +638,40 @@ pub struct RuntimeApiMethodParamMetadata {
     pub ty: u32,
 }
 
+/// Metadata of custom types with custom values, basically the same as `frame_metadata::v15::CustomMetadata<PortableForm>>`.
+#[derive(Debug, Clone)]
+pub struct CustomMetadata {
+    pub(crate) map: BTreeMap<String, frame_metadata::v15::CustomValueMetadata<PortableForm>>,
```
jsdw, line 644: I don't think the `pub(crate)` is necessary (since sub-modules can access stuff in parent ones anyway, and it's not needed anywhere else)

metadata/src/lib.rs
```diff
@@ -631,6 +638,40 @@ pub struct RuntimeApiMethodParamMetadata {
...
+pub struct CustomMetadataValue<'a> {
+    type_id: u32,
+    data: &'a [u8],
+}
+
+impl<'a> CustomMetadataValue<'a> {
+    /// the scale encoded value
+    pub fn bytes(&self) -> &[u8] {
+        &self.data
+    }
```
jsdw, line 667:
```suggestion
    pub fn bytes(&self) -> &'a [u8] {
        &self.data
    }
```

Just to tie the lifetime to that of the bytes and not that of this `CustomMetadataValue` instance

subxt/src/custom_values/custom_value_adress.rs
jsdw, line 1: Typo in the file name, should be `custom_value_address.rs` :)

subxt/src/custom_values/custom_value_adress.rs
```diff
@@ -0,0 +1,21 @@
...
+/// custom values from the metadata.
+pub trait CustomValueAddress {
+    /// The type of the custom value.
+    type Target: DecodeWithMetadata;
+
+    /// the name (key) by which the custom value can be accessed in the metadata.
+    fn name(&self) -> &str;
+}
+
+impl CustomValueAddress for str {
```
jsdw, line 15: Maybe for consistency we just impl this on `&str` so that we don't need the `+ ?Sized` bound below? I'm a bit torn; it doesn't really matter either way afaik :)

subxt/src/custom_values/custom_values_client.rs
```diff
@@ -0,0 +1,138 @@
...
+                call_enum_ty: unit_id,
+                event_enum_ty: unit_id,
+                error_enum_ty: unit_id,
+            },
+            custom: frame_metadata::v15::CustomMetadata {
+                map: {
+                    let mut m = BTreeMap::new();
+                    m.insert("Person".to_string(), person_value_metadata);
+                    m
+                },
```
jsdw, line 107: Just FYI a nice alternate way to declare this is:

```suggestion
                map: BTreeMap::from_iter([
                    ("Person".to_string(), person_value_metadata)
                ]),
```

jsdw:
Just some minor comments but looks great overall; nice one!

metadata/src/lib.rs
tadeohepperle: Oh yeah true, nice catch!

subxt/src/book/usage/custom_values.rs
```diff
@@ -0,0 +1,54 @@
...
+//! let api = OnlineClient::<PolkadotConfig>::new().await?;
+//! let custom_value_client = api.custom_values();
+//!
+//! // Now the `at()` function already decodes the value into the Foo type:
+//! let foo = custom_value_client.at(&static_address)?;
+//! ```         
+//!
+//! Note: Names of custom values are converted to __snake_case__ to produce a valid function name during code generation.
+//! If there are multiple values where the names would be equal when converted to __snake_case__, functions might not be statically generated for some of them, because of naming conflicts.
+//! Make sure names in the custom values of your metadata differ significantly.
```
jsdw, line 54:
```suggestion
```

Just because I think it's covered above, and users often won't have the ability to influence chain metadata anyway :)

subxt/src/book/usage/custom_values.rs
```diff
@@ -0,0 +1,54 @@
...
+//! ```rust,no_run
+//! #[subxt::subxt(runtime_metadata_path = "some_metadata.scale")]
+//! pub mod interface {}
+//!
+//! let static_address = interface::custom().foo();
+//!
+//! let api = OnlineClient::<PolkadotConfig>::new().await?;
+//! let custom_value_client = api.custom_values();
+//!
+//! // Now the `at()` function already decodes the value into the Foo type:
```
jsdw, line 48:
```suggestion
//! // Now, the `at()` function decodes the value into the correct type:
```

subxt/src/book/usage/custom_values.rs
jsdw, line 1: Lovely docs; nice one!

subxt/src/book/usage/custom_values.rs
```diff
@@ -0,0 +1,54 @@
...
+//! }
+//!
+//! let api = OnlineClient::<PolkadotConfig>::new().await?;
+//! let custom_value_client = api.custom_values();
+//! let foo_dynamic = custom_value_client.at("foo")?;
+//! let foo: Foo = foo_dynamic.as_type()?;
+//!
+//! ```
+//!
+//! Alternatively we also provide a statically generated api for custom values:
```
jsdw, line 37: I guess the logic for this is a followup PR, but as long as we don't do another release before the static stuff merges too then it's ok to leave the docs here :)

jsdw (approved):
Looks great to me!

subxt/src/book/usage/custom_values.rs
jsdw: Two small things you'll need to do to make tests pass:

- import metadata from the right location (look at other doc examples for this)
- wrap the doc stuff in a hidden `#[tokio::main] async fn main()` thing (use `#` to hide this; check out other examples to see :))

lexnv: Lovely docs indeed! 🚀

lexnv (approved):
LGTM! Amazing job here!

PR #1107, opened 2023-08-04
Make ExtrinsicParams more flexible, and introduce signed extensions 

jsdw:
Introduce `subxt::config::signed_extensions`. This exposes a trait, `SignedExtension`, implementations for the usual signed extensions, and an `AnyOf` wrapper which composes signed extensions into a single thing which implements `ExtrinsicParams` and uses metadata to determine which extension to apply and in what order.

Make some changes to `ExtrinsicParams` to accomodate this, namely providing an `OfflineClient` (so that we can get hold of metadata and generally anything else that might be useful), and splitting the `encode` methods to an `ExtrinsicParamsEncoder` trait (necessary so that we can `Box<dyn>` what we need to be able to dynamically decide which signed extensions to use etc.

There's now a `subxt::config::DefaultExtrinsicParams` which is just an `AllOf` containing all of the known signed extensions (that emit data, at least; we can ignore the others). This is used in `PolkadotConfig` and `SubstrateConfig` (rendering the two almost identical), and should generally work across various chains.

I updated the docs and examples to show how to write a new signed extension or just fully custom `ExtrinsicParams`.


subxt/src/config/era.rs
jsdw, line 1: `Era` is useful behind the scenes, but it's not a great interface and so I've hidden it from public view.

examples/wasm-example/src/services.rs
```diff
@@ -4,7 +4,7 @@ use js_sys::Promise;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use std::fmt::Write;
-use subxt::ext::codec::Encode;
+use subxt::ext::codec::{ Compact, Encode };
```
niklasad1, line 7:
```suggestion
use subxt::ext::codec::{Compact, Encode};
```

examples/wasm-example/src/services.rs
```diff
@@ -143,7 +143,7 @@ pub async fn extension_signature_for_partial_extrinsic(
     let spec_version = encode_to_hex_reverse(&api.runtime_version().spec_version);
     let transaction_version = encode_to_hex_reverse(&api.runtime_version().transaction_version);
     let mortality_checkpoint = encode_to_hex(&api.genesis_hash());
-    let era = encode_to_hex(&subxt::config::extrinsic_params::Era::Immortal);
+    let era = encode_to_hex(&0u8);
```
niklasad1, line 146: I guess `0` is `Era::Immortal` would be nice with a comment to explain that, this is just an example right?

subxt/src/config/default_extrinsic_params.rs
```diff
@@ -0,0 +1,132 @@
...
+    tip: u128,
+    tip_of: u128,
+    tip_of_asset_id: Option<u32>,
+}
+
+impl<T: Config> Default for DefaultExtrinsicParamsBuilder<T> {
+    fn default() -> Self {
+        Self {
+            mortality_checkpoint_hash: None,
+            mortality_checkpoint_number: 0,
```
niklasad1, line 39: default == immortal?

niklasad1: could be cleaner to add `Option<Mortality>` instead of the three mortality fields

niklasad1 (approved):
Looks good and I like the new API where the Era is not exposed anymore.

Some minor questions

examples/wasm-example/src/services.rs
jsdw: Yup just an example; I'll add a comment :)

subxt/src/config/default_extrinsic_params.rs
jsdw: Yup, good point, that's def cleaner!

subxt/src/config/era.rs
jsdw: I actually re-exposed it again from `utils` (but it's still not exposed in any Subxt APIs, just used internally. The reason I did this was just so that people would be able to properly encode mortal transactions in eg the `wasm-example` using the Era type (so I went back to using that there too)

lexnv (approved):

tadeohepperle (commented):
I like the changes a lot, will make many things much easier.

PR #1111, opened 2023-08-07
Add `storage_version()` and `runtime_wasm_code()` to storage

tadeohepperle:
This PR adds two methods `storage_version(pallet_name)` and `runtime_wasm_code` to the storage to fetch the respective data from the node. I also wanted to add other keys (see mod `well_known_keys`) but retrieving data with them did not work (at least with the polakdot node I tried it with). I left them in the PR as private to maybe activate them in the future if necessary.


subxt/src/storage/storage_type.rs
```diff
@@ -237,6 +241,43 @@ where
...
+        // construct the storage key. This is done similarly in `frame_support::traits::metadata::StorageVersion::storage_key()`.
+        pub const STORAGE_VERSION_STORAGE_KEY_POSTFIX: &[u8] = b":__STORAGE_VERSION__:";
+        let mut key_bytes: Vec<u8> = vec![];
+        key_bytes.extend(&sp_core_hashing::twox_128(pallet_name.as_ref().as_bytes()));
+        key_bytes.extend(&sp_core_hashing::twox_128(
+            STORAGE_VERSION_STORAGE_KEY_POSTFIX,
+        ));
+
+        // fetch the raw bytes and decode them into the StorageVersion struct:
+        let storage_version_bytes = self.fetch_raw(&key_bytes).await?.ok_or(format!(
```
jsdw, line 265: Nit: `ok_or_else` here toa void formatting the string unless there's an error (I'd expect clippy to spot this but havent checked)

subxt/src/storage/utils.rs
```diff
@@ -37,3 +37,50 @@ pub(crate) fn storage_address_root_bytes<Address: StorageAddress>(addr: &Address
     write_storage_address_root_bytes(addr, &mut bytes);
     bytes
 }
+
+/// this should match the constants in the module `sp_core::storage::well_known_keys`
+pub mod well_known_keys {
```
jsdw, line 42: I'd probably keep this in `subxt::storage` and not utils (it's storage specific I think?).

Is there a use in exposing any of this stuff at all?  I'd be inclined to just inline what you need into the `runtime_wasm_code` function for now, with a pointer to where you got the key from there so we can add more if need be or whatever.

Alternately we could expose this `WellKnownKey<R>` type in `subxt::storage` and then a single `api.storage().fetch_well_known_key(key: WellKnownKey<R>) -> Result<R, Error>` type function to fetch them (it sortof depends whether any of the other keys are actually useful anywhere; that might be worth finding out!).

tadeohepperle: This is in `subxt::storage::utils` not in the general `subxt::utils`. I wanted to expose a "nicer" interface in the hopes that many other well known keys might work in a similar fashion, such that we can abstract over their types and keys with a function like `fetch_well_known_key()` on the storage. 
But the other keys do not seem to work as intended and I think you are right, it is less bloat to just hardcode the `":code"` in the right place and remove the other stuff here. Also first I thought the returned value was scale decoded but that is also not the case apparently.

jsdw: Ok that makes sense to me; let's keep it simple then until we see a use for allowing other keys!

subxt/src/storage/storage_type.rs
```diff
@@ -237,6 +241,43 @@ where
             })
         }
     }
+
+    /// the storage version of a pallet.
```
lexnv, line 245: nit: I prefer upper case for documentation and for consistency, but totally up to you

subxt/src/storage/utils.rs
lexnv: It would be interesting to investigate what is causing the failure in Polkadot for the other known keys :D

jsdw: I agree!

testing/integration-tests/src/full_client/storage/mod.rs
```diff
@@ -125,3 +125,33 @@ async fn storage_n_map_storage_lookup() -> Result<(), subxt::Error> {
...
+async fn storage_pallet_storage_version() -> Result<(), subxt::Error> {
+    let ctx = test_context().await;
+    let api = ctx.client();
+
+    // cannot assume anything about version number, but should work to fetch it
+    let _version = api
+        .storage()
+        .at_latest()
+        .await?
+        .storage_version("System")
```
lexnv, line 148: Crazy thought here; could we use this to do a quick version validation? Similar to our hashing, I guess that would require substrate to constantly modify these numbers appropriately with changes.

tadeohepperle: When would that version validation happen? I guess we would need to make about 30 concurrent calls to validate all pallets. Or we might just validate the pallet whenever a storage query/call is made for a specific pallet?

jsdw: We already validate the keys and value types on a per-entry basis so probably what we have is already better (I imagine the version might increment for every storage change, but not really sure what it represents exactly?)

subxt/src/storage/storage_type.rs
```diff
@@ -237,6 +240,47 @@ where
...
+    }
+
+    /// Fetches the Wasm code of the runtime.
+    pub async fn runtime_wasm_code(&self) -> Result<Vec<u8>, Error> {
+        // note: this should match the `CODE` constant in `sp_core::storage::well_known_keys`
+        const CODE: &[u8] = b":code";
+        let data = self.fetch_raw(CODE).await?.ok_or(format!(
+            "Unexpected: entry for well known key \"{}\" not found",
+            std::str::from_utf8(CODE).expect("qed;")
+        ))?;
```
jsdw, line 281:
```suggestion
        let data = self.fetch_raw(CODE).await?.ok_or_else(|| format!(
            "Unexpected: entry for well known key \"{}\" not found",
            std::str::from_utf8(CODE).expect("qed;")
        ))?;
```

subxt/src/storage/storage_type.rs
```diff
@@ -237,6 +240,47 @@ where
...
+        Ok(storage_version)
+    }
+
+    /// Fetches the Wasm code of the runtime.
+    pub async fn runtime_wasm_code(&self) -> Result<Vec<u8>, Error> {
+        // note: this should match the `CODE` constant in `sp_core::storage::well_known_keys`
+        const CODE: &[u8] = b":code";
+        let data = self.fetch_raw(CODE).await?.ok_or(format!(
+            "Unexpected: entry for well known key \"{}\" not found",
+            std::str::from_utf8(CODE).expect("qed;")
```
jsdw, line 280: If you wanted to simplify this you could just have:

```rust
        const CODE: &str = ":code";
        let data = self.fetch_raw(CODE).await?.ok_or_else(|| format!(
            "Unexpected: entry for well known key \"{CODE}\" not found",
```

subxt/src/utils/mod.rs
```diff
@@ -76,3 +76,19 @@ unsafe impl<T> Sync for PhantomDataSendSync<T> {}
...
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Encode,
+    Decode,
+    Debug,
+    scale_encode::EncodeAsType,
+    scale_decode::DecodeAsType,
+)]
+pub struct StorageVersion(u16);
```
jsdw, line 94: Since we return this to the user we should either:

1. Make the `u16` pub (that's probably what they care about but currently it's inaccessible, or
2. Just return a `u16` from `storage_version` (that's my preference; I don't think this struct adds anything)

jsdw (approved):
This looks good to me now; nice one :)

subxt/src/storage/storage_type.rs
```diff
@@ -237,6 +239,42 @@ where
...
+
+        // construct the storage key. This is done similarly in `frame_support::traits::metadata::StorageVersion::storage_key()`.
+        pub const STORAGE_VERSION_STORAGE_KEY_POSTFIX: &[u8] = b":__STORAGE_VERSION__:";
+        let mut key_bytes: Vec<u8> = vec![];
+        key_bytes.extend(&sp_core_hashing::twox_128(pallet_name.as_ref().as_bytes()));
+        key_bytes.extend(&sp_core_hashing::twox_128(
+            STORAGE_VERSION_STORAGE_KEY_POSTFIX,
+        ));
+
+        // fetch the raw bytes and decode them into the StorageVersion struct:
```
lexnv, line 260: nit: Does this comment needs updating?

lexnv (approved):

PR #1117, opened 2023-08-10
Codegen for custom values in metadata

tadeohepperle:
This PR adds static addresses for custom values, accessible via codegen.
So for example, if there is a custom value with name "Foo" referring to a type `struct Foo { n: u8, b: bool }` and the attached value is scale encoded (e.g. [5,0] for Foo), then we get back a decoded Foo from `custom_value_client.at(&static_address)?`:

```rs
#[subxt::subxt(runtime_metadata_path = "some_metadata.scale")]
pub mod interface {}
let static_address = interface::custom().foo();
let api = OnlineClient::<PolkadotConfig>::new().await?;
let custom_value_client = api.custom_values();
// Now the `at()` function already decodes the value into the Foo type:
let foo = custom_value_client.at(&static_address)?;
```

After some back and forth, I decided not to add a test for this, because it would involve creating some artificial metadata scale or json file, using it in the subxt macro and then using the interface with it. I think if people are moving to using custom types or they appear in polkadot in the future we should add an integration test for it.


subxt/src/custom_values/custom_value_address.rs
```diff
@@ -19,3 +22,27 @@ impl CustomValueAddress for str {
         self
     }
 }
+
+/// A static address to a custom value.
+pub struct StaticAddress<R> {
+    name: &'static str,
+    phantom: PhantomData<R>,
+}
```
jsdw, line 30: The main thing to add here is the validation hash stuff (see how static constant addresses are defined). Then it can be checked in the `at` call (look at constants again for inspiration; it's pretty much the same idea :))

codegen/src/api/custom_values.rs
```diff
@@ -0,0 +1,104 @@
...
+                use super::#types_mod_ident;
+
+                pub struct CustomValuesApi;
+
+                impl CustomValuesApi {
+                    #( #custom_value_fns )*
+                }
+            }
+
+
```
lexnv, line 74: Some extra spaces here

```suggestion
```

codegen/src/api/custom_values.rs
```diff
@@ -0,0 +1,104 @@
...
+        )
+    });
+
+    if custom_value_fns.is_empty() {
+        None
+    } else {
+        Some(quote!(
+            pub fn custom() -> custom::CustomValuesApi {
```
lexnv, line 60: With this implementation, I wonder if the metadata doesn't expose any custom values, then the users will not have the `fn custom()` exposed.

Wouldn't it be more consistent to have the `pub fn custom()` either way? Similar to how we expose `pub fn tx() -> TransactionApi`, then it would be an implementation detail that the `CustomValuesApi` doesn't expose anything

codegen/src/api/custom_values.rs
```diff
@@ -0,0 +1,104 @@
...
+                custom::CustomValuesApi
+            }
+
+            pub mod custom{
```
lexnv, line 64: Would this work without placing the `CustomValuesApi` under the `pub mod custom` module? 

Then `ConstantsApi`, `StorageApi`, `TxApi` would be placed on the same indentation/path level as our newly added `CustomValuesApi`

lexnv (commented):
Lovely PR! Left a few minor things wrt consistency. 

Would it be possible to have a custom metadata-generated test under `integration-tests`? I'm not sure how hard it would be since substrate doesn't expose atm any custom fields, would love to hear your thoughts 🙏

codegen/src/api/custom_values.rs
```diff
@@ -0,0 +1,104 @@
...
+                pub struct CustomValuesApi;
+
+                impl CustomValuesApi {
+                    #( #custom_value_fns )*
+                }
+            }
+
+
+        )
+    });
```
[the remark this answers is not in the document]
niklasad1: I agree with Alex that `if custom_value_fns.is_empty()` is easier for me to understand here but I always struggle to understand then/then_some on bool but you could also have another let binding such as:

``rust
let has_custom_vals = !custom_value_fns.is_empty();
has_custom_val.then(|| ... );
```

**Comment by jsdw** [comment]
I can think of a couple of ways to test this but I'm not sure which is cleanest offhand:

1. Generate a metadata bundle in our `artifacts` folder which has some arbitrary custom values added.
   - There are other tests showing how to construct an instance of a type registry with some types in, so one could do that and then build some metadata with _only_ custom types in and nothing else (for one thing, because merging types is a pain).
   - This could be used in some test by pointing the macro at it and checking that you can pull the custom values out.
   - It could also be used in example docs to show custom values working.
   - Bit of a pain because you need some code to generate that file (at least, I'd prefer this to having some opaque pre-generated file we had no code to re-create. This is essentially a script but would need to be in Rust. Wish it was more "ready" but  maybe https://crates.io/crates/cargo-script-mvs could be used to write us a rust script to generate).
2. Make a UI test; there's already a helper to generate arbitrary metadata and then build a UI test from it IIRC, so that shouldn't be super hard to do.
   - Easier because you don't need a script somewhere to generate the metadata bundle.
   - Worse because we can't use this also for good doc examples.

So yeah probably it's worth doing 1 to generate some metadata and use that for test/doc example purposes. When Substrate has custom metadata appearing in it for whatever reason, we can move to using that instead if we prefer.

File: metadata/src/utils/validation.rs [hunk 6]
```diff
@@ -503,6 +534,7 @@ impl<'a> MetadataHasher<'a> {
             metadata,
             specific_pallets: None,
             specific_runtime_apis: None,
+            include_custom_values: true,
```
jsdw, line 537: Cool!

Let's add an `pub fn ignore_custom_values(&mut self) -> &mut Self { self.include_custom_values = false; self }` maybe as well so that people can opt out of hashing custom values :)

subxt/src/custom_values/custom_value_address.rs
```diff
@@ -19,3 +27,33 @@ impl CustomValueAddress for str {
...
+/// A static address to a custom value.
+pub struct StaticAddress<R> {
+    name: &'static str,
+    hash: [u8; 32],
+    phantom: PhantomData<R>,
+}
+
+impl<R> StaticAddress<R> {
+    /// Creates a new StaticAddress.
+    pub fn new(name: &'static str, hash: [u8; 32]) -> Self {
```
jsdw, line 40: For consistency with eg constants, can we call this `new_static` and make it `#[doc(hidden)]`. The reason for this is that "regular" users won't be able to provide a meaningful `hash` argument; only our static codegen can do that (and we reserve the right to change how we hash things, so allowing a user to save and provide hashes will just be a recipe for issues anyway :))

Let's also copy how constants is implemented by making the hash optional (ie `Option<[u8; 32]>`) and adding an `unvalidated()` function so that users of the static codegen have the option to opt out of the validation if they want to YOLO it.

metadata/src/lib.rs
```diff
@@ -641,35 +642,62 @@ pub struct RuntimeApiMethodParamMetadata {
...
+    pub fn name(&self) -> &str {
+        self.name
+    }
+
+    /// Calculates the hash for the CustomValueMetadata.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.type_id` is not registered in the provided type registry
+    pub fn get_hash(&self, registry: &PortableRegistry) -> [u8; HASH_LEN] {
```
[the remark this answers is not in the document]
jsdw: Ah, and, have a look at how we get the hashes for eg runtime APIs above. I think we should:

- Rename this to `hash` for consistency
- Keep a copy of the portable registry internally, so the user doesn't need to provide one (which doesn't make much sense since this is the metadata which should have all of that, and opens the door for inconsistency :))

testing/ui-tests/src/custom_values.rs
```diff
@@ -0,0 +1,70 @@
...
+        extrinsic,
+        ty: unit_ty,
+        apis: vec![],
+        outer_enums: OuterEnums {
+            call_enum_ty: unit_ty,
+            event_enum_ty: unit_ty,
+            error_enum_ty: unit_ty,
+        },
+        custom: CustomMetadata {
+            map: BTreeMap::from_iter([("Foo".into(), foo_value_metadata)]),
```
jsdw, line 65: Tiny suggestion; also add a "foo", just to make sure that nothing blows up when the same name is provided (at least, the same that would be code genned).

testing/ui-tests/src/lib.rs
```diff
@@ -51,6 +52,12 @@ fn ui_tests() {
             .build(dispatch_errors::metadata_array_dispatch_error()),
     );
 
+    t.pass(
+        m.new_test_case()
+            .name("custom_values_foo")
+            .build(custom_values::metadata_custom_values_foo()),
+    );
```
jsdw, line 59: I'm happy with this for now! 

I'll raise an issue for a future PR to discuss how to test this more fully (eg here we've tested that the codegen doesn't fail in a fairly simple case, but it'd be good to also test that we can fetch custom values via the generated interface). It might be a bit fiddly because we'd want to do generate some metadata with custom values in and then use the subxt macro on it to generate the interface and allow us to test it. Will ponder :)

testing/ui-tests/src/custom_values.rs
jsdw: Perhaps also a value or two with weird names to check that codegen doesn't choke on odd naming (eg "2foo" or "@hello")

testing/ui-tests/src/lib.rs
jsdw: I opened #1134, so after this merges we can add that :)

jsdw:
Looking good overall; just a bunch of small things to fix up re consistency or whatever, and I raised #1134 to do a little followup piece of work to spit out custom metadata for more thorough testing :)

subxt/src/custom_values/custom_values_client.rs
```diff
@@ -30,11 +30,15 @@ impl<T: Config, Client: OfflineClientT<T>> CustomValuesClient<T, Client> {
...
+
+        // 2. Attempt to decode custom value:
         let metadata = self.client.metadata();
-        let custom_value = metadata
-            .custom()
+        let custom = metadata.custom();
+        let custom_value = custom
             .get(address.name())
-            .ok_or_else(|| MetadataError::CustomValueNameNotFound(address.name().to_string()))?;
+            .expect("custom value is present, checked above; qed");
```
jsdw, line 41: It's only checked above if validation is enabled (ie static values with validation passed). If somebody passes some arbitrary `str` address it'll fail. So best to stick to the error you were returning before :)

subxt/src/custom_values/custom_values_client.rs
```diff
@@ -43,6 +47,31 @@ impl<T: Config, Client: OfflineClientT<T>> CustomValuesClient<T, Client> {
...
+        &self,
+        address: &Address,
+    ) -> Result<(), Error> {
+        let metadata = self.client.metadata();
+        if let Some(actual_hash) = address.validation_hash() {
+            let custom = metadata.custom();
+            let custom_value = custom
+                .get(address.name())
+                .ok_or_else(|| MetadataError::CustomValueNameNotFound(address.name().into()))?;
+            // Because of the check above this cannot panic:
```
jsdw, line 64: Nit: this comment can go away (we wouldn't expect a panic anyway now)

metadata/src/utils/validation.rs
jsdw: This also still should be added

jsdw (approved):
Awesome stuff, good job @tadeohepperle!

lexnv (approved):
👍

PR #1122, opened 2023-08-14
refactor(rpc): support default port in URL

niklasad1:
Updates jsonrpsee to v0.20 to support the default port number in URLs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ getrandom = { version = "0.2", default-features = false }
 hex = "0.4.3"
 heck = "0.4.1"
 impl-serde = { version = "0.4.0" }
-jsonrpsee = { version = "0.16" }
+jsonrpsee = { version = "0.20" }
 pretty_assertions = "1.4.0"
 primitive-types = { version = "0.12.1", default-features = false, features = ["codec", "scale-info", "serde"] }
 proc-macro-error = "1.0.4"
--- a/codegen/src/utils/fetch_metadata.rs
+++ b/codegen/src/utils/fetch_metadata.rs
@@ -6,7 +6,7 @@ use crate::error::FetchMetadataError;
 use codec::{Decode, Encode};
 use jsonrpsee::{
     async_client::ClientBuilder,
-    client_transport::ws::{Uri, WsTransportClientBuilder},
+    client_transport::ws::{Url, WsTransportClientBuilder},
     core::{client::ClientT, Error},
     http_client::HttpClientBuilder,
     rpc_params,
@@ -46,7 +46,7 @@ impl std::str::FromStr for MetadataVersion {
 
 /// Returns the metadata bytes from the provided URL, blocking the current thread.
 pub fn fetch_metadata_bytes_blocking(
-    url: &Uri,
+    url: Url,
     version: MetadataVersion,
 ) -> Result<Vec<u8>, FetchMetadataError> {
     tokio_block_on(fetch_metadata_bytes(url, version))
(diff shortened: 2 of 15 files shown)


codegen/src/utils/fetch_metadata.rs
```diff
@@ -97,29 +94,29 @@ pub async fn fetch_metadata_hex(
 }
 
 async fn fetch_metadata_ws(
-    url: &Uri,
+    url: Url,
     version: MetadataVersion,
 ) -> Result<Vec<u8>, FetchMetadataError> {
     let (sender, receiver) = WsTransportClientBuilder::default()
-        .build(url.to_string().parse::<Uri>().unwrap())
```
niklasad1, line 104: I removed this, it doesn't make sense to parse to URL twice instead better to just clone the URL?! :)

Thus, all APIs are changed to accept to URL by value instead of reference...

Cargo.toml
```diff
@@ -54,7 +54,7 @@ getrandom = { version = "0.2", default-features = false }
 hex = "0.4.3"
 heck = "0.4.1"
 impl-serde = { version = "0.4.0" }
-jsonrpsee = { version = "0.16" }
+jsonrpsee = { version = "0.20" }
```
jsdw, line 57: Wierd; I updated jsonrpsee to 0.19 not long ago; I wonder where those changes went!

jsdw: (Maybe I got distracted and never actually merged those changes, whoops!)

codegen/src/utils/fetch_metadata.rs
```diff
@@ -46,15 +46,15 @@ impl std::str::FromStr for MetadataVersion {
...
+    url: Url,
     version: MetadataVersion,
 ) -> Result<Vec<u8>, FetchMetadataError> {
     tokio_block_on(fetch_metadata_bytes(url, version))
 }
 
 /// Returns the raw, 0x prefixed metadata hex from the provided URL, blocking the current thread.
 pub fn fetch_metadata_hex_blocking(
-    url: &Uri,
+    url: Url,
```
jsdw, line 57: I wonder why passing the reference no longer works with `Url` but did with `Uri`? (this trickled up to meaning ectra clones and stuff in other places; no big deal but curious :))

codegen/src/utils/fetch_metadata.rs
jsdw: Hehe yeah, no idea why we were doing that :D

codegen/src/utils/fetch_metadata.rs
```diff
@@ -97,29 +94,29 @@ pub async fn fetch_metadata_hex(
...
 
 async fn fetch_metadata_http(
-    url: &Uri,
+    url: Url,
     version: MetadataVersion,
 ) -> Result<Vec<u8>, FetchMetadataError> {
     let client = HttpClientBuilder::default()
         .request_timeout(Duration::from_secs(180))
-        .build(url.to_string())?;
+        .build(url)?;
```
jsdw, line 119: Aah this is why the cloning was needed I guess!

subxt/src/client/lightclient/builder.rs
```diff
@@ -172,28 +172,26 @@ async fn fetch_url(url: impl AsRef<str>) -> Result<serde_json::Value, Error> {
...
-        if url.scheme_str() != Some("ws") && url.scheme_str() != Some("wss") {
+        if url.scheme() != "ws" && url.scheme() != "wss" {
             return Err(Error::LightClient(LightClientError::InvalidScheme));
         }
 
         let (sender, receiver) = ws_transport(url).await?;
 
-        Ok(ClientBuilder::default()
-            .max_notifs_per_subscription(4096)
+        Ok(Client::builder()
```
jsdw, line 189: nice!

jsdw (approved):
Awesome!

codegen/src/utils/fetch_metadata.rs
niklasad1: Yeah, if look at by comment above you can see that we are parsing an "already parsed URL" using its String representation :D

lexnv (approved):

PR #1126, opened 2023-08-18
Introduce Backend trait to allow different RPC (or other) backends to be implemented

jsdw:
Previously, Subxt looked something like:

```
high level APIs <- RPC calls
```

This PR introduces a `Backend` trait inbetween those; a trait which provides the minimal set of things that the high level APIs need (while trying to cater for what the new RPCs can provide). It then implements that trait for the "legacy" API methods, and migrates the high level code to using the trait rather than directly calling methods (in fact; making this impossible). So now we have:

```
high level APIs <- backend implementation <- RPC calls
```

(Where the RPC calls are just an implementation detail of the backend implementation)

I'd suggest to look at the `subxt/backend` folder first to see what this trait looks like, then the `subxt/backend/legacy` to see the `LegacyBackend` implementation, which is based on the old RPC methods.

The idea is that we can add a `v2` backend which is based on the new methods.

The examples are handy to look at next to see what's changed. Much of the rest of the code is fallout from adding this new backend thing!

There are a number of breaking changes here (mostly I'll cover below in the notes), but the high level APIs haven't changed significantly. Manually calling RPC methods is now more difficult, but still possible via creating an `RpcClient` manually, and using that to instantiate either/both of `LegacyRpcMethods` to make RPC calls and `OnlineClient` to do high level Subxt things. 

Random Notes:
- I've removed all of the `chainHead` RPC stuff at the moment. The plan is to put this stuff back as part of the new backend.
- `block.body().await.extrinsics()` is now just `block.extrinsics().await`; the body of a block only contains extrinsics anyway (the new API inparticular _only_ returns extrinsics, and not justifications, which we didn't expose anyway).
- Filling in gaps in finalized blocks is now handled by the backend, whose job it is to return a full stream.
- Things now take an `RpcClient` instead of an `Arc<impl RpcClientT>`. `RpcClient::from_url` replaces `default_rpc_client()`, or you can instantiate `RpcClient` from an underlying trait. I've done renames so that the raw trait stuff is better separated now (eg `Raw*` names) to make the difference clearer. One reason for this all is that calling RPC methods requires an RPC client, so it's easier if people can create a single `RpcClient` quite easily and then use that in the places it's needed).
- We have a new `BlockRef` type that, in the new backend, can try to "hold onto" a block, ie keep it from being unpinned, if it's alive. Things that may need to get details at a particular block should now keep the corresponding `BlockRef` around until they've finished. 
- Storage changes include removing the explicit "page_count" from the user API; the backend will now determine the page size to use if applicable, renaming `fetch_keys` to `fetch_keys_raw` (it returns raw bytes like fetch_raw), and making `iter()` hand back a proper `Stream` which should make it more ergonomic (and changes results from being `Result<Option<T>>` to `Option<Result<T>>`). Currently it fetches keys in batches but values individually; we can look to improve that again though if needbe.
- No more `client.rpc()`, because this is now an implementation detail of a specific `Backend`. Instead, you can `OnlineClient::from_backend` to pass a backend which has been instantiated with some RPC client that you can also use elsewhere.
- `tx.dry_run()` is now `tx.validate()` and uses a runtime call for validating (in part because the new RPCs won't expose a dry run call). This returns a different result though, leading to breakage here, so I've also kept the RPC call around so that one can manually call that instead if they need the old behaviour.
- `tx.submit()` used to directly call a submit RPC which handed the hash back. Now we delegate to `submit_and_watch` and return the hash once the first non-error status comes back, since the new APIs don't (currently) have a way to just submit a TX anyway.
- The underlying `TransactionProgress` events are altered to better fit the new API results (omitting a couple of details that the old API can't replicate). The high level API around submitting and watching is otherwise unchanged. 
- Stream-like things always have a `next()` method on them instead of eg `next_item()` now, which is mostly just a convenience to avoid needing to import `StreamExt` for this common op.


subxt/src/backend/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/// The status of the transaction.
+///
+/// If the status is [`TransactionStatus::InFinalizedBlock`], [`TransactionStatus::Error`],
+/// [`TransactionStatus::Invalid`] or [`TransactionStatus::Dropped`], then no future
+/// events will be emitted.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TransactionStatus<Hash> {
+    /// Transaction is part of the future queue.
```
tadeohepperle, line 310: I like the new `TransactionStatus`!

subxt/src/backend/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+use async_trait::async_trait;
+use codec::{Decode, Encode};
+use futures::{Stream, StreamExt};
+use std::pin::Pin;
+use std::sync::Arc;
+
+/// Prevent the backend trait being implemented externally.
+#[doc(hidden)]
+pub(crate) mod sealed {
+    pub trait Sealed {}
```
tadeohepperle, line 24: What is the reason to put `Sealed` into a separate `mod`? Would `pub(crate) trait Sealed {}` not suffice?

subxt/src/backend/legacy/rpc_methods.rs
```diff
@@ -0,0 +1,532 @@
...
+            .client
+            .request("state_getMetadata", rpc_params![at])
+            .await?;
+        let metadata = Metadata::decode(&mut &bytes[..])?;
+        Ok(metadata)
+    }
+
+    /// Fetch system health
+    pub async fn system_health(&self) -> Result<SystemHealth, Error> {
+        self.client.request("system_health", rpc_params![]).await
```
tadeohepperle, line 97: I did not know these RPC methods were available, maybe we should add an example, that showcases all the little things we can fetch from a node?

tadeohepperle (approved):
I looked over the entire thing, looks good. From my understanding we will add a newer backend soon as a replacement of the `LegacyBackend`? If that is the case, what will be the mechanism by which it is determined what backend is used when an `OnlineClient` is constructed?

subxt/src/backend/mod.rs
jsdw: Good question! This is a bit of a hack, because if you have `pub (crate) trait Sealed` on it's own then you run into:

```
crate-private trait `Sealed` in public interface
can't leak crate-private trait
```

But having the trait be fully `pub` (but in a module that people can't access, so the thing is pub but nobody can name it) circumvents this. It's weird, but it's a generally established pattern that crates use for this purpose :)

(I have only skimmed it but maybe eg https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/ is interesting; it also talks about another approach via methods that can't be implemented :))

subxt/src/backend/mod.rs
jsdw: It's basically much more similar to the one that the new APIs will return, so making this change now means that hopefully we won't need to change anything else as we implement the new backend :)

subxt/examples/blocks_subscribing.rs
```diff
@@ -24,8 +23,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         println!("  Extrinsics:");
 
         // Log each of the extrinsic with it's associated events:
-        let body = block.body().await?;
-        for ext in body.extrinsics().iter() {
+        let extrinsics = block.extrinsics().await?;
```
lexnv, line 26: Nice simplification for the end user!

If (ever) justification turns to be a thing in the future we could expose it on the same level as `extrinsics` (ie block.extrinsics())

subxt/src/tx/tx_client.rs
```diff
@@ -452,50 +455,323 @@ where
...
 }
+
+impl ValidationResult {
+    fn try_from_bytes(bytes: Vec<u8>) -> Result<ValidationResult, crate::Error> {
+        // TaggedTransactionQueue_validate_transaction returns this:
+        // https://github.com/paritytech/substrate/blob/0cdf7029017b70b7c83c21a4dc0aa1020e7914f6/primitives/runtime/src/transaction_validity.rs#L210
+        // We copy some of the inner types and put the three states (valid, invalid, unknown) into one enum,
+        // because from our perspective, the call was successful regardless.
+        if bytes[0] == 0 {
+            // ok: valid (more detail is available here, but ).
```
lexnv, line 574: nit: `ok: valid (more detail is available here, but we only care about the enum variant).`

subxt/src/tx/tx_client.rs
```diff
@@ -452,50 +455,323 @@ where
...
     }
 }
+
+impl ValidationResult {
+    fn try_from_bytes(bytes: Vec<u8>) -> Result<ValidationResult, crate::Error> {
+        // TaggedTransactionQueue_validate_transaction returns this:
+        // https://github.com/paritytech/substrate/blob/0cdf7029017b70b7c83c21a4dc0aa1020e7914f6/primitives/runtime/src/transaction_validity.rs#L210
+        // We copy some of the inner types and put the three states (valid, invalid, unknown) into one enum,
+        // because from our perspective, the call was successful regardless.
+        if bytes[0] == 0 {
```
lexnv, line 573: nit: In case we somehow don't receive an `Result<ValidTransaction, TransactionValidityError>` and end-up with some empty bytes, would it be beneficial to do a `bytes.get(0).expect("Invalid wire format for ValidationResult")`?

testing/integration-tests/src/full_client/client/mod.rs
```diff
@@ -442,227 +253,28 @@ async fn unsigned_extrinsic_is_same_shape_as_polkadotjs() {
...
-    Ok(())
-}
-
-#[tokio::test]
-async fn chainhead_unstable_follow() {
-    let ctx = test_context().await;
-    let api = ctx.client();
-
-    // Check subscription with runtime updates set on false.
-    let mut blocks = api.rpc().chainhead_unstable_follow(false).await.unwrap();
```
lexnv, line 470: I think this is fine since we plan to add them back under a dedicated backend for the Spec V2.

In the meanwhile, I remember this is used by [polakadot-interceptor here](https://github.com/paritytech/polkadot-introspector/blob/7284fc638d241d23063f4a1607fb158c76a856ff/essentials/src/api/subxt_wrapper.rs#L693-L701).  However I think, the same could be achieved with the new Subxt API (for reference @AndreiEres ) which will allow seamless transition between `chainHead` and the legacy RPC.

IIRC this was used to create a metric about the duration time between two reported blocks (either best / finalized), we could work together here to find an appropriate solution with the new API (we might merge the 2 streams for best and finalized)

lexnv (approved):
Amazing work here! 👍

testing/integration-tests/src/full_client/client/mod.rs
AndreiEres: Thank you for mentioning me! I should've expected it by `unstable` mark in `chainHead` call :-) 

So I see that I should switch our subscription in introspector to new API, right? Which one, btw?

lexnv: After we release the next subxt version, merging the [Backend::stream_best_block_headers](https://github.com/paritytech/subxt/pull/1126/files#diff-17bc33d8a1d22f8fcb082c5eff935dfa21b562c9d9133b520c357e9975e48bdaR170) and [Backend::stream_finalized_block_headers](https://github.com/paritytech/subxt/pull/1126/files#diff-17bc33d8a1d22f8fcb082c5eff935dfa21b562c9d9133b520c357e9975e48bdaR183) should produce the same information as the `chainHead`

jsdw: Would this call also work (if all you care about is the time between _any_ new block)?

```rust
api.blocks().subscribe_all().await
```

or 

```rust
api.blocks().subscribe_best().await
```

If you care about all new "best" blocks.

subxt/src/backend/mod.rs
tadeohepperle: Oh that is very interesting!

subxt/examples/blocks_subscribing.rs
jsdw: Yup definitely; if they are a thing then they won't be in the block body I guess anyway (since the body of a block afaik is just the extrinsics). I'm not actually sure where they live offhand; maybe theres a storage or runtime API call to fetch them (I haven't looked!)

subxt/src/tx/tx_client.rs
jsdw: Hehe I feel like I do this quite often; get distracted and forget to finish comments!

testing/integration-tests/src/full_client/client/mod.rs
AndreiEres: Thank you

PR #1127, opened 2023-08-18
Fix `subxt metadata` fallback code to work with default opts

### LINKED ISSUE #1125: `subxt metadata` does not work when talking to older nodes
I'm trying to download the metadata from a blockchain built with Substrate. I have a local node running and http://localhost:9933 is accessible: 

```
$ curl -H "Content-Type: application/json" -d '{"id":"1", "jsonrpc":"2.0", "method": "state_getMetadata", "params":[]}' http://localhost:9933 
{"jsonrpc":"2.0","result":"0x6d6574610e5906000c1c73705f636f72651863727970746f2c4163636f756e7449643332000004000401205b75383b2033325d0000040000032000000008000800000503000c08306672616d655f73797374656d2c4163636f756e74496e666f0814496e64657801102c4163636f756e74446174610114001401146e6f6e6365100114496e646578000124636f6e73756d657273100120526566436f756e7400012470726f7669646572

... truncated ...
```

Trying subxt fails:

```
$ subxt metadata --format bytes 
Error: 
   0: Other error: The node can only return version 14 metadata using the legacy API but you've asked for something else

Location:
   /home/gluwa/.cargo/registry/src/index.crates.io-6f17d22bba15001f/subxt-cli-0.31.0/src/utils.rs:73

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

jsdw:
Closes #1125

### DIFF
--- a/codegen/src/utils/fetch_metadata.rs
+++ b/codegen/src/utils/fetch_metadata.rs
@@ -201,14 +201,18 @@ async fn fetch_metadata(
         client: &impl ClientT,
         version: MetadataVersion,
     ) -> Result<Vec<u8>, FetchMetadataError> {
-        if !matches!(version, MetadataVersion::Version(14)) {
+        // If the user specifically asks for anything other than version 14 or "latest", error.
+        if !matches!(
+            version,
+            MetadataVersion::Latest | MetadataVersion::Version(14)
+        ) {
             return Err(FetchMetadataError::Other(
                 "The node can only return version 14 metadata using the legacy API but you've asked for something else"
                     .to_string(),
             ));
         }
 
-        // Fetch the metadata at that version:
+        // Fetch the metadata.
         let metadata_string: String = client
             .request("state_call", rpc_params!["Metadata_metadata", "0x"])
             .await?;
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -24,15 +24,11 @@ pub struct FileOrUrl {
     pub file: Option<PathBuf>,
     /// Specify the metadata version.
     ///
-    ///  - unstable:
+    ///  - "latest": Use the latest stable version available.
+    ///  - "unstable": Use the unstable metadata, if present.
+    ///  - a number: Use a specific metadata version.
     ///
-    ///    Use the latest unstable metadata of the node.
-    ///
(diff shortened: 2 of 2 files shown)


codegen/src/utils/fetch_metadata.rs
```diff
@@ -201,14 +201,18 @@ async fn fetch_metadata(
...
         version: MetadataVersion,
     ) -> Result<Vec<u8>, FetchMetadataError> {
-        if !matches!(version, MetadataVersion::Version(14)) {
+        // If the user specifically asks for anything other than version 14 or "latest", error.
+        if !matches!(
+            version,
+            MetadataVersion::Latest | MetadataVersion::Version(14)
+        ) {
             return Err(FetchMetadataError::Other(
                 "The node can only return version 14 metadata using the legacy API but you've asked for something else"
```
niklasad1, line 210: shouldn't this message say that it supports "v14" or "latest"?

jsdw: Only V14 can be provided from this legacy route, so whether they ask for V14 or "latest that node has" that's what they will get back, so I htink it's all good?

lexnv (approved):
Looks good to me, I wonder if it's worth checking that the `Latest` version from the node is 14 / or if the API to fetch the version returns `MethodNotIMplemented`?

Ideally, in the future we'd have the V15 nodes. The `inner_legacy` method could still fetch V14 when the V15 code path fails (with maybe some network error). I think that's fine since it should be considered a fallback :D

codegen/src/utils/fetch_metadata.rs
niklasad1: ok, I see didn't understand that :P

niklasad1 (approved):

jsdw:
> Looks good to me, I wonder if it's worth checking that the Latest version from the node is 14 / or if the API to fetch the version returns MethodNotIMplemented?

My reasoning was that if we are trying this legacy path, then we have already failed to call the "new" methods and so the node only supports V14 (please do correct me if I'm wrong though :))

PR #1137, opened 2023-08-24
Unstable Backend: add back the old chainHead RPCs/tests

jsdw:
This PR isn't as big as it looks; new `rustfmt` understands `let ... else` statements now, and otherwise seemed to want to reformat quite a lot! See below for actual changes.

- I put back the chainHead methods and tests.
- I also added back any legacy RPCs that I removed in the last PR; I decided that there's probably enough breaking changes already, and so want to give people using those RPCs a slightly smoother migration.
- I also added some other of the new APIs including to submit transactions, and the now stable `chainSpec` methods.

The only real changes in terms of files are:
- `subxt/src/backend/unstable/` - add unstable RPC methods back
- `testing/integration-tests/src/full_client/client/legacy_rpcs.rs` - for the tests

I also removed the `#[deny(unused_dependencies)]` thing from the integration-tests crate, because IMO it's not really important there and we had more and more `use foo as _;` things depending on different features being enabled or not.

### DIFF
--- a/metadata/src/lib.rs
+++ b/metadata/src/lib.rs
@@ -704,10 +704,6 @@ impl<'a> CustomValueMetadata<'a> {
     }
 
     /// Calculates the hash for the CustomValueMetadata.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `self.type_id` is not registered in the provided type registry
     pub fn hash(&self) -> [u8; HASH_LEN] {
         let mut cache = HashMap::new();
         get_custom_value_hash(self, &mut cache)
--- a/subxt/src/backend/legacy/rpc_methods.rs
+++ b/subxt/src/backend/legacy/rpc_methods.rs
@@ -73,6 +73,35 @@ impl<T: Config> LegacyRpcMethods<T> {
         Ok(data.into_iter().map(|b| b.0).collect())
     }
 
+    /// Query historical storage entries
+    pub async fn state_query_storage(
+        &self,
+        keys: impl IntoIterator<Item = &[u8]>,
+        from: T::Hash,
+        to: Option<T::Hash>,
+    ) -> Result<Vec<StorageChangeSet<T::Hash>>, Error> {
+        let keys: Vec<String> = keys.into_iter().map(to_hex).collect();
+        let params = rpc_params![keys, from, to];
+        self.client
+            .request("state_queryStorage", params)
+            .await
+            .map_err(Into::into)
+    }
+
+    /// Query historical storage entries
+    pub async fn state_query_storage_at(
+        &self,
+        keys: impl IntoIterator<Item = &[u8]>,
+        at: Option<T::Hash>,
+    ) -> Result<Vec<StorageChangeSet<T::Hash>>, Error> {
+        let keys: Vec<String> = keys.into_iter().map(to_hex).collect();
+        let params = rpc_params![keys, at];
+        self.client
+            .request("state_queryStorageAt", params)
+            .await
+            .map_err(Into::into)
+    }
+
     /// Fetch the genesis hash
     pub async fn genesis_hash(&self) -> Result<T::Hash, Error> {
         let block_zero = 0u32;
@@ -156,6 +185,32 @@ impl<T: Config> LegacyRpcMethods<T> {
         let params = rpc_params![hash];
         let block = self.client.request("chain_getBlock", params).await?;
         Ok(block)
+    }
+
+    /// Reexecute the specified `block_hash` and gather statistics while doing so.
+    ///
+    /// This function requires the specified block and its parent to be available
+    /// at the queried node. If either the specified block or the parent is pruned,
+    /// this function will return `None`.
+    pub async fn dev_get_block_stats(
+        &self,
+        block_hash: T::Hash,
+    ) -> Result<Option<BlockStats>, Error> {
+        let params = rpc_params![block_hash];
+        let stats = self.client.request("dev_getBlockStats", params).await?;
+        Ok(stats)
+    }
+
+    /// Get proof of storage entries at a specific block's state.
+    pub async fn state_get_read_proof(
+        &self,
+        keys: impl IntoIterator<Item = &[u8]>,
+        hash: Option<T::Hash>,
+    ) -> Result<ReadProof<T::Hash>, Error> {
+        let keys: Vec<String> = keys.into_iter().map(to_hex).collect();
+        let params = rpc_params![keys, hash];
+        let proof = self.client.request("state_getReadProof", params).await?;
+        Ok(proof)
     }
 
     /// Fetch the runtime version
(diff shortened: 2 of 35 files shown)


subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+pub struct NewBlock<Hash> {
+    /// The hash of the new block.
+    pub block_hash: Hash,
+    /// The parent hash of the new block.
+    pub parent_block_hash: Hash,
+    /// The runtime version of the new block.
+    ///
+    /// # Note
+    ///
+    /// This is present only if the `with_runtime` flag is set for
```
tadeohepperle, line 465: Do you mean with `with_runtime` the argument `runtime_updates: bool` that is passed to `chainhead_unstable_follow()`? I think we should use a uniform name adressing it throughout the code. 
Since it has an effect on the data in many struct, we could also think about making the `FollowEvent` generic over this flag (maybe with const generics). Just an idea. :)

tadeohepperle (approved):
Looks very nice, just left a few comments :)

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+                "chainHead_unstable_unpin",
+                rpc_params![subscription_id, hash],
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    /// Return the genesis hash.
+    pub async fn chainspec_v1_genesishash(&self) -> Result<T::Hash, Error> {
```
niklasad1, line 255:
```suggestion
    pub async fn chainspec_v1_genesis_hash(&self) -> Result<T::Hash, Error> {
```

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+        let hash = self
+            .client
+            .request("chainSpec_v1_genesisHash", rpc_params![])
+            .await?;
+        Ok(hash)
+    }
+
+    /// Return a string containing the human-readable name of the chain.
+    pub async fn chainspec_v1_chainname(&self) -> Result<String, Error> {
```
niklasad1, line 264:
```suggestion
    pub async fn chainspec_v1_chain_name(&self) -> Result<String, Error> {
```

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn can_deserialize_apis_from_tuple_or_object() {
```
niklasad1, line 835: :+1:

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+    }
+}
+
+fn to_hex(bytes: impl AsRef<[u8]>) -> String {
+    format!("0x{}", hex::encode(bytes.as_ref()))
+}
+
+/// Attempt to deserialize either a string or integer into an integer.
+/// See <https://github.com/paritytech/json-rpc-interface-spec/issues/83>
+pub(crate) mod unsigned_number_as_string {
```
niklasad1, line 724: add a simple test for this as well?

subxt/src/config/mod.rs
```diff
@@ -82,7 +82,7 @@ pub trait Hasher {
 }
 
 /// This represents the block header type used by a node.
-pub trait Header: Sized + Encode {
+pub trait Header: Sized + Encode + Decode {
```
niklasad1, line 85: why is `Decode` needed now?

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -35,19 +35,6 @@ use subxt::{
 };
 use subxt_metadata::Metadata;
 
-// We don't use these dependencies.
-use assert_matches as _;
```
niklasad1, line 39: :+1:

metadata/src/lib.rs
```diff
@@ -704,10 +704,6 @@ impl<'a> CustomValueMetadata<'a> {
     }
 
     /// Calculates the hash for the CustomValueMetadata.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `self.type_id` is not registered in the provided type registry
```
niklasad1, line 710: Is this not true anymore?

Can't find anything in this PR about this :P

niklasad1 (approved):
Looks good

subxt/src/backend/unstable/rpc_methods.rs
jsdw: Yeah; we should prolly call it `with_runtime` elsewhere too because the spec calls it `withRuntime`!

I guess generics could be used to "guarantee" that we get a RuntimeEvent if we ask for it, and not if we don't, maybe? But the generics might complicate things more than it's worth because they would bleed out to anything else that cared about this stuff :)

metadata/src/lib.rs
lexnv: I think the `pub fn get_type_hash` would still panic if the `id` passed as param is not present in the `PortableRegistry` 🤔 

I belive the `CustomValueMetadata` could only panic in practice here if the metadata is broken, since that's where our value is coming from

subxt/src/backend/legacy/rpc_methods.rs
```diff
@@ -268,6 +323,49 @@ impl<T: Config> LegacyRpcMethods<T> {
...
+        let params = rpc_params![key_type, suri, Bytes(public)];
+        self.client.request("author_insertKey", params).await?;
+        Ok(())
+    }
+
+    /// Generate new session keys and returns the corresponding public keys.
+    pub async fn author_rotate_keys(&self) -> Result<Vec<u8>, Error> {
+        let bytes: Bytes = self
+            .client
+            .request("author_rotateKeys", rpc_params![])
```
lexnv, line 342: Hmm, are we adding those back to the legacy API since they might be useful for validators? 🤔

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -0,0 +1,890 @@
...
+    /// runtime API calls at this block ([`chainhead_unstable_call`](UnstableRpcMethods::chainhead_unstable_call)).
+    ///
+    /// # Note
+    ///
+    /// When the user is no longer interested in a block, the user is responsible
+    /// for calling the [`chainhead_unstable_unpin`](UnstableRpcMethods::chainhead_unstable_unpin) method.
+    /// Failure to do so will result in the subscription being stopped by generating the `Stop` event.
+    pub async fn chainhead_unstable_follow(
+        &self,
+        runtime_updates: bool,
```
lexnv: nit: I think this is now named `with_runtime` from [spec](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/chainHead_unstable_follow.md)

subxt/src/config/mod.rs
jsdw: In the old API methods, header came back as JSON, but in the new one it comes back as hex encoded SCALE encoded bytes instead :)

metadata/src/lib.rs
jsdw: The comment was left over from when this call was taking in a `PortableRegistry` as an argument (which might then have different types from what he metadata knew about and thus could panic). That was since removed but I noticed that the comment wasn't updated so I tweaked it :)

lexnv (approved):

subxt/src/backend/unstable/rpc_methods.rs
jsdw: Maybe you're right; I went back and forth on this naming between "what feels nicest" and "what's more technically correct" :)

subxt/src/backend/unstable/rpc_methods.rs
jsdw: Yup I should def do this!

subxt/src/backend/unstable/rpc_methods.rs
tadeohepperle: Yeah I agree that we should not complicate it more.

subxt/src/backend/unstable/rpc_methods.rs
jsdw: Yup; I renamed it :)

subxt/src/backend/unstable/rpc_methods.rs
jsdw: (I did the renames :))

PR #1146, opened 2023-08-28
ci: Apply cargo fmt

lexnv:


niklasad1 (approved):
new rust version? :)

lexnv:
Yep, rustfmt now can handle `let Some(..) = {} else {}` properly

tadeohepperle (approved):

PR #1147, opened 2023-08-29
Add proper test Custom values 

### LINKED ISSUE #1134: Custom Values: add proper tests of the interface
In #1117 we add a statically generated interface for custom values. We test that the codegen doesn't blow up, but should also test that generated interface itself works.

Probably we want to 

1. write a script that can spit out metadata that contains a bunch of custom values in (the custom values are fully contained in the metadata, so it doesn't matter about the rest of this metadata; it can just be empty).
2. Use the subxt macro pointed at this `artifacts/metadata_with_custom_values.scale` to generate an interface in some test file
3. Try fetching values via static and dynamic interface to check that it all works as hoped.

(1) is a bit tricky; currently we have a `scripts/artifacts.sh` file to generate them. I'd lean towards adding a small `testing/generate_custom_metadata` binary crate that can spit it out, and then having this script call `cargo run` on it eg `cargo run --bin testing-generate-custom-metadata > artifacts/metadata_with_custom_values.scale`.

tadeohepperle:
fixes #1134.

I added a test that uses some generated metadata to get custom values via an `OfflineClient`:
```rs
// static query:
let foo_address = node::custom().foo();
let foo = api.custom_values().at(&foo_address)?;
assert_eq!(foo, expected_foo);

// dynamic query:
let foo_value = api.custom_values().at("Foo")?;
let foo: Foo = foo_value.as_type()?;
assert_eq!(foo, expected_foo);
```


testing/generate-custom-metadata/src/lib.rs
```diff
@@ -1,15 +1,14 @@
-// Copyright 2019-2023 Parity Technologies (UK) Ltd.
```
lexnv, line 1: nit: I think we could keep the copyright here (everywhere else)

codegen/src/api/custom_values.rs
```diff
@@ -47,14 +47,15 @@ fn generate_custom_value_fn(
     if fn_names_taken.contains(&fn_name) {
         return None;
     }
-    let fn_name_ident = format_ident!("{fn_name}");
+    // if the fn_name would be an invalid ident, return None:
```
lexnv, line 50: I do wonder, would this happen when the function name is comming from the chain metadata (substrate/polkadot)?
I believe the `ident` parsing would fail when the input is not a valid variable name or keyword.
We could probably take this approach everywhere else in the codegen, since I guess `Ident::new` would panic for those cases too.

tadeohepperle: For custom values I wanted to be extra sure, because people can put in there whatever they want. I think the rest of the metadata is at least valid in terms of identifiers, because they are already used as identifiers on the substrate side, where the metadata is created. But we could definitely use `syn::parse_str::<syn::Ident>` instead of `format_ident!` in a few places.

testing/generate-custom-metadata/src/main.rs
```diff
@@ -0,0 +1,14 @@
...
+// see LICENSE for license details.
+
+use codec::Encode;
+
+/// can be called from the root of the project with: `cargo run --bin generate-custom-metadata`.
+/// Generates a "./artifacts/metadata_with_custom_values.scale" file.
+fn main() {
+    let metadata_prefixed = generate_custom_metadata::metadata_custom_values_foo();
+    let bytes = metadata_prefixed.encode();
+    std::fs::write("./artifacts/metadata_with_custom_values.scale", bytes)
```
jsdw, line 12: Can we add a call to this to `/scripts/artifacts.sh` (ie `cargo run --bin generate-custom-metadata > artifacts/metadata_with_custom_values.scale`), so that running that script updates all of our artifacts :)

(also, let's either allow the caller to decide whether to write the file to, or write the file to stdout, so that the file location can be specified in `artifacts.sh` so it's clear from that script what is being generated)

tadeohepperle: Oh yes, that is a great idea!

codegen/src/api/custom_values.rs
jsdw: It's a good point that we should only use `format_ident` when we are confident that the value must be an ident :) Good catch changing this to `parse_str` here!

testing/generate-custom-metadata/src/lib.rs
```diff
@@ -48,7 +54,7 @@ pub fn metadata_custom_values_foo() -> RuntimeMetadataPrefixed {
     let unit_ty = registry.register_type(&meta_type::<()>());
 
     // Metadata needs to contain this DispatchError, since codegen looks for it.
-    registry.register_type(&meta_type::<crate::utils::dispatch_error::ArrayDispatchError>());
+    registry.register_type(&meta_type::<dispatch_error::ArrayDispatchError>());
 
     let metadata = RuntimeMetadataV15 {
```
jsdw, line 59: Below, we add some `CustomMetadata` with a calid Foo value inside it.

Could we also add a custom value with a random type ID that isn't in use (eg `u32::MAX`); people might opt to do this too and we want to be sure that we can still access the bytes and that nothing will panic (just Err) if we try decoding them.

jsdw:
Looks good; just a couple of small comments! I'm not super keen on the dispatch error types being exported from the new generate-custom-values crate (it's a bit of a random place for them), but I'm not too bothered for now (I think ultimately a `test-utils` crate to dump various helper code like this is probably the way to go ultimately)

testing/generate-custom-metadata/src/lib.rs
tadeohepperle: In that case we just test if the bytes are returned from the dynamic interface right? E.g. we do not generate any static custom value address. Or should we generate a static function that has a return type of `Vec<u8>`?

codegen/src/types/mod.rs
```diff
@@ -117,6 +117,11 @@ impl<'a> TypeGenerator<'a> {
             .clone()
     }
 
+    /// Returns type registry.
+    pub fn types(&self) -> &PortableRegistry {
+        self.type_registry
+    }
```
jsdw, line 123: Nit: Instead of exposing this in this TypeGenerator, let's just pass the `PortableRegistry` into the function instead (it's accessible as something like `self.metadata.types()` in the parent function). Not a big deal though!

metadata/src/utils/validation.rs
```diff
@@ -405,15 +405,28 @@ pub fn get_custom_metadata_hash(custom_metadata: &CustomMetadata) -> [u8; HASH_L
...
     custom_value: &CustomValueMetadata,
     cache: &mut HashMap<u32, CachedHash>,
 ) -> [u8; HASH_LEN] {
-    concat_and_hash3(
-        &hash(custom_value.name.as_bytes()),
-        &get_type_hash(custom_value.types, custom_value.type_id(), cache),
-        &hash(custom_value.bytes()),
-    )
+    let name_hash = hash(custom_value.name.as_bytes());
+    let bytes_hash =hash(custom_value.bytes());
```
jsdw, line 416:
```suggestion
    let bytes_hash = hash(custom_value.bytes());
```

jsdw, line 416: The goal of this validation stuff is to make sure that the code generated from some metadata is still compatible with some node (and the node metadata) so that it can still be used without issue.

With that in mind, we don't actually need to check that `custom_value.bytes()` is identical; it doesn't matter if the bytes are different so long as they still decode OK into the generated type that the codegen is expecting. So let's remove the hashing of bytes. Otherwise looks good offhand (though this stuff always needs some thought and I obv missed this last time :))

subxt/src/custom_values/custom_value_address.rs
```diff
@@ -10,6 +10,9 @@ use crate::metadata::DecodeWithMetadata;
 pub trait CustomValueAddress {
     /// The type of the custom value.
     type Target: DecodeWithMetadata;
+    /// Should be set to `Yes` for Dynamic values and static values that have a valid type.
+    /// Should be `()` for custom values, that have an invalid type id.
+    type Decodable;
```
jsdw, line 15: For consistency with storage, how about `IsDecodable`?

subxt/src/custom_values/custom_values_client.rs
```diff
@@ -48,6 +48,24 @@ impl<T: Config, Client: OfflineClientT<T>> CustomValuesClient<T, Client> {
...
+
+        // 2. Return the underlying bytes:
+        let metadata = self.client.metadata();
+        let custom = metadata.custom();
+        let custom_value = custom
+            .get(address.name())
+            .ok_or_else(|| MetadataError::CustomValueNameNotFound(address.name().to_string()))?;
+
+        Ok(custom_value.bytes().to_vec())
+    }
```
jsdw, line 67: Looks good! Shame about needing to `to_vec()`, but I can see it's because the metadata is cloned and not referenced; maybe something we can address elsewhere at some point :)

testing/ui-tests/Cargo.toml
```diff
@@ -10,8 +10,10 @@ publish = false
 
 [dev-dependencies]
 trybuild = { workspace = true }
+hex = { workspace = true }
 scale-info = { workspace = true, features = ["bit-vec"] }
 frame-metadata ={ workspace = true }
 codec = { package = "parity-scale-codec", workspace = true, features = ["derive", "bit-vec"] }
-subxt = { workspace = true, features = ["native", "jsonrpsee"] }
+subxt = { workspace = true, features = ["native", "jsonrpsee", "substrate-compat"] }
```
jsdw, line 17: Where is `substrate-compat` needed? Offhand I can't see any use for it but I'm likely to be missing something :)

jsdw:
Looking good, and I like the new `bytes_at()` approach for valiues without valid type IDs!

I left a few comments which are mostly just little things, but with these small bits looked at I'd be happy to see this merge:

- Decodable -> IsDecodable
- Don't hash the bytes in the validation check (name and type should suffice)
- Confirm whether `substrate-compat` feature is needed in the subxt dev-dep
- Check on the test failure; I'm not quite sure what that's about so would be good to understand it :)

tadeohepperle:
> Confirm whether `substrate-compat` feature is needed in the subxt dev-dep

You are right, it is not necessary, I changed an import to `config::substrate::H256` and now it works without it. I found the cause of the test failure (older custom value metadata) and now it should be okay.

jsdw (approved):
The changes look great to me; nice one!

scripts/artifacts.sh
```diff
@@ -17,5 +17,7 @@ cargo run --bin subxt metadata --version 15 > artifacts/polkadot_metadata_full.s
 cargo run --bin subxt codegen --file artifacts/polkadot_metadata_full.scale | rustfmt > testing/integration-tests/src/full_client/codegen/polkadot.rs
 # generate a metadata file that only contains a few pallets that we need for our examples.
 cargo run --bin subxt metadata --file artifacts/polkadot_metadata_full.scale --pallets "Balances,Staking,System,Multisig,Timestamp,ParaInherent" > artifacts/polkadot_metadata_small.scale
-# generate a metadata file that only contains no pallets
+# generate a metadata file that contains no pallets
 cargo run --bin subxt metadata --file artifacts/polkadot_metadata_full.scale --pallets "" > artifacts/polkadot_metadata_tiny.scale
+# generate a metadata file that only contains some custom metadata
+cargo run --bin generate-custom-metadata > artifacts/metadata_with_custom_values.scale
```
lexnv, line 23: nit: Could use an extra line at the end of the script

testing/ui-tests/Cargo.toml
```diff
@@ -10,8 +10,10 @@ publish = false
 
 [dev-dependencies]
 trybuild = { workspace = true }
+hex = { workspace = true }
 scale-info = { workspace = true, features = ["bit-vec"] }
 frame-metadata ={ workspace = true }
 codec = { package = "parity-scale-codec", workspace = true, features = ["derive", "bit-vec"] }
 subxt = { workspace = true, features = ["native", "jsonrpsee"] }
 subxt-metadata = { workspace = true }
+generate-custom-metadata = { path = "../generate-custom-metadata" }
```
lexnv, line 19: nit: extra line

lexnv (approved):
LGTM! Nice one!

Issue #1140, opened 2023-08-27
Strange behavior calling a ink! contract using subxt


gcp-development:
I am using this [example](https://github.com/paritytech/subxt/blob/master/testing/integration-tests/src/full_client/frame/contracts.rs#L146) to call a contract.

I am using the below instructions to pass an argument:
(Note:I have try the name of the function also.)
```
 let mut input_data = Vec::<u8>::new();
 input_data.append(&mut (&blake2_256("0x633aa551".as_bytes())[..]).to_vec());
```
My ABI json of the function that I am calling.(This is the [flipper](https://github.com/paritytech/ink-examples/tree/main/flipper) example)

```
    "messages": [
      {
        "args": [],
        "default": false,
        "docs": [
          " Flips the current value of the Flipper's boolean."
        ],
        "label": "flip",
        "mutates": true,
        "payable": false,
        "returnType": {
          "displayName": [
            "ink",
            "MessageResult"
          ],
          "type": 1
        },
        "selector": "0x633aa551"
      }
    ]
```

Probably I am doing something wrong , but I am not able to make a call to that function. 

It gives an error like this:
(this is only the first block)

![image](https://github.com/paritytech/subxt/assets/76512851/793b7c6c-c4a8-4415-b81a-c69a8b8f0d5c)

gcp-development:
Hi, @tadeohepperle 

I have created a repo in[ here](https://github.com/gcp-development/dummy-contract).

I am "cruising" in a ubuntu. 

![image](https://github.com/paritytech/subxt/assets/76512851/6c22f5b3-9f91-4210-865a-b282fb73acb3)


``
rustup show
``

![image](https://github.com/paritytech/subxt/assets/76512851/bb5b0aff-8036-4982-8cd3-12be7c50b715)


Cheers,

lexnv:
Hey,

I'm running into some issues reproducing this. From my side, the failure comes from `let uri = SecretUri::from_str(DEVICE_ACCOUNT_SEEDS).expect("valid URI");`.

Could you provide the complete `ModuleError` in text? (from the issue description)

Thanks,
Alex

gcp-development:
Hi @lexnv ,

could you tell me what is the error that is coming up for you, please ?

The dump of the error is in [here](https://github.com/gcp-development/dummy-contract/blob/main/error.txt)

Cheers,

gcp-development:
Hello, @lexnv, @tadeohepperle, @jsdw 

I hope all is well.

[...]

PR #1153, opened 2023-09-07
Custom `Debug` impl for `DispatchError`

ascjones:
Currently the `Debug` implementation prints the whole `metadata` which fills the screen and doesn't help much:

image

This PR adds a custom `Debug` impl which instead resolves to virtual `pallet` and `error` fields similar to the `Display` impl. To know what the actual error is extremely useful :smile:

See also https://github.com/paritytech/subxt/issues/1140#issuecomment-1706195275

### DIFF
--- a/subxt/src/error/dispatch_error.rs
+++ b/subxt/src/error/dispatch_error.rs
@@ -122,7 +122,7 @@ pub enum TransactionalError {
 }
 
 /// Details about a module error that has occurred.
-#[derive(Clone, Debug, thiserror::Error)]
+#[derive(Clone, thiserror::Error)]
 #[non_exhaustive]
 pub struct ModuleError {
     metadata: Metadata,
@@ -141,6 +141,22 @@ impl PartialEq for ModuleError {
 }
 
 impl Eq for ModuleError {}
+
+/// Custom `Debug` implementation, ignores the very large `metadata` field, using it instead (as
+/// intended) to resolve the actual pallet and error names. This is much more useful for debugging.
+impl Debug for ModuleError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Ok(details) = self.details() else {
+            return f
+                .write_str("Unknown pallet error (pallet and error details cannot be retrieved)");
+        };
+        f.debug_struct("ModuleError")
+            .field("bytes", &self.bytes)
+            .field("pallet", &details.pallet.name())
+            .field("error", &details.variant.name)
+            .finish()
+    }
+}
 
 impl std::fmt::Display for ModuleError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


subxt/src/error/dispatch_error.rs
```diff
@@ -142,6 +142,22 @@ impl PartialEq for ModuleError {
 
 impl Eq for ModuleError {}
 
+/// Custom `Debug` implementation, ignores the very large `metadata` field, using it instead (as
+/// intended) to resolve the actual pallet and error names. This is much more useful for debugging.
+impl Debug for ModuleError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Ok(details) = self.details() else {
+            return f
+                .write_str("Unknown pallet error (pallet and error details cannot be retrieved)");
```
jsdw, line 151: Maybe better to have?:

```rust
write!(f, "ModuleError(<Unknown: pallet and error details cannot be retrieved>)")
```

So that it plays well if it ends up as part of a larger Debug call with other types etc involved

jsdw: I might also be tempted to havea similar format below like:

```
write!(f, "ModuleError(<{pallet_name}::{variant_name}>")
```

To be consistent with above and be clear that we aren't returning the internal struct layout, and instead something a bit custom? :)

jsdw: (But thank you; I love this! The default debug impl was stupid :D

ascjones: In both cases I would still expect to see the `bytes` though? How would that look?

ascjones: I've just added the `bytes` to the error in case the details cannot be resolved.

jsdw (approved):
Looks great, thank you!

subxt/src/error/dispatch_error.rs
jsdw: Sorry for the slow reply; was off on a team retreat! Looks great to me!

subxt/src/error/dispatch_error.rs
```diff
@@ -142,6 +142,26 @@ impl PartialEq for ModuleError {
...
+                pallet_name = details.pallet.name(),
+                variant_name = details.variant.name,
+            ),
+            Err(_) => format!(
+                "Unknown pallet error '{bytes:?}' (pallet and error details cannot be retrieved)",
+                bytes = self.bytes
+            ),
+        };
+
+        write!(f, "ModuleError(<{details}>)")
```
lexnv, line 161: nit: Would it be beneficial to have this syntax `ModuleError<..>` also for the `Display` impl?

lexnv (approved):

subxt/src/error/dispatch_error.rs
ascjones: Fine by me, wdyt @jsdw?

jsdw: I think the display impl can be "prettier" (it doesn't need to fit in with the rust like syntax that everything else spits out for debug impls), but perhaps we should indeed standardise these; maybe Debug should print `ModuleError(<{details}>)` as it does here and Display should just print `{details}`, so an inner func to write details and then Debug just wraps to make fit better with rusty syntax?

ascjones: Done

PR #1154, opened 2023-09-08
Make the codegen more accessible for running in WASM

tadeohepperle:
This PR makes some adjustments in subxt-codegen to make it WASM-compilable.
- added a `fetch_metadata` feature flag (enabled by default, when disabled, no jsonrpsee, frame-metadata and tokio are needed as dependencies anymore which makes it easier to compile to WASM)
- added a utility function `value_ty` on storage entries, which makes the code more concise in a few places.
-  expose `types()` function on the `RuntimeTypeGenerator` to have easier access to the metadata.


metadata/src/lib.rs
```diff
@@ -395,6 +395,23 @@ pub enum StorageEntryType {
     },
 }
 
+impl StorageEntryType {
+    /// returns the type of the value.
```
lexnv, line 399: small nit: Could start with uppercase here to be consistent with docs / fn below

metadata/Cargo.toml
```diff
@@ -14,7 +14,9 @@ homepage.workspace = true
 description = "Command line utilities for checking metadata compatibility between nodes."
 
 [dependencies]
-codec = { package = "parity-scale-codec", workspace = true, features = ["derive"] }
+codec = { package = "parity-scale-codec", workspace = true, features = [
+    "derive",
```
lexnv, line 18: dq: I do like the prior formatting more since I find one-liners easier to follow. Did you find any good formatters for .toml or its just personal preference?

codegen/Cargo.toml
```diff
@@ -12,19 +12,32 @@ documentation = "https://docs.rs/subxt-codegen"
 homepage.workspace = true
 description = "Generate an API for interacting with a substrate node from FRAME metadata"
 
+[features]
+default = ["fetch_metadata"]
+fetch_metadata = ["dep:jsonrpsee", "dep:tokio", "dep:frame-metadata"]
+web = ["getrandom/js"]
```
lexnv, line 18: Is the `web` feature compatible with `fetch_metadata`?

tadeohepperle: Ah, no it isn't. I'm gonna make sure we get a compile error when they are combined.

metadata/Cargo.toml
tadeohepperle: Oh, I use the [`Even Better TOML`](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml) VSCode extension for formatting. Let me revert it.

codegen/src/api/mod.rs
```diff
@@ -233,7 +237,7 @@ impl RuntimeGenerator {
     /// unique path, so that types with matching paths don't end up overwriting each other
     /// in the codegen. We ignore any types with generics; Subxt actually endeavours to
     /// de-duplicate those into single types with a generic.
-    fn ensure_unique_type_paths(metadata: &mut Metadata) {
+    pub fn ensure_unique_type_paths(metadata: &mut Metadata) {
```
jsdw, line 240: Why is this pub? (It's called on `new` above, so does exposing it help with something?)

codegen/src/types/mod.rs
```diff
@@ -262,6 +262,11 @@ impl<'a> TypeGenerator<'a> {
         self.derives.default_derives()
     }
 
+    /// Returns the type registry.
+    pub fn types(&self) -> &PortableRegistry {
+        self.type_registry
+    }
```
jsdw, line 268: This looks familiar :) is it needed here?

codegen/src/api/storage.rs
```diff
@@ -173,6 +170,7 @@ fn primitive_type_alias(type_path: &TypePath) -> TokenStream {
     quote!(#type_path)
 }
 
+#[cfg(feature = "fetch_metadata")]
```
jsdw, line 173: I can't quite see why this is needed offhand; we don't fetch any metadata below?

tadeohepperle: Oh, it is just used here, because the `frame_metadata` crate that is used in the tests, is only a dependency if the "fetch_metadata" feature is enabled. Otherwise we do not need it as a dependency.

codegen/src/types/mod.rs
jsdw: Okeydoke; I'm happy to leave it be then!

codegen/src/api/mod.rs
jsdw: Ah I see!

Perhaps we should expose this function on our `Metadata` then instead, so that optionally one can make this be the case there too. It's a bit of a weird one; I pondered just having it run on the `Metadata` automatically but wasn't entirely comfortable with altering these things by default.

So yeah, let's add it on `Metadata` and then just call it here as well as wherever you need it :)

codegen/Cargo.toml
```diff
@@ -12,19 +12,30 @@ documentation = "https://docs.rs/subxt-codegen"
...
-frame-metadata = { workspace = true }
+frame-metadata = { workspace = true, optional = true }
 heck = { workspace = true }
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true }
 scale-info = { workspace = true }
 subxt-metadata = { workspace = true }
-jsonrpsee = { workspace = true, features = ["async-client", "client-ws-transport-native-tls", "http-client"] }
+jsonrpsee = { workspace = true, features = [
```
niklasad1, line 29: please revert the formatting of the features

codegen/Cargo.toml
niklasad1: can you explain why `getrandom` is needed? It's not clear to me why it's needed :P

codegen/Cargo.toml
```diff
@@ -12,19 +12,30 @@ documentation = "https://docs.rs/subxt-codegen"
 homepage.workspace = true
 description = "Generate an API for interacting with a substrate node from FRAME metadata"
 
+[features]
+default = ["fetch_metadata"]
+fetch_metadata = ["dep:jsonrpsee", "dep:tokio", "dep:frame-metadata"]
```
niklasad1, line 17: I prefer kebab-case as naming convention for optional features i.e, `fetch-metadata` in this case

codegen/src/api/storage.rs
niklasad1: Just make `frame-metadata` a non-optional dev-dependency and remove the added feature should work?

codegen/Cargo.toml
jsdw: Yep, let's stick with dashes over underscores for consistency :)

codegen/Cargo.toml
jsdw: I'm not 100% if it's definitely needed here; @tadeohepperle now that the `fetch-metadata` feature flag is a thing, could you confirm whether we need to care about the `getrandom/js` feature?

Looking at `cargo tree -i getrandom` I think it's only poulled in when jsonrpsee is enabled anyway, so we shouldn't need to care otherwise :)

jsdw (approved):
I think we can remove the `getrandom` dep entirely now (see comment above) but otherwise looks great!

codegen/Cargo.toml
tadeohepperle: @jsdw @niklasad1 I removed the `getrandom` dependency. Originally I wanted to have jsonrpsee available and needed this `getrandom/js` feature in the WASM, but later figured it might be unnecessary.

metadata/Cargo.toml
niklasad1: @tadeohepperle can you fix this as well?

niklasad1 (approved):

PR #1161, opened 2023-09-12
chainHead based backend implementation

jsdw:
This PR results in an `UnstableBackend`, which implements the `Backend` trait and can be used in place of the `LegacyBackend` to drive Subxt with the [new (and currently unstable) RPC APIs](https://github.com/paritytech/json-rpc-interface-spec/).

The files are structured roughly as so:

- `subxt/src/backend/unstable/follow_stream.rs`: A `Stream` whose goal is to remain subscribed to `chainHead_follow`. It will re-subscribe if the subscription is ended for any reason, and it will return the current `subscription_id` as an event, along with follow events.
- `subxt/src/backend/unstable/follow_stream_unpin.rs`: A `Stream` which builds on the above, and handles pinning. It replaces any block hash seen in the follow events with a `BlockRef` which, when all clones are dropped, will lead to an "unpin" call for that block hash being queued. It will also automatically unpin any blocks that exceed a given max age, to try and prevent the underlying stream from ending (and _all_ blocks from being unpinned as a result). Put simply, it tries to keep every block pinned as long as possible until it's no longer used anywhere.
- `subxt/src/backend/unstable/follow_stream_driver.rs`: A `Stream` which builds on the above, and allows multiple subscribers to obtain events from the single underlying subscription (each being provided an `Initialised` message and all new blocks since then, as if they were each creating a unique `chainHead_follow` subscription). This is the "top" layer and the one that's interacted with elsewhere.

Each of these layers is independently unit tested to check that they are doing what we expect. A key consideration is that we want all of this to compile to WASM, and so we avoid using tokio and such anywhere.

We also have `subxt/src/backend/unstable/storage_items.rs`: A wrapper around a stream of storage events which will handle continuing and stopping correctly, and stream each `StorageResult` back one at a time (rather than in groups).

From these, we then implement the actual `Backend` trait in `subxt/src/backend/unstable/mod.rs` to pull everything together.

To test this all, we add an "unstable-backend-client" feature flag to our `integration-tests` crate. When this is enabled, all of the integration tests are ran using the new `UnstableBackend` instead of the default `LegacyBackend`. All tests should pass with both backends (excepting CI issues at the time of writing). We tweak CI to run these tests, too.

Other misc changes were made to be compatible with the newest `substrate-node` binary (locally at least), namely `tranfer` => `transfer_allow_death`.


subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -0,0 +1,488 @@
...
+impl<Hash: BlockHash> Default for Shared<Hash> {
+    fn default() -> Self {
+        Shared(Arc::new(Mutex::new(SharedState {
+            next_id: 1,
+            done: false,
+            subscribers: HashMap::new(),
+            current_init_message: None,
+            current_subscription_id: None,
+            seen_runtime_events: HashMap::new(),
+            block_events_from_last_finalized: VecDeque::new(),
```
tadeohepperle, line 189: Nit: Why do we set `next_id: 1` instead of having it initialized to 0? 
I think we can also derive the `Default` trait for `SharedState`, to use `SharedState { next_id: 1, ..Default::default() }` or `Default::default()` here.

subxt/src/config/mod.rs
```diff
@@ -65,6 +56,40 @@ pub trait Config: Sized + Send + Sync + 'static {
...
+    + Send
+    + Sync
+    + Decode
+    + AsRef<[u8]>
+    + Serialize
+    + DeserializeOwned
+    + Encode
+    + PartialEq
+    + std::cmp::Eq
+    + std::cmp::PartialEq
```
tadeohepperle, line 72: I think this trait bound might have PartialEq twice?

subxt/src/config/mod.rs
```diff
@@ -65,6 +56,40 @@ pub trait Config: Sized + Send + Sync + 'static {
...
+        + Send
+        + Sync
+        + Decode
+        + AsRef<[u8]>
+        + Serialize
+        + DeserializeOwned
+        + Encode
+        + PartialEq
+        + std::cmp::Eq
+        + std::cmp::PartialEq
```
tadeohepperle, line 88: Same as above, probably one of the `PartialEq`s can go.

testing/integration-tests/src/full_client/client/mod.rs
```diff
@@ -242,10 +242,10 @@ async fn unsigned_extrinsic_is_same_shape_as_polkadotjs() {
     // - start local substrate node.
     // - open polkadot.js UI in browser and point at local node.
     // - open dev console (may need to refresh page now) and find the WS connection.
-    // - create a balances.transfer to ALICE with 12345 and "submit unsigned".
+    // - create a balances.transferAllowDeath to ALICE (doesn't matter who from) with 12345 and "submit unsigned".
     // - find the submitAndWatchExtrinsic call in the WS connection to get these bytes:
     let expected_tx_bytes = hex::decode(
-        "b004060700d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0f0090c04bb6db2b"
+        "b004060000d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d0f0090c04bb6db2b"
```
tadeohepperle, line 248: Question: could we also arrive at these bytes constructing a call via subxt-signer and the static interface?

tadeohepperle (approved):
Great work, thanks a lot for your explanation last week, without it I would have had a hard time seeing what is going on. I just have a few comments.

subxt/src/backend/unstable/follow_stream.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+}
+
+enum InnerStreamState<Hash> {
+    /// We've just created the stream; we'll start Initializing it
+    New,
+    /// We're fetching the inner subscription. Move to Ready when we have one.
+    Initializing(FollowEventStreamFut<Hash>),
+    /// Report back the subscription ID here, and then start ReceivingEvents.
+    Ready(Option<(FollowEventStream<Hash>, String)>),
```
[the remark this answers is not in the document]
jsdw: Hehe, I sat pondering this for a bit myself!

The reason for the option is just because, below, we want to take ownership of those two values to put them in other places, but we only have a `&mut` reference to them. So the question is, if I want to take ownership of them, I have to replace the originals with something.. but what?! With an `Option` I can replace it with `None` using `option.take()` :)

I'll look at this again though because I can maybe do better here and it'd be lovely to remove the option!

subxt/src/backend/unstable/follow_stream_driver.rs
jsdw: For no particular reason other than to avoid `0` being used, which sometimes could be seen as a special or meaningful value. But there would be no harm in this being initialized to 0 either really; just a habit I adopted a bit!

testing/integration-tests/src/full_client/client/mod.rs
jsdw: We wouldn't even need to sign anything :)

This call is really just checking that, for a given random unsigned extrinsic, subxt (via the static interface above) produces the same bytes that somehting like Polkadot.js produces. ie we want to compare the Subxt bytes with some non-subxt bytes and if they align, we gain some confidence that subxt constructed them properly :)

subxt/src/config/mod.rs
jsdw: Oh yeah, so it does! Good spot!

testing/integration-tests/src/full_client/client/mod.rs
jsdw: Aah I see, yup we could have done that too :) But at least by checking the whole hash against a different client, we get a bit of extra reassurance that _everything_ subxt does is in line at least with the other client; any part we do using Subxt code itself will no longer be "compared"

subxt/src/backend/legacy/mod.rs
```diff
@@ -356,6 +352,7 @@ impl<T: Config> Stream for StorageFetchDescendantKeysStream<T> {
 
             // We have some keys to hand back already, so do that.
             if let Some(key) = this.keys.pop_front() {
+                println!("Storage entry: {key:?}");
```
lexnv, line 355: nit: Maybe `log::debug!` although I feel they are left-overs from debugging :D

testing/integration-tests/src/full_client/client/unstable_rpcs.rs
```diff
@@ -279,20 +273,26 @@ async fn transaction_unstable_submit_and_watch() {
         .transaction_unstable_submit_and_watch(&tx_bytes)
         .await
         .unwrap();
-
+    println!("about to wait for tx progress");
```
lexnv, line 276: nit: we could remove these 2 extra printlns

subxt/src/backend/legacy/mod.rs
jsdw: Oooh yup, whoops!!

testing/integration-tests/src/full_client/client/unstable_rpcs.rs
jsdw: Yup; I completely forgot to go clean up my debug logs by the looks of it! :)

testing/test-runtime/build.rs
```diff
@@ -63,7 +63,7 @@ async fn run() {
 
     // Save metadata to a file:
     let out_dir = env::var_os("OUT_DIR").unwrap();
-    let metadata_path = Path::new(&out_dir).join("metadata.scale");
+    let metadata_path = Path::new(&out_dir).join("test_node_runtime_metadata.scale");
```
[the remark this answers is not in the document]
jsdw: Honestly I wasn't sure; I was just running into weird issues and decided to rename it as a bit of a last ditch thing and it seemed to work afterwards. Probably it was something local to my system getting confused, but no harm in making the name harder to collide with anything else :)

subxt/src/events/events_client.rs
```diff
@@ -58,10 +58,10 @@ where
         // return a Future that's Send + 'static, rather than tied to &self.
         let client = self.client.clone();
         async move {
-            // If a block ref isn't provided, we'll get the latest best block to use.
+            // If a block ref isn't provided, we'll get the latest finalized block to use.
             let block_ref = match block_ref {
                 Some(r) => r,
-                None => client.backend().latest_best_block_ref().await?,
+                None => client.backend().latest_finalized_block_ref().await?,
```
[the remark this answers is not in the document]
jsdw: Yeah exactly; when I went over the code, it felt like using the current best block for this stuff by default was weird for that exact reason that it may well be pruned. So yeah, this will hopefully be a little more reliable, and of course the user can still provide a block hash that they prefer :)

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -0,0 +1,634 @@
...
+                }
+                FollowStreamMsg::Event(FollowEvent::NewBlock(details)) => {
+                    // One bigger than our parent, and if no parent seen (maybe it was
+                    // unpinned already), then one bigger than the last finalized block num
+                    // as a best guess.
+                    let parent_rel_block_num = this
+                        .pinned
+                        .get(&details.parent_block_hash)
+                        .map(|p| p.rel_block_num)
+                        .unwrap_or(this.rel_block_num);
```
lexnv, line 127: This might be a good place to catch rpc-v2 implementation inconsistencies in the wild, but realize that we might not want to polute users with this.
What would you think about having a `log::warning` about the rpc not reporting parent hashes that we expect to exist?

lexnv:
Lovely PR and as always amazing work 👍

lexnv (approved):

subxt/src/backend/unstable/follow_stream_unpin.rs
jsdw: I like this idea, buttt if the parent block is unpinned already then it will not exist when we look it up here, so I don't think I can so easily add a log that won't lead to spurious warnings

PR #1175, opened 2023-09-25
Add Web Feature to Subxt Codegen

### LINKED ISSUE #1169: `subxt-codegen`: support `web` and `fetch-metadata` together? Tidy `web`.
https://github.com/paritytech/subxt/pull/1154 makes `subxt-codegen` play nicely in WASM. It currently does this by simply hiding all of the `jsonrpsee` related bits behind a `fetch-metadata` feature. This feature isn't compatible with the `web` feature that is also added at the moment.

Since `jsonrpsee` can be compiled to WASM, we could support `fetch-metadata` and `web` together if we wanted.

Secondly, if `subxt` has its `web` feature enabled, we should ensure to enable the `web` feature in `subxt-codegen` too, just so that `subxt-codegen` knows to handle this. (easiest way might just be to have a web feature in `subxt-macro` which jsut enables it in `subxt-codegen`, and then in `subxt` we have `web = [..., "subxt-macro/web"]`.

This isn't very pressing, since everything works as-is right now; just a nice to have to consistify things a little.

tadeohepperle:
Fixes: #1169


jsdw (approved):
Nice! (I was a bit premature here, btu I think all that needs doing is removing that compile_error! now)

lexnv (approved):
Nice! Looks good and should be ready for merge after the `compile_error!` fix 👍

niklasad1 (approved):
nice

jsdw (approved):

PR #1177, opened 2023-09-26
try composite action to avoid dupe use-substrate code

jsdw:
Create a local "composite action" to avoid duplicating a bunch of code and make using a substrate binary in CI nice and easy.


niklasad1 (approved):

tadeohepperle (approved):
Nice, this makes it much shorter :+1:

.github/workflows/actions/use-substrate/README.md
```diff
@@ -0,0 +1,3 @@
+# use-substrate
+
+This action downloads the substrate binary produced from the `build-substrate` workflow and puts it into the `$PATH` so that it can be used by the following steps.
```
niklasad1, line 3: I don't the `following steps` part of the README and I think the last pattern after `$PATH` can removed anyway

jsdw: Yeah that's fair :)

PR #1181, opened 2023-09-28
Introduce `sp_arithmetic` as `subxt::ext::sp_arithmetic` and implement default substitues behind `substrate-compat` feature.

### LINKED ISSUE #1170: Provide implementations of `Permill`, `Perbill` etc in Subxt?
We fixed decoding into the generated `Permill` etc types a short while ago, but the types also have a bunch of custom logic and such. Currently, to work with them properly, one would have to pull in `sp_arithmetic` and convert the decoded parts into the proper types manually.

Perhaps we could either:
- Pull in `sp-arithmetic` and just substitute for the "proper" types. (sp-arithmetic doesn't look like it depends on much, so I think this would be fairly doable), or
- Make some concise implementations of the types ourselves in Subxt if pulling in the above is a pain for some reason. (I think we'd rather not do this)
- Leave things as they are and people will have to manually convert the types to use them.

Types are defined here: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/arithmetic/src/per_things.rs#L1864

tadeohepperle:
Fixes #1170 
This PR adds  `sp_arithmetic` (16.0.0) as a dependency to subxt that is exposed unser `subxt::ext::sp_arithmetic`. 

Currently I am facing the problem, that the types from `sp_arithmetic` do not implement `EncodeAsType` and `DecodeAsType` and we cannot easily implement them because of the orphan rule.

Draft PR. Incomplete.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4243,6 +4243,7 @@ dependencies = [
  "scale-value",
  "serde",
  "serde_json",
+ "sp-arithmetic",
  "sp-core",
  "sp-core-hashing",
  "sp-keyring",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ tokio-util = "0.7.9"
 sp-core = { version = "21.0.0", default-features = false }
 sp-core-hashing = { version = "9.0.0", default-features = false }
 sp-runtime = "24.0.0"
+sp-arithmetic = "16.0.0"
 sp-keyring = "24.0.0"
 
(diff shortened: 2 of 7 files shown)


tadeohepperle:
Needs some more thought because the types from sp_arithmetic do not implement EncodeAsType/DecodeAsType. Which results in invalid code being generated, when they are used as fields of generated structs that derive EncodeAsType. Closing this PR for now.

PR #1182, opened 2023-09-29
Allow getting account nonce at arbitrary blocks, too

jsdw:
We have an `account_nonce` function on `client.tx()` already, which fetches the account nonce based on the latest finalized block. In some cases (eg when using `manual-seal` in testing which doesn't produce finalized blocks), we'd like to be able to get the acocunt nonce for other blocks. So this just exposes it on the "blocks" API for that purpose, sharing the logic.

### DIFF
--- a/subxt/src/blocks/mod.rs
+++ b/subxt/src/blocks/mod.rs
@@ -14,3 +14,6 @@ pub use crate::backend::BlockRef;
 pub use block_types::Block;
 pub use blocks_client::BlocksClient;
 pub use extrinsic_types::{ExtrinsicDetails, ExtrinsicEvents, Extrinsics, StaticExtrinsic};
+
+// We get account nonce info in tx_client, too, so re-use the logic:
+pub(crate) use block_types::get_account_nonce;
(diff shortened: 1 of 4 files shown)


subxt/src/blocks/mod.rs
```diff
@@ -14,3 +14,6 @@ pub use crate::backend::BlockRef;
 pub use block_types::Block;
 pub use blocks_client::BlocksClient;
 pub use extrinsic_types::{ExtrinsicDetails, ExtrinsicEvents, Extrinsics, StaticExtrinsic};
+
+// We get account nonce info in tx_client, too, so re-use the logic:
+pub(crate) use block_types::get_account_nonce;
```
ascjones, line 19:
```suggestion
pub use block_types::get_account_nonce;
```
I would like to use this function from my lib please :pray:

jsdw: Oh but you can via `client.blocks().at_latest().account_nonce()` or `client.blocks().at(block_hash).account_nonce()` (there is one downside currently which is that it downloads the block header though, so yeah I'm not completely sold on this either; I forgot about that)

niklasad1 (approved):

lexnv (approved):

ascjones (approved):

tadeohepperle (approved):

PR #1191, opened 2023-10-05
fix for when runtime API field name is _

jsdw:
Runtime APIs are methods which take arguments, but Subxt converts these argument names into struct field names. Mostly this is ok, but "_" is a valid argument name and not a valid struct name (I can't think of any other cases of valid argument names that arent valid struct names offhand; can you?).

Anyway, we fix this specific case in the runtime API codegen


lexnv (approved):
Nice! Lovely test!

tadeohepperle (approved):
Looks good! :+1:

PR #1193, opened 2023-10-06
Add `subxt` feature in  `subxt-signer` crate to default features

### LINKED ISSUE #1189: Docs: clarify "subxt" feature in subxt-signer
It's come up a couple of times now where people have used subxt-signer with subxt and have run into an error about the signer trait not being implemented.

This is resolved by adding the "subxt" feature on the subxt-signer dep to tell it to play nicely with subxt.

We should either:
- make the feature be enabled by default because 99% of the time it'll be used with subxt anyway (I think this is my preference), or
- Clarify in the docs that the subxt feature should be enabled.

tadeohepperle:
Fixes #1189 .

Small PR to simplify use of subxt-signer to avoid confusion.

### DIFF
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -15,7 +15,7 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa"]
+default = ["sr25519", "ecdsa", "subxt"]
 
 # Pick the signer implementation(s) you need by enabling the
 # corresponding features. Note: I had more difficulties getting
--- a/signer/Cargo.toml
+++ b/signer/Cargo.toml
@@ -15,7 +15,7 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa", "subxt"]
+default = ["sr25519", "ecdsa", "subxt", "native"]
 
 # Pick the signer implementation(s) you need by enabling the
 # corresponding features. Note: I had more difficulties getting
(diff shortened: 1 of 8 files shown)


jsdw (approved):
Looks good to me! A nice little QoL tweak

niklasad1 (approved):

lexnv (approved):
Nice! 👍

jsdw:
I wonder whether, to fix the feature flag compile issue, we should add no-default-features to the subxt dep in subxt-signer (it shouldn't care, after all, which subxt features are enabled). That might lead to other issues htough; needs some playing with!

signer/Cargo.toml
```diff
@@ -15,7 +15,7 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa"]
+default = ["sr25519", "ecdsa", "subxt", "native"]
```
jsdw, line 18: I don't really understand this change; it means that signer by default will be "native" and then to be web you need to disable default features and then select eg subxt,web?

I think it'd be better to do as subxt and ask the depends library to pick which one it wants explicitly. This way, libraries can (and should) defer the choice unless it's relevant to them :)

tadeohepperle: @jsdw My reasoning was that in 90% of cases people will just want to use it with native anyway. But you are right, I will change it to be similar to subxt in that an explicit choice has to be made.

jsdw: Yeah, I get that; my main issue is that feature flags are additive and so the choice to enable web or native should be made at the last possible moment; enabling it by default makes it more likely that something depending on subxt somewhere in the tree and not thinking about features will screw up somebody trying to compile instead to web or whatever :)

PR #1198, opened 2023-10-09
Make web and native feature explicit opt in in signer crate

tadeohepperle:
When using the  `subxt-signer` crate with the `subxt` feature enabled, now one has to explicitly opt in to the `web` or the `native` feature. Failure to do so, yields a compile error.

(Changes some of the things I did in https://github.com/paritytech/subxt/pull/1193)


jsdw:
Oooohhh, this is my bad; I thought that with Subxt you had to explicitly opt in to both, but when I checked just now  Ican see that Subxt enables the "native" feature by default already, because otherwise `cargo check` is a pain etc.

So actually I was wrong about this and I don't think you need to change that after all; sorry!

jsdw:
So yup, I think the only things I'd change are: #1200. Sorry for sending you down the wrong track!

PR #1199, opened 2023-10-09
Batch fetching storage values again

### LINKED ISSUE #1195: Regression in speed of fetching child storage values in 0.32
As I understand, when iterating over a map in the storage, this constant
`const STORAGE_FETCH_PAGE_SIZE: u32 = 32;`
is used to determine how many keys should be fetched at once. 

This being a constant is quite limiting -- in fact for maps of size a few thousands, this becomes slow because of this constant being too small. Ideally it should be possible to pass this into `iter()` or somewhere, so that one can configure the page size depending on factors like size of the key and value, instead of using a constant.

jsdw:
When the Backend trait was added, I moved to fetching storage values one by one because the batch function we used was a little confusing and it seemed simpler at the time.

This PR adds back the fetching of storage values in batches, and improves the docs on the relevant RPC methods to clarify their behaviour.

Closes #1195

### DIFF
--- a/subxt/src/backend/legacy/mod.rs
+++ b/subxt/src/backend/legacy/mod.rs
@@ -78,7 +78,7 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
             done: Default::default(),
             keys_fut: Default::default(),
             pagination_start_key: None,
-            page_size: Default::default()
+            page_size: Default::default(),
         };
 
         let keys = keys.flat_map(|keys| {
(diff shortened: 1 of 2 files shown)


lexnv (approved):

tadeohepperle (approved):

niklasad1 (approved):
LGTM

Makes sense to me that you removed it in the first place because the code is quite complicated

jsdw:
Before I merge this I'll just see if its worth adding a test or something to make sure that the storage values returned are as expected.

PR #1200, opened 2023-10-09
Wee tidy to subxt-signer flags

jsdw:
Subxt-signer probably shouldn't bother trying to set the corresponding subxt feature flags when it's enabled; just expect them to be set for each crate independently. And compile_error regardless of whether subxt is enabled

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,8 +65,8 @@ jobs:
             - name: Cargo check subxt-signer
               run: |
                   cargo check -p subxt-signer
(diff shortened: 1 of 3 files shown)


niklasad1 (approved):

jsdw:
Aah, I see why subxt-signer sets the subxt features now, so maybe this makes no sense either. I think it's ok as it is then!

lexnv (approved):

sergejparity (approved):

tadeohepperle (approved):
Okay, makes sense. :+1:

PR #1203, opened 2023-10-11
Clone + Debug on Payloads/Addresses, and compare child storage results

jsdw:
For consistency, I added proper Clone + Debug impls on the payload/address types. I also added a test for #1199 to make sure that the new batch storage method aligns with the unstable backend results, and tweaked the test stuff a bit to make it easier to add such tests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -241,8 +241,8 @@ jobs:
         name: Test (WASM)
         runs-on: ubuntu-latest
         env:
-            # Set timeout for wasm tests to 120 secs; default is too low sometimes.
-            WASM_BINDGEN_TEST_TIMEOUT: 120
+            # Set timeout for wasm tests to be much bigger than the default 20 secs.
+            WASM_BINDGEN_TEST_TIMEOUT: 300
 
         steps:
             - uses: actions/checkout@v4
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -240,6 +240,9 @@ jobs:
     wasm_tests:
         name: Test (WASM)
         runs-on: ubuntu-latest
+        env:
+            # Set timeout for wasm tests to 120 secs; default is too low sometimes.
+            WASM_BINDGEN_TEST_TIMEOUT: 120
 
         steps:
             - uses: actions/checkout@v4
--- a/subxt/src/custom_values/custom_value_address.rs
+++ b/subxt/src/custom_values/custom_value_address.rs
@@ -42,6 +42,26 @@ pub struct StaticAddress<ReturnTy, IsDecodable> {
     phantom: PhantomData<(ReturnTy, IsDecodable)>,
 }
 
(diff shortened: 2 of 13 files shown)


testing/integration-tests/src/utils/node_proc.rs
```diff
@@ -18,6 +19,12 @@ pub struct TestNodeProcess<R: Config> {
     // Keep a handle to the node; once it's dropped the node is killed.
     proc: SubstrateNode,
 
+    // Lazily construct these when asked for,
```
lexnv, line 22: nit: s/,/.

lexnv (approved):

testing/integration-tests/src/utils/node_proc.rs
jsdw, line 22:
```suggestion
    // Lazily construct these when asked for.
```

subxt/src/custom_values/custom_value_address.rs
```diff
@@ -42,6 +42,26 @@ pub struct StaticAddress<ReturnTy, IsDecodable> {
     phantom: PhantomData<(ReturnTy, IsDecodable)>,
 }
 
+impl<ReturnTy, IsDecodable> Clone for StaticAddress<ReturnTy, IsDecodable> {
```
niklasad1, line 45: Why is the manual Clone + Debug impls needed?

It should be possible to derive those I think

jsdw: Because Clone/Debug should be possible regardless of some of the generic params, and the default derives assume that they all impl Clone/debug too.

Buttt, I've been avoiding using derivative and maybe I should just use that instead to automate these better!

niklasad1 (approved):

subxt/src/custom_values/custom_value_address.rs
jsdw: I mean that if you have:

```rust
#[derive(Clone)]
struct Foo<T> {
    val: usize,
    marker: PhantomData<T>
}
```

Then the `Clone` impl will be like:

```rust
impl <T: Clone> Clone for Foo<T> { ... }
```

Even though it's irrelevant whether `T` impls `Clone` or not. So really we want:

```rust
impl <T> Clone for Foo<T> { ... }
```

I just wrote them manually but should probably switch these to using derivative to save some loc :)

jsdw: I switched to using derivative to derive these things instead to get rid of all of the boilerplate!

sergejparity (approved):

.github/workflows/rust.yml
```diff
@@ -240,6 +240,9 @@ jobs:
     wasm_tests:
         name: Test (WASM)
         runs-on: ubuntu-latest
+        env:
+            # Set timeout for wasm tests to 120 secs; default is too low sometimes.
+            WASM_BINDGEN_TEST_TIMEOUT: 120
```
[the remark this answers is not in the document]
jsdw: I haven't fully debugged this yet; previous runs are all failing, and locally increasing the timeout seemed to fix it but not in CI :(

PR #1209, opened 2023-10-16
Support decoding signed extensions

### LINKED ISSUE #1124: Support decoding signed extensions from extrinsics in blocks
I tried looking in the calls, e.g., `polkadot::Call::Balances`, `polkadot::Call::Staking`, etc. 
I reviewed extrinsic events, such as `Transfer`, `Nominate`, and many others. But I was unable to find this information.

Does `subxt` provide such functionality? If not, what is the way to retrieve this information?
I assume it is possible because many Polkadot [explorers](https://polkadot.subscan.io/extrinsic/16711660-2) provide this information.

tadeohepperle:
Fixes #1124 
This PR adds an option to dynamically decode signed extensions from extrinsics.
Until now we just skipped over any signed extensions (Signed Extra bytes).
There are two convenience functions for `tip()` and `nonce()`. When you have ExtrinsicDetails you can not do the following:
```rs
let extrinsic: ExtrinsicDetails;
let signed_extensions = extrinsic.signed_extensions()?;
// tip and nonce
let tip: u128 = signed_extensions.tip()?;
let nonce: u64 = signed_extensions.nonce()?;
// iterate over signed extensions
for s in signed_extensions.iter() {
     let s = s.unwrap(); // error if decoding failed. Ends iteration.
     let name = s.name();    
     let scale_value = s.value();
     let bytes = s.bytes();
}

// or find a specific one:
let mortality = signed_extensions.find("CheckMortality")?;
...
```


subxt/src/blocks/extrinsic_types.rs
```diff
@@ -217,12 +232,10 @@ where
         // Skip over the first byte which denotes the version and signing.
         let cursor = &mut &bytes[1..];
 
-        let mut address_start_idx = 0;
-        let mut address_end_idx = 0;
-
-        if is_signed {
-            address_start_idx = bytes.len() - cursor.len();
-
+        let signed = if !is_signed {
```
jsdw, line 235: Nit: maybe call this `signed_details` for clarity.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -155,12 +160,8 @@ pub struct ExtrinsicDetails<T: Config, C> {
...
     /// Extrinsic bytes.
     bytes: Arc<[u8]>,
-    /// True if the extrinsic payload is signed.
-    is_signed: bool,
-    /// The start index in the `bytes` from which the address is encoded.
-    address_start_idx: usize,
-    /// The end index of the address in the encoded `bytes`.
-    address_end_idx: usize,
+    /// Some if the extrinsic payload is signed.
+    signed: Option<SignedExtrinsicDetails>,
```
jsdw, line 164: Nit: call `signed_details` for clarity?

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -326,8 +347,100 @@ where
...
+                    name,
+                    metadata,
+                    phantom: PhantomData,
+                });
+            }
+
+            Ok(ExtrinsicSignedExtensions {
+                bytes: extra_bytes,
+                signed_extensions,
+                ty_id: extra_ty_id,
```
jsdw, line 431: We don't need this type ID; instead of accessing and checking this type above, prefer `metadata.extrinsic().signed_extensions()`. This gives a vec of signed extension details that you can iterate to decode each of them :)

Then that type ID also doesn't need to propagate to any other types etc; it's not needed here.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +671,101 @@ impl<T: Config> ExtrinsicEvents<T> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ExtrinsicSignedExtensions<'a, T: Config> {
```
jsdw, line 675: Is there any use for having `T: Config` here? I don't think so at the moment, so maybe it can be removed from this and the child structs at the mo?

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +671,101 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    signed_extensions: Vec<ExtrinsicSignedExtension<'a, T>>,
+    bytes: &'a [u8],
+    ty_id: u32,
+    metadata: &'a Metadata,
```
jsdw, line 679: Prefer `metadata.clone()` I think to avoid the lifetime here (it's an Arc anyway).

Since the bytes are already `Arc<[u8]>` to be cheaply shared, we could avoid all lifetimes and just keep the bytes and start/end indexes as above. I'm torn on this one. It would make it easier for people to store away ExtrinsicSignedExtensions etc though :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +671,101 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+#[derive(Debug, Clone)]
+pub struct ExtrinsicSignedExtension<'a, T: Config> {
+    bytes: &'a [u8],
+    ty_id: u32,
+    name: &'a str,
+    metadata: &'a Metadata,
+    phantom: PhantomData<T>,
+}
+
+impl<'a, T: Config> ExtrinsicSignedExtensions<'a, T> {
```
jsdw, line 692: I'd be tempted to remove`decoded`, `bytes` and `signed_extensions` for now, and tweak `find`. The goal I think is to try to be consistent with `Extrinsics` here.

So I reckon the interface here should be something like:

```rust
// iterate and return each ExtrinsicSignedExtension (one can
// find by name or whatever using this interface):
iter()

// find and decode a specific signed extension just like we do for Extrinsics (leave this 
// for a separate PR since it's tricky and needs more work/thought):
find<S: SignedExtension>() -> Result<Option<S>, Error>

// The nice convenience helpers, which internally just self.iter().find() or whatever:
tip()
nonce()
```

subxt/src/config/polkadot.rs
```diff
@@ -11,6 +11,7 @@ use crate::SubstrateConfig;
 pub use primitive_types::{H256, U256};
 
 /// Default set of commonly used types by Polkadot nodes.
+#[derive(Debug)]
```
jsdw, line 14: We shouldn't need this; this is because deriving Debug on `Foo<T: Config>` will automatically expect it on T, too. We work around this by either using Derivative to say that we don't care about `T`, or manually implementing something to ignore T :)

subxt/src/config/substrate.rs
```diff
@@ -14,6 +14,7 @@ pub use primitive_types::{H256, U256};
 /// Default set of commonly used types by Substrate runtimes.
 // Note: We only use this at the type level, so it should be impossible to
 // create an instance of it.
+#[derive(Debug)]
```
jsdw, line 17: As above; not needed.

subxt/src/config/polkadot.rs
jsdw: (It's more annoying to do it this way, but more "correct" as well :))

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +671,101 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+#[derive(Debug, Clone)]
+pub struct ExtrinsicSignedExtensions<'a, T: Config> {
+    signed_extensions: Vec<ExtrinsicSignedExtension<'a, T>>,
+    bytes: &'a [u8],
+    ty_id: u32,
+    metadata: &'a Metadata,
+    phantom: PhantomData<T>,
+}
+
+#[derive(Debug, Clone)]
```
jsdw, line 683: I expect this can't be cloned and needs `derivative` impls (see some other structs in Subxt) to do this properly because `T != Clone`, but would be interested to see!)

jsdw: But actually, we Don't need to have the `T` in these structs at all anyway, so might all be a moot point!

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: We could do that, but for the `?` operator used for error handling inside the block this would mean, we have 
```rs
let signed_details = is_signed
            .then(|| -> Result<SignedExtrinsicDetails, Error> {
                // use of `?` to handle decoding errors...
                ..
                Ok(SignedExtrinsicDetails {..}) 
            }).transpose()?;
```
I personally find it a bit more confusing.

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: Hmm, maybe that is better. But I don't like about it that the `ExtrinsicSignedExtensions` would then contain a pointer to the bytes of the entire extrinsic instead of just having the bytes relating to the actual extra fields of signed extensions. I think this would then require storing start and end indices as well?

subxt/src/blocks/extrinsic_types.rs
jsdw: Good point; let's leave it as is then :)

subxt/src/blocks/extrinsic_types.rs
jsdw: Yeah; if you had a pointer to all of the bytes you'd need indices as well. I guess it's a choice between user ergonomics (Arc of bytes + indices) versus optimal performance (lifetimes + borrow the bytes we need).

I'm sortof happy either way for now; we could stick to lifetimes and move to Arc+indices if it proves to be annoying for people :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -668,7 +668,7 @@ impl<'a> ExtrinsicSignedExtensions<'a> {
 
     /// The tip of an extrinsic, extracted from the ChargeTransactionPayment signed extension.
     pub fn tip(&self) -> Option<u128> {
-        let tip = self.find("ChargeTransactionPayment")?;
+        let tip = self.find("ChargeAssetTxPayment")?;
```
jsdw, line 671: Actually, both this and ChargeTransactionPayment could be used and contain a tip. Only one should ever be used at a time, and the tip I think is encoded at the start of either, so you can decode the start bytes in the same way for both.

So I'd try finding both and use whichever you find first :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -326,8 +342,41 @@ where
...
+            .as_ref()
+            .map(|e| &self.bytes[e.address_end_idx..e.signature_end_idx])
+    }
+
+    /// Return only the bytes of the extra fields for this signed extrinsic.
+    ///
+    /// # Note
+    ///
+    /// Returns `None` if the extrinsic is not signed.
+    pub fn extra_bytes(&self) -> Option<&[u8]> {
```
niklasad1, line 366: dq: does this include just `extra` or `extra | additional_signed`? 

See https://docs.rs/sp-runtime/latest/src/sp_runtime/generic/unchecked_extrinsic.rs.html#197 for context

Regardless, I think it would be good to document that

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    ty_id: u32,
+    identifier: &'a str,
+    metadata: Metadata,
+}
+
+impl<'a> ExtrinsicSignedExtensions<'a> {
+    /// Get a certain signed extension by its name.
+    pub fn find(&self, signed_extension: impl AsRef<str>) -> Option<ExtrinsicSignedExtension> {
+        self.iter()
+            .find_map(|e| e.ok().filter(|e| e.name() == signed_extension.as_ref()))
```
niklasad1, line 628: I don't find this super easy to read but yeah are you using Option::filter here and my hunch was that an iterator was returned by find_map.

I guess `find_map` here fixes a bunch of boiler plate instead using flatten etc...

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    }
+
+    pub fn iter(&'a self) -> impl Iterator<Item = Result<ExtrinsicSignedExtension<'a>, Error>> {
+        let signed_extension_types = self.metadata.extrinsic().signed_extensions();
+        let num_signed_extensions = signed_extension_types.len();
+        let mut index = 0;
+        let mut byte_start_idx = 0;
+
+        std::iter::from_fn(move || {
```
niklasad1, line 637: never used this but neat to avoid allocation an Vec for this

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+                let cursor = &mut &self.bytes[byte_start_idx..];
+                if let Err(err) = scale_decode::visitor::decode_with_visitor(
+                    cursor,
+                    ty_id,
+                    self.metadata.types(),
+                    scale_decode::visitor::IgnoreVisitor,
+                )
+                .map_err(|e| Error::Decode(e.into()))
+                {
+                    index = num_signed_extensions; // (such that None is returned in next iteration)
```
niklasad1, line 652: Why not decode all things in the signed_ext? 

Couldn't be useful to get all the things in the signed_ext? Such as you may want to know everything that failed or just filter the ones that were successful?

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+                    metadata: self.metadata.clone(),
+                }))
+            }
+        })
+    }
+
+    /// The tip of an extrinsic, extracted from the ChargeTransactionPayment signed extension.
+    pub fn tip(&self) -> Option<u128> {
+        let tip = self
+            .find("ChargeTransactionPayment")
```
niklasad1, line 672: Would be cool have these in the metadata, I'm not sure whether some chains overrides the transaction_payment pallet and adds their own impl probably not.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+            .or_else(|| self.find("ChargeAssetTxPayment"))?;
+        // Note: ChargeAssetTxPayment might have addition information in it (asset_id).
+        //       But both should start with a compact encoded u128, so this decoding is fine.
+        let tip = Compact::<u128>::decode(&mut tip.bytes()).ok()?.0;
+        Some(tip)
+    }
+
+    /// The nonce of the account that submitted the extrinsic, extracted from the CheckNonce signed extension.
+    pub fn nonce(&self) -> Option<u64> {
```
niklasad1, line 681: This may fail on certain runtimes if the `CheckNonce` isn't part of the SignedExtra.

Might worth a comment for it and advise to use the dynamic API then.

niklasad1: Should be nice to have for the general usecase...

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -227,3 +229,61 @@ async fn fetch_block_and_decode_extrinsic_details() {
     assert_eq!(ext.value, 10_000);
     assert!(tx.is_signed());
 }
+
+#[tokio::test]
+async fn decode_signed_extensions_from_blocks() {
```
niklasad1, line 234: Would be cool to test with different runtimes so both `ChargeAssetTxPayment` and `ChargeTransactionPayment` works but probably annoying to support here.

Could be done in another PR

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: @niklasad1 I could definitely do that, but chances are if one fails to decode, the bytes from here onward are probably not representing what the user thinks they are. I think it is unlikely any following signed extensions are decoded correctly in that case. I was trying to mimic the API of `Extrinsics`, where we have a similar return on error policy in the the `iter()` function.

subxt/src/blocks/extrinsic_types.rs
jsdw: Yeah; the idea of these "quick" helpers is that they work in the overwhelmingly common case, but may not always work (and should fail gracefully eg returning `None` here if the extension isn't available)

subxt/src/blocks/extrinsic_types.rs
jsdw: Yeah, it's just for a common case that will most likely work, and should fail gracefully and return `None` if the relevant extension isn't found :)

testing/integration-tests/src/full_client/blocks/mod.rs
jsdw: Yeah; I pondered this too but not sure how I'd test this?! It'd be a pain introducing the ability to spawn up a different node just for this

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    pub fn iter(&'a self) -> impl Iterator<Item = Result<ExtrinsicSignedExtension<'a>, Error>> {
+        let signed_extension_types = self.metadata.extrinsic().signed_extensions();
+        let num_signed_extensions = signed_extension_types.len();
+        let mut index = 0;
+        let mut byte_start_idx = 0;
+
+        std::iter::from_fn(move || {
+            if index == num_signed_extensions {
+                None
+            } else {
```
jsdw, line 640: Nit; about the indentation in the `else` by doing `return None` here instead; an early exit when done :)

subxt/src/blocks/extrinsic_types.rs
niklasad1: okey, I see uniform API makes sense to me :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+pub struct ExtrinsicSignedExtension<'a> {
+    bytes: &'a [u8],
+    ty_id: u32,
+    identifier: &'a str,
+    metadata: Metadata,
+}
+
+impl<'a> ExtrinsicSignedExtensions<'a> {
+    /// Get a certain signed extension by its name.
+    pub fn find(&self, signed_extension: impl AsRef<str>) -> Option<ExtrinsicSignedExtension> {
```
jsdw, line 626: To be consistent with how Extrinsics/Events are decoded, I think `find` should look more like:

```rust
fn find<S: SignedExtension>() -> Result<Option<S>, Error>
```

For "dynamically finding things", one can `.iter().find(..)` anyway.

I'd suggest just removing this for now until we have a nice way to find and decode based on the `subxt::tx::signed_extensions::*` types.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    /// The nonce of the account that submitted the extrinsic, extracted from the CheckNonce signed extension.
+    pub fn nonce(&self) -> Option<u64> {
+        let nonce = self.find("CheckNonce")?;
+        let nonce = Compact::<u64>::decode(&mut nonce.bytes()).ok()?.0;
+        Some(nonce)
+    }
+}
+
+impl<'a> ExtrinsicSignedExtension<'a> {
+    /// The extra bytes associated with the signed extension.
```
jsdw, line 689:
```suggestion
    /// The bytes representing this signed extension.
```

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,117 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+        self.identifier
+    }
+
+    /// The type id of the signed extension.
+    pub fn type_id(&self) -> u32 {
+        self.ty_id
+    }
+
+    /// DecodedValueThunk representing the type of the extra of this signed extension.
+    pub fn decoded(&self) -> Result<DecodedValueThunk, Error> {
```
jsdw, line 705: Let's remove thid `decoded` fn; it doesn't add much on top of `value()` below imo :)

testing/integration-tests/src/full_client/blocks/mod.rs
jsdw: I guess one way to unit test this would be crafting some custom metadata with whatever signed extensions we want, and then creating an `OfflineClient` (note the _offline_), setting the metadata on it to be what we want, and then building an extrinsic (could be dynamically constructed). Then, one can manually make the bytes into an `ExtrinsicDetails` and try to decode each of the signed extensions from it.

This way, we could test a few different combinations of signed extensions (eg if Nonce/Tip ones don't exist, or if both the tip ones exist etc).

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -227,3 +229,61 @@ async fn fetch_block_and_decode_extrinsic_details() {
...
+
+    let transaction2 = submit_transfer_extrinsic_and_get_it_back!(5678);
+    let extensions2 = transaction2.signed_extensions().unwrap();
+    let nonce2 = extensions2.nonce().unwrap();
+    let tip2 = extensions2.tip().unwrap();
+
+    assert_eq!(nonce1, 0);
+    assert_eq!(tip1, 1234);
+    assert_eq!(nonce2, 1);
+    assert_eq!(tip2, 5678);
```
jsdw, line 288: I'd like us to also test that iterating over signed extensions works; right now we have no way to really know if it's actually decoding them as we expect :)

The set of signed extensions on a node is fairly fixed, so we can just sanity test that each of the expected ones is decoded and that the names are right or something like that (in addition to the above).

jsdw:
Looking good; I resolved all of the comments from myself that you've addressed so just worth looking over the others; nothing major from me now! Will be lovely to have this!

On testing: one way to unit test this would be crafting some custom metadata with whatever signed extensions we want, and then creating an `OfflineClient` (note the _offline_), setting the metadata on it to be what we want, and then building an extrinsic (could be dynamically constructed). Then, one can manually make the bytes into an `ExtrinsicDetails` and try to decode each of the signed extensions from it.

This way, we could test a few different combinations of signed extensions (eg if Nonce/Tip ones don't exist, or if both the tip ones exist etc).

But yes, I don't mind this being a follow-up PR because it'll take a little work to do nicely!

testing/integration-tests/src/full_client/blocks/mod.rs
tadeohepperle: I added code for testing this to the end of the test :)

jsdw (approved):
Great job, looks good to me!

I wondered why missing docs on a function weren't being picked up in CI, and noticed that the new structs weren't exposed in the module, so I did that and extended a few lifetimes while in the area :)

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -227,3 +229,81 @@ async fn fetch_block_and_decode_extrinsic_details() {
...
 }
+
+#[tokio::test]
+async fn decode_signed_extensions_from_blocks() {
+    let ctx = test_context().await;
+    let api = ctx.client();
+    let alice = dev::alice();
+    let bob = dev::bob();
+
+    macro_rules! submit_transfer_extrinsic_and_get_it_back {
```
lexnv, line 240: dq: What are the benefits of this macro over a plain function? Looks good otherwise!

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,116 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+                ty_id,
+                identifier: extension.identifier(),
+                metadata,
+            }))
+        })
+    }
+
+    /// The tip of an extrinsic, extracted from the `ChargeTransactionPayment` or
+    /// `ChargeAssetTxPayment` signed extension, depending on which is present.
+    pub fn tip(&self) -> Option<u128> {
```
lexnv, line 662: Might be worth extending these functions to also fetch:
- transaction version (CheckTxVersion)
- spec version (CheckSpecVersion)
- tx genesis (CheckGenesis)
- tx weigth (CheckWeight)

Totally up to you tho

lexnv:
LGTM! 

What do you think about extending the [blocks_subscribing](https://github.com/paritytech/subxt/blob/2b21f8dc8cdec34a122794a0b590f5c0303e2e62/subxt/examples/blocks_subscribing.rs) to also decode the extensions? 🤔

subxt/src/blocks/extrinsic_types.rs
jsdw: Personally I'd just start with a couple of obvious ones that people def want to begin with, since they all rely on specific signed extensions existing to work :)

testing/integration-tests/src/full_client/blocks/mod.rs
tadeohepperle: There were just some really stupid lifetime issues when using a Closure.
And I cannot use a normal function easily because then I need to specify all types.
And the type of `api` here depends on the `unstable-light-client` feature flag. And the type of `api` would be a generic type of the return value of that function as well. So I would have to provide two implementations of such a function depending on the feature flag. Also I cannot put the api a trait object easily, because OnlineClientT is not object safe. So in the end I just went with this macro.

jsdw:
> What do you think about extending the [blocks_subscribing](https://github.com/paritytech/subxt/blob/2b21f8dc8cdec34a122794a0b590f5c0303e2e62/subxt/examples/blocks_subscribing.rs) to also decode the extensions? 🤔

I don't think that this is a blocker for merging, but would be nice to show this stuff in an example somewhere!

subxt/src/blocks/extrinsic_types.rs
niklasad1: yeah, let's do in a follow-up PR if needed

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -558,6 +607,116 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+        })?;
+
+        // Note: ChargeAssetTxPayment might have addition information in it (asset_id).
+        // But both should start with a compact encoded u128, so this decoding is fine.
+        let tip = Compact::<u128>::decode(&mut tip.bytes()).ok()?.0;
+        Some(tip)
+    }
+
+    /// The nonce of the account that submitted the extrinsic, extracted from the
+    /// CheckNonce signed extension.
```
niklasad1, line 676:
```suggestion
    /// CheckNonce signed extension.
    ///
    /// Returns `None` if the signed extension was not found or decoding it failed.
```

niklasad1 (approved):
nice

subxt/examples/blocks_subscribing.rs
```diff
@@ -34,6 +34,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                println!("      Signed Extensions:");
+                for signed_extension in signed_extensions.iter() {
+                    let signed_extension = signed_extension?;
+                    let name = signed_extension.name();
+                    let value = signed_extension.value()?.to_string();
+                    println!("        {name}: {value}");
+                }
+            } else {
+                println!("      No Signed Extensions");
+            }
```
jsdw, line 47: Just because this is an example, I wonder whether we could make this a bit less ugly, ie removing the `else` branch and moving it after the other `println`'s or something?

lexnv (approved):

PR #1212, opened 2023-10-19
Update smoldot to 0.12

lexnv:
Update smoldot to 0.12 and smoldot-light to 0.10 to benefit from the latest updates and performance improvements.

This PR changes the `SubxtPlatform` implementation to align with the latest `PlatformRef` from smoldot.

A new module named `with-buffers` is added to facilitate the low-level implementation.

The module is added to ease the transition to the smoldot `with-buffers`, that at the moment is not compatible with wasm due to the `std::time::Instant`. Ideally, smoldot will provide support for these buffers and the said code could be entirely removed from subxt.


niklasad1:
Seems to some issue with the gloo-net version in the WASM tests...

tadeohepperle (approved):
:+1:

lightclient/src/platform/wasm_helpers.rs
```diff
@@ -40,87 +30,381 @@ pub fn sleep(duration: Duration) -> Delay {
...
+        /// to [`WithBuffers::read_write_access`].
+        read_write_wake_up_after: Option<Instant>,
+    }
+
+    impl<T> WithBuffers<T> {
+        /// Initializes a new [`WithBuffers`] with the given socket.
+        ///
+        /// The socket must still be open in both directions.
+        pub fn new(socket: T) -> Self {
+            let read_buffer_reasonable_capacity = 65536; // TODO: make configurable?
```
niklasad1, line 92: could be a constant

lightclient/src/platform/wasm_helpers.rs
```diff
@@ -40,87 +30,381 @@ pub fn sleep(duration: Duration) -> Delay {
...
+                read_write: read_write::ReadWrite {
+                    now,
+                    incoming_buffer: mem::take(this.read_buffer),
+                    expected_incoming_bytes: if !*this.read_closed { Some(0) } else { None },
+                    read_bytes: 0,
+                    write_bytes_queued,
+                    write_buffers: mem::take(this.write_buffers),
+                    write_bytes_queueable: if !*this.write_closed {
+                        // Limit outgoing buffer size to 128kiB.
+                        // TODO: make configurable?
```
niklasad1, line 148: I think it's fine to hard-code this because it will likely to be replaced the next smoldot version

niklasad1: As always, it's preferred to open issues for TODOs

lightclient/src/platform/wasm_helpers.rs
```diff
@@ -40,87 +30,381 @@ pub fn sleep(duration: Duration) -> Delay {
...
+                }
+            });
+
+            // Grow the read buffer in order to make space for potentially more data.
+            this.read_buffer.resize(this.read_buffer.capacity(), 0);
+
+            future::poll_fn(move |cx| {
+                if this.error.is_some() {
+                    // Never return.
+                    return Poll::Pending;
```
niklasad1, line 210: it's never woken up again this?

niklasad1 (approved):

jsdw (approved):
Nice! (I'm assuming the `with_buffer` module is basically copied and will go away, which will be lovely and really simplify things!)

lightclient/src/platform/wasm_helpers.rs
lexnv: I believe the smoldot will drop the future entirely if it returns poll::pending, this is similar to smoldot with-buffers

lightclient/src/platform/wasm_platform.rs
```diff
@@ -50,6 +52,10 @@ impl PlatformRef for SubxtPlatform {
         super::wasm_helpers::now()
     }
 
+    fn fill_random_bytes(&self, buffer: &mut [u8]) {
+        rand::RngCore::fill_bytes(&mut rand::thread_rng(), buffer);
```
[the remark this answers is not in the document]
lexnv: Have looked at these crates and under the hood the `rand::OsRand` calls directly into `getrandom::getrandom`.
The `getrandom` can fail internally, and similarly to `rand::OsRand::fill_bytes` I've panicked on failure. Which I believe should be an indicator of low system entropy.
That is to say, we don't require any extra dependencies. Thanks for the suggestion 👍

lexnv:
Indeed, the `with_buffer` module will go away with the next smoldot deployment (which should include https://github.com/smol-dot/smoldot/pull/1250). For the meanwhile, our `with_buffer` module is a close copy of smoldot. 

Thanks for the reviews!

PR #1225, opened 2023-10-25
Tidy `subxt-codegen` crate interface

jsdw:
This crate interface has always been a bit all over the place, being intended primarily for internal use. However, that's not been clear, and so it has been used.

This PR:
- Hides or removes most of the codegen interface, exposing only one builder struct which allows for code to be generated using the same options provided via the `#[subxt]` macro or `subxt codegen` CLI tool. The aim here is to provided a supported third means to generate the interface; via a rust library instead of macro/CLI.
- Exposes a few things in a hidden `__private` module **that should not be relied on**, just because `subxt-explorer` is currently relying on them.
- Makes the `fetch-metadata` flag very much opt-in (the codegen stuff should just be pure code really; we only have these fns here because they would need to live in a new crate otherwise).
- Removes `CratePath` and some other bits that were sortof pointless (and I didn't want to expose such a thing in the public interface).

With this, I'd be happy saying that we supported people using the `codegen` crate directly. Ultmately, the type generation code should all be moved into a separate crate and the overall interfaces tidied up etc. This PR will allow that to happen without affecting the `subxt-codegen` crate (hopefully).


cli/src/commands/codegen.rs
```diff
@@ -132,67 +132,77 @@ fn codegen(
...
+            .map_err(|e| eyre!("Cannot parse type substitution for path {from_str}: {e}"))?;
+        codegen.set_type_substitute(from, to);
     }
 
-    let should_gen_docs = !no_docs;
+    let metadata = subxt_metadata::Metadata::decode(&mut &*metadata_bytes)
+        .map_err(|e| eyre!("Cannot decode the provided metadata: {e}"))?;
+    let code = codegen
+        .generate(metadata)
+        .map_err(|e| eyre!("Cannot generate code: {e}"))?;
```
tadeohepperle, line 204: This is so much cleaner now :+1:

codegen/src/fetch_metadata.rs
```diff
@@ -44,22 +49,34 @@ impl std::str::FromStr for MetadataVersion {
...
 }
 
+// /// Returns the raw, 0x prefixed metadata hex from the provided URL, blocking the current thread.
+// pub fn fetch_metadata_from_url_hex_blocking(
+//     url: Url,
+//     version: MetadataVersion,
+// ) -> Result<String, FetchMetadataError> {
+//     tokio_block_on(fetch_metadata_hex(url, version))
+// }
+
```
tadeohepperle, line 79: Can this be removed?

codegen/src/fetch_metadata.rs
```diff
@@ -83,15 +100,15 @@ pub async fn fetch_metadata_bytes(
...
-}
+// /// Returns the raw, 0x prefixed metadata hex from the provided URL.
+// pub async fn fetch_metadata_hex(
+//     url: Url,
+//     version: MetadataVersion,
+// ) -> Result<String, FetchMetadataError> {
+//     let bytes = fetch_metadata_bytes(url, version).await?;
+//     let hex_data = format!("0x{}", hex::encode(bytes));
+//     Ok(hex_data)
+// }
```
tadeohepperle, line 111: This is also not needed I guess?

codegen/src/lib.rs
```diff
@@ -2,65 +2,267 @@
...
+use api::RuntimeGenerator;
+use proc_macro2::TokenStream as TokenStream2;
+use std::collections::HashMap;
+
+// We expose these only because they are currently needed in subxt-explorer.
+// Eventually we'll move the type generation stuff out into a separate crate.
+#[doc(hidden)]
+pub mod __private {
+    pub use crate::types::{DerivesRegistry, TypeDefGen, TypeGenerator, TypeSubstitutes};
+}
```
tadeohepperle, line 35: Thanks for keeping it exposed, later we can have a cleaner solution.

codegen/src/types/mod.rs
```diff
@@ -334,55 +334,3 @@ impl Module {
         &self.root_mod
     }
 }
-
-/// A newtype wrapper which stores the path to the Subxt crate.
-#[derive(Debug, Clone)]
-pub struct CratePath(syn::Path);
```
tadeohepperle, line 340: I like that this got removed, it was a bit annoying to work with.

macro/src/lib.rs
```diff
@@ -85,61 +89,60 @@ pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
...
-    let universal_attributes = args.attributes_for_all_types.unwrap_or_default();
-    derives_registry.extend_for_all(
-        universal_derives,
-        universal_attributes.iter().map(|a| a.0.clone()),
-    );
+    // Use the item module that the macro is on:
+    codegen.set_target_module(item_mod);
+
+    // Use the provided crate path:
+    if let Some(crate_path) = args.crate_path {
```
tadeohepperle, line 98: The args used in the macro are very similar to the args used in the CLI, but I think it would be difficult/not worth it to unify them somehow?

macro/src/lib.rs
```diff
@@ -85,61 +89,60 @@ pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
...
+        codegen.set_type_substitute(sub.path, sub.with);
+    }
 
-    let runtime_api_generator = GenerateRuntimeApi::new(item_mod, crate_path)
-        .derives_registry(derives_registry)
-        .type_substitutes(type_substitutes)
-        .generate_docs(should_gen_docs)
-        .runtime_types_only(args.runtime_types_only);
+    // Do we want to fetch unstable metadata? This only works if fetching from a URL.
+    let unstable_metadata = args.unstable_metadata.is_present();
```
tadeohepperle, line 145: What is the reasoning behind allowing unstable metadata only from URLs?

tadeohepperle (approved):
Great PR, looks good to me, I just had a few small comments/questions :)

codegen/src/fetch_metadata.rs
jsdw: Ooh good points; I forgot to remove the code I commented out!

macro/src/lib.rs
jsdw: Yeah; I mean ideally they would be about identical. The types will vary a bit just because they come from different sources (eg strings from a terminal vs syn::Paths from a lib call or whatever)

macro/src/lib.rs
jsdw: Oh; this is just because if you provide the metadata, you can't pick what version it will be, but if you fetch from a URL you can decide what to fetch. No real need for this option at the mo since it wouldn't actually work, but when we start work on V16 metadata at some point it'll become useful again to try it out in Subxt :)

testing/integration-tests/src/full_client/codegen/documentation.rs
```diff
@@ -145,18 +138,10 @@ fn check_root_attrs_preserved() {
...
-        .generate_runtime(
-            item_mod,
-            derives,
-            type_substitutes,
-            CratePath::default(),
-            true,
-        )
+    let mut codegen = CodegenBuilder::new();
+    codegen.set_target_module(item_mod);
+    let generated_code = codegen
```
lexnv, line 143: This looks way better! Nice!

lexnv (approved):
LGTM! This API improves a lot the user experience for our codgen crate!

PR #1226, opened 2023-10-26
Static decoding of signed extensions.

tadeohepperle:
Builds on top of #1209 which should probably be merged first.

This PR adds a new trait `SignedExtensionDecoder` that all signed extensions have to implement. It specifies how the `extra` and `additional` parts of a `SignedExtension` should be encoded (and decoded for `extra`).
 Implementing this trait on some type, auto-implements ExtrinsicParamsEncoder on that type as well.
 The trait looks like this: 
```rs
pub trait SignedExtensionDecoder {
    type Extra: Encode + DecodeAsType;
    type Additional: Encode;
    fn extra(&self) -> &Self::Extra;
    fn additional(&self) -> &Self::Additional;
}
```
To create a new signed extension you now just have to fill out the fields of this trait and specify a `const NAME: &'static str` for the extension in the `SignedExtension` trait.

This PR makes use of this by providing a `find<S: SignedExtension>()` function on the `ExtrinsicsSignedExtensions`. 
I also included a test to confirm that this works.


subxt/src/config/signed_extensions.rs
```diff
@@ -10,19 +10,45 @@
...
+impl<S: SignedExtensionDecoder + 'static> ExtrinsicParamsEncoder for S {
+    fn encode_extra_to(&self, v: &mut Vec<u8>) {
+        self.extra().encode_to(v);
+    }
+
+    fn encode_additional_to(&self, v: &mut Vec<u8>) {
+        self.additional().encode_to(v);
+    }
+}
+
```
jsdw, line 51: I'll have to think about this a bit!

On the one hand, I do like the symmetry between `extra` and `additional`, and the `find` API is lovely!

On the other hand, I have some reservations still:
- `ExtrinsicParamsEncoder` is in a weird place now. The blanket impl means that nobody can create their own impl of it now I think (so if I wanted to impl `ExtrinsicParams` myself and ignore the signed extension stuff, I couldn't). It also just doesn't fit nicely into the picture now and is relegated to being a hack for object safety.
- Previously, `ExtrinsicParamsEncoder` allowed one to be very flexible about how they encode extra+additional properties, because how the encoding happens is hidden behind a method. They could use plain old `codec::Encode`, or they could write a signed extension which stores the metadata and then uses that to more intelligently encode properties based on metadata. 
  - There are definite uses for this; `ChargeAssetTxPayment` for instance can be generic over asset location, and metadata would allow somebody to be smart and write a version of it that worked in a few places. S
  - Since a signed extension can save the metadata, maybe this is ok though; they can just have a custom `Encode` impl which internally uses metadata to do something smarter. Hmmm.
- Currently we have `type Extra = Encode + DecodeAsType` which is a weird asymmetry, though I can't see a way around that offhand (Encode doesn't need the metadata perhaps, but decoding does).
- The requirement to point to a separate type for the `Additional` thing too exists only the symmetry, but does it actually benefit us or just make life a little harder for no reason?

I think the main thing in my head to resolve then is the weird place that `ExtrinsicParamsEncoder` is in, and whether it's still possible to allow a user to implement `ExtrinsicParams` themselves without having to go via signed extensions (I'm not sure how much this would be useful, but it's a nice escape hatch to allow users to add their very own abstractions instead if our SignedExtension stuff isn't working for them).

I do still wonder whether it is better to just add a `type DecodedExtra: DecodeAsType` bound to `SignedExtension` and rely on that; this doesn't have the symmetry (but maybe that's misleading anyway), but it does mean that `ExtrinsicParams` can still be implemented normally, avoids needing _another_ trait here (there are already 3 :D) and leaves `ExtrinsicParamsEncoder` with a clear purpose other than to be an object safety hack. I think it'd be good to see what that option looks like as well, perhaps in a separate PR, for comparison :)

tadeohepperle:
Closing this PR, as #1235 is the better approach probably.

PR #1227, opened 2023-10-27
ChargeAssetTxPayment: support providing u32 or MultiLocation in default impl

### LINKED ISSUE #1162: `ChargeAssetTxPayment`: support providing u32 _or_ MultiLocation in default impl

tadeohepperle:
fixes #1162


jsdw:
Offhand the approach here looks good to me; keeps it simple and shows the user how to support MultiLocations via generated code or whatever :)

subxt/src/config/mod.rs
```diff
@@ -51,6 +51,9 @@ pub trait Config: Sized + Send + Sync + 'static {
 
     /// This type defines the extrinsic extra and additional parameters.
     type ExtrinsicParams: ExtrinsicParams<Self>;
+
+    /// Asset Id
```
jsdw, line 55:
```suggestion
    /// This is used to identify an asset in the ChargeAssetTxPayment signed extension.
```

jsdw (approved):
Looks good!

There's still a bit of me that is wondering whether it's better to expose `AssetId` as a config param like this, or to have an `AssetId` generic on `ChargeAssetTxPayment` and default it to u32, but then people can change it there instead.

Given that the signed extension and need to configure this feels like a fairly common thing, I do think that the approach in this PR is best for now though!

subxt/src/config/mod.rs
```diff
@@ -51,6 +51,9 @@ pub trait Config: Sized + Send + Sync + 'static {
...
+    type AssetId: Debug + Encode;
```
jsdw, line 56: This _could_ also impl `EncodeAsType` possibly (same as a couple of the others above like singature I think), which would at least mean that if somebody tried setting a tip with an asset ID, and it was configured wrong, we'd get an `EncodeAsType` error back (eg because it failed to encode an `Option<u32>` into an `Option<MultiLocation>`) rather than have some error from the node that it failed to decode the tx provided.

For now I'd leave it as is, but we should probably consider moving some things in Config to `EncodeAsType` instead of `Encode` so that we get this extra protection.

lexnv: Might be worth an issue to not forget about it 🙏

lexnv (approved):
Nice one!

PR #1235, opened 2023-10-30
Static Decoding of Signed Extensions: Simple Approach

tadeohepperle:
Instead of introducing a new trait, like in #1226, this PR just adds an associated `type Decoded: DecodeAsType` to the SignedExtension trait.

Results should be the same as in #1226.


subxt/examples/blocks_subscribing.rs
```diff
@@ -34,6 +34,17 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                for signed_extension in signed_extensions.iter() {
+                    let signed_extension = signed_extension?;
+                    let name = signed_extension.name();
+                    let value = signed_extension.value()?.to_string();
+                    println!("        {name}: {value}");
+                }
+            } else {
+                println!("      No Signed Extensions");
+            }
             println!("      Bytes: {bytes_hex}");
```
jsdw, line 48:
```suggestion
            println!("    Extrinsic #{idx}:");
            println!("      Bytes: {bytes_hex}");
```

It's below already and looks nicer there :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -605,19 +609,19 @@ impl<T: Config> ExtrinsicEvents<T> {
...
+    metadata: Metadata,
+    _marker: std::marker::PhantomData<T>,
 }
 
-impl<'a> ExtrinsicSignedExtensions<'a> {
+impl<'a, T: Config> ExtrinsicSignedExtensions<'a, T> {
     /// Returns an iterator over each of the signed extension details of the extrinsic.
     /// If the decoding of any signed extension fails, an error item is yielded and the iterator stops.
-    pub fn iter(&self) -> impl Iterator<Item = Result<ExtrinsicSignedExtension<'a>, Error>> {
+    pub fn iter(&'a self) -> impl Iterator<Item = Result<ExtrinsicSignedExtension<'a, T>, Error>> {
```
jsdw, line 621: I think you added this `&'a self` lifetime back because metadata is no longer referenced, but you should be able to get rid of it (esp since you `self.metadata.clone()` below still).

If `Metadata` needs to be owned in `ExtrinsicSignedExtensions` (even though other bits are borrowed) then you could:
- `let metadata = self.metadata.clone()` at the top
- then don't reference `self` in the `iter::from_fn` below.

But since everything else is borrowed, I'd borrow the Metadata in `ExtrinsicSignedExtensions` and then all of the data has the `'a` lifetime and that lifetime flows into the iterated stuff too. (Basically, we should either fully commit to lifetimes or avoid them completely here I reckon, and fully committing seems like the good way to go unless we run into some case where they are a pain)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -648,21 +652,35 @@ impl<'a> ExtrinsicSignedExtensions<'a> {
...
                 ty_id,
                 identifier: extension.identifier(),
-                metadata,
+                metadata: self.metadata.clone(),
+                _marker: std::marker::PhantomData,
             }))
         })
     }
 
+    fn find_by_name(&self, name: impl AsRef<str>) -> Option<ExtrinsicSignedExtension<'_, T>> {
```
jsdw, line 661: Could this take `&str` instead of `impl AsRef<str>`?

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -707,12 +726,24 @@ impl<'a> ExtrinsicSignedExtension<'a> {
...
 
+    /// DecodedValueThunk representing the type of the extra of this signed extension.
+    fn decoded(&self) -> Result<DecodedValueThunk, Error> {
+        let decoded_value_thunk = DecodedValueThunk::decode_with_metadata(
+            &mut &self.bytes[..],
+            self.ty_id,
+            &self.metadata,
+        )?;
+        Ok(decoded_value_thunk)
+    }
```
jsdw, line 737: It looks like some things, like this and the lifetime stuff on `ExtrinsicSignedExtensions` were lost again compared to the base PR that these build off; would be good to merge the underlying branch or rebase :)

subxt/src/config/signed_extensions.rs
```diff
@@ -213,13 +223,16 @@ impl<T: Config> ExtrinsicParamsEncoder for CheckMortality<T> {
...
 /// The [`ChargeAssetTxPayment`] signed extension.
-#[derive(Debug)]
+#[derive(Debug, Encode, DecodeAsType)]
 pub struct ChargeAssetTxPayment {
-    tip: Compact<u128>,
-    asset_id: Option<u32>,
+    /// Tip
+    pub tip: Compact<u128>,
+    /// Asset Id
+    pub asset_id: Option<u32>,
```
jsdw, line 235: here and below I think I wouldn't make these Pub and instead expose `tip()` and `asset_id()` getters where `tip` returns `u128` and the `Compact` thing is hidden away (it's an implementation detail) :)

subxt/src/config/signed_extensions.rs
```diff
@@ -271,18 +284,20 @@ impl<T: Config> ExtrinsicParams<T> for ChargeAssetTxPayment {
 
 impl ExtrinsicParamsEncoder for ChargeAssetTxPayment {
     fn encode_extra_to(&self, v: &mut Vec<u8>) {
-        (self.tip, self.asset_id).encode_to(v);
+        self.encode_to(v);
```
jsdw, line 287: Personally I prefer the prev approach and not making the struct `Encode` (it feels like an implementation detail that's better to not expose). Same as below

subxt/src/config/signed_extensions.rs
```diff
@@ -271,18 +284,20 @@ impl<T: Config> ExtrinsicParams<T> for ChargeAssetTxPayment {
...
+    type Decoded = Self;
 }
 
 /// The [`ChargeTransactionPayment`] signed extension.
-#[derive(Debug)]
+#[derive(Debug, Encode, DecodeAsType)]
 pub struct ChargeTransactionPayment {
-    tip: Compact<u128>,
+    /// Tip
+    pub tip: Compact<u128>,
```
jsdw, line 300: As above I'd expose a `tip()` getter that returns `u128` to hide the `Compact` implementation detail.

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -286,18 +287,35 @@ async fn decode_signed_extensions_from_blocks() {
...
 
     assert_eq!(nonce1, 0);
+    assert_eq!(nonce1, nonce1_static);
     assert_eq!(tip1, 1234);
+    assert_eq!(tip1, tip1_static);
     assert_eq!(nonce2, 1);
+    assert_eq!(nonce2, nonce2_static);
     assert_eq!(tip2, 5678);
-
+    assert_eq!(tip2, tip2_static);
```
jsdw, line 318: it'd be really good I think to also test getting the mortality (just because that returns an enum and would be good to check that it decodes OK! I'm not sure off the top of my head whether the encoded impl is consistent with this or not))

subxt/src/blocks/extrinsic_types.rs
tadeohepperle: You are right, some stuff was lost in the merge, sorry. But this `decoded` function I kept on purpose. It is just private now, but used in the public `as_type()` and `value()` functions.

jsdw: In the other PR I'd updated to directly use `Value::decode_as_type(..)` which should be better than `DecodedValueThunk::decode_with_metadata(..)`, and also I'd remove the  `value()` function and just return a `Value` from this one (I'd have a look at the other PR because that was good here :))

testing/integration-tests/src/full_client/blocks/mod.rs
tadeohepperle: I added a check at the end of the test to verify Mortality decodes and is `Era::Immortal`.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -707,12 +727,24 @@ impl<'a> ExtrinsicSignedExtension<'a> {
...
     pub fn value(&self) -> Result<DecodedValue, Error> {
-        let value =
-            DecodedValue::decode_as_type(&mut &self.bytes[..], self.ty_id, self.metadata.types())?;
+        self.decoded()?.to_value()
+    }
 
-        Ok(value)
+    /// Decodes the `extra` bytes of this Signed Extension into a static type.
+    pub fn as_type<E: DecodeAsType>(&self) -> Result<E, Error> {
+        self.decoded()?.as_type::<E>().map_err(Into::into)
```
jsdw, line 747: Please re-apply the changes that weren't properly merged here:

https://github.com/paritytech/subxt/pull/1209/commits/64b5c90852bfca7dfa00c8dc80e2818f05075bdb#diff-68ce0867464cd9e4eb9e44abb6a5f2b227d8cd4d75f237d17b77e0d98166c3ebR712

testing/integration-tests/src/full_client/blocks/mod.rs
jsdw: Nice, thank you!

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -709,9 +728,12 @@ impl<'a> ExtrinsicSignedExtension<'a> {
...
-        let value =
-            DecodedValue::decode_as_type(&mut &self.bytes[..], self.ty_id, self.metadata.types())?;
+        self.as_type()
+    }
 
+    /// Decodes the `extra` bytes of this Signed Extension into a static type.
+    pub fn as_type<E: DecodeAsType>(&self) -> Result<E, Error> {
+        let value = E::decode_as_type(&mut &self.bytes[..], self.ty_id, self.metadata.types())?;
         Ok(value)
     }
```
jsdw, line 738: So for eg events we have `pub fn as_event<E: StaticEvent>(&self) -> Result<Option<E>, Error>`, and then for extrinsics we have `pub fn as_extrinsic<E: StaticExtrinsic>(&self) -> Result<Option<E>, Error>`.

I'd be inclined to add something here like:

```rust
pub fn as_signed_extra<S: SignedExtension>(&self) -> Result<Option<S::Decoded>, Error> {
    // Have sanity check that identifier lines up with `SignedExtension` given and then the 
    // usual decode logic, `Ok(None)` if name doesn't match (like the others).
}
```

This would be consistent with the above, and also be a little stricter than `as_type`, only allowing something that looks like the correct signed extension type to even be tried in the first place. I have no real objections to keeping `as_type` around too, but probably would remove it for now in favour of people defining and using more proper `SignedExtension` impls here :)

jsdw:
All looks good now; just one comment to change `as_type` into `as_signed_extra` and add the extra check that the name lines up (see https://github.com/paritytech/subxt/pull/1235#discussion_r1380570221); this will be consistent then with the methods on events and extrinsics, and a little safer than allowing any `as_type` :)

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -669,7 +669,7 @@ impl<'a, T: Config> ExtrinsicSignedExtensions<'a, T> {
     /// If the Signed Extension is found, but decoding failed, `Some(Err(err))` is returned.
     pub fn find<S: SignedExtension<T>>(&self) -> Option<Result<S::Decoded, Error>> {
```
jsdw, line 670: Nit: Maybe this should be `Result<Option<..>` instead, actually?

jsdw (approved):

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -649,25 +653,38 @@ impl<'a> ExtrinsicSignedExtensions<'a> {
...
-            e.ok().filter(|e| {
-                e.name() == "ChargeTransactionPayment" || e.name() == "ChargeAssetTxPayment"
-            })
-        })?;
-
-        // Note: ChargeAssetTxPayment might have addition information in it (asset_id).
-        // But both should start with a compact encoded u128, so this decoding is fine.
-        let tip = Compact::<u128>::decode(&mut tip.bytes()).ok()?.0;
-        Some(tip)
+        // Note: the overhead of iterating twice should be negligible.
```
lexnv, line 680: We could probably modify the `find_by_name` function to take in a filter instead of the name, but yea we have a few extras anyway

lexnv (approved):
LGTM! Nice one 👍

tadeohepperle:
Since #1227 merged, I am having some trouble implementing `scale_decode::IntoVisitor` for `ChargeAssetTxPayment<T: Config>` manually. Just deriving `DecodeAsType` on `ChargeAssetTxPayment` does not work anymore because of the Generics, even though I add a `DecodeAsType` trait bound to the `AssetId` type of the config.

It tells me, that `Option<T::AssetId>` also does not implement `IntoVisitor` and `DecodeAsType` which seems strange to me, because `Option<G>` should implement `DecodeAsType` and so should `Option<T::AssetId>` should do as well. Need to investigate this more next week.

jsdw:
> Just deriving DecodeAsType on ChargeAssetTxPayment does not work anymore because of the Generics

Maybe it's not documented (I didn't check) but I think `DecodeAsType` derive macro supports a top level `trait_bounds` attribute which you can set to "" or something to manually tell it that `T: Config` doesn't need to impl `IntoVisitor`!

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -272,38 +274,64 @@ async fn decode_signed_extensions_from_blocks() {
...
+
     assert_eq!(extensions2.iter().count(), expected_signed_extensions.len());
     for (e, expected_name) in extensions2.iter().zip(expected_signed_extensions.iter()) {
         assert_eq!(e.unwrap().name(), *expected_name);
     }
+
+    // check that era decodes:
+    for extensions in [&extensions1, &extensions2] {
+        let era: Era = extensions
+            .find::<CheckMortality<SubstrateConfig>>()
```
jsdw, line 332: It's a shame to have to write the `Config` bound in like this, but unavoidable as it stands!

(We can probably come up with something to avoid this at some point, but all good for now!)

tadeohepperle: I agree.

PR #1237, opened 2023-10-31
tx: Remove `wait_for_in_block` helper method

lexnv:
The `wait_for_in_block` waits until the transaction stream produces a `TxStatus::InBestBlock` event.

Transitioning to the latest RPC-V2, mainly using the `transaction` and `chainHead`, there is no strong guarantee that a new block reported by the `transaction` method will ever be reported by the `chainHead`.
This edge case could happen when the `transaction` API encounters a best block that is immediately pruned, possibly leading to the `chainHead` not gaining visibility for the said pruned block. 

To offer stronger guarantees, the subxt `submit_transaction` ensures that a block reported by `TxStatus::InFinalizedBlock` is accompanied by a `chainHead::Finalized` event.

Users who may still want the same behavior for testing purposes could then:

```rust
let mut progress: TxProgress = ..;

while let Some(status) = self.next().await {
    match status? {
        TxStatus::InBestBlock(s) | TxStatus::InFinalizedBlock(s) => return Ok(s),

        // Error scenarios; return the error.
        TxStatus::Error { message } | TxStatus::Invalid { message } | TxStatus::Dropped { message } => {
            // treat error
        }
        // Ignore anything else and wait for next status event:
        _ => continue,
    }
}
```

Motivated by: https://github.com/paritytech/subxt/pull/1236

cc @ascjones this might be of interest for contracts

### DIFF
--- a/subxt/src/tx/tx_progress.rs
+++ b/subxt/src/tx/tx_progress.rs
@@ -71,37 +71,6 @@ where
         StreamExt::next(self).await
     }
 
-    /// Wait for the transaction to be in a block (but not necessarily finalized), and return
-    /// an [`TxInBlock`] instance when this happens, or an error if there was a problem
-    /// waiting for this to happen.
-    ///
-    /// **Note:** consumes `self`. If you'd like to perform multiple actions as the state of the
-    /// transaction progresses, use [`TxProgress::next()`] instead.
-    ///
-    /// **Note:** transaction statuses like `Invalid`/`Usurped`/`Dropped` indicate with some
-    /// probability that the transaction will not make it into a block but there is no guarantee
-    /// that this is true. In those cases the stream is closed however, so you currently have no way to find
-    /// out if they finally made it into a block or not.
-    pub async fn wait_for_in_block(mut self) -> Result<TxInBlock<T, C>, Error> {
-        while let Some(status) = self.next().await {
-            match status? {
-                // Finalized or otherwise in a block! Return.
-                TxStatus::InBestBlock(s) | TxStatus::InFinalizedBlock(s) => return Ok(s),
-                // Error scenarios; return the error.
-                TxStatus::Error { message } => return Err(TransactionError::Error(message).into()),
-                TxStatus::Invalid { message } => {
-                    return Err(TransactionError::Invalid(message).into())
-                }
-                TxStatus::Dropped { message } => {
-                    return Err(TransactionError::Dropped(message).into())
-                }
-                // Ignore anything else and wait for next status event:
-                _ => continue,
-            }
-        }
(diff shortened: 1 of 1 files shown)


jsdw:
I think this is a good way to go; we don't remove the ability to wait for whatever event somebody wants (via the lower level stream of events one can wait for as above), but we take away the assumption that `wait_for_in_block` will Just Work under normal circumstances (which it occasionally won't, especially via the new APIs).

jsdw (approved):

tadeohepperle (approved):

lexnv:
> I still find this useful for instance on some scenarios such as the staking miner then one could select "best blocks" instead of "finalized block" but yeah it can be a footgun

Yep, this is still possible to do, although using a lower-level approach.
It might be possible that staking miner runs into a similar issue as detected by the CI, where it waits for the "BestBlock" but that might not (ever) be propagated to chainHead and cause a failure while using the RPCs methods with that block. Also, there's no guarantee that the current "BestBlock" will be finalized, since another "BestBlock { hash: None }" could indicate that the prior reported "BestBlock" is now pruned.

This is in response to https://github.com/paritytech/subxt/pull/1236 just to keep the stronger guarantees around the RPC v2, waiting for @ascjones to confirm this is ok from their side 🙏

niklasad1:
> Yep, this is still possible to do, although using a lower-level approach.
It might be possible that staking miner runs into a similar issue as detected by the CI, where it waits for the "BestBlock" but that might not (ever) be propagated to chainHead and cause a failure while using the RPCs methods with that block. Also, there's no guarantee that the current "BestBlock" will be finalized, since another "BestBlock { hash: None }" could indicate that the prior reported "BestBlock" is now pruned.

Yepp, it's dangerous to use but "might be faster" than finalized blocks so in some applications it might be worth the risk (probably not).... 

It's probably better for subxt just to support finalized blocks and for anything else one can implement custom logic for specific behavior but it's tricky to write such thing because the actual behavior when a subscription is not really clearly documented, see https://substrate.stackexchange.com/questions/7892/when-will-i-stop-receiving-transactionstatus-updates-for-a-transaction-submitted

jsdw:
> It's probably better for subxt just to support finalized blocks and for anything else one can implement custom logic for specific behavior but it's tricky to write such thing because the actual behavior when a subscription is closed or not.

In Subxt, if you use `.next()` on the TX progress events, rather than eg `wait_for_in_block` or whatever, then the stream will close itself at the right time and things anyway, so a naive impl doesn't need to worry about anything; it's just got a choice about whether it wants to report on any of the specific errors or just report a general "stream has closed without tx being in block" error :)

PR #1238, opened 2023-11-02
lightclient: Add support for multi-chain usecase

### LINKED ISSUE #1213: LightClient: Optimize multi-chain usecase
Subxt spawns a background task for handling communication with the lightclient (smoldot).

Currently, users who want to target multiple chains must construct one smoldot instance per chain.
This leads to spawning and handling multiple background tasks.

To optimize this process, subxt could reuse the same background task to handle multi chains.
Investigate if this is feasible without breaking changes in the subxt-API; and extend the communication protocol to support the `AddChain` command.

lexnv:
The subxt lightclient support enforced users to spawn one smoldot instance per chain.
With this PR, users can reuse the same smoldot instance to communicate with multiple chains and parachains.

API design for multi chain usecase:
- Users are responsible for creating a low level smoldot instance, which allows complete control over registering chains (users can configure internal buffers, maximum number of subscription, add multiple chains, remove chains etc).
- Chains are also reported to the `RawLightClientBuilder` object in terms of chain ID and RPC object that produces method responses
- The `RawLightClientBuilder` is turned into a `LightClient` and the smoldot instance is spawned on a background task that handles requests and multiplexes responses back to the internal RPC objects of subxt
- A `LightClient` instance is mapped to one chain. To interact with a different chain, users can call `LightClient::target_chain` method. This fetches under the hood the needed info of the `OnlineClient` (metadata, spec, runtime) and receives a custom `Config`. The newly created light client communicates with the same background instance of smoldot.

```rust
// Smoldot instance.
let client = subxt_lightclient::Client::new(...);
// Raw subxt builder.
let mut builder = LightClient::raw_builder();

// Configure smoldot.
let chain = client.add_chain(...);

// Make subxt aware of the chain IFF
// - the chain has the RPC functionality enabled 
// - the chain should be used by the subxt LightClient
builder.add_chain(chain.chain_id, chain.rpc_object);

// Other configuration ...

// Spawn the smoldot on the background task and target chain x
let api_chain_x = builder.build(client, chian_x_id).await?;

// Target chain Y (if prior added).
let api_chain_y = api_chain_x.target_chain(chian_y_id).await?;

// Use subxt clients api_chain_x and api_chain_y
```

The changes include:
- Adding support for building a light client instance from a smoldot instance via `RawLightClientBuilder`
- Ability to target different chains / parachains (and reuse the same smoldot instance) via `target_chain` API
- Extend internal the frontend - backend protocol to support multiple chains and multiplex smoldot responses accordingly
- Created a low level example which subscribes to finalized blocks of polkadot relay chain and asset hub parachain
- Added 2 demo chain specs to reach Polkadot and PolkadotAssetHub
- Improved the readability and added granular detail for tracing logs

Review notes:
- I would suggest looking first at the example to get a feel of the API [subxt/examples/unstable_light_client_parachains.rs](https://github.com/paritytech/subxt/compare/lexnv/light-client-multi-chain?expand=1#diff-29f43e490eb317f43418664b312ae2eaa084bae2a4ed349a0846532586cebddc) then at the background implementation for multichain use-cases [lightclient/src/background.rs](https://github.com/paritytech/subxt/compare/lexnv/light-client-multi-chain?expand=1#diff-ef0162345401f6161a0d2e589a9d9532557776431a839a1ee73b003c6f02b02a)

Next steps (possibly in another PR):
- Add subxt book entry about the raw light client
- Maybe expose the `subxt_lightclient::fetch_spec` via a `ChainSpecBuilder` to avoid demo-chain-specs artifacts and the need to keep them in sync (currently taken from smoldot)
- Extend the low-level raw client builder with easier to use API, but more restrictive (maybe abstract the smoldot instance entirely)

Closes: https://github.com/paritytech/subxt/issues/1213.


artifacts/demo_chain_specs/polkadot.json
jsdw, line 1: I wonder if it's possible to obtain a more condensed version of this file; there's one here for polkadot which is much smaller:

https://github.com/paritytech/substrate-connect/blob/main/projects/extension/assets/chainspecs/polkadot.json

It'd be great to know how to programmatically get hold of the smallest ones possible!

subxt/examples/unstable_light_client_parachains.rs
```diff
@@ -0,0 +1,109 @@
...
+        .expect("Light client configured with json rpc enabled; qed");
+    let parachain_chain_id = assethub_connection.chain_id;
+
+    // Add the chain to the light client builder.
+    builder.add_chain(parachain_chain_id, parachain_json_rpc_responses);
+
+    println!("Added AssetHub parachain");
+
+    // Step 4. Turn the smoldot client into a subxt light client using the builder.
+    let polkadot_api = builder
```
jsdw, line 76: As a nit, I think I'd find the example a little easier to follow if all of the builder stuff was in one place, like:

```rust
let polkadot_api = builder
    .add_chain(polkadot_chain_id, polkadot_json_rpc_responses)
    .add_chain(parachain_chain_id, parachain_json_rpc_responses)
    .build(client, polkadot_chain_id)
    .await?;
```

I guess that `client` is used to make requests, but we also need to hold onto those `json_rpc_responses` for each chain, but looking at this I do wonder if it's possible to do any better here, eg avoid those `.add_chain` steps because that information can be extracted from the `client` on the `.build` step or something (I guess I'll know better when I read the rest of the PR :D)

jsdw: `Client` can be indexed on a chain Id for instance to get some stuff (I didn't check what); I wonder if that helps at all:

https://docs.rs/smoldot-light/latest/src/smoldot_light/lib.rs.html#1097

subxt/src/client/light_client/mod.rs
```diff
@@ -54,31 +55,52 @@ pub enum LightClientError {
...
 #[derive(Derivative)]
 #[derivative(Clone(bound = ""))]
-pub struct LightClient<T: Config>(OnlineClient<T>);
+pub struct LightClient<T: Config> {
+    client: OnlineClient<T>,
+    raw_rpc: rpc::LightClientRpc,
+}
 
 impl<T: Config> LightClient<T> {
+    // TODO: Add `raw_builder()`.
```
jsdw, line 64: No longer a todo :)
```suggestion
```

subxt/src/client/light_client/mod.rs
```diff
@@ -115,11 +137,36 @@ impl<T: Config> LightClient<T> {
...
+        &self,
+        chain_id: subxt_lightclient::ChainId,
+    ) -> Result<LightClient<TChainConfig>, crate::Error> {
+        let raw_rpc = self.raw_rpc.target_chain(chain_id);
+        let rpc_client = RpcClient::new(raw_rpc.clone());
+
+        let client = OnlineClient::<TChainConfig>::from_rpc_client(rpc_client).await?;
+
+        Ok(LightClient { client, raw_rpc })
+    }
```
jsdw, line 164: There's something I find a little odd about this being on the `LightClient` itself (and also I think it's the reason the `LightClient` needs to carry around the `raw_rpc` thing).

I wonder whether it'd make sense to instead have something like:

```rust
// Instead of `build` spitting out a subxt client immediately, it spits out a RawLightClient:
let raw_light_client: RawLightClient = RawLightClientBuilder::new().add_chain(..).add_chain(..).build(client);

// Turn `RawLightClient` into a subxt client as many times as we want like this:
let light_client: LightClient<PolkadotConfig> = raw_light_client.for_chain(chain_id);
let light_client2: LightClient<SubstrateConfig> = raw_light_client.for_chain(other_chain_id);
```

lightclient/src/client.rs
```diff
@@ -11,14 +13,38 @@ use tokio::sync::{mpsc, mpsc::error::SendError, oneshot};
...
+    to_backend: mpsc::UnboundedSender<FromSubxt>,
+}
+
+impl RawLightClientRpc {
+    /// Construct a [`LightClientRpc`] that can communicated with the provided chain.
+    ///
+    /// # Note
+    ///
+    /// This uses the same underlying instance created by [`LightClientRpc::new_from_client`].
+    pub fn for_chain(&self, chain_id: smoldot_light::ChainId) -> LightClientRpc {
```
niklasad1, line 32: I found the naming of this a bit odd,  

I would prefer `connect_to_chain(id: ChainId)`

niklasad1: Another question is it possible to connect the same chain more than once?

lightclient/src/client.rs
```diff
@@ -46,22 +72,52 @@ impl LightClientRpc {
...
+    }
+
+    /// Constructs a new [`RawLightClientRpc`] from the raw smoldot client.
+    ///
+    /// Receives a list of RPC objects as a result of calling `smoldot_light::Client::add_chain`.
+    /// This [`RawLightClientRpc`] can target different chains using [`RawLightClientRpc::for_chain`] method.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if being called outside of `tokio` runtime context.
```
niklasad1, line 94: what about WASM i.e, `wasm_bindgen_futures::spawn_local`?

subxt/examples/unstable_light_client_parachains.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // Step 4. Turn the smoldot client into a raw client.
+    let raw_light_client = RawLightClient::builder()
+        .add_chain(polkadot_chain_id, polkadot_json_rpc_responses)
+        .add_chain(parachain_chain_id, parachain_json_rpc_responses)
+        .build(client)
+        .await?;
+
+    // Step 5. Obtain a client to target the parachain.
+    let polkadot_api: LightClient<PolkadotConfig> =
```
niklasad1, line 75: This looks like the polkadot relay to me? 

Maybe move the comment a few lines above the parachain_api?

subxt/examples/unstable_light_client_parachains.rs
```diff
@@ -0,0 +1,101 @@
...
+        })
+        .expect("Light client chain added with valid spec; qed");
+    let parachain_json_rpc_responses = assethub_connection
+        .json_rpc_responses
+        .expect("Light client configured with json rpc enabled; qed");
+    let parachain_chain_id = assethub_connection.chain_id;
+
+    // Step 4. Turn the smoldot client into a raw client.
+    let raw_light_client = RawLightClient::builder()
+        .add_chain(polkadot_chain_id, polkadot_json_rpc_responses)
```
niklasad1, line 69: hmm, so this a little complicated to first "add the chain" and then after call "for_chain".

I don't follow why adding the chain isn't sufficient AFAIU the `polkadot_json_rpc_responses` is just a stream that needs to polled or spawned in background task

lightclient/src/background.rs
```diff
@@ -43,59 +47,58 @@ pub enum FromSubxt {
...
-    /// The ID of the chain used to identify the chain protocol (ie. substrate).
-    ///
-    /// Note: A single chain is supported for a client. This aligns with the subxt's
-    /// vision of the Client.
-    chain_id: ChainId,
-    /// Unique ID for RPC calls.
-    request_id: usize,
+    client: smoldot_light::Client<TPlatform, TChain>,
+    /// Unique IDs for RPC calls.
+    rpc_chain_id: UniqueChainID,
```
niklasad1, line 62: I find the explanation and the naming hard to understand.
My suggestion is to name this `request_id_per_chain` or something similar

However, I don't understand why the request IDs has been groped by chain because it will just be incremented on every RPC call, no?

lightclient/src/background.rs
```diff
@@ -43,59 +47,58 @@ pub enum FromSubxt {
...
             oneshot::Sender<MethodResponse>,
             mpsc::UnboundedSender<Box<RawValue>>,
         ),
     >,
     /// Map the subscription ID to the frontend `Sender`.
-    subscriptions: HashMap<usize, mpsc::UnboundedSender<Box<RawValue>>>,
+    ///
+    /// The subscription ID is entirely generated by the node (smoldot). Therefore, it is
+    /// possible for two distinct subscriptions of different chains to have the same subscription ID.
+    subscriptions: HashMap<(usize, smoldot_light::ChainId), mpsc::UnboundedSender<Box<RawValue>>>,
```
niklasad1, line 82: Is subscription ID generation an monotonic increasing counter in smoldot? 

For instance in jsonrpsee it's a randomly generated such that you can't really assume that the subscription ID is unique but unlikely to happen.

lightclient/src/background.rs
```diff
@@ -446,3 +473,23 @@ impl std::str::FromStr for RpcResponse {
         })
     }
 }
+
+/// Generate unique IDs for RPC calls that are distinct for each chain.
+#[derive(Default)]
+struct UniqueChainID {
```
niklasad1, line 479: As said previously I think this can be removed but your choice

lightclient/src/client.rs
lexnv: Yep, its possible to use the same chain ID to create multiple instances of the light client

lightclient/src/client.rs
lexnv: Ahh good point, this is not limited to tokio will update the documentation here

subxt/examples/unstable_light_client_parachains.rs
lexnv: The `rpc_stream` needs to be provided while spawning the light-client at the moment.

It might be possible to introduce another subxt frontend -> subxt backend to propagate this to the smoldot client from the background task and add the `rpc_stream` to the streams polled in the backend. That seemed a bit more complicated than this, would love to hear your opinion on this 🙏

lightclient/src/background.rs
lexnv: Yep indeed, for RPC calls we control the request ID and the smoldot / node needs to provide the same ID in the response. Initially I avoided at all costs adding another dimension (chainID) to the internal mapping. 

However, for subscriptions we need to make one plain RPC method call, and the smoldot / node is free to give us any number in response to identify a subscription ID. 
And since that code was already there, seemed like an easy thing to reuse the logic and have unique IDs wrt to calling different chains. But that could be reverted if you think simplifies things :D

lightclient/src/background.rs
lexnv: I believe to be so since the first subscription ID is zero, although Id have to double check.
I've had some issues initially where two different chains created one subscription each, then the smoldot would provide sub_id = 0 for both and the multiplexing forwarded all responses to one chain instead.

lightclient/src/background.rs
lexnv: Yep, this turned out to be quite small in the end. Worth refactoring it into the main background struct, thanks for the feedback here and above 🙏

subxt/examples/unstable_light_client_parachains.rs
lexnv: I believe regarding the json_rpc_responses object, if the user drops this object, than the backend might not provide any responses:

```rust
pub struct JsonRpcResponses {
    /// Receiving side for responses.
    ///
    /// As long as this object is alive, the JSON-RPC service will continue running. In order
    /// to prevent that from happening, we destroy it as soon as the
    /// [`JsonRpcResponses::public_api_chain_destroyed`] is notified of the destruction of
    /// the sender.
    inner: Option<json_rpc_service::Frontend>,

    /// Notified when the [`PublicApiChain`] is destroyed.
    public_api_chain_destroyed: pin::Pin<Box<event_listener::EventListener>>,
}
```

Extracted from [smolodot](https://github.com/smol-dot/smoldot/blob/c5ec15dd8279f08796fa2271221ad061ef77f381/light-base/src/lib.rs#L306-L310).

Wrt https://docs.rs/smoldot-light/latest/src/smoldot_light/lib.rs.html#1097, I believe that returns the user data that we provided as `()` for default purposes. I couldn't dig anything that's related to the frontend channel that gets responses back

subxt/examples/unstable_light_client_parachains.rs
niklasad1: It's fine for now but perhaps worth writing up an issue to look into.

My point is initializing and using the API is a bit complicated but 
looks to mainly because expose some of smoldots APIs.

subxt/examples/unstable_light_client_parachains.rs
```diff
@@ -0,0 +1,101 @@
...
+            },
+            // The chain specification of the asset hub parachain mentions that the identifier
+            // of its relay chain is `polkadot`.
+            potential_relay_chains: [polkadot_chain_id].into_iter(),
+            database_content: "",
+            user_data: (),
+        })
+        .expect("Light client chain added with valid spec; qed");
+    let parachain_json_rpc_responses = assethub_connection
+        .json_rpc_responses
```
niklasad1, line 63: Is it possible to write a wrapper for this API?

It's quite annoying to do this unwrap because the RPC client should always be enabled in smoldot?

lightclient/src/background.rs
niklasad1: Ok, my point is that HashMap could be removed but what you did it's quite clean anyway so I don't mind it.

I just want some better docs here.

subxt/examples/unstable_light_client_parachains.rs
lexnv: Yep, I was thinking of creating a bit of a higher level API as a followup to this.

Ideally we could provide:
- The highest level API used for most people out there, but with some API limitations (ie multichain usage). And this is already merged in subxt. Btw we need to handle this under the hood, its just the way smoldot exposes that API at the moment depending if the RPC was enabled or not.



```rust
let api = LightClient::<PolkadotConfig>::builder()
        .bootnodes([
            "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp",
        ])
        .build_from_url("ws://127.0.0.1:9944")
        .await?;
```

- A low level API that provides the most flexibility and allows users to utilize 100% of smoldot features. This is included in this PR, although the tradeoff is that we have some extra method calls on our builder to insert the `chain_id, rpc-responses` objects.

- I was also thinking of providing something in between were we tackle the multichain use-case together with parachain use-case. And sort of evolve the API in a natural way when smoldot adds additional features; or users may need something easier to use than the low level API. 

```rust

/// Naming tbd, suggesting that's somewhere in the middle between super high level and super low level API.
let client: MediumLightClient  = MediumLightClient::builder()
    .add_chain(
         AddChianBuilder::with_identifier( "polkadot" )
             .with_spec( include_str!( "../assets/polkadot_chain_spec.json")
    )
    .add_chain(
         AddChainBuilder::with_identifier( "asset-hub")
            .relay_chain( "polkadot")
            .from_url("ws://127.0.0.1:9944")
     )
    .build().await;
    
let polkadot_api = client.for_chain( "polkadot");
```


Something like that, but I didn't put a proper thought into the API. We still need a way to identify the chains while adding them to the client. And another approach here, would be to have a `fn add_chain(&mut self) -> SubxtChainDetails` and abstract that bit away; as opposed to allowing users to provide strs / tokens.

Another thing worth exploring would be extending the `LightClientBuilder`, maybe we could have a `LightClient::builder().from_token(PolkadotToken)` etc

Still believe having the ability for users to turn a raw light client into something that subxt can use under the scene would cover the most complex use-cases :D

lightclient/src/background.rs
```diff
@@ -43,59 +47,61 @@ pub enum FromSubxt {
...
-    client: smoldot_light::Client<PlatformType>,
-    /// The ID of the chain used to identify the chain protocol (ie. substrate).
-    ///
-    /// Note: A single chain is supported for a client. This aligns with the subxt's
-    /// vision of the Client.
-    chain_id: ChainId,
-    /// Unique ID for RPC calls.
-    request_id: usize,
+    client: smoldot_light::Client<TPlatform, TChain>,
+    /// Generates an unique monotonically increasing ID for each chian.
```
niklasad1, line 61:
```suggestion
    /// Generates an unique monotonically increasing ID for each chain.
```

niklasad1 (approved):
nice and great to have this for sure

as I said in the review, this API is low-level and would be nice with 
some high-level API because one would need understand some configurations 
from smoldot to use it.

subxt/examples/unstable_light_client_parachains.rs
jsdw: Yup, I'm generally for the idea too of having: 
- a simple high level API (like we do)
- an API that allows one to manually configure a `smoldot::Client` or whatever however they need and connect that to subxt (like this PR)
- maybe something inbetween which a `smoldot::Client` could be turned into as one way of building it, or it could be built via some nice higher level APIs that avoid the duplication of passing chain IDs around and such

But yeah it does need some thought and experimentation to find the ideal interface for this stuff I think; we shouldn't rush into it :)

lightclient/src/client.rs
jsdw: I guess I like `for_chain` better than `connect_to_chain` because I think it's already connected, so all we are doing I believe is "selecting" a chain from those that are available to work with. (maybe `seelct_chain` would also work for that reason; I'm easy either way)

lightclient/src/client.rs
```diff
@@ -31,7 +57,14 @@ impl LightClientRpc {
...
+    /// ### Native
+    ///
+    /// Panics when called outside of `tokio` runtime context for native context.
+    ///
+    /// ### Web
+    ///
+    /// If smoldot panics, then the promise created will be leaked. For more details, see
+    /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
     pub fn new(
         config: smoldot_light::AddChainConfig<'_, (), impl Iterator<Item = smoldot_light::ChainId>>,
```
jsdw, line 69: Nit: I wonder whether `impl IntoIterator` would give a bit more flexibility here, if possible?

lightclient/src/client.rs
```diff
@@ -46,22 +79,59 @@ impl LightClientRpc {
...
+    ///
+    /// Panics when called outside of `tokio` runtime context for native context.
+    ///
+    /// ### Web
+    ///
+    /// If smoldot panics, then the promise created will be leaked. For more details, see
+    /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
+    pub fn new_from_client<TPlat>(
+        client: smoldot_light::Client<TPlat>,
+        chains: impl Iterator<Item = AddedChain>,
```
jsdw, line 111: Same here; maybe `Intoiterator` would make it a little easier to use?

lightclient/src/client.rs
```diff
@@ -46,22 +79,59 @@ impl LightClientRpc {
...
+    /// Constructs a new [`RawLightClientRpc`] from the raw smoldot client.
+    ///
+    /// Receives a list of RPC objects as a result of calling `smoldot_light::Client::add_chain`.
+    /// This [`RawLightClientRpc`] can target different chains using [`RawLightClientRpc::for_chain`] method.
+    ///
+    /// ## Panics
+    ///
+    /// ### Native
+    ///
+    /// Panics when called outside of `tokio` runtime context for native context.
```
jsdw, line 103: I wonder if there's any sensible way to hide tokio, so that users don't need to care, and we can potentially remove it as a dependency in the future. Possibly not though; nothing comes to mind offhand!

subxt/src/client/light_client/builder.rs
```diff
@@ -88,7 +89,14 @@ impl<T: Config> LightClientBuilder<T> {
...
     ///
-    /// Panics if being called outside of `tokio` runtime context.
+    /// ### Native
+    ///
+    /// Panics when called outside of `tokio` runtime context for native context.
+    ///
+    /// ### Web
+    ///
+    /// If smoldot panics, then the promise created will be leaked. For more details, see
+    /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
```
jsdw, line 99:
```suggestion
    /// The panic behaviour depends on the feature flag being used:
    ///
    /// ### Native
    ///
    /// Panics when called outside of a `tokio` runtime context.
    ///
    /// ### Web
    ///
    /// If smoldot panics, then the promise created will be leaked. For more details, see
    /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
```

(Am I right in understanding that for the `web` feature flag, it won't panic at all (under any normal operaiton), and if smoldot panics then it will just never return anything (hence the promise might live forever?)

subxt/src/client/light_client/builder.rs
```diff
@@ -116,7 +124,14 @@ impl<T: Config> LightClientBuilder<T> {
...
```
jsdw, line 134: Same suggestion as above

subxt/src/client/light_client/builder.rs
```diff
@@ -147,10 +162,68 @@ impl<T: Config> LightClientBuilder<T> {
...
+        rpc_responses: subxt_lightclient::JsonRpcResponses,
+    ) -> Self {
+        self.chains.push(subxt_lightclient::AddedChain {
+            chain_id,
+            rpc_responses,
+        });
+        self
+    }
+
+    /// Construct a [`RawLightClient`] from a raw smoldot instance.
```
jsdw, line 200: Just to check, is this a `smoldot` instance or specifically just a `smoldot_lightclient::Client` instance?

subxt/src/client/light_client/builder.rs
lexnv: Yep, that is my understanding too. Our wasm module will "panic", which is translated into a javascript exception; and everything will continue to "work" as normal, although we'll not have a smoldot on a background task anymore 


> Note that in wasm panics are currently translated to aborts, but “abort” in this case means that a JavaScript exception is thrown. The wasm module is still usable (likely erroneously) after Rust panics.
> If the future provided panics then the returned Promise will not resolve. Instead it will be a leaked promise. This is an unfortunate limitation of wasm currently that’s hoped to be fixed one day!

subxt/src/client/light_client/builder.rs
lexnv: The `subxt_lightclient::Client` in this case is a re-export of `smoldot_light::Client`. Will modify the docs to say `client` instead of instance here and everywhere else, thanks!

lightclient/src/lib.rs
```diff
@@ -33,8 +33,13 @@ mod platform;
...
 
-pub use client::LightClientRpc;
-pub use smoldot_light::{AddChainConfig, AddChainConfigJsonRpc, ChainId};
+pub use client::{AddedChain, LightClientRpc, RawLightClientRpc};
+pub use smoldot_light::{
+    platform::PlatformRef, AddChainConfig, AddChainConfigJsonRpc, ChainId, Client, JsonRpcResponses,
+};
+
+#[cfg(feature = "native")]
+pub use smoldot_light::platform::default::DefaultPlatform;
```
jsdw, line 42: I'm generally thinking if it's better to expose all of the `smoldot` things behind it's own name; when reading the PR I wasn't sure which of the types actually came from `smoldot` and which were our own.

I think the ideal place to be here is:
- `subxt_lightclient` doesn't need to ever be explicitly depended on; it's just internal logic that Subxt leans on and re-exports as needed.
- It's easy to tell which types are from `smoldot_lightclient` and which types are from our crates (this way, it's super obvious in the example for instance that the RawLightClientBuilder is taking `smoldot` args for instance).

So maybe here we could expose all of the smoldot stuff via a `subxt_lightclient::smoldot::...` and perhaps in subxt we expose our own types in `subxt::client::light_client`, and maybe the smoldot types via either `subxt::ext::smoldot` (or `smoldot_light`.. not sure which makes more sense) or maybe `subxt::client::light_client::smoldot` (or `smoldot_light`) if we prefer to export specific types only and want to namespace them into the light client stuff.

subxt/src/client/light_client/mod.rs
```diff
@@ -51,10 +52,57 @@ pub enum LightClientError {
...
+
+    /// Target a different chain identified by the provided chain ID for requests.
+    ///
+    /// The provided chain ID is provided by the `smoldot_light::Client::add_chain` and it must
+    /// match one of the `smoldot_light::JsonRpcResponses` provided in [`Self::new_from_client`].
+    ///
+    /// # Note
+    ///
+    /// This uses the same underlying instance created by [`Self::new_from_client`].
```
jsdw, line 86: I can't see a `Self::new_from_client` here any more? :)

subxt/src/client/light_client/mod.rs
```diff
@@ -115,11 +163,16 @@ impl<T: Config> LightClient<T> {
     pub fn runtime_api(&self) -> RuntimeApiClient<T, Self> {
         <Self as OfflineClientT<T>>::runtime_api(self)
     }
+
+    /// Returns the chain ID of the current light-client.
+    pub fn chain_id(&self) -> subxt_lightclient::ChainId {
+        self.raw_rpc.chain_id()
```
jsdw, line 169: Does it make more sense now to save just the actual `chain_id` in `LightClient` instead of the `raw_rpc` thing now?

subxt/src/client/light_client/rpc.rs
```diff
@@ -28,7 +44,16 @@ impl LightClientRpc {
     ///
     /// ## Panics
     ///
-    /// Panics if being called outside of `tokio` runtime context.
+    /// The panic behaviour depends on the feature flag being used:
+    ///
+    /// ### Native
+    ///
+    /// Panics when called outside of a `tokio` runtime context.
+    ///
```
jsdw, line 52: Same comment as others :)

subxt/src/client/light_client/builder.rs
jsdw: See my other comment re renaming exports a bit :)

jsdw (approved):
This looks great! Nice clean code :)

I left some comments, mostly small and one regarding jigging some exports around a bit but happy to approve and let you merge when you're happy with everything!

subxt/src/client/light_client/rpc.rs
lexnv: Think this should be updated now 🤔 Might be a github race somewhere

subxt/examples/unstable_light_client_parachains.rs
lexnv: Yep that's a good point, raised https://github.com/paritytech/subxt/issues/1253 to handle it later, thanks!

niklasad1:
@lexnv 

I just realized that creating a "connection" to the same chain more than twice might end up in weird race-conditions because the background task has a `HashMap<(request_id, ChainId)>` which may overwrite a previous request with "the same (request_id; ChainId)" but a different sender...

I don't know whether it makes sense restrict that one can only the connect to the same chain once or refactor that work with several connections....

artifacts/demo_chain_specs/polkadot.json
lexnv: I have created this issue that I ll tackle shortly 👍 

- https://github.com/paritytech/subxt/issues/1263

lightclient/src/lib.rs
lexnv: Yep that makes sense to have different exports.

I modified the exports as:
- subxt_lightclient::smoldot { .. } where smoldot and smoldot_light objects are exported
- then I exported this in subxt::client::light_client::smoldot

PR #1260, opened 2023-11-13
Use `scale-typegen` as a backend for the codegen

### LINKED ISSUE #1247: Handling skipped type params in `subxt-codegen`
### Background context ###

The bridges team uses `subxt-codegen` in order to generate runtime types from metadata. We then use these types in order to submit bridge-related extrinsics to the chains.

[Here](https://github.com/paritytech/parity-bridges-common/blob/c20fc8325678c9c9447412e072bf9463d37a7a43/tools/runtime-codegen/src/main.rs#L110) is the code.

We substitute the types that are extrinsic arguments with the original types, instead of the ones generated by `subxt-codegen`. We encoutered an error recently with one of these types (`sp_runtime::generic::header::Header`). It has 2 type parameters (`pub struct Header<Number: Copy + Into<U256> + TryFrom<U256>, Hash: HashT>`), but recently a [#[scale_info(skip_type_params(Hash))]](https://github.com/paritytech/polkadot-sdk/blob/cd2d5d2579f12dff6d6999b5c9e1f011dcd19a36/substrate/primitives/runtime/src/generic/header.rs#L37) has been added to the original structure. So `subxt-codegen` skips the `Hash`, leading to generated code that goes like:

```
				pub enum Call {
					#[codec(index = 0)]
					submit_finality_proof {
						finality_target: ::std::boxed::Box<
							runtime_types::sp_runtime::generic::header::Header<
								::core::primitive::u32,
							>,
						>,
						justification: ...
					},
```

But, since we substitute it with the original type which has 2 type params, we get compilation errors:

```
error[E0107]: struct takes 2 generic arguments but 1 generic argument was supplied
    --> relays/client-bridge-hub-westend/src/codegen_runtime.rs:1133:31
     |
1133 | ...                   ::sp_runtime::generic::Header<::core::primitive::u32>,
     |                                              ^^^^^^ ---------------------- supplied 1 generic argument
     |                                              |
     |                                              expected 2 generic arguments
     |
note: struct defined here, with 2 generic parameters: `Number`, `Hash`
    --> /home/bparity/.cargo/git/checkouts/polkadot-sdk-cff69157b985ed76/1cf7d3a/substrate/primitives/runtime/src/generic/header.rs:41:12
     |
41   | pub struct Header<Number: Copy + Into<U256> + TryFrom<U256>, Hash: HashT> {
     |            ^^^^^^ ------                                     ----
help: add missing generic argument
     |
1133 |                             ::sp_runtime::generic::Header<::core::primitive::u32, Hash>,
```

### Request ###

I was wondering if it would make sense to keep the original definition of the structure (including the skipped type params) in such cases, for compatibility with type substitutions, even if the skipped typed params wouldn't actually be used if the type is not substituted.

I would be happy to help with the implementation if you don't have time.

### LINKED ISSUE #1268: codegen: Recursive derives for types
Extend the `derive_for_type` to recursively add the derives on all inner fields. 

For example, to derive Clone on `MultiLocation` one would need to do the following:

```rust
derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone"),
derive_for_type(path = "xcm::v2::multilocation::Junctions", derive = "Clone"),
derive_for_type(path = "xcm::v2::junction::Junction", derive = "Clone"),
derive_for_type(path = "xcm::v2::NetworkId", derive = "Clone"),
derive_for_type(path = "xcm::v2::BodyId", derive = "Clone"),
derive_for_type(path = "xcm::v2::BodyPart", derive = "Clone"),
derive_for_type(
    path = "bounded_collections::weak_bounded_vec::WeakBoundedVec",
    derive = "Clone"
)
```

Ideally, developers could type just:
`derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone", recursive),`

tadeohepperle:
This PR uses [`scale-typegen`](https://github.com/paritytech/scale-typegen) to generate the `runtime_types` module and structs for calls, events, etc. 
Let's see if the pipeline works, I manually checked the differences in the `polkadot.rs` file to make sure the differences are valid (almost none).

## Recursive Derives

PR also includes adjustments to the subxt macro to allow for recursive derives. Shown in the `setup_config_custom.rs` example.

Fixes #1268
Fixes #1247 (indirectly)

## Type Aliases

I tried to reconcile the conflicts between the master and this branch, and implemented the alias modules a bit differently, I hope it does not break anything.


cli/src/commands/codegen.rs
```diff
@@ -201,6 +203,10 @@ fn codegen(
         .map_err(|e| eyre!("Cannot decode the provided metadata: {e}"))?;
     let code = codegen
         .generate(metadata)
+        .map_err(|e| {
```
niklasad1, line 206: remove?

codegen/src/api/mod.rs
```diff
@@ -377,28 +342,36 @@ where
         .variants
         .iter()
         .map(|var| {
-            let struct_name = variant_to_struct_name(&var.name);
+            let mut type_params = TypeParameters::from_scale_info(&[]);
+            let composite_ir_kind =
+                type_gen.create_composite_ir_kind(&var.fields, &mut type_params)?;
+            // let fields = CompositeDefFields::from_scale_info_fields(
```
niklasad1, line 348: remove?

codegen/src/api/mod.rs
```diff
@@ -377,28 +342,36 @@ where
...
-
-            Ok((var.name.to_string(), struct_def))
+            let struct_name = variant_to_struct_name(&var.name);
+            let composite_ir = CompositeIR::new(
+                syn::parse_str(&struct_name).expect("enum variant name is valid ident"),
+                composite_ir_kind,
+                type_gen.docs_from_scale_info(&var.docs),
+            );
+
+            // let docs = should_gen_docs.then_some(&*var.docs).unwrap_or_default();
```
niklasad1, line 362: remove?

codegen/src/api/runtime_apis.rs
```diff
@@ -131,13 +135,11 @@ pub fn generate_runtime_apis(
     type_gen: &TypeGenerator,
     types_mod_ident: &syn::Ident,
     crate_path: &syn::Path,
-    should_gen_docs: bool,
+    _should_gen_docs: bool,
```
niklasad1, line 138: remove?!

subxt/examples/setup_config_custom.rs
```diff
@@ -5,15 +5,10 @@ use subxt_signer::sr25519::dev;
...
-    derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone"),
-    derive_for_type(path = "xcm::v2::multilocation::Junctions", derive = "Clone"),
-    derive_for_type(path = "xcm::v2::junction::Junction", derive = "Clone"),
-    derive_for_type(path = "xcm::v2::NetworkId", derive = "Clone"),
-    derive_for_type(path = "xcm::v2::BodyId", derive = "Clone"),
-    derive_for_type(path = "xcm::v2::BodyPart", derive = "Clone"),
     derive_for_type(
-        path = "bounded_collections::weak_bounded_vec::WeakBoundedVec",
-        derive = "Clone"
+        path = "xcm::v2::multilocation::MultiLocation",
```
niklasad1, line 9: :muscle:

testing/integration-tests/src/full_client/codegen/polkadot.rs
lexnv, line 1: It seems that the changes have not affected the aliases and other code generations (except for additional docs, which should be fine). Just to double check, is this file generated from the latest polkadot with all the code changes included right? 🤔

codegen/src/lib.rs
```diff
@@ -167,8 +167,16 @@ impl CodegenBuilder {
         &mut self,
```
lexnv, line 167: We'd need to adjust the docs of this function a bit, the warning section specifically

codegen/src/lib.rs
```diff
@@ -181,11 +189,19 @@ impl CodegenBuilder {
...
```
lexnv, line 189: Same doc adjustment here, since we now have the recursive parameter

cli/src/commands/codegen.rs
```diff
@@ -168,7 +168,8 @@ fn codegen(
             .map_err(|e| eyre!("Cannot parse derive for type {ty_str}: {e}"))?;
         let derive = syn::parse_str(&derive)
             .map_err(|e| eyre!("Cannot parse derive for type {ty_str}: {e}"))?;
-        codegen.add_derives_for_type(ty, std::iter::once(derive));
+        // Note: recursive derives and attributes not supported in the CLI => recursive: false
```
lexnv, line 171: Would we want to support recursive derives in the CLI? Might be worth a separate issue to remind us of it

codegen/src/api/calls.rs
```diff
@@ -21,74 +21,63 @@ use subxt_metadata::PalletMetadata;
...
-                CompositeDefFields::Named(ref named_fields) => named_fields
+    let (call_structs, call_fns): (Vec<_>, Vec<_>) = variant_names_and_struct_defs
+        .into_iter()
+        .map(|var| {
+            let (call_fn_args, call_args): (Vec<_>, Vec<_>) = match &var.composite.kind {
+                CompositeIRKind::Named(named_fields) => named_fields
                     .iter()
                     .map(|(name, field)| {
-                        let call_arg = if field.is_boxed() {
+                        // Note: fn_arg_type this is relative the type path of the type alias when prefixed with `types::`, e.g. `set_max_code_size::New`
```
lexnv, line 44: Could there be a variant of the metadata where this does not correspond to the expected alias? Or it would be different from the previous `format_ident!("{}", name.to_string().to_upper_camel_case())`?

codegen/src/api/constants.rs
```diff
@@ -36,9 +36,7 @@ use super::CodegenError;
 pub fn generate_constants(
     type_gen: &TypeGenerator,
     pallet: &PalletMetadata,
```
lexnv, line 38: Same for the docs here

codegen/src/api/errors.rs
```diff
@@ -20,8 +19,8 @@ pub fn generate_error_type_alias(
         return Ok(quote!());
     };
 
-    let error_type = type_gen.resolve_type_path(error_ty);
-    let error_ty = type_gen.resolve_type(error_ty);
+    let error_type = type_gen.resolve_type_path(error_ty)?;
+    let error_ty = type_gen.resolve_type(error_ty)?;
     let docs = &error_ty.docs;
     let docs = should_gen_docs
```
lexnv, line 25: Could this be infered from the `type_gen.settings()`?

codegen/src/api/mod.rs
```diff
@@ -114,24 +109,20 @@ impl RuntimeGenerator {
     pub fn generate_runtime(
         &self,
         item_mod: syn::ItemMod,
-        derives: DerivesRegistry,
-        type_substitutes: TypeSubstitutes,
+        derives: scale_typegen::DerivesRegistry,
+        type_substitutes: scale_typegen::TypeSubstitutes,
         crate_path: syn::Path,
         should_gen_docs: bool,
```
lexnv, line 115: Is there any particular reason for having this? It is only passed to the error generation, while others use the `type_gen` settings

codegen/src/api/errors.rs
tadeohepperle: Oh, totally! Good catch!

codegen/src/api/mod.rs
tadeohepperle: What do you mean? Having `scale_typegen::DerivesRegistry` specified fully, is not necassary, I should just import `DerivesRegistry`. Maybe I understand the question wrong.

testing/integration-tests/src/full_client/codegen/polkadot.rs
tadeohepperle: Yes, I ran the `polkadot.rs` file generation command after implementing all the code changes in this PR. :)

cli/src/commands/codegen.rs
tadeohepperle: Hmm, maybe. Or maybe we just make everything recursive by default in the CLI, I think it is already quite hard to specify derives in the CLI.

codegen/src/api/mod.rs
lexnv: Ops, I forgot to type `should_gen_docs`

lexnv:
The code looks clean and simplifies the code quite a lot! Nice one! 👍 

Just some nits about the documentation and inferring the type alias paths

codegen/src/api/mod.rs
tadeohepperle: Ah okay, the `should_gen_docs` we actually need because the typegen settings are not built up yet. They are only built in this very function using the `crate_path`, `should_gen_docs`, substitutes and derives :)

codegen/src/api/calls.rs
tadeohepperle: Good question, when calling `generate_structs_from_variants`, which calls `generate_type_alias_mod` for each such composite, for each field, the field.type_path is replaced by `#alias_mod_name::#alias_name`, where `alias_name` is either `format_ident!("{}", name.to_string().to_upper_camel_case())` (named fields) or `format_ident!("Field{}", i)` for unnamed fields, so the logic should be very similar to what you implemented before. I cannot think of a case right now where this is not fulfilled.

cli/src/commands/codegen.rs
niklasad1: Would be great if you could open an issue for it :)

codegen/src/api/mod.rs
```diff
@@ -12,20 +12,20 @@ mod events;
 mod runtime_apis;
 mod storage;
 
+use scale_typegen::typegen::ir::type_ir::{CompositeFieldIR, CompositeIR, CompositeIRKind};
```
niklasad1, line 15: would be nice to export all these types from the root in `scale_typegen` all the types are already unique and I don't think for instance `scale_typegen::typegen::ir::type_ir::CompositeFieldIR` is necessary and `scale_typegen::CompositeFieldIR` would be sufficient.

Nothing to address here just a comment

tadeohepperle: Good point! We already expose some other types at the root, I did not anticipate that we would need easy access to these structs.

codegen/src/api/mod.rs
```diff
@@ -378,84 +342,91 @@ where
...
-                docs,
-                crate_path,
-                Some(alias_module_name.clone()),
-            )?;
-
-            let type_aliases = TypeAliases::new(fields, types_mod_ident.clone(), alias_module_name);
-
-            Ok((var.name.to_string(), struct_def, type_aliases))
+            let mut composite = CompositeIR::new(
+                syn::parse_str(&struct_name).expect("enum variant name is valid ident"),
```
niklasad1, line 350: This should be infallible

codegen/src/api/custom_values.rs
```diff
@@ -60,6 +59,7 @@ fn generate_custom_value_fn(
     let (return_ty, decodable) = if type_is_valid {
         let return_ty = type_gen
             .resolve_type_path(custom_value.type_id())
+            .expect("expected to contain type id")
```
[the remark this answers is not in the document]
tadeohepperle: The `CustomValueMetadata` that this oparates on should always point to a valid type id, I expect this to never fail, except if someone corrupted the metadata.

codegen/src/api/mod.rs
tadeohepperle: Yes, unless the metadata is somehow corrupted. Should I formulate the `expect` message differently here?

codegen/src/api/storage.rs
```diff
@@ -71,25 +58,28 @@ pub fn generate_storage(
...
     let alias_name = format_ident!("{}", storage_entry.name().to_upper_camel_case());
     let alias_module_name = format_ident!("{snake_case_name}");
     let alias_storage_path = quote!( types::#alias_module_name::#alias_name );
 
     let storage_entry_map = |idx, id| {
         let ident: Ident = format_ident!("_{}", idx);
-        let ty_path = type_gen.resolve_type_path(id);
+        let ty_path = type_gen
+            .resolve_type_path(id)
+            .expect("type should be present");
```
niklasad1, line 82: Infallible? Is it possible to for some type to not be in the PortableRegistry? I'm just asking wheter subxt should provide a better compile error than the expect.

niklasad1: Before I think subxt generated a Codegen Error

codegen/src/api/mod.rs
```diff
@@ -378,84 +342,91 @@ where
...
+/// }
+/// ```
+/// And the type alias module will look like this:
+/// ```ignore
+/// pub mod set_max_code_size {
+///     use super::runtime_types;
+///     pub type New = ::core::primitive::u32;
+/// }
+/// ```
+pub fn generate_type_alias_mod(
```
niklasad1, line 393: Is this used for type substitution or when is it used?

niklasad1 (approved):
Looks great to me

codegen/src/api/mod.rs
niklasad1: I prefer what we use in parity with a explanation/"semi-formal" proof why it can't panic such `"enum variant is a valid; qed"` that is very clear to me :)

codegen/src/api/custom_values.rs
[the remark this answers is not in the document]
niklasad1: ok, I just think it would be nice with a little better message if this panics i.e, ends up with compilation error.

codegen/src/api/storage.rs
tadeohepperle: Should also be infallible, if we assume valid metadata.

codegen/src/api/mod.rs
tadeohepperle: This is used in the `generate_structs_from_variants` function above that we use to generate special functions for `events` and `calls`. For each of these calls/events we also generate a struct that bundles together the arguments needed for the call / event query. Because some of these arguments are very long complicated types, we do not use the types directly, instead we generate a small module next to this the struct for each call / event. The module has the same name as the struct, just in snake case. It contains a type alias for each field of the struct. This `generate_type_alias_mod` function is responsible for generating these little modules for each call / event. 
But it does not change much in terms of usage, the type aliases are basically just syntactic sugar. No types are actually changed/substituted.

niklasad1: got it, thanks

cli/src/commands/diff.rs
```diff
@@ -309,12 +309,12 @@ fn storage_differences<'a>(
         |e| {
             pallet_metadata_1
                 .storage_hash(e.name())
-                .expect("storage entry should be present")
+                .expect("storage entry is in medadata; qed")
```
niklasad1, line 312:
```suggestion
                .expect("storage entry is in metadata; qed")
```

lexnv (approved):
Nice one! 👍

PR #1264, opened 2023-11-14
config: Add `SkipCheckIfFeeless` signed extension

### LINKED ISSUE #1262: Subxt integration tests failed against latest Substrate build.

lexnv:
This PR adds the `SkipCheckIfFeeless` to the signed extensions of subxt.

The extension was introduced in substrate to skip calculating feels if the extrinsics are feeless.

From [substrate](https://github.com/paritytech/polkadot-sdk/blob/05ddbf57a154f350e0015d3264543bcda09153cf/substrate/frame/transaction-payment/skip-feeless-payment/src/lib.rs#L80), the `SkipCheckIfFeeless` is simply a wrapper over another signed extension.

```rust
/// A [`SignedExtension`] that skips the wrapped extension if the dispatchable is feeless.
#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
#[scale_info(skip_type_params(T))]
pub struct SkipCheckIfFeeless<T: Config, S: SignedExtension>(pub S, sp_std::marker::PhantomData<T>);

```

This extension will forward for encoding whatever extension it wraps.
The extension has implications for the `pre_dispatch` method where the inner extension is conditionally disabled depending on whether the transaction is feeless or not. As well for the `post_dispatch` to generate an event specific to the feeless extrinsics named `FeeSkipped`.

From the subxt perspective, we need to implement our `SignedExtension` trait for the `SkipCheckIfFeeless`. Doing so will propagate the extension name that is correlated for decoding purposes and it fixes errors such as:

```bash
 Error: ExtrinsicParams(UnknownSignedExtension("SkipCheckIfFeeless"))
 ```

Closes https://github.com/paritytech/subxt/issues/1262

### DIFF
--- a/subxt/examples/setup_config_custom.rs
+++ b/subxt/examples/setup_config_custom.rs
@@ -3,7 +3,19 @@ use subxt::client::OfflineClientT;
 use subxt::config::{Config, ExtrinsicParams, ExtrinsicParamsEncoder};
 use subxt_signer::sr25519::dev;
 
-#[subxt::subxt(runtime_metadata_path = "../artifacts/polkadot_metadata_full.scale")]
+#[subxt::subxt(
+    runtime_metadata_path = "../artifacts/polkadot_metadata_full.scale",
+    derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone"),
+    derive_for_type(path = "xcm::v2::multilocation::Junctions", derive = "Clone"),
+    derive_for_type(path = "xcm::v2::junction::Junction", derive = "Clone"),
+    derive_for_type(path = "xcm::v2::NetworkId", derive = "Clone"),
+    derive_for_type(path = "xcm::v2::BodyId", derive = "Clone"),
+    derive_for_type(path = "xcm::v2::BodyPart", derive = "Clone"),
+    derive_for_type(
+        path = "bounded_collections::weak_bounded_vec::WeakBoundedVec",
+        derive = "Clone"
+    )
+)]
 pub mod runtime {}
 use runtime::runtime_types::xcm::v2::multilocation::MultiLocation;
 
--- a/subxt/examples/setup_config_signed_extension.rs
+++ b/subxt/examples/setup_config_signed_extension.rs
@@ -81,8 +81,8 @@ impl ExtrinsicParamsEncoder for CustomSignedExtension {
 pub fn custom(
     params: DefaultExtrinsicParamsBuilder<CustomConfig>,
 ) -> <<CustomConfig as Config>::ExtrinsicParams as ExtrinsicParams<CustomConfig>>::OtherParams {
-    let (a, b, c, d, e, f, g) = params.build();
-    (a, b, c, d, e, f, g, ())
+    let (a, b, c, d, e, f, g, h) = params.build();
+    (a, b, c, d, e, f, g, (), ())
 }
 
 #[tokio::main]
--- a/subxt/examples/setup_config_signed_extension.rs
+++ b/subxt/examples/setup_config_signed_extension.rs
@@ -1,4 +1,5 @@
 use codec::Encode;
+use scale_encode::EncodeAsType;
 use subxt::client::OfflineClientT;
 use subxt::config::signed_extensions;
 use subxt::config::{
@@ -11,6 +12,7 @@ pub mod runtime {}
 
 // We don't need to construct this at runtime,
 // so an empty enum is appropriate:
+#[derive(EncodeAsType)]
 pub enum CustomConfig {}
 
 impl Config for CustomConfig {
@@ -31,9 +33,10 @@ impl Config for CustomConfig {
             signed_extensions::CheckGenesis<Self>,
             signed_extensions::CheckMortality<Self>,
             signed_extensions::ChargeAssetTxPayment<Self>,
+            signed_extensions::ChargeTransactionPayment,
             signed_extensions::SkipCheckIfFeeless<
                 Self,
-                signed_extensions::ChargeTransactionPayment,
+                signed_extensions::ChargeAssetTxPayment<Self>,
             >,
             // And add a new one of our own:
             CustomSignedExtension,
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Review by tadeohepperle — APPROVED** [review]

**Comment by jsdw** [comment]
```rust
#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
#[scale_info(skip_type_params(T))]
pub struct SkipCheckIfFeeless<T: Config, S: SignedExtension>(pub S, sp_std::marker::PhantomData<T>);
```

This implies that the inner signed extension will be encoded into the tx payload; your code doesn't appear to cater for this?

Two cases to consider:
- How would a user configure the `ExtraParams` to use this, so that they can provide an instance of this with whatever signed ext wrapped inside it?
- How can we decode/skip over the bytes for this reliably, regardless of which signed extension it wraps (we can prob use metadata to skip over the relevant bytes?)

Anyway, back to being away hehe :)

subxt/src/config/default_extrinsic_params.rs
```diff
@@ -16,8 +16,7 @@ pub type DefaultExtrinsicParams<T> = signed_extensions::AnyOf<
         signed_extensions::CheckGenesis<T>,
         signed_extensions::CheckMortality<T>,
         signed_extensions::ChargeAssetTxPayment<T>,
-        signed_extensions::SkipCheckIfFeeless,
-        signed_extensions::ChargeTransactionPayment,
+        signed_extensions::SkipCheckIfFeeless<T, signed_extensions::ChargeTransactionPayment>,
```
niklasad1, line 19: This will break polkadot, westend and rococo.

AFAIU, only substrate-node has this so far

niklasad1: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/westend/src/lib.rs#L836

jsdw: AnyOf is clever, and will just ignore signed extensions that aren't used on a given node, so it's always good to add all of the common ones to this list (other comment notwithstanding) :)

subxt/examples/setup_config_signed_extension.rs
```diff
@@ -31,7 +31,10 @@ impl Config for CustomConfig {
             signed_extensions::CheckGenesis<Self>,
             signed_extensions::CheckMortality<Self>,
             signed_extensions::ChargeAssetTxPayment<Self>,
-            signed_extensions::ChargeTransactionPayment,
```
jsdw, line 34: prob worth keeping this in the default list so it'll keep working with chains that don't wrap the charge asset signed ext (we'll just ignore it if they do)

subxt/src/config/default_extrinsic_params.rs
jsdw: Ooh and yeah, let's not remove the charge asset one from the list, which I just noticed and maybe that ls what you were saying on second read :)

subxt/src/config/signed_extensions.rs
```diff
@@ -370,6 +371,95 @@ impl<T: Config> SignedExtension<T> for ChargeTransactionPayment {
...
+    }
+}
+
+impl<T, S> ExtrinsicParamsEncoder for SkipCheckIfFeeless<T, S>
+where
+    T: Config,
+    S: SignedExtension<T>,
+{
+    fn encode_extra_to(&self, v: &mut Vec<u8>) {
+        self.0.encode_extra_to(v);
```
jsdw, line 422: Since this can be used with different exts inside, I wonder how easily we could make it use encode_as_type to encode the inner extension based on the metadata. That'd offer extra safety because then we'd get an error if we try to encode an inner signed ext that doesn't match what the node is expecting, whereas without metadata we'd just blindly encode and hope for the best (and prolly get a super obscure decode error back from the node if it's wrong!)

It could enforce that the signed ext impls EncodeAsType too maybe! Not sure :)

subxt/src/config/default_extrinsic_params.rs
niklasad1: All right, thanks.

So you are saying that we shall have both `SkipCheckIfFeeLess` and `ChargeTransactionPayment` here, correct?

niklasad1 (approved):
nice

subxt/examples/setup_config_custom.rs
```diff
@@ -3,7 +3,19 @@ use subxt::client::OfflineClientT;
 use subxt::config::{Config, ExtrinsicParams, ExtrinsicParamsEncoder};
 use subxt_signer::sr25519::dev;
 
-#[subxt::subxt(runtime_metadata_path = "../artifacts/polkadot_metadata_full.scale")]
+#[subxt::subxt(
+    runtime_metadata_path = "../artifacts/polkadot_metadata_full.scale",
+    derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone"),
```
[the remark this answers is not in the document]
lexnv: Nope :( I've raised https://github.com/paritytech/subxt/issues/1268 to track this in the future. 

https://github.com/paritytech/subxt/blob/eeacf4b041a287b33f665e9b72fd37bd0218e1d2/subxt/src/lib.rs#L231-L235

subxt/src/config/signed_extensions.rs
```diff
@@ -370,6 +382,201 @@ impl<T: Config> SignedExtension<T> for ChargeTransactionPayment {
...
+    type Error = ExtrinsicParamsError;
+
+    fn new<Client: OfflineClientT<T>>(
+        nonce: u64,
+        client: Client,
+        other_params: Self::OtherParams,
+    ) -> Result<Self, Self::Error> {
+        let other_params = match other_params.0 {
+            Some(other_params) => other_params,
+            None => <S as ExtrinsicParams<T>>::OtherParams::default(),
```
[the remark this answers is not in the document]
lexnv: Have changed with `wrap_or_default()`, thanks for the suggestion 🙏

Issue #1190, opened 2023-10-05
rpc: add retry logic to the default WS client


niklasad1:
Apparently users are having issues with that RPC client connection closes https://github.com/paritytech/jsonrpsee/issues/949#issuecomment-1747401886 and it's not possible to detect that easily in subxt.

Suggestions:
1. Provide an abstraction for "a reconnecting WS client" re-established active subscriptions and re-send requests that were not finished. The downside it that it's possible to loose a subscription notifications when reconnecting.  Lost notifications which may cause to submit a transaction twice for something `author_submitAndWatch` if the `Finalized event` is lost. 
2. Add example code how to deal with lost connections in subxt, it's unfortunately not super easy to implement because to RpcError variant has to be downcasted to a JsonrpseeError and check if it's `JsonrpseeError::RestartedNeeded`. Sadly this is not possible to do without restarting the subxt client.

There may be more similar issues with the recently added  RPC spec v2 implementation.
Other ideas are welcome  

/cc @jsdw @lexnv

ara-selini:
Hi, coming from https://github.com/paritytech/jsonrpsee/issues/949 

I am using a simple block subscriber now. To simulate a network error, I turn off my wifi.  The code does not provide any errors or even `None`s -- it just keeps awaiting for the next element. Both for `subxt` and `jsonrpsee`. 

When using `jsonrpsee`, it is possible to detect failures similarly to the code snippet you shared in the above link but doing that in `subxt` is pretty difficult e.g. this code **doesn't** work because `subxt` takes ownership of the client:

```
        let ws_client = WsClientBuilder::default().build(node_url).await.unwrap();
        tokio::spawn(async move {
            ws_client.on_disconnect().await;
            println!("Disconnected");
        });

        let rpc_client = RpcClient::new(ws_client);
        let client = OnlineClient::<PolkadotConfig>::from_rpc_client(rpc_client).await?;
```

So it sounds like detecting disconnects is difficult in `subxt` which means I can't go ahead with it in sensitive production code. Are there any temporary fixes (or other libraries) you would recommend in the meantime? Alternatively, I might just go ahead with implementing the functionality I need with reconnecting on top of `jsonrpsee` but that wouldn't be elegant.

niklasad1:
@ara-selini 

You need to put the ws-client in an `Arc`for that to work but that doesn't seems to work in subxt v0.32

@jsdw 

[...]

PR #1277, opened 2023-11-21
impl RpcClientT for Arc<T> and Box<T>

niklasad1:
With the RPC client refactoring it's not possible to create a subxt::RpcClient from `Arc<WsClient>` which this PR fixes.

See my comment here https://github.com/paritytech/subxt/issues/1190#issuecomment-1820508297


jsdw (approved):
LGTM (potentially we could impl `RpcClientT` for `Arc<T>`, `Box<T>`, `Rc<T>` to be a bit more generally applicable)

lexnv (approved):

jsdw (commented):
One final thing; move the generic `RpcClientT` impls out of the `jsonrpsee_impl` mod (which is only conditionally active) and put next to the `trait RpcClientT`?

PR #1278, opened 2023-11-21
 cli: Command to fetch chainSpec and optimise its size

### LINKED ISSUE #1263: cli: Command to fetch chainSpec and optimise its size
Add a cli command that:
- fetches the chainSpec needed to start a light client from a running node
- add an extra config to enable / disable chainSpec size optimization

The size optimization can be done by hashing the `genesis.raw` entries, obtaining as a result a `stateRootHash` that will drastically reduce the chainSpec size; and allow for faster light client initialization.

Then, this CLI tool together with `scripts/artifacts.sh` can run in a cronjob to automatically update:
- demo chain specs
- metadata files
- codegen `polkadot.rs` file

lexnv:
This PR adds a new CLI `chain-spec` command.

```bash
cargo run -- chain-spec --url wss://rpc.polkadot.io:443 \
  --output-file artifacts/demo_chain_specs/polkadot.json \
  --state-root-hash \
  --remove-substitutes
```

###  URL
The spec is fetched from the address specified as `--url` by making a raw RPC call to the method `sync_state_genSyncSpec` with parameters `true` (obtains the raw entry).

### Output File

The ` --output-file` is optional and represents the path at which the chainSpec is written, by default `stdout` is used.

### State Root Hash

The `--state-root-hash` replaces `genesis.raw` with `genesis.stateRootHash`, by decoding the provided chainSpec and computing the `Blake2` hash of the entries. Smoldot helpers are used for simplicity, given that our main goal is to have a compatible chainSpec with the light-client.

This should be used for optimization purposes, both in terms of reducing the size of the chainSpec, as well as improving the synchronization time of the light-client with the target chain. When a `genesis.raw` entry is provided to the light client, the light client would need to compute the `stateRootHash` internally. 

### Remove Substitutes

This option removes the `codeSubstitutes` entry of the chainSpec. At this moment, the light client does not utilize this entry. Enabling this option would lead to a smaller size chainSpec, which is desirable to generate a minimal working chainSpec, especially for resource constraint environments.

## Testing Done

The `--state-root-hash` command is tested against the output of the smoldot client.

## Limitations

Fetching the chainSpec relies on the legacy RPC method `sync_state_genSyncSpec`, which is not exposed by all substrate-based chains. Such is the case of the parachain `PolkadotAssetHub`. 

Ideally, we would need a `chainSpec_getSpec` that provides this information for all substrate chains. Initial spec issue that is an extension of this:
- https://github.com/paritytech/json-rpc-interface-spec/issues/68

Closes: https://github.com/paritytech/subxt/issues/1263


cli/src/commands/chain_spec/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// The url of the substrate node to query for metadata for codegen.
+    #[clap(long)]
+    url: Url,
+    /// Write the output of the command to the provided file path.
+    #[clap(long, short, value_parser)]
+    output_file: Option<PathBuf>,
+    /// Replaced the genesis raw entry with a stateRootHash to optimize
+    /// the spec size and avoid the need to calculate the genesis storage.
+    ///
+    /// This option is enabled with the `chain-spec-root-hash` feature.
```
tadeohepperle, line 25: Nit: Here I think the feature mentioned in the docs should be "chain-spec" instead of "chain-spec-root-hash".

cli/src/commands/chain_spec/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    let url = opts.url;
+
+    let mut spec = fetch::fetch_chain_spec(url).await?;
+
+    let mut output: Box<dyn Write> = match opts.output_file {
+        Some(path) => Box::new(std::fs::File::create(path)?),
+        None => Box::new(output),
+    };
+
+    #[cfg(feature = "chain-spec")]
```
tadeohepperle, line 102: I am a bit torn about the `chain-spec` feature. It is the first feature flag introduced to the subxt CLI tool and having feature flags on the CLI tool means that people can have different experiences using the subxt CLI tool, depending on which features they installed it with. However I understand that smoldot is quite a big dependency that we do not want to expose unnecessarily for all users.
We could also guard the entire `chainspec` subcommand behind the `"chain-spec"` feature flag, then we do not need so many `#[cfg(feature = "chain-spec")]` instances throughout the code. But if we keep it like this, I would probably rename it to something more clear like `chain-spec-pruning`.

tadeohepperle (approved):
All in all a great addition to the CLI :) Just not really sure about the use of the feature flag.

cli/src/commands/chain_spec/fetch.rs
```diff
@@ -0,0 +1,63 @@
...
+    client_transport::ws::WsTransportClientBuilder,
+    core::{client::ClientT, Error},
+    http_client::HttpClientBuilder,
+};
+use std::time::Duration;
+use subxt_codegen::fetch_metadata::Url;
+
+/// Returns the node's chainSpec from the provided URL.
+pub async fn fetch_chain_spec(url: Url) -> Result<serde_json::Value, FetchSpecError> {
+    async fn fetch_ws(url: Url) -> Result<serde_json::Value, Error> {
```
niklasad1, line 16: This is clean but my hunch is that we have duplicated this in a few places now, maybe a util/helper crate for this is worth it? 

No need to address it in this PR but just a thought, worth filing an issue IMO :)

cli/src/commands/chain_spec/fetch.rs
```diff
@@ -0,0 +1,63 @@
...
+    let spec = match url.scheme() {
+        "http" | "https" => fetch_http(url).await.map_err(FetchSpecError::RequestError),
+        "ws" | "wss" => fetch_ws(url).await.map_err(FetchSpecError::RequestError),
+        invalid_scheme => Err(FetchSpecError::InvalidScheme(invalid_scheme.to_owned())),
+    }?;
+
+    Ok(spec)
+}
+
+/// Error attempting to fetch chainSpec.
```
niklasad1, line 53: this error is specific for the JSON-RPC connection and I would be super happy with some util crate that instead :)

cli/src/commands/chain_spec/mod.rs
niklasad1: Would be nice to get rid of the feature completely  :)

niklasad1 (approved):
Looks good but I agree with Tadeo that it would be nice to get rid of the `chain-spec` feature flag if the number of dependencies are "reasonable"

lexnv:
| | With Feature Flag chain-spec | No Feature Flag |
|-|-|-|
|Dependencies| 520 | 707 |
|Build Time |  29.26s  | 35.41s |

I believe having a feature-flag here might be worth the effort, but would love to hear your thoughts on this 🙏 

Other revenues to avoid the feature-flag:
- Add a feature-flag to smoldot https://github.com/smol-dot/smoldot/blob/main/lib/Cargo.toml
- Publish the `smoldot::trie` as a separate crate
- Create a custom `subxt-chain-spec` tool crate that has minimal dependency/code to generate the `stateRootHash` 

I do like the third option, although it might take some time until we properly prioritize that

### Methodology 
For dependencies, I've used cargo tree. And for build time, I did a `cargo clean`, followed by running the CLI command for chainSpec.


### Util-crate

While at it, have raised an issue about the util-crate for the chain-spec: https://github.com/paritytech/subxt/issues/1293

niklasad1:
Ok, I think you made right choice with a feature flag all good from myside

PR #1283, opened 2023-11-22
Improve SignedExtension matching logic and remove SkipCheckIfFeeless bits

### LINKED ISSUE #1280: Alter `SignedExtension` and related to cater for more advanced exts like `SkipCheckIfFeeless`
`SkipCheckIfFeeless`  is a signed ext that wraps another, but externally just has the name `SkipCheckIfFeeless`.

The current impl of it on master will return an error if the ext is _not_ present on a chain, rather than being able to be silently ignored if not needed. But more than that, matching on just the name of the signed ext is bound to lead to future issues with this sort of ext in existence (if a chain has two exts wrapped in `SkipCheckIfFeeless`, we won't be able to match things up properly any more).

So, the goal here will be to modify the related traits so that signed extensions can search for themselves more intelligently, and also have the ability to return errors at the point of encoding (because they become more possible.

jsdw:
Closes #1280.

- `SkipCheckIfFeeless` is now transparent to the outside world in the latest polkadot-sdk, so remove logic specific to that.
- But, remove `SignedExtension::IDENTIFIER` and replace with `SignedExtension::matches(..)` function which is able to do much more custom/intelligent matching to work out which of our signed extensions to use at which point (in part, just in case we have more complex things to do later, and in part just to provide the flexibility now for anybody that wants it).
- Tidy some bounds (Debug/Clone only for the signed exts the user might handle ie the ones that are Decode targets) and no need for `EncodeAsType`/`DecodeAsType` on most of the exts now.
- Just return `ExtrinsicParamsError` rather than `type Error: Into<ExtrinsicParamsError>`.

### DIFF
--- a/subxt/src/blocks/extrinsic_types.rs
+++ b/subxt/src/blocks/extrinsic_types.rs
@@ -13,7 +13,7 @@ use crate::{
 };
 
 use crate::config::signed_extensions::{
-    ChargeAssetTxPayment, ChargeTransactionPayment, CheckNonce, SkipCheckIfFeeless,
+    ChargeAssetTxPayment, ChargeTransactionPayment, CheckNonce,
 };
 use crate::config::SignedExtension;
 use crate::dynamic::DecodedValue;
@@ -660,24 +660,22 @@ impl<'a, T: Config> ExtrinsicSignedExtensions<'a, T> {
         })
     }
 
-    fn find_by_name(&self, name: &str) -> Option<ExtrinsicSignedExtension<'_, T>> {
-        let signed_extension = self
-            .iter()
-            .find_map(|e| e.ok().filter(|e| e.name() == name))?;
-        Some(signed_extension)
-    }
-
     /// Searches through all signed extensions to find a specific one.
     /// If the Signed Extension is not found `Ok(None)` is returned.
     /// If the Signed Extension is found but decoding failed `Err(_)` is returned.
     pub fn find<S: SignedExtension<T>>(&self) -> Result<Option<S::Decoded>, Error> {
-        self.find_by_name(S::NAME)
-            .map(|s| {
-                s.as_signed_extra::<S>().map(|e| {
-                    e.expect("signed extra name is correct, because it was found before; qed.")
-                })
-            })
-            .transpose()
+        for ext in self.iter() {
+            let Ok(ext) = ext else { continue };
+            match ext.as_signed_extension::<S>() {
+                // We found a match; return it:
+                Ok(Some(e)) => return Ok(Some(e)),
+                // No error, but no match either; next!
+                Ok(None) => continue,
+                // Error? return it
+                Err(e) => return Err(e),
+            }
+        }
+        Ok(None)
     }
 
     /// The tip of an extrinsic, extracted from the ChargeTransactionPayment or ChargeAssetTxPayment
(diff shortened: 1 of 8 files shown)


subxt/src/blocks/extrinsic_types.rs
```diff
@@ -660,24 +660,22 @@ impl<'a, T: Config> ExtrinsicSignedExtensions<'a, T> {
...
     pub fn find<S: SignedExtension<T>>(&self) -> Result<Option<S::Decoded>, Error> {
-        self.find_by_name(S::NAME)
-            .map(|s| {
-                s.as_signed_extra::<S>().map(|e| {
-                    e.expect("signed extra name is correct, because it was found before; qed.")
-                })
-            })
-            .transpose()
+        for ext in self.iter() {
+            let Ok(ext) = ext else { continue };
```
niklasad1, line 668: The error is ignored here, can it be returned as an error if no other valid signed extension was found?

Or what's reason about that?

subxt/src/config/signed_extensions.rs
```diff
@@ -9,42 +9,45 @@
...
     /// Decoding from this type should be symmetrical to the respective
     /// `ExtrinsicParamsEncoder::encode_extra_to()` implementation of this signed extension.
     type Decoded: DecodeAsType;
+
+    /// This should return true if the signed extension matches the details given.
+    /// Often, this will involve just checking that the identifier given matches that of the
+    /// extension in question.
+    ///
+    /// The first match that returns true will be the entry that this signed extension
+    /// is used to encode values for. This takes `&mut self`, allowing the extension to
```
niklasad1, line 35: I don't understand the &mut self part, can you elaborate for my slow eyes?

subxt/src/config/signed_extensions.rs
```diff
@@ -258,22 +243,11 @@ impl<T: Config> ChargeAssetTxPayment<T> {
 }
 
 /// Parameters to configure the [`ChargeAssetTxPayment`] signed extension.
-#[derive(Debug)]
```
niklasad1, line 261: why?

niklasad1 (approved):
I like it, it simplifies the signed extension impls.

tadeohepperle (approved):
I don't really understand the reasons why SkipCheckIfFeeless was introduced and why we don't need it as a singed extension in subxt anymore.

niklasad1:
> I don't really understand the reasons why SkipCheckIfFeeless was introduced and why we don't need it as a singed extension in subxt anymore.

AFAIU, SkipCheckIfFeeless was introduced as "wrapper signed extension" that wrapped "ChargeTxPayment (not sure if that's correct but doesn't matter)", initially the name of the signed extension was "SkipCheckIfFeeless" but was reverted to propagate the the name of signed extension as it wrapped.

Such SkipCheckIfFeeLess<ChargeTxPayment>, its name signed extension name will be "ChargeTxPayment" and not "SkipCheckIfFeeLess"

subxt/src/config/signed_extensions.rs
jsdw: Ah whoops; it took `&mut self` in a prev iteration and I forgot to updatethe comment; good spot!

subxt/src/config/signed_extensions.rs
jsdw: it would never have worked anyway afaik, because `T: Config` doesn't impl `Debug`. Maybe I was a bit overzealous in removing a bunch of derives for these things though; I did initially replace them all with working ones and then backed out of it!

jsdw:
> I don't really understand the reasons why SkipCheckIfFeeless was introduced and why we don't need it as a singed extension in subxt anymore.

Yup, a couple of Substrate PRs made the name return the innet signed ext name, and the typeinfo match the inner typeinfo, so `SkipCheckIfFeeless` is now invisible to us and can be ignored again :)

lexnv (approved):

subxt/src/blocks/extrinsic_types.rs
jsdw: That's a good point; initially I just skipped any iteration error with the reaosning that from `find`s perspective, we only care about an error if we find a signed ext and then can't decode it into the type we were hoping, and any iteration errors were sortof unrelated.

But that was probably a bit silly, and anyway, if we do hit an iteration error, we'll bail from `iter` anyway, so we may as well just return it!

PR #1289, opened 2023-11-23
lightclient: Fix wasm socket closure called after being dropped

### LINKED ISSUE #1288: Fix errors that crop up when running the lightclient in WASM
To reproduce the errors:

```
cd testing/wasm-lightclient-tests 
WASM_BINDGEN_TEST_TIMEOUT=600 wasm-pack test --firefox
```

Then open a browser to the URL printed and open the dev console in the browser.

lexnv:
The errors reported by the wasm related to `Error closure invoked recursively or destroyed already` could happen whenever a `Closure` object is used after being dropped.

This PR makes the assumption that the JS realm socket created by `web_sys::WebSocket` is able to access the references of the registered `Closure` callbacks, even after the Rust object has been dropped due to the socket not being closed properly.

https://github.com/paritytech/subxt/blob/f3eb4cdf350eb1f83b5ec478be93321631adad9a/lightclient/src/platform/wasm_socket.rs#L103

This PR makes the following adjustments to fix and improve the code:
- The main fix is to use `Closure::once_into_js` for closures that can are called only once
  - This primitive will deallocate the memory of the closure after the call happens (eg. on_close, on_error, on_open callbacks are called once (https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#events)
  - The on_close and on_open are evident, however for the on_error the documentation implies it is called only once
  > [error](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event)
    Fired when a connection with a WebSocket has been closed because of an error, such as when some data couldn't be sent. Also available via the onerror property.

- poll_read verifies if the connected is not opened yet, then it saves the waker and returns `Poll::Pending`
- poll_close now closes the socket
- Socket is closed from `Drop` only if its internal state is open

Closes: https://github.com/paritytech/subxt/issues/1288

### DIFF
--- a/lightclient/src/platform/wasm_socket.rs
+++ b/lightclient/src/platform/wasm_socket.rs
@@ -63,10 +63,10 @@ struct InnerWasmSocket {
 ///
 /// These need to be kept around until the socket is dropped.
 type Callbacks = (
-    JsValue,
+    Closure<dyn FnMut()>,
     Closure<dyn FnMut(web_sys::MessageEvent)>,
-    JsValue,
-    JsValue,
+    Closure<dyn FnMut(web_sys::Event)>,
+    Closure<dyn FnMut(web_sys::CloseEvent)>,
 );
 
 impl WasmSocket {
@@ -89,7 +89,7 @@ impl WasmSocket {
             waker: None,
         }));
 
-        let open_callback = Closure::once_into_js({
+        let open_callback = Closure::<dyn FnMut()>::new({
             let inner = inner.clone();
             move || {
                 let mut inner = inner.lock().expect("Mutex is poised; qed");
@@ -120,7 +120,7 @@ impl WasmSocket {
         });
         socket.set_onmessage(Some(message_callback.as_ref().unchecked_ref()));
 
-        let error_callback = Closure::once_into_js({
+        let error_callback = Closure::<dyn FnMut(_)>::new({
             let inner = inner.clone();
             move |_event: web_sys::Event| {
                 // Callback does not provide useful information, signal it back to the stream.
@@ -134,7 +134,7 @@ impl WasmSocket {
         });
         socket.set_onerror(Some(error_callback.as_ref().unchecked_ref()));
 
-        let close_callback = Closure::once_into_js({
+        let close_callback = Closure::<dyn FnMut(_)>::new({
             let inner = inner.clone();
             move |_event: web_sys::CloseEvent| {
                 let mut inner = inner.lock().expect("Mutex is poised; qed");
@@ -242,8 +242,15 @@ impl AsyncWrite for WasmSocket {
 
 impl Drop for WasmSocket {
     fn drop(&mut self) {
-        if self.socket.ready_state() == web_sys::WebSocket::OPEN {
+        if self.socket.ready_state() == web_sys::WebSocket::OPEN
+            || self.socket.ready_state() == web_sys::WebSocket::CONNECTING
+        {
             let _ = self.socket.close();
         }
+
+        self.socket.set_onopen(None);
+        self.socket.set_onmessage(None);
+        self.socket.set_onerror(None);
+        self.socket.set_onclose(None);
     }
 }
--- a/lightclient/src/platform/wasm_socket.rs
+++ b/lightclient/src/platform/wasm_socket.rs
@@ -171,7 +171,7 @@ impl AsyncRead for WasmSocket {
         let mut inner = self.inner.lock().expect("Mutex is poised; qed");
         inner.waker = Some(cx.waker().clone());
 
-        if self.socket.ready_state() == web_sys::WebSocket::Connecting {
+        if self.socket.ready_state() == web_sys::WebSocket::CONNECTING {
             return Poll::Pending;
         }
 
@@ -225,8 +225,18 @@ impl AsyncWrite for WasmSocket {
         Poll::Ready(Ok(()))
     }
 
-    fn poll_close(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
-        Poll::Ready(Ok(()))
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
+        if self.socket.ready_state() == web_sys::WebSocket::CLOSED {
+            return Poll::Ready(Ok(()));
+        }
+
+        if self.socket.ready_state() != web_sys::WebSocket::CLOSING {
+            let _ = self.socket.close();
+        }
+
+        let mut inner = self.inner.lock().expect("Mutex is poised; qed");
+        inner.waker = Some(cx.waker().clone());
+        Poll::Pending
     }
 }
 
--- a/lightclient/src/platform/wasm_socket.rs
+++ b/lightclient/src/platform/wasm_socket.rs
@@ -230,7 +230,7 @@ impl Drop for WasmSocket {
     fn drop(&mut self) {
         let inner = self.inner.lock().expect("Mutex is poised; qed");
 
-        if inner.state == ConnectionState::Opened {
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):
LGTM :)

tadeohepperle (approved):
Looks good, very nice!

lightclient/src/platform/wasm_socket.rs
```diff
@@ -134,9 +134,9 @@ impl WasmSocket {
         });
         socket.set_onerror(Some(error_callback.as_ref().unchecked_ref()));
 
-        let close_callback = Closure::<dyn FnMut(_)>::new({
+        let close_callback = Closure::once_into_js({
```
jsdw, line 137: The docs for `once_into_js` say:

```
Convert a FnOnce(A...) -> R into a JavaScript Function object.

If the JavaScript function is invoked more than once, it will throw an exception.

Unlike Closure::once, this does not return a Closure that can be dropped before the function is invoked to deallocate the closure. The only way the FnOnce is deallocated is by calling the JavaScript function. If the JavaScript function is never called then the FnOnce and everything it closes over will leak.
```

So, since some of these callbacks might never fire, I think we'd be gradually leaking them here for any connection we make that doesn't fire one of the callbacks.

I wonder if the errors we got happened because previously, if `Self` is dropped before one of the JS websocket callbacks is fired, then it would have decallocated the closures and we'd get an error in JS land instead.

But anyway, from reading [the spec](https://websockets.spec.whatwg.org/#interface-definition), it sounds like we:
- May never fire `onopen` if connection can't be opened (but will fire it max once)
- May never fire `onerror` or `onmessage` if no errors/messages sent (not certain if `onerror` will only fire once or not)
- Will always eventualy fire `onclose` (max once).

So if I read right, I wonder whether we can do some cleanup in `onclose` closure to ensure that all of the other closures are dropped. Then, we can otherwise just leak them all so that nothing can lead them to being dropped otherwise.

lightclient/src/platform/wasm_socket.rs
```diff
@@ -171,6 +171,10 @@ impl AsyncRead for WasmSocket {
         let mut inner = self.inner.lock().expect("Mutex is poised; qed");
         inner.waker = Some(cx.waker().clone());
 
+        if self.socket.ready_state() == web_sys::WebSocket::CONNECTING {
+            return Poll::Pending;
```
jsdw, line 175: So I think this is ook because any of the callbacks will wake the waker when JS clals them, ensuring we call this `poll_read` again.

lightclient/src/platform/wasm_socket.rs
```diff
@@ -221,16 +225,24 @@ impl AsyncWrite for WasmSocket {
...
+        Poll::Pending
     }
 }
 
 impl Drop for WasmSocket {
     fn drop(&mut self) {
-        let inner = self.inner.lock().expect("Mutex is poised; qed");
-
-        if inner.state == ConnectionState::Opened {
+        if self.socket.ready_state() == web_sys::WebSocket::OPEN {
```
jsdw, line 245: Would it be safer to say `if ready_state != CLOSING then close()`? Or maybe just call `close()` regardless to ensure our `onclose` handler always fires and can tidy up (I think nothing happens if you try closing when it's already closing anyway but not sure!)

lightclient/src/platform/wasm_socket.rs
lexnv: That makes sense!

I've found another way, which I believe to be more straight forward than performing clean-up on the `onclose`:
- keep the `Closure::new()` around for each callback
- while dropping the object:
  - close the socket if in: OPEN or CONNECTING states
  - set closures on the web_sys::WebSocket as None

This shall clean-up all memory associated with the rust closures, and deny the javascript real access to the dropped objects

lightclient/src/platform/wasm_socket.rs
lexnv: I've extended this to do clean-up for `OPEN` and `CONNECTING`, since we might generate console errors if we attempt to close a socket that's already closed 👍

lexnv: Attempting to close the socket regardless of its state will generate on `--chrome`:

```bash
WebSocket connection to 'ws://polkadot-boot.dwellir.com:30334/' failed: WebSocket is closed before the connection is established.
```

The same error is generated if the socket is in the `CONNECTING` state, note that `--firefox` is not printing this warning.

lexnv:
I've changed the following to this PR:

- Closures are created with `new` and the socket callbacks are set to `None` in the Drop to avoid recursive calls and leaking closures (prior done with `Closure::once_into_js` that might never be invoked)
- Closing the socket is performed on `drop` if the state is not already CLOSING
  - this generates an warning in the Chrome browser (https://github.com/paritytech/subxt/pull/1289#discussion_r1406152272); in balance with the alternative of possibly leaking the socket I believe this ok
  - Firefox does not present the warning, which might be specific to Chrome since the documentation states that `close` does nothing in the unexpected states: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close
  
  
The bootnode that was causing issues seems to be online now, which makes testing a bit more difficult.

Alternatives for testing:
- For manual testing we could populate the bootnodes with invalid ports / DNS; after this PR is applied the `Uncaught errors` are resolved:
```bash
        .bootnodes([
            "/dns/polkadot-boot.dwellir.com/tcp/1/ws/p2p/12D3KooWKvdDyRKqUfSAaUCbYiLwKY8uK3wDWpCuy2FiDLbkPTDJ",
            "/dns/polkadot-boot.dwellir.com/tcp/1/wss/p2p/12D3KooWKvdDyRKqUfSAaUCbYiLwKY8uK3wDWpCuy2FiDLbkPTDJ",
            "/dns/dot14.rotko.net/tcp/1/p2p/12D3KooWPyEvPEXghnMC67Gff6PuZiSvfx3fmziKiPZcGStZ5xff",
        ])
```
- For automated testing; I did not find a way to capture the browser's console programmatically in rust
   - I have tried creating hooks to panics and errors, although they seem to only capture the WASM module / errors coming from Rust
 ```rust
     let errors = Box::leak(Box::new(Arc::new(Mutex::new(Vec::with_capacity(16)))));
    std::panic::set_hook(Box::new(|info: &std::panic::PanicInfo| {
        errors
            .lock()
            .expect("Single threaded; qed")
            .push(info.to_string());
    }));
```
  - We might need a different system for testing this to ensure that we report back zero errors or that `!error.contains("closure invoked recursively")`. Considering the complexity of doing that, we could takle it at a later time

niklasad1:
> - Closing the socket is performed on `drop` if the state is not already CLOSING
>  - this generates an warning in the Chrome browser (https://github.com/paritytech/subxt/pull/1289#discussion_r1406152272); in balance with the alternative of possibly leaking the socket I believe this ok
>  - Firefox does not present the warning, which might be specific to Chrome since the documentation states that `close` does nothing in the unexpected states: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close

Fine be me

> For manual testing we could populate the bootnodes with invalid ports / DNS; after this PR is applied the Uncaught errors are resolved:

Can we add a GHA cronjob that queries the bootnodes once per day to detect  that?

> For automated testing; I did not find a way to capture the browser's console programmatically in rust

> I have tried creating hooks to panics and errors, although they seem to only capture the WASM module / errors coming from Rust

I guess we could to setup a CI that runs these in chrome and firefox seems a bit overkill for now.

niklasad1:
An alternative is to wrap https://docs.rs/gloo-net/latest/gloo_net/websocket/futures/struct.WebSocket.html and implement AsyncRead and AsyncWrite for SplitStream (stream, sink).

The implementations looks very similar without the option stuff that I don't really understand, is it because you don't want to box the closures and leak them?

niklasad1:
> Do you mean the option that we set for populating the callbacks?

I was wrong I missed that that part...

PR #1290, opened 2023-11-24
CLI Tool Explore Command: Runtime APIs, Events, Colorized outputs, scale-typegen integration

tadeohepperle:
**Warning: Do not merge** (we need to wait for a full scale-typgen release before we can integrate this PR)

This PR restructures the `subxt explore` CLI command. You can now explore pallet events: `subxt explore pallet balances events`:
image
And you can explore runtime apis: `subxt explore api metadata` by adding the `--execute` (`-e`) flag to a runtime api method, e.g. `subxt explore api core version -e` you can view the return-value of the runtime-api call: 
image

Upon starting the explore command (`subxt explore`) you get shown a list of all available pallets and runtime apis. When You only type `subxt explore pallet` you only see available pallets. For outputs I use ANSI coloring which is already part of the [color_eyre](https://crates.io/crates/color-eyre) crate, so no new dependencies needed to be added. I added a dependency though for [indoc](https://docs.rs/indoc/latest/indoc/), which makes the code a lot cleaner.

### Testing
Manually tested, and tests have been adjusted.


niklasad1:
@tadeohepperle Can you merge/rebase this PR to master to make it easier to review after that you merged the scale-typegen PR?

cli/src/commands/explore/mod.rs
```diff
@@ -1,159 +1,223 @@
...
+/// Show the type of a specific event
+/// ```text
+/// subxt explore pallet Balances events frozen
+/// ```
+/// ## Runtime APIs
+/// Show the input and output types of a runtime api:
+/// ```text
+/// subxt explore api core version
+/// ```
+/// Execute a runtime API call with the `--execute` (`-e`) flag, to see the return value:
```
niklasad1, line 79: Is it possible to get the params in the runtime api call as well?

cli/src/commands/explore/mod.rs
```diff
@@ -1,159 +1,223 @@
...
-    let Some(pallet_name) = opts.pallet else {
-        let available_pallets = print_available_pallets(&metadata);
-        writeln!(output, "Usage:",)?;
-        writeln!(output, "    subxt explore <PALLET>",)?;
-        writeln!(output, "        explore a specific pallet",)?;
-        writeln!(output, "\n{available_pallets}",)?;
-        return Ok(());
-    };
+    #[allow(non_snake_case)]
+    let PALLET: String = "<PALLET>".blue().to_string();
```
niklasad1, line 127: I don't get why you use capital letters on PALLET and RUNTIME_API, can you explain?

cli/src/commands/explore/mod.rs
```diff
@@ -1,159 +1,223 @@
...
-            // if the metadata came from some url, we use that same url to make storage calls against.
-            let node_url = opts.file_or_url.url.map(|url| url.to_string());
-            explore_storage(command, &metadata, pallet_metadata, node_url, output).await
+        PalletOrRuntimeApi::Api(opts) => {
+            let Some(name) = opts.name else {
+                let runtime_apis = runtime_apis_as_string(&metadata);
+                writedoc! {output, "
+                Usage:
+                    subxt explore api {RUNTIME_API}
+                        explore a specific runtime api trait
```
niklasad1, line 182:
```suggestion
                        explore a specific runtime api
```

cli/src/commands/explore/mod.rs
```diff
@@ -187,44 +287,139 @@ pub mod tests {
...
-        assert!(output.unwrap().starts_with("Usage:\n    subxt explore Balances calls <CALL>\n        explore a specific call within this pallet\n\nAvailable <CALL>'s in the \"Balances\" pallet:\n"));
-        let output = run_against_file("Balances storage").await;
-        assert!(output.unwrap().starts_with("Usage:\n    subxt explore Balances storage <STORAGE_ENTRY>\n        view details for a specific storage entry\n\nAvailable <STORAGE_ENTRY>'s in the \"Balances\" pallet:\n"));
-        let output = run_against_file("Balances constants").await;
-        assert!(output.unwrap().starts_with("Usage:\n    subxt explore Balances constants <CONSTANT>\n        explore a specific call within this pallet\n\nAvailable <CONSTANT>'s in the \"Balances\" pallet:\n"));
+        let output = run_against_file("pallet Balances calls")
+            .await
+            .unwrap()
+            .strip_ansi();
+        // dbg!(output);
```
niklasad1, line 350:
```suggestion
```

Cargo.toml
```diff
@@ -89,6 +83,7 @@ scale-decode = "0.10.0"
 scale-encode = "0.5.0"
 serde = { version = "1.0.195" }
 serde_json = { version = "1.0.108" }
+strip-ansi-escapes = "0.2.0"
```
niklasad1, line 86: This dependency looks unused to me and is only used for tests in the CLI crate

cli/src/commands/explore/mod.rs
tadeohepperle: I use these strings like constants, but the `blue()` function is not const. So I use these uppercase variable names to not confuse it with something else.

I wish Rust would automatically have functions const by default, much like a type is Send as long as it does not have any non Send fields. Maybe they can put all of const evaluation behind a `Const` auto-trait in the future that all functions implement except the ones doing non-const stuff.

cli/src/commands/explore/mod.rs
```diff
@@ -1,159 +1,223 @@
...
+/// subxt explore pallet Balances events
+/// ```
+/// Show the type of a specific event
+/// ```text
+/// subxt explore pallet Balances events frozen
+/// ```
+/// ## Runtime APIs
+/// Show the input and output types of a runtime api:
+/// ```text
+/// subxt explore api core version
```
niklasad1, line 77: Not sure whether we should explain that `core is the name of runtime API` and `version` is a method on it.

cli/src/commands/explore/mod.rs
```diff
@@ -187,44 +287,139 @@ pub mod tests {
...
+        "};
+        assert_eq_start!(output, start);
+        // check that we can see methods of a runtime api:
+        let output = run_against_file("api metadata").await.unwrap().strip_ansi();
+
+        let start = formatdoc! {"
+            Description:
+                The `Metadata` api trait that returns metadata for the runtime.
+            
+            Usage:
```
niklasad1, line 416: ok, good answers one of my questions above

cli/src/commands/explore/pallets/calls.rs
```diff
@@ -61,38 +73,53 @@ pub fn explore_calls(
...
+    writedoc! {output, "
+    Encoded call data:
+        {hex_bytes}
+    "}?;
+
     Ok(())
 }
 
-fn print_available_calls(pallet_calls: &TypeDefVariant<PortableForm>, pallet_name: &str) -> String {
+fn available_calls_string(
```
niklasad1, line 119: hmm, maybe `pallet_calls_to_string`

cli/src/commands/explore/pallets/constants.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let Some(constant_name) = command.constant else {
+        writeln!(output, "{}", usage())?;
+        return Ok(());
+    };
+
+    // if specified constant is wrong, show user the constants to choose from (but this time as an error):
+    let Some(constant) = pallet_metadata
+        .constants()
+        .find(|constant| constant.name().to_lowercase() == constant_name.to_lowercase())
```
niklasad1, line 41: https://doc.rust-lang.org/std/string/struct.String.html#method.eq_ignore_ascii_case

cli/src/commands/explore/mod.rs
tadeohepperle: Yes, I called the params `INPUT_VALUE`. I should probably add an example here in the docs, do you know an API where the params are not so complicated (maybe just a number)?

image

cli/src/commands/explore/pallets/events.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let Some(event_name) = command.event else {
+        writeln!(output, "{}", usage())?;
+        return Ok(());
+    };
+
+    // if specified event is wrong, show user the events to choose from (but this time as an error):
+    let Some(event) = event_variants
+        .iter()
+        .find(|event| event.name.to_lowercase() == event_name.to_lowercase())
```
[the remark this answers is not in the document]
tadeohepperle: Oh no, I should have thought about this earlier...

cli/src/commands/explore/mod.rs
niklasad1: cool, INPUT value sounds good to me.
maybe "StakingApi::nominations_quota" it just takes "Balance" as input.

I wonder how useful it's with the input values printed here, I would be happy just with the type as you printed out:

```rust
EraStakersPageCount {
   era: u32,
   account: AccountId32
}
```

niklasad1 (commented):
Nice, I like it.

Not sure about `-e` though in the CLI but it's really hard to format the inputs if there any in the runtime API call.

For subxt explorer the `-e` would be really neat feature to have but perhaps you already have it in there? I haven't checked it lately :(

tadeohepperle:
> For subxt explorer the `-e` would be really neat feature to have but perhaps you already have it in there? I haven't checked it lately :(

Good idea, I added an [issue for that in subxt explorer](https://github.com/paritytech/subxt-explorer/issues/13) now, we currently only have that for storage entries. Maybe for the CLI we remove the `-e` flag and just execute always if no params are required, otherwise we don't provide the execution option? But that would not be good for api calls where execution would result in thousands of bytes being spit out (e.g. metadata api).

niklasad1:
> Maybe for the CLI we remove the -e flag and just execute always if no params are required, otherwise we don't provide the execution option? But that would not be good for api calls where execution would result in thousands of bytes being spit out (e.g. metadata api).

I think we should keep the `-e` flag anyway even if it's difficult to use with "input params" but lets hear what Alex and James thinks :)

tadeohepperle:
@jsdw That formatting error is weird, on my machine it went fine in the VSCode and in the native Terminal (Manjaro Linux). I will look into it though.

image

image

tadeohepperle:
I finally figured out how to submit the parameters needed for a dynamic runtime api call in the CLI: 
- take the input as a `&str`
- parse it into a scale value
- encode it to bytes using the params type id
- decode it from the bytes using the params type id

IIRC @niklasad1 was doing something similar a few months back somewhere.

Now we can execute runtime apis like this:

image

niklasad1 (approved):

cli/Cargo.toml
```diff
@@ -40,5 +40,12 @@ scale-value = { workspace = true }
 syn = { workspace = true }
 jsonrpsee = { workspace = true, features = ["async-client", "client-ws-transport-native-tls", "http-client"] }
 tokio = { workspace = true, features = ["rt-multi-thread"] }
+scale-typegen-description = { workspace = true }
+heck = { workspace = true }
+indoc = "2"
```
lexnv, line 45: We could probably move the `indoc` adn `strip-ansi` to the workspace Cargo.toml to make sure we can easily update it everywhere. Just in case we'll add it to multipl crates in the future. Leaving this up to you

tadeohepperle: Good idea, but I would prefer to do it later when we really need it another crate of the workspace.

jsdw: Let's do it now for consistency, because I think every other dependency is now listed in the workspace thing first (except for non-workspace crates where we can't). Makes it mostly easy to see the entire set of crates we use in one place at least!

cli/src/commands/explore/pallets/events.rs
```diff
@@ -0,0 +1,88 @@
...
+    pallet_metadata: PalletMetadata,
+    metadata: &Metadata,
+    output: &mut impl std::io::Write,
+) -> color_eyre::Result<()> {
+    let pallet_name = pallet_metadata.name();
+    let event_variants = pallet_metadata.event_variants().unwrap_or(&[]);
+
+    let usage = || {
+        let events = events_to_string(event_variants, pallet_name);
+        formatdoc! {"
```
jsdw, line 25: This does make it easier to see how things will be formatted; good find!

cli/src/commands/explore/pallets/storage.rs
```diff
@@ -0,0 +1,214 @@
...
+};
+
+#[derive(Debug, Clone, Args)]
+pub struct StorageSubcommand {
+    storage_entry: Option<String>,
+    #[clap(required = false)]
+    trailing_args: Vec<String>,
+}
+
+pub async fn explore_storage(
```
jsdw, line 32: With runtime APIs, we have an -e flag to actually try to execute a runtime API call. I feel like we should do the same for executing a storage request, so that by default we don't try to make a request but we can if we want. What do you think?

cli/src/commands/explore/runtime_apis/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+    if !doc_string.is_empty() {
+        writedoc! {output, "
+        Description:
+        {doc_string}
+        
+        "}?;
+    }
+
+    #[allow(non_snake_case)]
+    let INPUT_VALUE: String = "<INPUT_VALUE>".blue().to_string();
```
jsdw, line 97: Personally I'd just name it something like `blue_input_placeholder` or something (likewise elsewhere) and avoid needing to add the allow

cli/src/commands/explore/runtime_apis/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        formatdoc! {"
+        The method expects an {INPUT_VALUE} with this shape:
+        {fields_description}
+    
+        For example you could provide this {INPUT_VALUE}:
+        {fields_example}"}
+    };
+
+    let execute_usage = || {
```
jsdw, line 125: This could just be a block?

cli/src/commands/explore/mod.rs
```diff
@@ -1,159 +1,223 @@
...
+/// subxt explore pallet Grandpa calls note_stalled { "delay": 5, "best_finalized_block_number": 5 }
 /// # Encoded call data:
 /// # 0x2c0411020500000005000000
-/// subxt explore Balances calls transfer  "{ \"dest\": v\"Raw\"((255, 255, 255)), \"value\": 0 }"
+/// subxt explore pallet Balances calls transfer  "{ \"dest\": v\"Raw\"((255, 255, 255)), \"value\": 0 }"
 /// # Encoded call data:
 /// # 0x24040607020cffffff00
 /// ```
-/// ## Constants
+/// ### Constants
```
jsdw, line 50: I'd ensure a newline above and below each heading (this and below are missing one or both), and in general add newlines before and after the code blocks in this whole section (and anywhere else) just to make it more readable as plain text :)

cli/src/commands/explore/mod.rs
jsdw: As I said elsewhere, I'd instead call them eg `blue_input_placeholder` and then no need for the opt-out allow. I don't think it's worth "breaking the rules" for this case :)

jsdw:
Minor comments but I think they should be addressed, and then for me this is good to go! The colouring thing on my laptop is strange but we can sort that out later if it keeps coming up.

cli/src/commands/explore/runtime_apis/mod.rs
tadeohepperle: Oh, good catch, I made it a closure to use it twice below, but used it only once. I am now replacing the `usage()` below by the second `execute_usage()` as intended.

cli/src/commands/explore/pallets/storage.rs
tadeohepperle: Okay, I added an `-e` flag for storage entries as well and made the API very similar to the runtime apis.

jsdw (approved):
Awesome stuff, nicely done :)

PR #1294, opened 2023-11-28
request allow to use subxt scale encoded types in wasm32 std not in browser

dzmitry-lahoda:
I tried to compile wasm32 std crate which is not browser with subxt generated client.

if i set native subxt  feature, get js error.

if set web, too many non existing host functions requested (browser specific).

setting none is not option.

so request is option to have None, and generate only SUBX type to create TXes, in wasm32 std env.

please note cannot reference runtimes directly for various reason, on is that SP fails to compile wasm32 std with js error.

why?

using in CosmWasm contracts to form Parachain Transactions. Also I guess in pallets.

Currently people manually cut parts of subxt and paste them - bad maintainance.


dzmitry-lahoda:
Use case is CosmWasm contract in Piccasso to call any pallet, for this it to encode call for this need subxt client to compile in wasm32 std.

tadeohepperle:
That is interesting, if I understand you correctly, you want to compile subxt for the `wasm32-unknown-unknown` target but doing so with the `web` feature flag enabled is not an option for you because this would lead to code that assumes it is running in a browser, having access to Browser APIs that are not applicable in your case. These likely stem from the `web` implementations in the [`jsonrpsee`](https://github.com/paritytech/jsonrpsee) crate. We cannot build subxt clients without `jsonrpsee` currently.

What functionality do you want to get out of subxt for your usecase? Would it be an option for you to just use the `subxt-codegen` crate? We are also working on a [`scale-typegen`](https://github.com/paritytech/scale-typegen) crate that can be used without subxt if you only want to generate runtime types. It is more lightweight and compiles to wasm fine.

dzmitry-lahoda:
I see. 

I used subxt cli, which uses codegen and it still generates networking code as it seems. So may be need more parameters?

New crate is good, just need some doc/examplehl how to clean compile for wasm32 std.

yes, thank you

dzmitry-lahoda:
true, i just need to encode tx/extrinsic. I will send myself. target is wasm std
 std. std, but without web sys exports and imports embeeded into wasm file, as they require from host web exports.
 CosmWasm has its own request and reponse messages, basically i will encode tx and send it as msg to parachain host of cosmwasm contract.

I am ok do not have any backend or rcpclient abstraction, so contract to parachain host is such.

dzmitry-lahoda:
I have sent pr as in the end check of wasm to be ci check.

PR #1298, opened 2023-11-30
storage: Fix partial key storage iteration 

### LINKED ISSUE #1251: Iterating over StorageDoubleMap with partial keys not working
Problem :
Trying to fetch values from a StorageDoubleMap by supplying only a partial key, the result still returns all values in storage

### Example :

Code : 
```
pub async fn calculate_pool_data(pool_id : u32, era: u32) -> Result<(), Box<dyn std::error::Error>> {
	let relay_client = RelayClient::from_url("wss://rpc.relay.blockchain.enjin.io:443").await?;

	// get unbonding members
	let unbonding_members = {
		let query = relay::storage().nomination_pools().unbonding_members_iter1(6u32);
		let mut results = relay_client.storage().at_latest().await?.iter(query).await?;

		let mut unbonding_members = HashMap::new();
		while let Some(Ok((key, prefs))) = results.next().await {
			unbonding_members.insert(key, prefs);
		}
		unbonding_members
	};

	println!("Unbonding members {:?}", unbonding_members.len());

	Ok(())

}
```

Output : 
```
Unbonding members 109
```

Expected Output :
1 (there is only one unbonding member)

Screenshot 2023-11-10 at 9 03 51 AM

lexnv:
This PR fixes the partial key storage iteration.

The storage key constructed by partial inputs was converted to the root storage key entry.
This meant that any partial input parameters provided to the key were completely ignored; and resulted in a wrongful behavior of iterating over all storage entries, rather than just the ones starting with the partial key.

To mitigate this behavior, the address bytes take into account the provided parameter inputs.
This loosens the expectations of `utils::storage_address_bytes`, which now only returns errors if the provided input types exceed the number of types from the metadata. Before this PR, the function expected an identical match in terms of the number of inputs.

### Testing Done

- Manual testing with the metadata from "wss://rpc.relay.blockchain.enjin.io:443"
- Added an unit test that creates two assets, and checks that partial iterations result in the expected number of items

Closes: https://github.com/paritytech/subxt/issues/1251

### DIFF
--- a/subxt/src/storage/storage_address.rs
+++ b/subxt/src/storage/storage_address.rs
@@ -172,6 +172,12 @@ where
                     .resolve(*key_ty)
                     .ok_or(MetadataError::TypeNotFound(*key_ty))?;
 
+                // If the provided keys are empty, the storage address must be
+                // equal to the storage root address.
+                if self.storage_entry_keys.is_empty() {
+                    return Ok(());
+                }
+
                 // If the key is a tuple, we encode each value to the corresponding tuple type.
                 // If the key is not a tuple, encode a single value to the key type.
                 let type_ids = match &ty.type_def {
--- a/subxt/src/storage/storage_address.rs
+++ b/subxt/src/storage/storage_address.rs
@@ -190,7 +190,7 @@ where
                     }
                     hash_bytes(&input, &hashers[0], bytes);
                     Ok(())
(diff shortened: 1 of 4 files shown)


tadeohepperle (approved):
Ah nice, great work! :)

niklasad1 (approved):

PR #1299, opened 2023-12-01
cargo: Switch to workspace lints

lexnv:
In this PR, the recommended rust best practices are followed (since rust 1.74.0):
- declares the lints at the workspace level.
- remove a couple of lints that are now declared as unstable.

- the optimization profile of the light-client is moved at the root level


niklasad1 (commented):
Can you bump the rust-version to v1.74.0 in the workspace Cargo.toml as well? 

Otherwise, it looks good to me and nice to re-use the lints between the crates in subxt.

Cargo.toml
```diff
@@ -36,6 +36,29 @@ repository = "https://github.com/paritytech/subxt"
...
+path_statements = "deny"
+patterns_in_fns_without_body = "deny"
+unconditional_recursion = "deny"
+unused_allocation = "deny"
+unused_comparisons = "deny"
+unused_parens = "deny"
+while_true = "deny"
+trivial_casts = "deny"
+trivial_numeric_casts = "deny"
+unused_crate_dependencies = "deny"
```
niklasad1, line 55: I suggest that we remove these unused crate lints and use `cargo machete` instead

lexnv: I like this a lot! Already discovered few unused crates, with the tradeoff of some false-positive subxt dependency used by our build.rs for the testing crate! Have also added a CI step for it!

niklasad1 (approved):
nice :muscle:

tadeohepperle (approved):
This is nice!

PR #1306, opened 2023-12-04
Bugfix: trimmed metadata hash comparison fails in `is_codegen_valid_for`

### LINKED ISSUE #1286: codegen: is_codegen_valid_for doesn't work as expected for "trimmed metadata"

tadeohepperle:
Fixes #1286

## What is the bug:

Metadata can be trimmed with `Metadata::retain(..)`. This is done by the CLI tool, e.g. `subxt metadata --url wss://rpc.polkadot.io --pallets "ElectionProviderMultiPhase,System"` retains only the specified pallets.
When comparing the hash of trimmed metadata (metadata that contains only a limited number of pallets/runtime apis) with the hash of the original metadata using the hashers `only_these_pallets` / `only_these_runtime_apis`, the hashes were different. This comparison was made in the generated `is_codegen_valid_for` function.

## Reason for the bug

in `Metadata::retain(..)` 3 things happen:
- the unneeded pallets/apis are removed at the top level
- all types that only these pallets/apis used are removed as well
- the top level enums (call, event and error) are Variant types. The variants relating to unneeded pallets/apis are removed. This happens in `retain_pallets_in_runtime_outer_types()`.

The last point is what caused the bug:
When specifying `only_these_pallets` for a MetadataHasher, the variant types of the top level enums (call, event and error) were modified to discard unneded pallets before the hash is calculated. HOWEVER the same types can be included in other places of the metadata. In @niklasad1 case for the staking-miner, the `"Events"` storage entry of the `"System"` pallet, had a type of `Vec<Event>`, where `Event` is the event top level type. But we did not go through all types of the metadata checking for each one if they are a top level variant and modifying it accordingly before computing its hash.
==> This caused the bug. 

## Solution

To solve the bug, we need to change how `specific_pallets` is used in the `MetadataHasher::hash(..)` function. 
We compute the hashes of the 3 outer enums first, trimming them down to only the variants that refer to the pallets in  `specific_pallets`. In the rest of the function we often compute the hash of some type. Every time this happens, we pass the type ids and hashes of the 3 outer enums down in these recursive hash computations. Every time an outer enum type is encountered we just use the hash computed at the start.

Other change: I severly limited all caching (used to prevent infinite recursion). Now a new cache is created every time we resolve the hash of some type. -> Less Cache reuse than before -> Worse Performance but better consistency.

### Alternative solution (not implemented)

We could also just clone and trim the metadata when calling `MetadataHasher::hash()` and specific pallets/apis are set.
This is probably less performant, but maybe in the end easier and more consistent.

## Testing

Added a test to confirm this works and manually tested it for the example Niklas provided.

### DIFF
--- a/metadata/src/utils/validation.rs
+++ b/metadata/src/utils/validation.rs
@@ -5,15 +5,15 @@
 //! Utility functions for metadata validation.
 
 use crate::{
-    CustomMetadata, CustomValueMetadata, ExtrinsicMetadata, Metadata, OuterEnumsMetadata,
-    PalletMetadata, RuntimeApiMetadata, RuntimeApiMethodMetadata, StorageEntryMetadata,
-    StorageEntryType,
+    CustomMetadata, CustomValueMetadata, ExtrinsicMetadata, Metadata, PalletMetadata,
+    RuntimeApiMetadata, RuntimeApiMethodMetadata, StorageEntryMetadata, StorageEntryType,
 };
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, TypeDefVariant, Variant};
 use std::collections::HashMap;
 
 // The number of bytes our `hash` function produces.
 pub(crate) const HASH_LEN: usize = 32;
+pub type Hash = [u8; HASH_LEN];
 
 /// Internal byte representation for various metadata types utilized for
 /// generating deterministic hashes between different rust versions.
@@ -30,13 +30,13 @@ enum TypeBeingHashed {
 }
 
 /// Hashing function utilized internally.
-fn hash(data: &[u8]) -> [u8; HASH_LEN] {
+fn hash(data: &[u8]) -> Hash {
     sp_core_hashing::twox_256(data)
 }
 
 /// XOR two hashes together. Only use this when you don't care about the order
 /// of the things you're hashing together.
-fn xor(a: [u8; HASH_LEN], b: [u8; HASH_LEN]) -> [u8; HASH_LEN] {
+fn xor(a: Hash, b: Hash) -> Hash {
     let mut out = [0u8; HASH_LEN];
     for (idx, (a, b)) in a.into_iter().zip(b).enumerate() {
         out[idx] = a ^ b;
@@ -52,7 +52,7 @@ macro_rules! count_idents {
 }
 macro_rules! concat_and_hash_n {
     ($name:ident($($arg:ident)+)) => {
-        fn $name($($arg: &[u8; HASH_LEN]),+) -> [u8; HASH_LEN] {
+        fn $name($($arg: &Hash),+) -> Hash {
             let mut out = [0u8; HASH_LEN * count_idents!($($arg)+)];
             let mut start = 0;
             $(
@@ -75,8 +75,8 @@ fn get_field_hash(
     registry: &PortableRegistry,
     field: &Field<PortableForm>,
     cache: &mut HashMap<u32, CachedHash>,
-    precomputed: Option<&HashMap<u32, [u8; HASH_LEN]>>,
-) -> [u8; HASH_LEN] {
+    outer_enum_hashes: &OuterEnumHashes,
+) -> Hash {
     let field_name_bytes = match &field.name {
         Some(name) => hash(name.as_bytes()),
         None => [0u8; HASH_LEN],
@@ -84,7 +84,7 @@ fn get_field_hash(
 
     concat_and_hash2(
         &field_name_bytes,
-        &get_type_hash_recurse(registry, field.ty.id, cache, precomputed),
+        &get_type_hash_recurse(registry, field.ty.id, cache, outer_enum_hashes),
     )
 }
 
@@ -93,13 +93,16 @@ fn get_variant_hash(
     registry: &PortableRegistry,
     var: &Variant<PortableForm>,
     cache: &mut HashMap<u32, CachedHash>,
-    precomputed: Option<&HashMap<u32, [u8; HASH_LEN]>>,
-) -> [u8; HASH_LEN] {
+    outer_enum_hashes: &OuterEnumHashes,
+) -> Hash {
     let variant_name_bytes = hash(var.name.as_bytes());
     let variant_field_bytes = var.fields.iter().fold([0u8; HASH_LEN], |bytes, field| {
         // EncodeAsType and DecodeAsType don't care about variant field ordering,
         // so XOR the fields to ensure that it doesn't matter.
-        xor(bytes, get_field_hash(registry, field, cache, precomputed))
+        xor(
+            bytes,
+            get_field_hash(registry, field, cache, outer_enum_hashes),
+        )
     });
 
     concat_and_hash2(&variant_name_bytes, &variant_field_bytes)
@@ -110,8 +113,8 @@ fn get_type_def_variant_hash(
     variant: &TypeDefVariant<PortableForm>,
     only_these_variants: Option<&[&str]>,
     cache: &mut HashMap<u32, CachedHash>,
-    precomputed: Option<&HashMap<u32, [u8; HASH_LEN]>>,
-) -> [u8; HASH_LEN] {
+    outer_enum_hashes: &OuterEnumHashes,
+) -> Hash {
     let variant_id_bytes = [TypeBeingHashed::Variant as u8; HASH_LEN];
     let variant_field_bytes = variant.variants.iter().fold([0u8; HASH_LEN], |bytes, var| {
         // With EncodeAsType and DecodeAsType we no longer care which order the variants are in,
@@ -121,7 +124,10 @@ fn get_type_def_variant_hash(
             .map(|only_these_variants| only_these_variants.contains(&var.name.as_str()))
             .unwrap_or(true);
         if should_hash {
-            xor(bytes, get_variant_hash(registry, var, cache, precomputed))
+            xor(
+                bytes,
+                get_variant_hash(registry, var, cache, outer_enum_hashes),
+            )
         } else {
             bytes
         }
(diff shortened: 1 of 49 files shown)


metadata/src/utils/validation.rs
```diff
@@ -75,6 +75,7 @@ fn get_field_hash(
     registry: &PortableRegistry,
     field: &Field<PortableForm>,
     cache: &mut HashMap<u32, CachedHash>,
+    precomputed: Option<&HashMap<u32, [u8; HASH_LEN]>>,
```
niklasad1, line 78: I find it a bit hard to distinguish between "cache and precomputed" here, to me those seems very similar and wonder if it makes sense to merge them to single type instead?

It looks like `cache` is an empty hashmap in plenty of places and precomputed is an Option<HashMap>... would be nice unify that

metadata/src/utils/validation.rs
```diff
@@ -262,20 +295,29 @@ fn get_extrinsic_hash(
...
+                outer_enums_hashes,
+            ),
         )
     }
 
     bytes
 }
 
 /// Obtain the hash representation of the `frame_metadata::v15::OuterEnums`.
+/// Also returns a HashMap containing the individual hashes of the 3 otuer enums.
```
niklasad1, line 315:
```suggestion
/// Also returns a HashMap containing the individual hashes of the 3 outer enums.
```

metadata/src/utils/validation.rs
```diff
@@ -262,20 +295,29 @@ fn get_extrinsic_hash(
...
+                signed_extension.additional_ty,
+                outer_enums_hashes,
+            ),
         )
     }
 
     bytes
 }
 
 /// Obtain the hash representation of the `frame_metadata::v15::OuterEnums`.
```
niklasad1, line 314: I would prefer:

```
/// Obtain the hash representation of the `frame_metadata::v15::OuterEnums` 
/// and individual hashes of the outer enums.
///
/// The individual hashes may be used to calculate recursive hash computations
///
```

jsdw:
Generally looks good to me, but just a few bits:

Right now we have `get_type_hash`, `get_type_hash_with_precomputed` and `get_type_hash_recurse`. My ideal state would be that there weren't so many ways to get a type hash, but failing that, the short named one should be the default IMO, so I'd rename `get_type_hash` => `get_type_hash_without_precomputed` and ``get_type_hash_with_precomputed` => `get_type_hash`.

As Niklas said above, the presumputed `Option<HashMap<..>>` sitting next to the cache is confusing I think. I wouldn't merge it with the cache (we want to make a cache only once and not provide a way to create a cache such that it'll be reused if possible). What I'd do is make a new type (which also doesn't need to be a hashmap; we could optimise it since there are exactly 3 entries). Maybe something a bit like:

```rust
struct TopLevelEnumHashes(Option<[(u8, [u8; HASH_LEN]); 3]>);

impl TopLevelEnumHashes {
    pub fn new(
        error_hash: (u8, [u8; HASH_LEN]),
        event_hash: (u8, [u8; HASH_LEN]),
        call_hash: (u8, [u8; HASH_LEN])
    ) -> Self { .. }

    pub fn empty() -> Self {
        Self(None)
    }

    pub fn get(&self, idx: u8) -> Option<[u8; HASH_LEN]> {
        self.0.and_then(|hashes| 
            hashes.iter().find(|(i, _hash)| idx == i).map(|(_i, hash)| hash)
        )
    }
}
```

Then it's clear in all of the function signatures that use it that it's different from the cache, and if you had an `empty()` version of it then you could maybe scrap the niche `get_type_hash_without_precomputed` and in those cases just pass empty in to them.

metadata/src/utils/validation.rs
jsdw: I agree that it's confusing, although I wouldn't merge it with the cache (see comment below :))

niklasad1: Also a type-alias for hash would nice to avoid repeating `[u8; HASH_LEN]`

jsdw:
Side note: weird test failure! I hope that doesn't relate to my unstable backend submit_extrinsic change! hopefully will work on a rerun but we should keep a close eye on that!

metadata/src/utils/validation.rs
```diff
@@ -262,20 +295,29 @@ fn get_extrinsic_hash(
...
+/// Also returns a HashMap containing the individual hashes of the 3 otuer enums.
 fn get_outer_enums_hash(
     registry: &PortableRegistry,
     enums: &OuterEnumsMetadata,
     only_these_variants: Option<&[&str]>,
-) -> [u8; HASH_LEN] {
+) -> ([u8; HASH_LEN], HashMap<u32, [u8; HASH_LEN]>) {
```
niklasad1, line 320: A separate type for the the `hash` + individual hashes for the outer enums hashes could maybe make sense.
Because documenting this is a little tricky

metadata/src/utils/validation.rs
```diff
@@ -262,53 +300,95 @@ fn get_extrinsic_hash(
...
-        } else {
-            get_type_hash(registry, id, &mut HashMap::new())
+/// Hash representations of the `frame_metadata::v15::OuterEnums`.
+pub struct OuterEnumHashes {
+    call_hash: (u32, Hash),
+    error_hash: (u32, Hash),
+    event_hash: (u32, Hash),
+}
+
+impl OuterEnumHashes {
```
jsdw, line 318: Small thing; perhaps move this to a separate `outer_enum_hashes.rs` file, just so that the "pub"s are respected :)

Looks great otherwise though!

metadata/src/utils/validation.rs
```diff
@@ -207,11 +224,34 @@ impl CachedHash {
...
+/// Obtain the hash representation of a `scale_info::Type` identified by id.
+fn get_type_hash_recurse(
     registry: &PortableRegistry,
     id: u32,
     cache: &mut HashMap<u32, CachedHash>,
-) -> [u8; HASH_LEN] {
+    outer_enum_hashes: &OuterEnumHashes,
+) -> Hash {
+    // If the type is part of `precomputed` hashes, return the precomputed hash instead:
+
```
niklasad1, line 250:
```suggestion
```

metadata/src/utils/validation.rs
```diff
@@ -207,11 +224,34 @@ impl CachedHash {
...
+/// Obtain the hash representation of a `scale_info::Type` identified by id.
+fn get_type_hash_recurse(
     registry: &PortableRegistry,
     id: u32,
     cache: &mut HashMap<u32, CachedHash>,
-) -> [u8; HASH_LEN] {
+    outer_enum_hashes: &OuterEnumHashes,
+) -> Hash {
+    // If the type is part of `precomputed` hashes, return the precomputed hash instead:
```
niklasad1, line 249: perhaps, tweak this comment now when "precomputed hashes" are called "outer_enum_hashes"?

niklasad1 (approved):
Looks great now

jsdw:
Almost ready to merge I think; couple of small comments from Niklas to check out, move the `outer_enum_hashes` to a new file and confirm whether the test actually failed before or not and it's all good then from me :)

metadata/src/utils/validation.rs
```diff
@@ -1080,4 +1175,144 @@ mod tests {
...
+            m
+        };
+
+        // test that the hashes are the same:
+        let hash = MetadataHasher::new(&metadata)
+            .only_these_pallets(&["First"])
+            .hash();
+        let hash_trimmed = MetadataHasher::new(&trimmed_metadata).hash();
+
+        assert_eq!(hash, hash_trimmed);
```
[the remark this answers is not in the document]
tadeohepperle: This test would have failed before the PR. The reason for the fail would have been that the metadata constructed here contains a `Vec<Events>` as a storage entry (which is an outer enum at the same time).

The recursive types issue we talked about is actually kind of seperate from this issue. It only happened to show that we had this recursive types issue in the solution I used in the first iteration of this PR (which I now closed): https://github.com/paritytech/subxt/pull/1301

So hopefully this PR solves both issues at once. We have always had a test for the [recursive types](https://github.com/paritytech/subxt/blob/tadeohepperle/bugfix-trimmed-metadata-enums/metadata/src/utils/validation.rs#L867), but it only passed because it did not cover critical cases, so it just happened to be a bad test.

metadata/src/utils/validation.rs
```diff
@@ -262,53 +300,95 @@ fn get_extrinsic_hash(
...
-            get_type_hash(registry, id, &mut HashMap::new())
+/// Hash representations of the `frame_metadata::v15::OuterEnums`.
+pub struct OuterEnumHashes {
+    call_hash: (u32, Hash),
+    error_hash: (u32, Hash),
+    event_hash: (u32, Hash),
+}
+
+impl OuterEnumHashes {
+    /// Constructs new `OuterEnumHashes` from metadata. If `only_these_variants` is set, the enums are stripped down to only these variants, before their hashes are calculated.
```
lexnv, line 319: nit; could break this doc on multiple lines

tadeohepperle:
Thank you guys for reviewing, hmm interestingly the Unstable Backend Test time-outed again after 6h.

jsdw:
> Thank you guys for reviewing, hmm interestingly the Unstable Backend Test time-outed again after 6h.

I'm going to look into that now a bit and see if I can find out what's going on there!

jsdw (approved):
Good stuff!

metadata/src/utils/validation/outer_enum_hashes.rs
jsdw, line 1: Personally I'd have just had `utils/outer_enum_hashes` rather than the extra folder but it's all good!

PR #1309, opened 2023-12-05
Make using insecure connections opt-in

### LINKED ISSUE #1239: Make using insecure http/ws connections an explicit opt-in
Currently, connecting to insecure `http` or `ws` endpoints is just as easy as connecting to secure `https`/`wss` once (via either the Subxt library or the CLI tool).

A suggestion from @uzeyirdestan (who has been looking at potential sexurity issues in Subxt) is to make it a more explicit action to "opt-in" to insecure connections.

So, let's define any HTTP/WS connection apart from to `localhost`/`127.0.0.1` as insecure.

Given this, one option would be something like:

1. On the CLI tool, require an `--allow-insecure` flag to be allowed to make an insecure connection (else, it will be rejected), and
2. In the library, for `OnlineClient`, add a `from_insecure_url` which allows insecure URLs, and prevent them via `from_url` 
3. In the library, for `LightClient` connections to URLs, have something like `build_from_insecure_url` to complement `build_from_url`.
4. The macro also allows fetching metadata from any URL via `from_runtime_url`, but that's already meant to be dev only, so just make sure we document that it should be avoided in general.

Have I missed any other places where we establish a connection?

tadeohepperle:
Fixes #1239

To use insecure (no SSL, ws:// or http:// connections) one now has to:
- use `OnlineClient::from_insecure_url` when constructing an online client
- use `LightClientBuilder::build_from_insecure_url` when creating a light client
- add the `--allow-insecure` flag to most commands in the CLI tool

Note: localhost is considered secure and does not need this opt in.


cli/src/utils.rs
```diff
@@ -110,3 +110,23 @@ impl FromStr for FileOrUrl {
         }
     }
 }
+
+pub fn validate_url_security(
+    file_or_url: Option<&Url>,
```
lexnv, line 115: nit: could rename this to url

subxt/src/utils/mod.rs
```diff
@@ -18,6 +18,9 @@ use derivative::Derivative;
 
 pub use account_id::AccountId32;
 pub use era::Era;
+
+#[cfg(feature = "jsonrpsee")]
```
niklasad1, line 22: You need `#[cfg(all(feature = "jsonrpsee", feature = "native"))]` for this to compile.

It might be better to just import `url::Url` directly to get rid of the conditional compilation here....

jsdw: Yup; that or just check that the string starts with `https://` or `wss://`

cli/src/commands/diff.rs
```diff
@@ -29,9 +29,15 @@ pub struct Opts {
     metadata_or_url_1: FileOrUrl,
     /// metadata file or node URL
     metadata_or_url_2: FileOrUrl,
+    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL enscryption
```
niklasad1, line 32:
```suggestion
    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL encryption
```

cli/src/commands/compatibility.rs
```diff
@@ -36,9 +38,16 @@ pub struct Opts {
     /// Defaults to latest.
     #[clap(long = "version", default_value = "latest")]
     version: MetadataVersion,
+    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL enscryption
```
niklasad1, line 41:
```suggestion
    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL encryption
```

cli/src/commands/codegen.rs
```diff
@@ -62,6 +62,9 @@ pub struct Opts {
     /// Defaults to `false` (default substitutions are provided).
     #[clap(long)]
     no_default_substitutions: bool,
+    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL enscryption
```
niklasad1, line 65:
```suggestion
    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL encryption
```

cli/src/commands/explore/mod.rs
```diff
@@ -71,6 +71,9 @@ pub struct Opts {
     pallet: Option<String>,
     #[command(subcommand)]
     pallet_subcommand: Option<PalletSubcommand>,
+    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL enscryption
```
niklasad1, line 74:
```suggestion
    /// Allow insecure URLs e.g. URLs starting with ws:// or http:// without SSL encryption
```

cli/src/commands/explore/mod.rs
```diff
@@ -210,11 +219,38 @@ pub mod tests {
...
+            .unwrap_err()
+            .to_string()
+            .contains("Request error"));
+
+        // also errors, but later, because localhost is considered secure, even without the flag
+        assert!(run("--url ws://localhost")
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("Request error"));
```
niklasad1, line 254: Would be good to assert that the error indicates that url in insecure and run with `--allow-insecure` to fix that.

IIRC, you added a descriptive error and just change the assert here should do it.

subxt/src/utils/mod.rs
```diff
@@ -47,6 +50,21 @@ pub(crate) fn strip_compact_prefix(bytes: &[u8]) -> Result<(u64, &[u8]), codec::
...
 
+/// A URL is considered secure if it uses a secure scheme ("https" or "wss") or is referring to localhost.
+///
+/// Returns an error if the the string could not be parsed into a URL.
+pub fn url_is_secure(url: &str) -> Result<bool, Error> {
+    let url = Url::parse(url).map_err(|e| Error::Rpc(RpcError::ClientError(Box::new(e))))?;
+
+    let secure_scheme = url.scheme() == "https" || url.scheme() == "wss";
+    let is_localhost = url
+        .host_str()
```
niklasad1, line 61: It may be better to use https://docs.rs/url/latest/url/struct.Url.html#method.host instead.
You could then match whether it's a ipv6, ipv4 or domain, https://docs.rs/url/latest/url/enum.Host.html, then the nice thing is that the Ipv6/Ipv4 types exposes `fn is_loopback()`

This is fine though and I don't think we care about ipv6 addresses anyway.

subxt/src/client/light_client/builder.rs
```diff
@@ -101,8 +103,21 @@ impl<T: Config> LightClientBuilder<T> {
     /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
     #[cfg(feature = "jsonrpsee")]
     pub async fn build_from_url<Url: AsRef<str>>(self, url: Url) -> Result<LightClient<T>, Error> {
-        let chain_spec = fetch_url(url.as_ref()).await?;
+        if !url_is_secure(url.as_ref())? {
+            return Err(Error::Rpc(RpcError::InsecureUrl(url.as_ref().into())));
+        }
+        self.build_from_insecure_url(url).await
```
niklasad1, line 109: just call `fn from_url` here and remove the `url check here`

niklasad1 (approved):
LGTM, added some minor suggestions

cli/src/commands/explore/mod.rs
tadeohepperle: The point of the test against `--url ws://localhost` is that it should be accepted by default, even though the scheme is insecure. That is why we expect a "Request error" here instead of an error that prompts the user to use "--allow-insecure".

I know the string checking is not super nice, but this is an end to end test in the CLI tool, which just returns a color_eyre::Result<String>, so I lose the exact error type.

jsdw:
This looks great!

A couple of places left that this should _probably_ also be added to:

- `RpcClient` struct: add `RpcClient::from_insecure_url` and make `from_url` not allow insecure.
- `#[subxt]` macro: rename `runtime_metadata_url` to `runtime_metadata_insecure_url`? it's a dev only option, so I don't really care about doing the proper check, but might be worth following the naming convention (and adding the word `insecure` might put people off using it a bit which would be a good thing anyway :))

jsdw (approved):

cli/src/commands/explore/mod.rs
```diff
@@ -210,11 +219,38 @@ pub mod tests {
...
+        run("--url wss://rpc.polkadot.io:443").await.unwrap();
+
+        // errors because the --allow-insecure is not set:
+        assert!(run("--url ws://rpc.polkadot.io:443")
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("is not secure"));
+
+        // also errors, but later, because of networking reasons:
```
niklasad1, line 242: Can you explain why it error because of networking reasons?
The port is wrong?

cli/src/commands/explore/mod.rs
```diff
@@ -210,11 +219,38 @@ pub mod tests {
...
+        assert!(run("--url ws://rpc.polkadot.io:443 --allow-insecure")
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("Request error"));
+
+        // also errors, but later, because localhost is considered secure, even without the flag
```
niklasad1, line 249: No node is running on port 80 or why does the call fail?

tadeohepperle: Yes, it fails because we don't have a node running locally.

PR #1312, opened 2023-12-07
DO NOT MERGE: testing large tx submission

jsdw:


lexnv:
The issue is no longer reproducible after adding `printlns` for debugging, most probably due to the delay added by those

PR #1321, opened 2023-12-15
fix(runtime update): wait until upgrade on chain

### LINKED ISSUE #1319: runtime upgrade stream API: fetched metadata is "outdated" when upgrade is applied
Currently, I'm using the runtime update stream API in subxt as follows:

```rust
async fn runtime_upgrade_task(api: ChainClient, tx: oneshot::Sender<Error>) {
    let update_stream = ...;

    loop {
        // if the runtime upgrade subscription fails then try establish a new one and if it fails quit.
        let update = match update_stream.next().await {
            Some(Ok(update)) => update,
            _ => {
                log::warn!(target: LOG_TARGET, "Runtime upgrade subscription failed");
                update_stream = match updater.runtime_updates().await {
                    Ok(u) => u,
                    Err(e) => {
                        let _ = tx.send(e.into());
                        return;
                    }
                };
                continue;
            }
        };

        let version = update.runtime_version().spec_version;
        match updater.apply_update(update) {
            Ok(()) => {
                if let Err(e) = epm::update_metadata_constants(&api).await {
                    let _ = tx.send(e);
                    return;
                }
                prometheus::on_runtime_upgrade();
                log::info!(target: LOG_TARGET, "upgrade to version: {} successful", version);
            }
            Err(e) => {
                log::debug!(target: LOG_TARGET, "upgrade to version: {} failed: {:?}", version, e);
            }
        }
    }
}
```

Thus, I'm expecting the metadata in the runtime upgrade to have been updated after `apply_update` returns

Logs from the code

```bash
# start-up
2023-12-13T16:49:56.195801Z  INFO polkadot-staking-miner: Connected to chain: westend
2023-12-13T16:50:26.197747Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::MaxWinners`: 1000
2023-12-13T16:50:26.197759Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::SignedPhase`: 10
2023-12-13T16:50:26.259158Z DEBUG polkadot-staking-miner: upgrade to version: 1005000 failed: SameVersion

# runtime upgrade,  expecting MaxWinners=13, SignedPhase=5
# the metadata doesn't contain the runtime upgrade
2023-12-13T16:54:30.843695Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::MaxWinners`: 1000
2023-12-13T16:54:30.843706Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::SignedPhase`: 10
2023-12-13T16:54:30.843791Z  INFO polkadot-staking-miner: upgrade to version: 1009000 successful

# restart the client, now it works as expected.
2023-12-13T16:55:38.808698Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::MaxWinners`: 13
2023-12-13T16:55:38.808709Z TRACE polkadot-staking-miner: updating metadata constant `ElectionProviderMultiPhase::SignedPhase`: 5
```

After some investigation, I was able to track down that the issue is really that the fetched metadata in the runtime upgrade doesn't really contain the latest metadata. The reason is that `state_subscribeStorage` emits an runtime upgrade event before it's actually applied, and [subxt's RuntimeUpgradeStream](https://github.com/paritytech/subxt/blob/master/subxt/src/client/online_client.rs#L430-#L460) should really assert that the runtime upgrade has occurred at the block or earlier that the metadata is fetched at (with the exception for the initial runtime upgrade)

With the following hack in subxt the issue goes away:

```git
diff --git a/Cargo.lock b/Cargo.lock
index b8b0ab7..febe2bf 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4695,7 +4695,9 @@ dependencies = [
  "libc",
  "mio",
  "num_cpus",
+ "parking_lot",
  "pin-project-lite",
+ "signal-hook-registry",
  "socket2 0.5.5",
  "tokio-macros",
  "windows-sys 0.48.0",
diff --git a/subxt/Cargo.toml b/subxt/Cargo.toml
index cc8ff58..d3196ee 100644
--- a/subxt/Cargo.toml
+++ b/subxt/Cargo.toml
@@ -96,6 +96,7 @@ tokio-stream = { workspace = true, optional = true }
 
 # Included if "web" feature is enabled, to enable its js feature.
 getrandom = { workspace = true, optional = true }
+tokio = { workspace = true, features = ["full"] }
 
 [dev-dependencies]
 bitvec = { workspace = true }
diff --git a/subxt/src/client/online_client.rs b/subxt/src/client/online_client.rs
index 338643e..73b02e3 100644
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -3,6 +3,7 @@
 
 use super::{OfflineClient, OfflineClientT};
+use crate::config::Header;
 use crate::custom_values::CustomValuesClient;
 use crate::{
     backend::{
@@ -437,11 +438,24 @@ impl<T: Config> RuntimeUpdaterStream<T> {
             Err(err) => return Some(Err(err)),
         };
 
+        // Sleep a few blocks to ensure that fetch metadata below 
+        // is the metadata from the runtime upgrade.
+        tokio::time::sleep(std::time::Duration::from_secs(5 * 6)).await;
+
         let latest_block_ref = match self.client.backend().latest_finalized_block_ref().await {
             Ok(block_ref) => block_ref,
             Err(e) => return Some(Err(e)),
         };
 
+        let header = self
+            .client
+            .backend()
+            .block_header(latest_block_ref.hash())
+            .await
+            .unwrap()
+            .unwrap();
+        tracing::info!("Fetching metadata at: {}", header.number().into());
+
         let metadata = match OnlineClient::fetch_metadata(
             self.client.backend(),
             latest_block_ref.hash(),

```

niklasad1:
Close #1319 

This PR ensures that the runtime upgrade from `state_subscribeRuntimeVersion` is enacted on a block, and only after that point the metadata is fetched and updated by subxt.

The implementation itself queries `system::lastRuntimeUpgrade` to ensure that, but there may be better alternatives. 

I also tested it by just waiting for one more block, but that didn't work when I tested it but this solution does.

### What is bad

It's not nice to depend on storage stuff in the subxt client but I haven't find a better way than to wait for around 5-10 blocks statically.

### Other alternatives

Subscribe to the storage key `system().last_runtime_upgrade()` would work as well

### DIFF
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -3,7 +3,6 @@
 // see LICENSE for license details.
 
 use super::{OfflineClient, OfflineClientT};
-use crate::config::Header;
 use crate::custom_values::CustomValuesClient;
 use crate::{
     backend::{
@@ -436,26 +435,6 @@ impl<T: Config> RuntimeUpdaterStream<T> {
             Err(err) => return Some(Err(err)),
         };
 
-        // Remove this block before merging.
-        {
-            let b = self
-                .client
-                .backend()
-                .latest_finalized_block_ref()
-                .await
-                .unwrap();
-
-            let h = self
-                .client
-                .backend()
-                .block_header(b.hash())
-                .await
-                .unwrap()
-                .unwrap();
-
-            tracing::info!("Runtime upgrade initiated at #{}", h.number().into());
-        }
-
         let at =
             match wait_runtime_upgrade_in_finalized_block(&self.client, &runtime_version).await? {
                 Ok(at) => at,
@@ -513,7 +492,7 @@ async fn wait_runtime_upgrade_in_finalized_block<T: Config>(
     };
 
     let block_ref = loop {
-        let (b, block_ref) = match block_sub.next().await? {
+        let (_, block_ref) = match block_sub.next().await? {
             Ok(n) => n,
             Err(err) => return Some(Err(err)),
         };
@@ -521,13 +500,12 @@ async fn wait_runtime_upgrade_in_finalized_block<T: Config>(
         let key: Vec<scale_value::Value> = vec![];
         let addr = crate::dynamic::storage("System", "LastRuntimeUpgrade", key);
 
-        // The storage `system::lastRuntimeUpgrade` should always have a version
-        //
-        // <https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/system/src/lib.rs#L958>
         let chunk = match client.storage().at(block_ref.hash()).fetch(&addr).await {
             Ok(Some(v)) => v,
             Ok(None) => {
-                continue;
+                // The storage `system::lastRuntimeUpgrade` should always exist.
+                // <https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/system/src/lib.rs#L958>
+                unreachable!("The storage item `system::lastRuntimeUpgrade` should always exist")
             }
             Err(e) => return Some(Err(e)),
         };
@@ -550,10 +528,7 @@ async fn wait_runtime_upgrade_in_finalized_block<T: Config>(
         // We are waiting for the chain to have the same spec version
         // as sent out via the runtime subscription.
         if spec_version == runtime_version.spec_version {
-            tracing::info!("block #{} new runtime", b.number().into());
             break block_ref;
-        } else {
-            tracing::info!("block #{} no new runtime", b.number().into());
         }
     };
 
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -540,7 +540,7 @@ async fn wait_runtime_upgrade_in_finalized_block<T: Config>(
         let Some(Ok(spec_version)) = scale_val
             .at("spec_version")
             .and_then(|v| v.as_u128())
-            .map(|v| u32::try_from(v))
+            .map(u32::try_from)
         else {
             return Some(Err(Error::Other(
                 "Decoding `RuntimeVersion::spec_version` as u32 failed".to_string(),
--- a/subxt/src/client/online_client.rs
+++ b/subxt/src/client/online_client.rs
@@ -3,6 +3,7 @@
 // see LICENSE for license details.
 
 use super::{OfflineClient, OfflineClientT};
+use crate::config::Header;
 use crate::custom_values::CustomValuesClient;
 use crate::{
     backend::{
@@ -437,17 +438,13 @@ impl<T: Config> RuntimeUpdaterStream<T> {
             Err(err) => return Some(Err(err)),
         };
 
-        let latest_block_ref = match self.client.backend().latest_finalized_block_ref().await {
-            Ok(block_ref) => block_ref,
-            Err(e) => return Some(Err(e)),
+        let at = match wait_for_runtime_upgrade(&self.client, &runtime_version).await {
+            Some(Ok(at)) => at,
+            Some(Err(err)) => return Some(Err(err)),
+            None => return None,
         };
 
-        let metadata = match OnlineClient::fetch_metadata(
-            self.client.backend(),
-            latest_block_ref.hash(),
-        )
-        .await
-        {
+        let metadata = match OnlineClient::fetch_metadata(self.client.backend(), at.hash()).await {
             Ok(metadata) => metadata,
             Err(err) => return Some(Err(err)),
         };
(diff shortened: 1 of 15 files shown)


subxt/src/client/online_client.rs
```diff
@@ -484,3 +481,49 @@ impl Update {
...
+            .and_then(|v| v.as_u128())
+            .expect("specVersion should exist on RuntimeVersion; qed")
+            as u32;
+
+        if spec_version >= runtime_version.spec_version {
+            break block;
+        }
+    };
+
+    Some(Ok(head))
```
jsdw, line 528: I'd return the `BlockRef` instead of the `T::Header` from this; we want to keep the `BlockRef` around for as long as we might want to fetch the block, to signal to the node to keep said block details around if it can.

`BlockRef` has ` .hash()` function that you can call as well to get the block hash :)

subxt/src/client/online_client.rs
```diff
@@ -484,3 +481,49 @@ impl Update {
...
+
+        let scale_val = match chunk.to_value() {
+            Ok(v) => v,
+            Err(e) => return Some(Err(e)),
+        };
+
+        let spec_version = scale_val
+            .at("spec_version")
+            .and_then(|v| v.as_u128())
+            .expect("specVersion should exist on RuntimeVersion; qed")
```
jsdw, line 520: Nit: since we're talking to a thing that could return any sort of nonsense it's probably best to return an actual error here

subxt/src/client/online_client.rs
```diff
@@ -437,17 +438,13 @@ impl<T: Config> RuntimeUpdaterStream<T> {
             Err(err) => return Some(Err(err)),
         };
 
-        let latest_block_ref = match self.client.backend().latest_finalized_block_ref().await {
-            Ok(block_ref) => block_ref,
-            Err(e) => return Some(Err(e)),
+        let at = match wait_for_runtime_upgrade(&self.client, &runtime_version).await {
```
jsdw, line 441: Nit; maybe rename to `wait_for_finalized_block_containing_update` or something like that

jsdw (approved):
Thanks @niklasad1; def an improvement on what we have! Just keep the `BlockRef` around until after we've fetched metadata :)

subxt/src/client/online_client.rs
```diff
@@ -437,17 +438,13 @@ impl<T: Config> RuntimeUpdaterStream<T> {
...
+            None => return None,
         };
 
-        let metadata = match OnlineClient::fetch_metadata(
-            self.client.backend(),
-            latest_block_ref.hash(),
-        )
-        .await
-        {
+        let metadata = match OnlineClient::fetch_metadata(self.client.backend(), at.hash()).await {
```
tadeohepperle, line 447: Why do we fetch the metadata from the backend here? I guess it is not part of the "LastRuntimeUpgrade" storage item for which we already got the scale_val below, that contains the spec_version?

tadeohepperle (approved):

subxt/src/client/online_client.rs
jsdw: Yep exactly; the storage item tells us that the runtime version has updated in this block (but doesn't contain the metadata too), and so once we see that a block is updated, we fetch the metadata at that block to use to update Subxt

subxt/src/client/online_client.rs
```diff
@@ -437,17 +438,13 @@ impl<T: Config> RuntimeUpdaterStream<T> {
...
         };
 
-        let latest_block_ref = match self.client.backend().latest_finalized_block_ref().await {
-            Ok(block_ref) => block_ref,
-            Err(e) => return Some(Err(e)),
+        let at = match wait_for_runtime_upgrade(&self.client, &runtime_version).await {
+            Some(Ok(at)) => at,
+            Some(Err(err)) => return Some(Err(err)),
+            None => return None,
         };
```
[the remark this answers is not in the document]
niklasad1: thanks

subxt/src/client/online_client.rs
```diff
@@ -484,3 +480,53 @@ impl Update {
...
+
+        let scale_val = match chunk.to_value() {
+            Ok(v) => v,
+            Err(e) => return Some(Err(e)),
+        };
+
+        let Some(Ok(spec_version)) = scale_val
+            .at("spec_version")
+            .and_then(|v| v.as_u128())
+            .map(|v| u32::try_from(v))
```
niklasad1, line 519:
```suggestion
            .map(u32::try_from)
```

subxt/src/client/online_client.rs
```diff
@@ -484,3 +480,53 @@ impl Update {
...
+            .at("spec_version")
+            .and_then(|v| v.as_u128())
+            .map(u32::try_from)
+        else {
+            return Some(Err(Error::Other(
+                "Decoding `RuntimeVersion::spec_version` as u32 failed".to_string(),
+            )));
+        };
+
+        if spec_version >= runtime_version.spec_version {
```
niklasad1, line 526:
```suggestion
        if spec_version > runtime_version.spec_version {
```

niklasad1:
> Or when we see this block with a new spec version, are we already too late (ie we should be using updated metadata to work with it, too)

AFAIU, the runtime upgrade has already been applied when this storage item is changed such that it may be possible that transaction(s) on the block before the metadata is "updated" may fail but after point it should work. However, that should be quite unlikely to occur but I can test it tomorrow.

PR #1332, opened 2023-12-29
 generate docs for feature gated items

niklasad1:
I noticed that the stuff hidden behind the `unstable-light-client` doesn't show up on docs.rs/subxt which this PR fixes.

`docsrs` is still unstable and needs to enabled if one wants to use it, https://github.com/rust-lang/rust/issues/43781


codegen/Cargo.toml
```diff
@@ -40,3 +40,10 @@ bitvec = { workspace = true }
 scale-info = { workspace = true, features = ["bit-vec"] }
 pretty_assertions = { workspace = true }
 frame-metadata = { workspace = true }
+
+[package.metadata.docs.rs]
+all-features = true
```
jsdw, line 45: Can we test this somehow? would `all-features = true` enable eg both `native` _and_ `web` and cause a compile error for instance, or does it avoid that?

subxt/src/tx/signer.rs
```diff
@@ -25,6 +25,7 @@ pub trait Signer<T: Config> {
 }
 
 #[cfg(feature = "substrate-compat")]
+#[cfg_attr(docsrs, doc(cfg(feature = "substrate-compat")))]
```
jsdw, line 28: This line that we have to add below each feature is sortof ugly! I'm not really familiar with it; what does it do? I also wonder whether we could avoid this line duplication somehow.

niklasad1: we could write a macro for it as we have in jsonrpsee

```rust
macro_rules! cfg_feature {
	($feature:literal, $($item:item)*) => {
		$(
			#[cfg(feature = $feature)]
			#[cfg_attr(docsrs, doc(cfg(feature = $feature)))]
			$item
		)*
	}
}

```

codegen/Cargo.toml
niklasad1: yeah, you are right I have to remove that

subxt/src/lib.rs
```diff
@@ -55,6 +57,10 @@ pub mod storage;
 pub mod tx;
 pub mod utils;
 
+// Macros useful below, but not to be exposed outside of the crate.
```
niklasad1, line 60:
```suggestion
// Internal helper macros
```

tadeohepperle (approved):

lexnv (approved):

PR #1336, opened 2024-01-03
CLI Tool: Allow using `--file -` to read metadata from stdin

### LINKED ISSUE #1326: subxt codegen should support the input of metadata through stdin
That would allow to pipe it in the following manner:

`parachain-node export-metadata | subxt codegen --file - | rustfmt --edition=2021 --emit-stdout > metadata.rs`

tadeohepperle:
fixes #1326

I tested it manually by running `subxt metadata | subxt codegen --file -` (which is not super useful obviously) but shows that it works for the requested usecase of running `parachain-node export-metadata | subxt codegen --file -`.

### DIFF
--- a/cli/src/utils.rs
+++ b/cli/src/utils.rs
@@ -21,7 +21,7 @@ pub struct FileOrUrl {
     pub url: Option<Url>,
     /// The path to the encoded metadata file.
     #[clap(long, value_parser)]
-    pub file: Option<PathBuf>,
+    pub file: Option<PathOrStdIn>,
     /// Specify the metadata version.
     ///
     ///  - "latest": Use the latest stable version available.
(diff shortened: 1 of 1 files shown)


cli/src/utils.rs
```diff
@@ -33,6 +33,26 @@ pub struct FileOrUrl {
...
+#[derive(Debug, Clone)]
+pub enum PathOrStdIn {
+    Path(PathBuf),
+    StdIn,
+}
+
+impl FromStr for PathOrStdIn {
+    type Err = <PathBuf as FromStr>::Err;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
niklasad1, line 47: I wonder whether we should trim the leading and trailing whitespaces before checking that the `--file is `-`

niklasad1: and a simple test for the `from_str` impl would be nice to have.

niklasad1 (approved):

cli/src/utils.rs
tadeohepperle: I think regarding trimming, `clap` already takes care of that before the args are given to us to parse, but I am not entirely sure.

niklasad1: Tadeo added trimming and tests for it :)

jsdw (approved):
LGTM, nice one!

PR #1339, opened 2024-01-05
Macro error messages validating metadata and making suggestions

### LINKED ISSUE #1269: codegen: Compile error propagation for deriving at invalid paths
The `derive_as_type` adds derives with best effort, if the path to the type is invalid or the type does not exist, the macro silently continues.

Ideally, we could provide back to users with an explicit compile time error with extra information:
- similar paths at which the type is found
- type is not present in the metadata
- path is resolvable and type is present, although a derive was previously stated (deduplicating)

For example, the following is currently accepted by the subxt macro:
```rust
derive_for_type(
    path = "runtime_types::xcm::v2::multilocation::MultiLocation",
    derive = "Clone",
)
```
This is invalid, because the path `derive_for_type` expects should not include `runtime_types`.

The right format should be `derive_for_type(path = "xcm::v2::multilocation::MultiLocation", derive = "Clone"),` and the macro could generate a compile error:

```rust

derive_for_type(path = ...)
^^^^  Type `MultiLocation` does not exist at path `runtime_types::xcm::v2::multilocation::MultiLocation`.

A type with the same name is present at:
xcm::v2::multilocation::MultiLocation
xcm::v3::multilocation::MultiLocation
```

tadeohepperle:
Fixes #1269

Branched from #1260  so  #1260 should be merged first.

This PR introduces checks to all user defined type paths that are passed to the subxt macro. This could be a custom derive, attribute or a substitute for a type. The macro now makes sure that the specified path exists in the type registry. If this is not fulfilled it errors and gives the user a list of valid paths that belong to a Type with the same name. Here is an example:

image


macro/src/lib.rs
```diff
@@ -139,20 +149,65 @@ pub fn subxt(args: TokenStream, input: TokenStream) -> TokenStream {
...
+            alternatives
+        );
+    }
+
+    fn pretty_path(path: &syn::Path) -> String {
+        path.to_token_stream().to_string().replace(' ', "")
+    }
+}
+
+/// Fetches metadata in a blocking manner, either from a url (not recommended) or from a file path.
```
niklasad1, line 204: I don't get why fetching metadata from an URL is not recommended? :)

macro/src/lib.rs
```diff
@@ -83,17 +88,21 @@ struct SubstituteType {
...
-        Ok(v) => v,
-        Err(e) => return TokenStream::from(e.write_errors()),
-    };
+    match _subxt(args, parse_macro_input!(input as syn::ItemMod)) {
+        Ok(e) => e,
+        Err(e) => e,
+    }
+}
+
+// Node: just an additonal function to make early returns easier.
```
niklasad1, line 97:
```suggestion
// Note: just an additional function to make early returns easier.
```

niklasad1 (approved):

macro/src/lib.rs
tadeohepperle: It is not recommended for users because the compilation blocks while fetching the metadata, which is a bit slow and (maybe flaky) compared to using metadata file. But yeah, maybe the docs of this function don't need to mention this.

macro/src/lib.rs
```diff
@@ -83,17 +88,21 @@ struct SubstituteType {
...
-        Err(e) => return TokenStream::from(e.write_errors()),
-    };
+    match _subxt(args, parse_macro_input!(input as syn::ItemMod)) {
+        Ok(e) => e,
+        Err(e) => e,
+    }
+}
+
+// Note: just an additional function to make early returns easier.
+fn _subxt(args: TokenStream, item_mod: syn::ItemMod) -> Result<TokenStream, TokenStream> {
```
lexnv, line 98: nit: or `subxt_inner`; the underscore leads me to wonder if we are going to use the function (similar to `let _unused = ..`)

macro/src/lib.rs
lexnv: Maybe we can say its not recommended for production use-cases. Do remember we added the ability to fetch the metadata from URL as a feature request from the `cargo-contract` team, that wanted to fetch a fresh metadata for their testing

macro/Cargo.toml
```diff
@@ -23,5 +23,7 @@ proc-macro = true
 codec = { package = "parity-scale-codec", workspace = true }
 darling = { workspace = true }
 proc-macro-error = { workspace = true }
-syn = { workspace = true }
+syn.workspace = true
```
lexnv, line 26: I do wonder, did this happen as a result of some linting extensions from your IDE?

We could revert this to keep consistency with the dependencies of this file (`syn = { workspace = true }`)

lexnv (approved):
Nice one! 👍 

Just some changes wrt the documentation, and importing the dependencies with a different format

PR #1347, opened 2024-01-08
chore(deps): bump jsonrpsee v0.20 to v0.21

niklasad1:
Mostly for nicer logging target and bump a bunch of dependencies.


subxt/src/backend/rpc/rpc_client.rs
```diff
@@ -229,9 +229,13 @@ impl<Res: DeserializeOwned> Stream for RpcSubscription<Res> {
...
 mod jsonrpsee_helpers {
     pub use jsonrpsee::{
-        client_transport::ws::{Receiver, Sender, Url, WsTransportClientBuilder},
-        core::{client::Client, Error},
+        client_transport::ws::{self, EitherStream, Url, WsTransportClientBuilder},
+        core::client::{Client, Error},
     };
+    use tokio_util::compat::Compat;
+
+    pub type Sender = ws::Sender<Compat<EitherStream>>;
```
niklasad1, line 237: we should these type aliases in jsonrpsee but didn't notice it before now

lexnv (approved):
Nice 👍

PR #1352, opened 2024-01-09
Weekly Cronjob fetching artifacts and generating polkadot.rs file.

### LINKED ISSUE #1296: workflow: Update artifacts automatically
Create a workflow that runs the `scripts/artifacts.sh` file to update:
- metadata scale
- polkadot.rs codegen interface from testing
- demo chain specs

The changes should be pushed as a commit for us to review, with weekly frequency.

The advantage of doing this is that we'll never have to worry about manually updating those files.
In the past, we have added features to the `codegen` that were not reflected in the artifacts files. 
Also, the demo `chainSpec` plays a huge role in the synchronization time of the light clients.

This should be done after:
- https://github.com/paritytech/subxt/issues/1263

tadeohepperle:
Fixes #1296

This PR adds a github action that should run weekly on Monday.
It should run a local substrate node in the CI runner and regenerate the artifacts  and the polkadot.rs file.
It then proceeds to create a PR based off the master branch that we can manually approve (running testing CI in that PR).

I tested the CI Action manually in this Subxt Fork, it seems to run as desired:
[Actions](https://github.com/tadeohepperle/subxt/actions) + [Generated PR](https://github.com/tadeohepperle/subxt/pull/2)

I made sure the integration tests ran locally for me with the generated `polkadot.rs`


.github/workflows/update-artifacts.yml
```diff
@@ -0,0 +1,62 @@
...
+      # Needs to be built before, so that it starts quickly when called by artifacts script.
+      - name: Build the CLI tool)
+        run: cargo build --bin subxt
+
+      # Runs two processes asynchronously:
+      # - one for the substrate node, timing out after 120s.
+      # - and one to fetch the metadata from it with the `scripts/artifacts.sh` script. This waits for 30s first to give the node enough time to start up.
+      - name: Fetch Artifacts
+        run: |
+          { timeout 120s substrate-node --dev --tmp; } & { sleep 30s && ./scripts/artifacts.sh; }
```
niklasad1, line 46: I would prefer if you could write a simple script/binary to actually is using the `substrate node runner` that actually parses the assigned rpc port and ensures that node has been started so we don't need the sleeps here before lauching `./scripts/artifacts.sh` here.

Recall, the substrate may assign a random port if 9944 is already used and then this will not work

.github/workflows/update-artifacts.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@3cf7f8cc28d1b4e7d01e3783be10a97d55d483c8 # v2.7.1
+
+      # Needs to be built before, so that it starts quickly when called by artifacts script.
```
niklasad1, line 37: This shouldn't be needed, the artifacts scrips is using `cargo run <args>` which will build the thingy but doesn't matter.

scripts/artifacts.sh
```diff
@@ -1,5 +1,5 @@
-# # Generates the 3 metadata files in `/artifacts` and the polkadot.rs file for integration tests
-#
+# # Generates the 4 metadata files in `/artifacts` and the polkadot.rs file for integration tests.
```
niklasad1, line 1: Maybe you remove this completely now :rocket:

scripts/artifacts.sh
```diff
@@ -1,5 +1,5 @@
...
+
 # This script is to be run from the root of the repository: `scripts/artifacts.sh`
 #
 # It expects a local polkadot node to be running a JSON-RPC HTTP server at 127.0.0.1:9933
```
niklasad1, line 5:
```suggestion
# It expects a local polkadot node to be running a JSON-RPC HTTP server at 127.0.0.1:9944
```

scripts/artifacts/src/main.rs
```diff
@@ -0,0 +1,89 @@
+use std::{
```
niklasad1, line 1: nice :)

niklasad1 (approved):
nice

scripts/artifacts.sh
tadeohepperle: This line or the entire bash script file?

scripts/artifacts.sh
```diff
@@ -1,15 +1,18 @@
...
+# It expects a local polkadot node to be running a JSON-RPC HTTP server at 127.0.0.1:9944
 # A local polkadot node can be run via:
 # ```
 # git clone https://github.com/paritytech/polkadot.git
 # cd polkadot 
 # cargo build --release
 # ./target/release/polkadot --dev --tmp
 # ```
+#
+# Instead of this script you can also use the binary crate "scripts/artifacts",
```
jsdw, line 12: Is there a good reason to keep both around? I reckon that if we are happy with the rsut crate that we just use that to avoid having two similar things around which may diverge in how they work :)

scripts/artifacts/src/main.rs
jsdw, line 1: Looks good!

scripts/artifacts.sh
tadeohepperle: Good point, the bash script can be used for spinning up a local node and generating artifacts from that. But I am not sure who is using it, and if it is needed we can make it a CLI option to the binary crate as well.

jsdw: We only wrote it for our own internal use, so if the rust script does what we need then we can ditch it

sergejparity (approved):

jsdw (approved):
Looks good to me; I guess we can verify that it works properly in CI when it merges :)

PR #1318, opened 2023-12-12
[DNM] Do not merge:  Subxt UnstableBackend: Timeout issue repro

lexnv:

PR #1356, opened 2024-01-11
backend(fix): Remove only finalized blocks from the event window

lexnv:
This PR ensures that the window of events we feed to every new subscription contains the appropriate `NewBlock` and `BestBlock` events.

Before this PR, the `block_events_from_last_finalized` window of events was cleared out on every `Finalized` event.
However, this approach could miss reporting `NewBlocks` that will later become `Finalized`. Therefore, the unstable driver could report a block as `Finalized` and never report it as `NewBlock` first.

Considering the following edge case:

```bash

      0x1 -----> 0x2 -----> 0x3 -----> 0x4

T0    init       new         new       new
T1               fin                 
```

- T0: The chainHead subscription is started, 0x1 0x2 0x3 0x4 are announced
  - 0x1 is the `Initialized` event
  - `block_events_from_last_finalized` contains: 0x2 0x3 0x4 as `NewBlocks`
- T1: 0x2 is reported as `Finalized` 
  - Before this PR the `block_events_from_last_finalized` is cleared our entirely

If a new subscription is started after T1; it will receive an `Initialized` event with the 0x2 block. However, it will never receive `0x3` and `0x4` as `NewBlocks`.  After around 6 seconds, the `0x3` is reported in a `Finalized` event. This breaks the expectations of the rpc-v2 spec.

To mitigate this behavior, the window of events is not entirely cleaned out. Instead, only blocks reported as `Finalized` or pruned are removed from the window.
The last finalized block will be reported as `Initialized` by our driver, therefore there is no need to report `NewBlock` and `BestBlock` events for it.  If the `Finalized` event reported multiple finalized hashes, we only care about the state at the head of the chain, therefore it is correct to remove those as well. Idem for the pruned hashes; they will never be reported again and we remove them from the window of events.

### Testing Done
- This has been discovered by https://github.com/paritytech/subxt/pull/1318
- For more details see [this comment](https://github.com/paritytech/subxt/pull/1318#issuecomment-1884792300)
- The `decode_signed_extensions_from_blocks` with the `panic` patch is passing after this PR
- Added an unit test to ensure we don't clear out events of interest

### DIFF
--- a/subxt/src/backend/unstable/follow_stream_driver.rs
+++ b/subxt/src/backend/unstable/follow_stream_driver.rs
@@ -165,9 +165,8 @@ struct SharedState<Hash: BlockHash> {
     done: bool,
     next_id: usize,
     subscribers: HashMap<usize, SubscriberDetails<Hash>>,
-    // Keep a buffer of all events from last finalized block so that new
-    // subscriptions can be handed this info first.
-    block_events_from_last_finalized: VecDeque<FollowEvent<BlockRef<Hash>>>,
+    /// Keep a buffer of all events that should be handed to a new subscription.
+    block_events_for_new_subscriptions: VecDeque<FollowEvent<BlockRef<Hash>>>,
     // Keep track of the subscription ID we send out on new subs.
     current_subscription_id: Option<String>,
     // Keep track of the init message we send out on new subs.
@@ -186,7 +185,7 @@ impl<Hash: BlockHash> Default for Shared<Hash> {
             current_init_message: None,
             current_subscription_id: None,
             seen_runtime_events: HashMap::new(),
-            block_events_from_last_finalized: VecDeque::new(),
+            block_events_for_new_subscriptions: VecDeque::new(),
         })))
     }
 }
@@ -251,7 +250,7 @@ impl<Hash: BlockHash> Shared<Hash> {
                 // New subscriptions will be given this init message:
                 shared.current_init_message = Some(ev.clone());
                 // Clear block cache (since a new finalized block hash is seen):
-                shared.block_events_from_last_finalized.clear();
+                shared.block_events_for_new_subscriptions.clear();
             }
             FollowStreamMsg::Event(FollowEvent::Finalized(finalized_ev)) => {
                 // Update the init message that we'll hand out to new subscriptions. If the init message
@@ -276,7 +275,7 @@ impl<Hash: BlockHash> Shared<Hash> {
                 }
 
                 // The blocks reported by the finalized event should not be reported again
-                // by the initialized event.
+                // by the initialized event. The pruned blocks are not of interest.
                 let to_remove: HashSet<Hash> = finalized_ev
                     .finalized_block_hashes
                     .iter()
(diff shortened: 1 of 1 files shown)


subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -275,8 +274,26 @@ impl<Hash: BlockHash> Shared<Hash> {
...
+                    .finalized_block_hashes
+                    .iter()
+                    .chain(finalized_ev.pruned_block_hashes.iter())
+                    .map(|h| h.hash())
+                    .collect();
+
+                shared
+                    .block_events_for_new_subscriptions
+                    .retain(|ev| match ev {
+                        FollowEvent::NewBlock(new_block_ev) => {
```
niklasad1, line 289: I don't follow why `NewBlock` and `BestBlockChanged` are the only events that causes events to be removed from the `block_events`? 

Can you explain :pray:?

niklasad1: or the other events removed in some other scenario?

subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -275,8 +274,26 @@ impl<Hash: BlockHash> Shared<Hash> {
...
+                            !to_remove.contains(&new_block_ev.block_hash.hash())
+                        }
+                        FollowEvent::BestBlockChanged(best_block_ev) => {
+                            !to_remove.contains(&best_block_ev.best_block_hash.hash())
+                        }
+                        _ => true,
```
niklasad1, line 295: Shouldn't this be false because we only want to keep `NewBlock` and `BestBlockChanged` around after the block was finalized?

subxt/src/backend/unstable/follow_stream_driver.rs
niklasad1, line 289: can you add a comment here,

> The last finalized block will be reported as Initialized by our driver, therefore there is no need to report NewBlock and BestBlock events for it. If the Finalized event reported multiple finalized hashes, we only care about the state at the head of the chain, therefore it is correct to remove those as well. Idem for the pruned hashes; they will never be reported again and we remove them from the window of events.

subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -251,7 +250,7 @@ impl<Hash: BlockHash> Shared<Hash> {
                 // New subscriptions will be given this init message:
                 shared.current_init_message = Some(ev.clone());
                 // Clear block cache (since a new finalized block hash is seen):
-                shared.block_events_from_last_finalized.clear();
+                shared.block_events_for_new_subscriptions.clear();
```
niklasad1, line 253: yeah, alright when `Initialized` event is generated we can conclude that no messages regarding the next block are missed.

niklasad1 (approved):

jsdw (approved):
I could only review on my phone but this looks like a good approach to me; good job!

lexnv:
Thanks for the review guys! I'll keep my eyes on the CI for a couple of days to make sure no other errors appear 🙏

tadeohepperle (approved):

PR #1358, opened 2024-01-12
 [Do not merge] UnstableBackend: Debug master timeout

lexnv:


lexnv:
The timeout is reproducible with origin/master https://github.com/paritytech/subxt/pull/1358/commits/641a45f911fa5bc052e7d9307d6844e1c15a617b, however it is not with the refactoring patch https://github.com/paritytech/subxt/pull/1358/commits/c0c08cdd5d5502c95b3a8d58b01d8db4030783c1.

jsdw:
I think that the net change in https://github.com/paritytech/subxt/commit/c0c08cdd5d5502c95b3a8d58b01d8db4030783c1 is that this optimisation is removed (and otherwise just a bit of tidy up), but I can't think why removing it would fix anything off hand; do you have any theories?

Without the optimisation, the difference is that we are storing new blocks (and their block refs, which may keep them alive when otherwise they would be cleaned up eventually) even when we know that the tx is in a finalized block and only care about looking for those.

The only thing I can think of is that ditching new blocks might lead to block refs incorrectly being pruned even when we see them finalized, and so keeping them around is preventing that bug?

jsdw:
Ah yup, I think you're right about the `this.unpin_blocks(cx.waker());` bit; that unpins _all_ dropped blocks, which may include new blocks that aren't finalized yet! (It seems I made this mistake in more than one place!)

We should instead only unpin any blocks in the pruned blocks list from that finalized event that are also in the current unpin HashSet because it's only those blocks that we'll def never see any mention of again.

We should also add a test which checks this; I tested that new blocks aren't unpinned until finalized, but obviously didn't think to test the case that new blocks which arent pruned are also not unpinned yet :)

Well done for spotting this @lexnv!

lexnv:
The fix has been deployed on origin/master

PR #1365, opened 2024-01-15
[Do not merge] UnstableBackend: Debug master timeout - part 2

lexnv:
Continues https://github.com/paritytech/subxt/pull/1358; because the CI steps no longer run there (probably hit limit)


lexnv:
The fix has been deployed on origin/master

PR #1367, opened 2024-01-16
backend(fix): Early unpin for pruned blocks with no active BlockRef 

lexnv:
This PR ensures that blocks reported by `NewBlock` stay alive until the block is either reported as `Finalized` or `Pruned`.

Before this PR, the unstable backend would call `chainHead_unpin` for blocks that do not have an active `BlockRef`.
That behavior could lead to a scenario where a `NewBlock` reference is being dropped; causing the block to be unpinned by the server before it is announced as `Finalized`.

```bash
T0.
new0x0    new0x1   new0x2 
     mark 0x0 ready for unpin
          mark 0x1 ready for unpin
               mark 0x2 ready for unpin
               
T1.
                                                    finalized0x0
```
In this example, the `Finalized 0x0` event would result in blocks reported by `NewBlock 0x1` and `NewBlock 0x2` to be unpinned.

To mitigate this, the PR introduces an `UnpinPolicy`:
- pinned blocks that are pinned for too long (`max_block_life`) - represents all blocks except the pruned blocks
- blocks that are reported as pruned (since we'll never encounter another event about them) _and_ that have no active `BlockRef` (ie, `NewBlock` / `BestBlock` event has been dropped)

The `UnpinPolicy` is needed to remove only dropped pruned blocks. Without this, we would either provide an invalid `BlockRef` to the pruned block (which breaks the user expectation); or we would clean pruned blocks on `max_block_life` (since a `BlockRef` is created with every block reported by the `Finalized` / `Pruned` event).

### Testing Done
- Adjusted testing to reflect unpinning for pruned blocks
- Testing is extended to support pruned blocks
- Added `pruned_blocks_should_get_unpinned` and `dropped_new_blocks_should_not_get_unpinned` to verify that pruned blocks are unpinned; and respectively that dropping new blocks does not unpin (because we still want the finalized event to be valid)

Fixes: https://github.com/paritytech/subxt/pull/1358
For extra details see: https://github.com/paritytech/subxt/pull/1358#issuecomment-1892185017

### DIFF
--- a/subxt/src/backend/unstable/follow_stream_unpin.rs
+++ b/subxt/src/backend/unstable/follow_stream_unpin.rs
@@ -315,7 +315,7 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
         entry.block_ref.clone()
     }
 
-    /// Unpin any blocks that are either too old, or have the unpin flag set and are in the list of pruned hashes.
+    /// Unpin any blocks that are either too old, or have the unpin flag set and are old enough.
     fn unpin_blocks(&mut self, waker: &Waker) {
         let unpin_flags = std::mem::take(&mut *self.unpin_flags.lock().unwrap());
         let rel_block_num = self.rel_block_num;
--- a/subxt/src/backend/unstable/follow_stream_unpin.rs
+++ b/subxt/src/backend/unstable/follow_stream_unpin.rs
@@ -174,11 +174,13 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                         })
                         .collect();
 
+                    let pruned_hashes = pruned_block_refs.iter().map(|block| block.hash());
+
                     // At this point, we also check to see which blocks we should submit unpin events
                     // for. When we see a block hash as finalized, we know that it won't be reported again
                     // (except as a parent hash of a new block), so we can safely make an unpin call for it
                     // without worrying about the hash being returned again despite the block not being pinned.
-                    this.unpin_blocks(cx.waker());
+                    this.unpin_blocks(pruned_hashes, cx.waker());
 
                     FollowStreamMsg::Event(FollowEvent::Finalized(Finalized {
                         finalized_block_hashes: finalized_block_refs,
@@ -292,8 +294,8 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
         entry.block_ref.clone()
     }
 
-    /// Unpin any blocks that are either too old, or have the unpin flag set and are old enough.
-    fn unpin_blocks(&mut self, waker: &Waker) {
+    /// Unpin any blocks that are either too old, or have the unpin flag set and are in the list of pruned hashes.
+    fn unpin_blocks(&mut self, pruned_hashes: impl IntoIterator<Item = Hash>, waker: &Waker) {
         let unpin_flags = std::mem::take(&mut *self.unpin_flags.lock().unwrap());
         let rel_block_num = self.rel_block_num;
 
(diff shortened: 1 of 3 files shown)


subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -286,13 +308,14 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
...
                     }),
+                    unpin_policy,
                 },
             });
 
         entry.block_ref.clone()
     }
 
-    /// Unpin any blocks that are either too old, or have the unpin flag set and are old enough.
+    /// Unpin any blocks that are either too old, or have the unpin flag set and are in the list of pruned hashes.
```
lexnv, line 318:
```suggestion
    /// Unpin any blocks that are either too old, or have the unpin flag set and are old enough.
```

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -109,7 +109,11 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                 FollowStreamMsg::Event(FollowEvent::Initialized(details)) => {
                     // The first finalized block gets the starting block_num.
                     let rel_block_num = this.rel_block_num;
-                    let block_ref = this.pin_block_at(rel_block_num, details.finalized_block_hash);
+                    let block_ref = this.pin_block_at(
+                        rel_block_num,
+                        details.finalized_block_hash,
+                        UnpinPolicy::MaxAge,
```
jsdw, line 115: I don't really understand the need for this `UnpinPolicy`; it looks like we are just pinning blocks for as long as possible and no longer unpinning blocks when dropped now?

I was sortof assuming that the change would be smaller; on line https://github.com/paritytech/subxt/pull/1367/files#diff-4ef5f554da480b7c19eac236d6fa5cc67ab6d6d4441eec62765dd55e90df428fR197 we unpin only the blocks listed as pruned blocks rather than unpin _all_ blocks in the unpin set? Perhaps I'm missing an important detail :)

lexnv: To unpin only blocks reported as [pruned](https://github.com/paritytech/subxt/pull/1367/files#diff-4ef5f554da480b7c19eac236d6fa5cc67ab6d6d4441eec62765dd55e90df428fR183) by the finalized event we would need:
- to pass the pruned items of the finalized to the `unpin_blocks`
- and have the pruned blocks be present in the [UnpinFlags](https://github.com/paritytech/subxt/pull/1367/files#diff-4ef5f554da480b7c19eac236d6fa5cc67ab6d6d4441eec62765dd55e90df428fR301) -- their reference is dropped

However, we would never see a `pruned` block present in the `UnpinFlags` because we are pinning them again when we see the `Finalized` event [here](https://github.com/paritytech/subxt/pull/1367/files#diff-4ef5f554da480b7c19eac236d6fa5cc67ab6d6d4441eec62765dd55e90df428fR189). And that would remove their entry from the [UnpinFlags here](https://github.com/paritytech/subxt/pull/1367/files#diff-4ef5f554da480b7c19eac236d6fa5cc67ab6d6d4441eec62765dd55e90df428fR301).

We could move `this.unpin_blocks(cx.waker())` before we pin the prunned blocks. And in doing so cleaning the pruned blocks right away, iff the user has dropped the `NewBlock` reference. However, in doing so, we would return a dummy `BlockRef` for the pruned block that has just been unpinned.

jsdw: Ah interesting, and another thought is that if you haven't yet dropped a block that appears as a pruned block at some point, and then drop it after, then it will never show up in the pruned blocks list again. (In that case though, one could make a note of all currently pinned blocks that have been pruned or something).

I'm not overly comfortable with having most blocks exist just for the max age (in that case I'd lean towards only ever pruning blocks that reach a max age and never trying to do any smarter unpinning). Lemme have a bit of a play too to get a better feel for this and experiment with some other unpinning approach :)

lexnv: We could probably do something similar to the implementation of the `chainHead` in substrate; where we keep track of blocks a bit longer.
That would mean transforming the HashSet for UnpinFlags to a stateful HashMap; where we keep track of unpins from NewBlock refs and Finalized/Pruned refs; and only submit unpin requests when the latter happens. 

Indeed having the blocks live a `max_life` would probably mean that we might get terminated since the substrate implementation is free to change the server `max_life`.

jsdw: Building off your branch here, this is the approach I am trying out:

https://github.com/paritytech/subxt/pull/1368

PR #1368, opened 2024-01-16
fix: don't unpin blocks that may show up again

jsdw:
This builds on #1367.

The intention behind the original unpin logic is that blocks that will never show up again in any future event can now be safely unpinned. The logic was flawed though; we'd unpin _all_ flagged blocks when a finalized event comes in, even if some of those may later appear in another finalized event.

The intention behind this fix is to track of which of our pinned blocks have been pruned by the backend, and only _actually_ unpin those blocks for real, leaving other blocks pinned until they eventually are pruned by the backend too.

### DIFF
--- a/subxt/src/backend/unstable/follow_stream.rs
+++ b/subxt/src/backend/unstable/follow_stream.rs
@@ -249,9 +249,9 @@ pub(super) mod test_utils {
     }
 
     /// A new block event
-    pub fn ev_new_block(parent: u64, n: u64) -> FollowEvent<H256> {
+    pub fn ev_new_block(parent_n: u64, n: u64) -> FollowEvent<H256> {
         FollowEvent::NewBlock(NewBlock {
-            parent_block_hash: H256::from_low_u64_le(parent),
+            parent_block_hash: H256::from_low_u64_le(parent_n),
             block_hash: H256::from_low_u64_le(n),
             new_runtime: None,
         })
@@ -266,12 +266,15 @@ pub(super) mod test_utils {
 
     /// A finalized event
     pub fn ev_finalized(
-        ns: impl IntoIterator<Item = u64>,
-        pruned: impl IntoIterator<Item = u64>,
+        finalized_ns: impl IntoIterator<Item = u64>,
+        pruned_ns: impl IntoIterator<Item = u64>,
     ) -> FollowEvent<H256> {
         FollowEvent::Finalized(Finalized {
-            finalized_block_hashes: ns.into_iter().map(H256::from_low_u64_le).collect(),
-            pruned_block_hashes: pruned.into_iter().map(H256::from_low_u64_le).collect(),
+            finalized_block_hashes: finalized_ns
+                .into_iter()
+                .map(H256::from_low_u64_le)
+                .collect(),
+            pruned_block_hashes: pruned_ns.into_iter().map(H256::from_low_u64_le).collect(),
         })
     }
 }
--- a/subxt/src/backend/unstable/follow_stream.rs
+++ b/subxt/src/backend/unstable/follow_stream.rs
@@ -265,10 +265,13 @@ pub(super) mod test_utils {
     }
 
     /// A finalized event
-    pub fn ev_finalized(ns: impl IntoIterator<Item = u64>) -> FollowEvent<H256> {
+    pub fn ev_finalized(
+        ns: impl IntoIterator<Item = u64>,
+        pruned: impl IntoIterator<Item = u64>,
+    ) -> FollowEvent<H256> {
         FollowEvent::Finalized(Finalized {
             finalized_block_hashes: ns.into_iter().map(H256::from_low_u64_le).collect(),
-            pruned_block_hashes: vec![],
+            pruned_block_hashes: pruned.into_iter().map(H256::from_low_u64_le).collect(),
         })
     }
 }
--- a/subxt/src/backend/unstable/follow_stream_unpin.rs
+++ b/subxt/src/backend/unstable/follow_stream_unpin.rs
@@ -315,7 +315,7 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
         entry.block_ref.clone()
     }
 
-    /// Unpin any blocks that are either too old, or have the unpin flag set and are in the list of pruned hashes.
+    /// Unpin any blocks that are either too old, or have the unpin flag set and are old enough.
     fn unpin_blocks(&mut self, waker: &Waker) {
         let unpin_flags = std::mem::take(&mut *self.unpin_flags.lock().unwrap());
         let rel_block_num = self.rel_block_num;
--- a/subxt/src/backend/unstable/follow_stream_unpin.rs
+++ b/subxt/src/backend/unstable/follow_stream_unpin.rs
@@ -159,7 +159,7 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                             // but if they don't, we just increment the num from the last finalized block
                             // we saw, which should be accurate.
                             //
-                            // `pin_final_block_at` indicates that the block will not show up in future events
+                            // `pin_unpinnable_block_at` indicates that the block will not show up in future events
                             // (They will show up as a parent block, but we don't care about that right now).
                             let rel_block_num = this.rel_block_num + idx + 1;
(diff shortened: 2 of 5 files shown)


subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -155,8 +158,11 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                             // These blocks _should_ exist already and so will have a known block num,
                             // but if they don't, we just increment the num from the last finalized block
                             // we saw, which should be accurate.
+                            //
+                            // `pin_final_block_at` indicates that the block will not show up in future events
```
lexnv, line 162: nit: `pin_unpinnable_block_at`?

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -168,16 +174,19 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                         .pruned_block_hashes
                         .into_iter()
                         .map(|hash| {
-                            // We should know about these, too, and if not we set their age to last_finalized + 1
+                            // We should know about these, too, and if not we set their age to last_finalized + 1.
+                            //
+                            // `pin_final_block_at` indicates that the block will not show up in future events.
```
lexnv, line 179: nit: pin_unpinnable_block_at?

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -271,11 +282,31 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
...
+        hash: Hash,
+        can_be_unpinned: bool,
+    ) -> BlockRef<Hash> {
         let entry = self
             .pinned
             .entry(hash)
-            // Only if there's already an entry do we need to clear any unpin flags set against it.
-            .and_modify(|_| {
+            // If there's already an entry, then clear any unpin_flags and update the
+            // pruned status (a pruned block can never become unpruned again, but vice versa
```
lexnv, line 306: nit: `can_be_unpinned` status

lexnv (approved):
Looks good to me! I think one adjustment should be made to `never_unpin_new_block_before_finalized` to make it pass (`drop(_i0)` the initialized event)

Just to double check I got this right, we store an extra bit of information in the pinning details, the `can_be_unpinned`.
We then set `can_be_unpinned = true` for `Finalized` and `Pruned` events. Then use this flag in combination with the `UnpinFlags` to make a call to `chainHead_unpin`.

And this works for the following edge-case:
- NewBlockRef kept around
- FinalizedBlockRef (can_be_unpinned = true) and block ref dropped immediately -> hash is not added to the `UnpinFlags` yet
- NewBlockRef is used as normal; then dropped and this time is added to the `UnpinFlags`
- the next finalized event will clear out the block generating an unpin

In other words, this is safe for `NewBlock` references that outlive `Finalized` references 👍

tadeohepperle (approved):
Great job figuring this out, the entire pinning/unpinning API feels quite complicated.

subxt/src/backend/unstable/follow_stream.rs
```diff
@@ -265,10 +265,13 @@ pub(super) mod test_utils {
     }
 
     /// A finalized event
-    pub fn ev_finalized(ns: impl IntoIterator<Item = u64>) -> FollowEvent<H256> {
+    pub fn ev_finalized(
+        ns: impl IntoIterator<Item = u64>,
+        pruned: impl IntoIterator<Item = u64>,
```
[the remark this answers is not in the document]
jsdw: numbers (or more verbosely, block numbers!)

jsdw:
> Looks good to me! I think one adjustment should be made to never_unpin_new_block_before_finalized to make it pass (drop(_i0) the initialized event)

I had a general pass over this test just now to clarify it all a little so hopefully it makes sense now :)

> Just to double check I got this right, we store an extra bit of information in the pinning details, the can_be_unpinned.
We then set can_be_unpinned = true for Finalized and Pruned events. Then use this flag in combination with the UnpinFlags to make a call to chainHead_unpin.

Yup; so we should only _actually_ unpin blocks that we've flagged we are no longer using (ie in the `UnpinFlags` set) _and_ that are marked as `can_be_unpinned` (because we don't expect them to occur meaningfully in any future events now (the exception is that they can come up as "parent hashes" of blocks, but it'd be a pain to handle that properly so we haven't bothered).

> NewBlockRef kept around

It should be yup (there's a test for this)

> FinalizedBlockRef (can_be_unpinned = true) and block ref dropped immediately -> hash is not added to the UnpinFlags yet

I think this will be ok; pinned but finalized blocks should now have `can_be_unpinned: true` set, and so when they are eventually dropped, then when the next finalization event occurs we'll see that they can be properly unpinned now and tell the backend to unpin.

> NewBlockRef is used as normal; then dropped and this time is added to the UnpinFlags

These blocks should now continue to be kept pinned until they are finalized or pruned, at which point they will now be unpinned on the next finalized event after that.

subxt/src/backend/unstable/follow_stream.rs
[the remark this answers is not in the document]
niklasad1: might be better to rename it to finalized because `pruned` is also block numbers ^^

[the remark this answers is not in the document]
jsdw: Yeahh, I'll make it somehting clearer now :)

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -271,11 +282,31 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
...
     /// been sent though, then the block will be unpinned.
     fn pin_block_at(&mut self, rel_block_num: usize, hash: Hash) -> BlockRef<Hash> {
+        self.pin_block_at_setting_unpinnable_flag(rel_block_num, hash, false)
+    }
+
+    /// Pin a block, or return the reference to an already-pinned block.
+    ///
+    /// This is the same as [`Self::pin_block_at`], except that it also marks the block as being unpinnable now,
+    /// which should be done for any block that will no longer be seen in future events.
+    fn pin_unpinnable_block_at(&mut self, rel_block_num: usize, hash: Hash) -> BlockRef<Hash> {
```
niklasad1, line 292: I'm not a fan of the naming `pin_unpinnable_block_at` it just confuses me but I don't have any good suggestion for it than `pin_too_old_block`

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -271,11 +282,31 @@ impl<Hash: BlockHash> FollowStreamUnpin<Hash> {
...
+        self.pin_block_at_setting_unpinnable_flag(rel_block_num, hash, true)
+    }
+
+    fn pin_block_at_setting_unpinnable_flag(
+        &mut self,
+        rel_block_num: usize,
+        hash: Hash,
+        can_be_unpinned: bool,
```
niklasad1, line 300: I would prefer an enum here called BlockStatus such as:

```rust
enum BlockStatus {
  /// The block has been seen in a finalized block or a pruned block, it's ok to be unpinned and future events are of no interest
  Processed
  /// Waiting for the block to be finalized or pruned.
  Pending
}
```

niklasad1 (approved):
The logic looks good to me but not sure what's going on in the CI/tests..

subxt/src/backend/unstable/follow_stream_unpin.rs
jsdw: Yeah; I couldn't think of a good name either! I'm trying to say "pin a block that can be unpinned" versus "pin a block that can't yet be unpinned"

subxt/src/backend/unstable/follow_stream_unpin.rs
jsdw: I guess I feel like the boolean is clear enough here (can the block be unpinned; true or false), and then the functions you call to pin a block hide it anyway (but their name is another issue :D)

jsdw: I had a quick go at adding an enum and it felt messier to me offhand so I left it as a bool for now :)

PR #1372, opened 2024-01-17
lightclient(fix): Ensure lightclient chainSpec is at least one block old

### LINKED ISSUE #1562: GrandPa warp sync times out on github CI
We have recently encountered a few subxt github CI timeouts with the light client.
GrandPa warp sync seems to be idle at the same block until the CI times out our testing.
This is happening in a flaky manner and could be reproduced by https://github.com/paritytech/subxt/pull/1346. 

Smoldot version is 0.15 and smoldot-light is 0.13 (from [Cargo.toml](https://github.com/paritytech/subxt/blob/c5948dc967e0e56a403de807bc104f0129c2ff2e/Cargo.toml#L104C1-L105C65)).
The test connects the smoldot client to the polkadot network from a freshly downloaded chainSpec; then performs a few subscriptions and method requests. 

Logs reported by the [subxt](https://github.com/paritytech/subxt/blob/745403ca29b0b4a1aac0360ad69d9b21962f7b08/testing/integration-tests/src/light_client/mod.rs#L147) test:

```bash
2024-01-08T13:22:48.947771Z  WARN json-rpc-polkadot: The JSON-RPC client has just called a JSON-RPC function from the legacy JSON-RPC API (chain_getFinalizedHead). Legacy JSON-RPC functions have loose semantics and cannot be properly implemented on a light client. You are encouraged to use the new JSON-RPC API <https://github.com/paritytech/json-rpc-interface-spec/> instead. The legacy JSON-RPC API functions will be deprecated and removed in the distant future.    
...

2024-01-08T13:22:58.947182Z  WARN sync-service-polkadot: GrandPa warp sync idle at block #18952945 (0x0xfde5…4433)    
...

2024-01-08T13:25:38.948843Z  WARN sync-service-polkadot: GrandPa warp sync idle at block #18952945 (0x0xfde5…4433)    
...

2024-01-08T13:37:58.956476Z  WARN sync-service-polkadot: GrandPa warp sync idle at block #18952945 (0x0xfde5…4433)    
 Error: The operation was canceled.
```

Logs extracted from: https://github.com/paritytech/subxt/actions/runs/7447975422/job/20261334692?pr=1346

Let me know if I could provide more details here, thanks! 🙏

lexnv:
This PR ensures that the provided chainSpec provided to the lightClient is at least one block older than the head of the chain.

Otherwise, the CI has detected that the lightClient does not make progress to sync to the head of the chain.

This should be reverted once we update smoldot with this fix: https://github.com/smol-dot/smoldot/issues/1562


jsdw:
I do wonder whether we may as well just do a similar thing here instead:

https://github.com/paritytech/subxt/blob/553343548e5680ac5353185bcb6970aa164da2ea/subxt/src/client/light_client/builder.rs#L250

Since whatever we do, we'll need it to work in general, and the light client stuff is still behind a feature flag anyways :)

subxt/src/client/light_client/builder.rs
```diff
@@ -251,10 +251,17 @@ async fn fetch_url(url: impl AsRef<str>) -> Result<serde_json::Value, Error> {
...
-    client
+    let result = client
         .request("sync_state_genSyncSpec", jsonrpsee::rpc_params![true])
         .await
-        .map_err(|err| Error::Rpc(crate::error::RpcError::ClientError(Box::new(err))))
+        .map_err(|err| Error::Rpc(crate::error::RpcError::ClientError(Box::new(err))));
+
+    // Sleep 8 seconds to ensure that a new block is produced for the substrate test node.
+    // Although the block production should be 6 seconds, the CI environment is sometimes slow.
+    // This is a temporary workaround for: https://github.com/smol-dot/smoldot/issues/1562.
```
jsdw, line 261: This is good for now.

One thing to keep in mind is that parachains have eg a 12 second block time atm, so in the general case we should probably do something like download chain spec, and then subscribe to finalized blocks and wait for the next one or something like that, to make this nice and robust to block times or slow CI or whatever!

jsdw: (it may also be that if we do this, we don't need tokio/time any more?)

subxt/Cargo.toml
```diff
@@ -51,7 +51,7 @@ unstable-metadata = []
 
 # Activate this to expose the Light Client functionality.
 # Note that this feature is experimental and things may break or not work as expected.
-unstable-light-client = ["subxt-lightclient", "tokio-stream"]
+unstable-light-client = ["subxt-lightclient", "tokio-stream", "tokio/time"]
```
jsdw, line 54: I guess we can revert this and the optional `tokio` dep below now? :)

jsdw (approved):
Awesome!

niklasad1 (approved):
:pray:

Issue #1366, opened 2024-01-15
CI: Reduce our monthly minutes usage


jsdw:
We need to reduce our monthly CI usage so that we don't exceed an org imposed (I think) cap on monthly CI minutes. Currently, Subxt is the worst offender.

The 45 minute timeout added to our tests should already help with this a lot (and perhaps we can further reduce this to say 30mint; we'd have to see! We don't want potentially successful jobs timing out).

Further, once we fix the unstable backend test timeouts (@lexnv is working on this; my hope is we can merge a fix in the next couple of days) we should reduce our usage some more.

Next, let's see whether using the faster `ubuntu-8-core` etc runners (which we can assume use a multiple of minutes depending on how many multiples of 4 cores they are, eg 8 core would be 2x minutes) actually are worth it where used, and revert to slower machines if the total time is not decreased significantly. We need to find a good balance here between test speed and minute use.

Perhaps we could also optimise our CI pipeline so that we run fast checks first (fmt and clippy for instance, but check the times these take to decide) in one block, and then the rest in another block (thanks for the idea @tadeohepperle). This will hopefully catch the quite frequent/basic fmt/clippy fails and prevent excess minutes being used until they are all fixed up, only running the long tests when the basics are all good. 

Related issues:
- https://github.com/paritytech/ci_cd/issues/919 for subxt over usage
- https://github.com/paritytech/ci_cd/issues/553 for monthly minutes showing subxt doing badly

jsdw:
For reference, per minute rats for runners are as follows (https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates):

Operating system |	vCPUs |	Per-minute rate (USD)
---|---|---
Linux  |	2	 |  $0.008
Linux  |	4	 |  $0.016
Linux  |	8	 |  $0.032
Linux  |	16	 |  $0.064
Linux  |	32	 |  $0.128
Linux  |	64	 |  $0.256
Windows  |	2	 |  $0.016
Windows  |	8	 |  $0.064
Windows  |	16	 |  $0.128
Windows  |	32	 |  $0.256
Windows  |	64	 |  $0.512
macOS  |	3 or 4	 |  $0.08
macOS  |	12	 |  $0.12
macOS  |	6 (M1)	 |  $0.16

lexnv:
We could further reduce the CI time by sharing a substrate-node across tests. Although we should consider this at a later time since the other proposed ideas are easier to implement.

[...]

PR #1373, opened 2024-01-17
ci: Reduce the light client timeout to 15 minutes

lexnv:
The light-client CI time is reduced from 30 minutes to 15 minutes.

This is part of an initiative to reduce the CI minutes consumed by subxt (related to: https://github.com/paritytech/subxt/issues/1366).

The light-client has been recently patched by: https://github.com/paritytech/subxt/pull/1372. Without this fix, the light-client tests would fail to sync to the head of the chain, consuming all minutes (30) until the timeout was hit.

The following datapoints show the light-client only needs around 5 minutes:
- https://github.com/paritytech/subxt/actions/runs/7555940911/job/20571947269
- https://github.com/paritytech/subxt/actions/runs/7554871260/job/20568625374
- https://github.com/paritytech/subxt/actions/runs/7540594780/job/20525526129

To stay on the safe side and not have the test fail due to CI overloading, 15 minutes should be enough time.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -218,7 +218,7 @@ jobs:
 
   light_client_tests:
     name: "Test (Light Client)"
-    runs-on: ubuntu-latest-16-cores
+    runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
       - name: Checkout sources
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -219,7 +219,7 @@ jobs:
   light_client_tests:
     name: "Test (Light Client)"
     runs-on: ubuntu-latest-16-cores
-    timeout-minutes: 30
+    timeout-minutes: 15
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

.github/workflows/rust.yml
```diff
@@ -219,7 +219,7 @@ jobs:
   light_client_tests:
     name: "Test (Light Client)"
     runs-on: ubuntu-latest-16-cores
```
niklasad1, line 221: I wonder if we really need such beefy machine for these tests ^^

lexnv: Good point, let me play around with these settings 👍

niklasad1: I think it's probably compiling the tests that are slow when I tried it locally.

jsdw (approved):

jsdw:
Cool; happy to merge and see what happens!

We might want to reduce the machine spec for other tests too!

PR #1374, opened 2024-01-17
Add logging to submit_transaction and unstable driver

jsdw:
- Bail and log some metrics if submit_transaction takes way too long or we see an unexpected "None".
- Try to ensure logging from the unstable backend driver if an error is emitted there during tests.
- A small thing I spotted: allow "unpin" futures to complete even if underlying stream ends (this shouldn't really be an issue though, since the underlying stream ending prob means there's no point, but may as well).
- ~~Only kick off test jobs if the quick `fmt` (and machete) checks pass. Hopefully this helps a little because we always forget `fmt`, and those ones are fast (clippy can be pretty slow so I left that off)~~


subxt/src/backend/unstable/mod.rs
```diff
@@ -480,24 +480,47 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
...
-                    match seen_block {
-                        FollowEvent::Initialized(ev) => {
-                            // Just in case this stream is really slow to start or something..
-                            seen_blocks.insert(
-                                ev.finalized_block_hash.hash(),
-                                (SeenBlockMarker::Finalized, ev.finalized_block_hash),
-                            );
-                        }
+                if let Poll::Ready(Some(follow_event)) = seen_blocks_sub.poll_next_unpin(cx) {
+                    last_seen_follow_event_str = Some(format!("{follow_event:?}"));
```
lexnv, line 521: nit: It could be possible by formatting to introduce a delay here which would make the issue not reproducible

jsdw: True, but saving the event without stringifying it would also involve a clone and would change the behavioru (ie keep something pinned that might otherwise be dropped) so stringifying felt best. We could just not save the events, but I wanted to see if they revealed anything :)

lexnv: I do remember trying a similar thing, but can't remember if I managed to reproduce the issue or not. What I did in the end was to save just the block::hash() with a plain `str` ("newblock", "finalized")

jsdw: I'd rather keep it simple I guess for now and just log whatever the event it; I suspect that the issue will crop up again; outputting to astderr or whatever is pricier but I doubt stringifying will make much difference offhand!

subxt/src/backend/unstable/follow_stream_driver.rs
jsdw, line 1: This is just a little tidy up; no logic changes

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -74,20 +74,26 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
...
             };
 
-            // No more progress to be made if inner stream done.
             let Some(ev) = ev else {
-                return Poll::Ready(None);
+                // if the stream is done, but `unpin_futs` are still pending, then
+                // return pending here so that they are still driven to completion.
+                // Else, return `Ready(None)` to signal nothing left to do.
+                return match unpin_futs_are_pending {
+                    true => Poll::Pending,
```
jsdw, line 94: Here we just return pending if there are still unpin futures to resolve, even if the main stream is done.

Shouldn't really make a difference because we know that if the main stream is done, it'll have stopped and no more things will be pinned, but it felt better to not cancel the unpin futs midway just in case.

subxt/Cargo.toml
```diff
@@ -33,7 +33,14 @@ native = [
...
 # Enable this for web/wasm builds.
 # Exactly 1 of "web" and "native" is expected.
-web = ["jsonrpsee?/async-wasm-client", "jsonrpsee?/client-web-transport", "getrandom/js", "subxt-lightclient?/web", "subxt-macro/web"]
+web = [
+    "jsonrpsee?/async-wasm-client",
+    "jsonrpsee?/client-web-transport",
+    "getrandom/js",
+    "subxt-lightclient?/web",
+    "subxt-macro/web",
+    "instant/wasm-bindgen"
```
tadeohepperle, line 42: Very nice to have this for WASM

tadeohepperle (approved):
Great to have this, let's see if the CI gives us more insights into the failures.

lexnv (approved):
Nice! 👍

PR #1357, opened 2024-01-11
Improve Signed Extension and Block Decoding Examples/Book

tadeohepperle:
Fixes #1310

The PR introduces a new example `polkadot_monitor.rs` that listens to the live polkadot network to monitor transfer transactions. I wrote more about signed extension static decoding and extrinsic decoding in the book and showcase these things in the added example.

Fixes #1314

We split some parts of the `setup_config_signed_extension.rs` example into a new example `setup_config_assethub.rs` that makes an effort to explain in a clearer way how to use a type generated from the metadata `MultiLocation` as part of the Subxt Config and how tips can be added to an extrinsic using  `MultiLocation` values. The book has been adjusted.

PR #1376, opened 2024-01-18
Return `ExtrinsicDetails` alongside decoded static extrinsics

### LINKED ISSUE #1359: Access to Signed Extensions and Events when looking for a StaticExtrinsic
## Problem

The `Extrinsics<T: Config, C: Client>` struct exposes a lot of functionalities that come in handy to learn more about an extrinsic. But it lacks a statically decoded representation of a specific Extrinsic type. Right now, you do not have  a nice way of e.g. finding all `TransferKeepAlive` extrinsics in a block together with their events and signed extensions. 

### Status Quo

If you have the `Extrinsics` of a block, you can find StaticExtrinsics with `extrinsics.find<E: StaticExtrinsic>()` returning an iterator over elements of the type E. But by doing this, you lose the access to the information stored in the (untyped) ExtrinsicDetails: signed extensions and events.

To get e.g. all `TransferKeepAlive` extrinsics in a block with their events and signed extensions, you would have to iterate over all extrinsics and then return a tuple for all the ones that are a `TransferKeepAlive`. This tuple would have the statically decoded transfer + the untyped ExtrinsicDetails:

```
extrinsics.iter().filter_map(|e| {
     if e.is_err(){
          return None;
     }
    let e = e.unwrap();
    match e.as_extrinsic::<TransferKeepAlive>(){
         Err(e) => None,
         Ok(transfer) => Some((transfer, e)) // the tuple contains all the information we need
    }
})
```

This is a bit cumbersome and could lead people to not use the utility functions like `Extrinsics::find<E: StaticExtrinsic>()`;

## Proposal

We could introduce a new type `StaticExtrinsicDetails` and use it as a return value whenever we try to find/decode static extrinsics.
```
struct StaticExtrinsicDetails<E: StaticExtrinsic>{
    details: ExtrinsicDetails,
    decoded: StaticExtrinsic,
}
```

Alternative:

We could also just extend `ExtrinsicDetails` by another type parameter that is `()` by default:

```
pub struct ExtrinsicDetails<T: Config, C, E = ()> {
    decoded: E // a StaticExtrinsic or () or DecodeAsType
    bytes: Arc<[u8]>,
    pallet_index: u8,
    variant_index: u8,
    ...
}
```

Then we could do:
```
impl<T: Config, C: OfflineClientT<T>> for ExtrinsicDetails<T, C, ()>  { // note: impl only for unit type 
    pub fn as_extrinsic<E: StaticExtrinsic>(&self) -> Result<Option<ExtrinsicDetails<T,C,E>>, Error> {
         // static decoding
    }

    pub fn as_root_extrinsic<E: DecodeAsType>(&self) -> Result<ExtrinsicDetails<T,C,E>>, Error> {
         // decoding
    }
}
```

tadeohepperle:
fixes #1359

We introduce a new wrapper type around a static extrinsic of type`E` that contains details about the extrinsic (e.g. signature, signed extensions, original bytes, metadata...)

```rs
pub struct FoundExtrinsic<T: Client, C, E> {
    pub details: ExtrinsicDetails,
    pub value: E
}
```


lexnv (approved):

jsdw (approved):
This looks awesome!

I'm surprised no book/examples needed changing; something to look at quickly maybe just to see that the example docs etc still make sense (clearly none were actually using the decoded extrinsic since no errors), but can be done separately!

tadeohepperle:
> I'm surprised no book/examples needed changing

Not anymore! Since I just merged [the PR with the additional examples](https://github.com/paritytech/subxt/pull/1357) to master, this PR now also needed an adjustment :)

jsdw:
Nice!

PR #1378, opened 2024-01-18
Return Pending rather than loop around if no new finalized hash in submit_transaction

jsdw:
I also brazenly decided to remove the debug logging (though kept the timeout), so that it can run a bit faster again. I'm hoping that it won't pose an issue any more, but ifg it does then we can add more logging back again :)

I had a go at reworkign the code a little to simplify it, but it turned out about the same in the end! I thought about a more state-machine like approach, but it would become very verbose or end up being split across multiple functions etc, so I didn't for now.

### DIFF
--- a/subxt/src/backend/unstable/mod.rs
+++ b/subxt/src/backend/unstable/mod.rs
@@ -480,61 +480,43 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
         // with chainHead_follow.
         let mut finalized_hash: Option<T::Hash> = None;
 
-        // Monitor usage to help track down an obscure issue.
+        // Record the start time so that we can time out if things appear to take too long.
         let start_instant = instant::Instant::now();
-        let mut last_seen_follow_event_str = None;
-        let mut last_seen_follow_event_time = None;
-        let mut last_seen_tx_event_str = None;
-        let mut last_seen_tx_event_time = None;
-        let mut num_polls: usize = 0;
-        let mut num_loops: usize = 0;
 
-        macro_rules! bail_with_stats {
-            ($msg:literal) => {{
-                let msg = $msg;
-                let err = format!(
-                    "submit_transaction error: {msg}. Please raise an issue. Details: \
-                     start_time={start_instant:?}, \
-                     num_polls={num_polls}, \
-                     num_loops={num_loops}, \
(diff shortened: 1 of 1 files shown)


lexnv (approved):

subxt/src/backend/unstable/mod.rs
```diff
@@ -480,61 +480,43 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
...
+                    Poll::Ready(None) => {
+                        return Poll::Ready(err_other("chainHead_follow stream ended unexpectedly"))
+                    }
+                    Poll::Ready(Some(follow_ev)) => Poll::Ready(follow_ev),
+                    Poll::Pending => Poll::Pending,
+                };
+                let follow_ev_is_pending = follow_ev_poll.is_pending();
+
+                // If there was a follow event, then handle it and loop around to see if there are more.
+                // We want to buffer follow events until we hit Pending, so that we are as uptodate as possible
```
niklasad1, line 515:
```suggestion
                // We want to buffer follow events until we hit Pending, so that we are as up-to-date as possible
```

niklasad1 (approved):

subxt/src/backend/unstable/mod.rs
```diff
@@ -577,29 +559,27 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
                         };
                         return Poll::Ready(Some(Ok(ev)));
                     } else {
-                        // Keep waiting for more finalized blocks until we find it (get rid of any other block refs
-                        // now, since none of them were what we were looking for anyway).
+                        // Not found it! If follow ev is pending, then return pending here and wait for
+                        // a new one to come in, else loop around and see if we get another one immediately.
```
lexnv, line 563: nit: We could probably remove the else blocks to get rid of some extra indentation; up to you 👍

jsdw: I did actually think abotu this, but you'd end up with a bit much for my liking in the new `else` thing eg

```rust
let Some((SeenBlockMarker::Finalized, block_ref)) = seen_blocks.remove(hash) else {
    seen_blocks.clear();
    if follow_ev_is_pending {
        return Poll::Pending;
    } else {
        continue;
    }
};

// ..
```

Unless you had another idea? :)

PR #1379, opened 2024-01-19
CLI tool: recursive derives and attributes

### LINKED ISSUE #1353: CLI tool codgen command: How to apply recursive derives and attributes?
Alex raised this question [here](https://github.com/paritytech/subxt/pull/1260#discussion_r1445935201) and after [the PR](https://github.com/paritytech/subxt/pull/1260) is merged, we need to decide how to implement recursive derives and attributes in the CLI tool. 

Right now, you specify derives and attributes like this:

```
subxt codegen --derive-for-type my_module::my_type=serde::Serialize --attributes-for-type my_module::my_type=#[allow(clippy::all)]
```

There are these options in my opinion (in order of my preference):
- A: Add new arguments `--derive-for-type-recursive` and `--attributes-for-type-recursive`.
- B: Make all derives recursive by default: This would make sense in most cases, but maybe not in all. 
- C: Introduce a new input format per derive, e.g. `subxt codegen --derive-for-type my_module::my_type=serde::Serialize/recursive` so we just adjust the `derive_for_type_parser` function to check for this flag. This requires some weird syntax though, that the user has to know about.
- D: Have no recursive derives in the CLI. People need to specify the derives manually (status quo)

I think A would be best, what do you guys think? @niklasad1 @lexnv @jsdw

tadeohepperle:
Fixes #1353

This PR introduces an optional appendix of comma separated values to the CLI arguments `derive-for-type` and `attributes-for-type`. Right now, the appendix arguments are only checked for containing `"recursive"`, but we could add more options in the future.

I verified that these commands work with and without the recursive flag:
```
subxt codegen --derive-for-type xcm::v2::multilocation::MultiLocation=Clone,recursive
subxt codegen --attributes-for-type "xcm::v2::multilocation::MultiLocation=#[foofoo],recursive"
```


jsdw:
Looks good to me, but it's branched off #1290 I think so hard to tell if I'm missing anything right now.

Probably best to avoid doing too much on top of large PRs until they merge, because it becomes a pain all of a sudden if said PRs have comments that need addressing, though in this case it's fairly small :)

jsdw (approved):
Looks great!

`str_equals_str_comma_maybe_recursive_parser` is a strange name, but I don't have any great suggestions for it!

cli/src/commands/codegen.rs
```diff
@@ -67,20 +69,52 @@ pub struct Opts {
...
+    .ok_or_else(|| String::from("Invalid pattern for `attributes-for-type`. It should be `type=attribute` like `my_type=serde::#[allow(clippy::all)]` or `type=attribute,recursive` like `my_type=serde::#[allow(clippy::all)], recursive`"))?;
+    Ok(AttributeForType {
+        type_path,
+        attribute,
+        recursive,
+    })
+}
+
+/// Parses a `&str` of the form `str1=str2` into `(str1, str2, false)` or `str1=str2,recursive` into `(str1, str2, true)`.
+fn str_equals_str_comma_maybe_recursive_parser(src: &str) -> Option<(String, String, bool)> {
```
lexnv, line 107: We could probably have a few unit-tests for this. Would `str1=str2,a,b,c,recursive` also be a valid syntax?

I'd also probably rename this to `type_parser` or something similar

lexnv (approved):
Nice one! 👍 

Would be good to have a few test scenarios for `str_equals_str_comma_maybe_recursive_parser` to make sure we only capture the intended form, overall looks good!

cli/src/commands/codegen.rs
```diff
@@ -213,3 +249,23 @@ fn codegen(
...
+        assert_eq!(type_map_parser("Foo=Bar"), Some(("Foo", "Bar", false)));
+        assert_eq!(
+            type_map_parser("Foo=Bar,recursive"),
+            Some(("Foo", "Bar", true))
+        );
+        assert_eq!(type_map_parser("Foo=Bar,a"), Some(("Foo", "Bar", false)));
+        assert_eq!(
+            type_map_parser("Foo=Bar,a,b,c,recursive"),
+            Some(("Foo", "Bar", true))
+        );
```
lexnv, line 269: Should we reject this format? 🤔 

Maybe we could also check that we expect this function to panic on these inputs since they are invalid

tadeohepperle: Good point, let's reject it. To keep the possibility open to add other attributes (similar to `recursive`) in the future I implemented a simple match over the trailing attributes:
```rs
let mut recursive = false;
for r in split_rest {
    match r {
        "recursive" => {
            recursive = true;
        }
        _ => return None,
    }
}
```

PR #1397, opened 2024-01-25
CLI: Return error on wrongly specified type paths

### LINKED ISSUE #1380: CLI tool: Error on misspecified derives, attributes and substitutes
#1339 Introduces better error messages for the macro, pointing out misspecified type paths, that do not exist in the metadata.
We should give the user feedback similar to this when using custom derives, attributes and substitutes in the CLI tool.

tadeohepperle:
Fixes #1380

The CLI tool now reports misspecified type paths for the `codegen` attributes `derive-for-type`, `attributes-for-type` and `substitute-type`. In case another type with the same name is found in the metadata the user is provided with that information.

image

### DIFF
--- a/cli/src/commands/codegen.rs
+++ b/cli/src/commands/codegen.rs
@@ -363,7 +363,7 @@ mod tests {
         ];
         for case in invalid_cases.iter() {
             let output = run(case).await;
-            // asset that we make suggestions pointing the user to the valid type
+            // assert that we make suggestions pointing the user to the valid type
             assert!(output.unwrap_err().to_string().contains(valid_type));
         }
     }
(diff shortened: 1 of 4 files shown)


lexnv (approved):
Nice 👍

cli/src/commands/codegen.rs
```diff
@@ -276,4 +326,45 @@ mod tests {
...
+        }
+
+        let invalid_cases = [
+            format!("--derive-for-type {invalid_type}=serde::Serialize"),
+            format!("--attributes-for-type {invalid_type}=#[allow(clippy::all)]"),
+            format!("--substitute-type {invalid_type}=my_module::MultiAddress"),
+        ];
+        for case in invalid_cases.iter() {
+            let output = run(case).await;
+            // asset that we make suggestions pointing the user to the valid type
```
niklasad1, line 366:
```suggestion
            // assert that we make suggestions pointing the user to the valid type
```

niklasad1 (approved):
nice, LGTM

PR #1400, opened 2024-01-26
chore(deps): Update smoldot to the latest version

lexnv:
This PR updates smoldot to 0.17 and smoldot-light to 0.15.

Changes to repo:
- AddedChain now receives a generic PlatformRef trait to accommodate smoldot changes

Pending fix from:
- https://github.com/smol-dot/smoldot/issues/1638

### Status Update 8 July

- Smoldot is now producing blocks with the legacy RPCS
- The latest version has a regression in syncing from 30seconds to 175seconds
- This means we are producing blocks, but the distance between the first finalized block (the block presented in the chainSpec) and the second finalized block is now 175seconds (until grandpa warp sync completes)

Pending fix:
- https://github.com/smol-dot/smoldot/issues/1896


niklasad1 (approved):

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -93,42 +126,61 @@ async fn storage_plain_lookup(api: &Client) -> Result<(), subxt::Error> {
...
+    // non_finalized_headers_subscription(&api).await?;
     finalized_headers_subscription(&api).await?;
-    runtime_api_call(&api).await?;
-    storage_plain_lookup(&api).await?;
-    dynamic_constant_query(&api).await?;
-    dynamic_events(&api).await?;
+    // runtime_api_call(&api).await?;
+    // storage_plain_lookup(&api).await?;
+    // dynamic_constant_query(&api).await?;
+    // dynamic_events(&api).await?;
```
jsdw, line 183: Should these still be commented out?

lexnv: Yep, I've added a few debug logs in https://github.com/paritytech/subxt/pull/1400/commits/f14bf9a47457de79475eb683962a0c8d3f6c3681.
This is to help reproduce the issue from smoldot faster https://github.com/smol-dot/smoldot/issues/1638. 

I suggest putting this PR on pending until another smoldot version is published on crates io :D

jsdw: Sounds like a plan :)

lightclient/src/background.rs
```diff
@@ -242,6 +248,12 @@ struct ActiveSubscription {
     unsubscribe_method: String,
 }
 
+fn trim_response(response: &str) -> &str {
+    const MAX_RESPONSE_SIZE: usize = 256;
+    let len = std::cmp::min(response.len(), MAX_RESPONSE_SIZE);
+    &response[..len]
```
niklasad1, line 254: nit, you could actually index into non-char boundary here because not all chars are ascii which leads to a panic..

https://github.com/paritytech/jsonrpsee/blob/master/core/src/tracing.rs#L101-#L124

jsdw: There are a couple of nightly only methods that would be perfect here, but in the absense of them one could do:

```rust
while !response.is_char_boundary(len) {
    len -= 1;
}
```

To find the appropriate place to slice a message that exceeds the length

lightclient/src/background.rs
```diff
@@ -176,10 +177,15 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
                         tracing::trace!(target: LOG_TARGET, "Smoldot RPC responses channel closed");
                         break;
                     };
+                    let (maybe_truncated, delim) = if back_message.len() > MAX_LOG_SIZE {
+                        (&back_message[0..MAX_LOG_SIZE], "...")
+                    } else {
+                        (&back_message[..], "")
+                    };
```
jsdw, line 184: Do we need to worry about invalid UTF boundaries here too?

(perhaps a little helper fn which trims a message to keep it under some max length and uses `is_char_boundary` internally would be ideal?)

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -34,54 +34,93 @@ use subxt_metadata::Metadata;
 
 type Client = OnlineClient<PolkadotConfig>;
 
+/// The Polkadot chainspec.
+const POLKADOT_SPEC: &str = include_str!("../../../../artifacts/demo_chain_specs/polkadot.json");
+
+/// Use the unstable backend for testing.
+const USE_UNSTABLE_BACKEND: bool = true;
```
jsdw, line 41: I guess it makes sense to test the light client with just the "unstable" backend, though we could have that light client test be a fn that takes in this bool and then run it twice to test both backends until the unstable one is stabilised? (and then remove this const)

jsdw (approved):
Couple of small nits but looks great to me; thanks for doing this!

lightclient/src/background.rs
lexnv: I've ended up with something similar to Niklas's suggestion, I found that a bit easier to follow :D

Cargo.toml
```diff
@@ -120,9 +120,11 @@ sc-executor = "0.40.0"
 sc-executor-common = "0.35.0"
 
 # Light client support:
-smoldot = { version = "0.16.0", default-features = false }
-smoldot-light = { version = "0.14.0", default-features = false }
+smoldot = { version = "0.18.0", default-features = false }
+smoldot-light = { version = "0.16.2", default-features = false }
+
```
niklasad1, line 125:
```suggestion
```

Cargo.toml
```diff
@@ -120,9 +120,11 @@ sc-executor = "0.40.0"
...
 tokio-stream = "0.1.15"
+
```
niklasad1, line 127:
```suggestion
```

niklasad1 (approved):
:rocket:

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -29,59 +29,100 @@
 
 use crate::utils::node_runtime;
 use codec::Compact;
+use futures::StreamExt;
+use std::sync::Arc;
+use subxt::backend::rpc::RpcClient;
+use subxt::backend::unstable::UnstableBackend;
+use subxt::OnlineClient;
```
niklasad1, line 36:
```suggestion
```

testing/integration-tests/src/light_client/mod.rs
```diff
@@ -129,5 +216,24 @@ async fn light_client_testing() -> Result<(), subxt::Error> {
     dynamic_constant_query(&api).await?;
     dynamic_events(&api).await?;
 
+    tracing::trace!("Light complete testing took {:?}", now.elapsed());
```
niklasad1, line 219:
```suggestion
    tracing::trace!("Light complete testing took {:?}", now.elapsed());
    Ok(())
```

PR #1408, opened 2024-01-29
LightClient: Unsubscribe from subscriptions

### LINKED ISSUE #1402: LightClient: Unsubscribe from subscriptions to avoid warnings
The unstable-light-client feature does not unsubscribe from subscriptions.
This behavior leads to subscriptions living forever.

In the context of the `chainHead` this could be detrimental and could lead in the future to users not being able to subscribe anymore. This is because the `chainHead` spec states that at least 2 subscriptions should be supported by the server per connection.

This has been tested locally with the: #1400 

```bash
test light_client::light_client_testing has been running for over 60 seconds
2024-01-26T13:44:24.492595Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:30.816682Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:36.269689Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:42.308706Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:48.451372Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:48.616638Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:44:54.568271Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
2024-01-26T13:45:00.501182Z  WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
```

In the following code, the RPC unsubscribe method name is ignored:
https://github.com/paritytech/subxt/blob/4408a8c5872c9745661a190c3aabb9477f56c0ae/subxt/src/client/light_client/rpc.rs#L105-L109

To mitigate this extend the background protocol to utilize the unsubscribe method.
Whenever the user drops the subscription in the frontend, the backend should initiate a call to the `unsubscribe_method`.
Because the request to the `unsubscribe_method` does not have a frontend user associated with; its response should be logged and dropped.

lexnv:
Previous to the PR, subscribing to a method would never call unsubscribe from that method, even when the user dropped the subscription.

This PR ensures that the appropriate `unsubscribe` method is being called to avoid consuming unnecessary memory, overloading the smoldot instance by never-ending subscriptions, and unwanted side-effects as imposing a maximum limit of subscriptions after which they are rejected.

To achieve this:
- the background <-> foreground protocol of subxt is extended to capture the unsubscribe method
- the unsubscribe method is passed down from the RPC layer to the background task for tracking
- the unsubscribe call is made when the user dropped the foreground subscription and smoldot produced a response for the said subscription

Running the subxt test suite for the light-client produced multiple warnings;

```bash
 WARN subxt-light-client-background: Subscription response id=1 chain=ChainId(0) is not tracked
```

With this PR, the smoldot acknowledges the unsubscribe method and stops the subscription:

```bash
DEBUG json-rpc-polkadot: json-rpc-request-queued; request={"jsonrpc":"2.0","id":"11", "method":"chain_unsubscribeNewHeads","params":["1"]}
```

Closes: https://github.com/paritytech/subxt/issues/1402


tadeohepperle (approved):
LGTM, nice work! :+1:

lightclient/src/background.rs
```diff
@@ -69,18 +71,33 @@ pub struct BackgroundTask<TPlatform: PlatformRef, TChain> {
...
     ///
     /// The subscription ID is entirely generated by the node (smoldot). Therefore, it is
     /// possible for two distinct subscriptions of different chains to have the same subscription ID.
-    subscriptions: HashMap<(usize, smoldot_light::ChainId), mpsc::UnboundedSender<Box<RawValue>>>,
+    subscriptions: HashMap<(usize, smoldot_light::ChainId), SubscriptionState>,
+}
+
+/// The state needed to resolve the subscription ID and send
+/// back the response to frontend.
+struct SubscriptionIdState {
```
niklasad1, line 84: I don't like the naming `SubscriptionIdState`, I would prefer `PendingSubscription` or something similar.

Also, would be nice to rename/remove and make a type `ActiveSubscription` that only can happen from the state `PendingSubscription -> ActiveSubscription`

niklasad1 (approved):

lightclient/src/background.rs
niklasad1: Ok, looks like `rejected subscription` are simply just passed to frontend and not inserted in the HashMap.

So, to me it looks like you can merge `SubscriptionState` and `SubscriptionStateId` to single struct and call it Subscription/ActiveSubscription

lightclient/src/background.rs
```diff
@@ -247,7 +274,7 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
                             "Cannot send method response to id={id} chain={chain_id:?}",
                         );
                     }
-                } else if let Some((sub_id_sender, sender)) =
+                } else if let Some(subscription_id_state) =
                     self.id_to_subscription.remove(&(id, chain_id))
```
jsdw, line 278: Nit (and not related to this PR): It feels more costly than necessary to remove the subscription from the map every time a message comes in and then put it back after (assuming success). I reckon it'd be more efficienct to loko up the subscription, but only remove it in case of an error (which we'd hope would be less common)

jsdw: Ah I see; this is what you do for actual subscription messages below, and this is just the inital response to the subscription if I am reading right, so it's only a one-time thing per subscription

lightclient/src/background.rs
```diff
@@ -287,22 +319,37 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
...
+                    // Nothing else to do, user is informed about the notification.
+                    return;
+                }
 
-                        // Remove the sender if the subscription dropped the receiver.
-                        self.subscriptions.remove(&(id, chain_id));
-                    }
-                } else {
+                // User dropped the receiver, unsubscribe from the method and remove internal tracking.
+                let Some(subscription_state) = self.subscriptions.remove(&(id, chain_id)) else {
```
jsdw, line 335: I'd have to check; is the _only_ kind of error that can be returned from the channel above one that indicates it was closed by the receiver?

jsdw (approved):
Looks good to me; good job spotting and fixing this!

lightclient/src/background.rs
lexnv: Yep, taken from https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedSender.html#method.send:

> If the receive half of the channel is [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html#method.close)d, either due to close being called or the [UnboundedReceiver](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html) having been dropped, this function returns an error. The error includes the value passed to send.

PR #1409, opened 2024-01-29
ci: Better job parallelism, cancel pipeline on failure and extra clippy checks

lexnv:
There are 3 main changes included in this PR:
- The CI pipeline is shuffled for better parallelism and faster build times
- Cancelling the pipeline if one of the CI steps fails (as tested by [this job](https://github.com/paritytech/subxt/actions/runs/7709721254/job/21011517365))
- Introduce better clippy for steps to account for `unstable-light-client` and `web` feature flags
  - As a result of this, the clippy warnings are also fixed

Change the CI parallelism for faster build times:

### Step 1
- fmt, machete

### Step 2
- clippy, check, wasm_check, docs

### Step 3
- all other tests

## Time Saved

We are saving an average of 10min per successful CI time. This is the time it takes the CI from starting the first job to finishing the last job, not the total amount of CI minutes used.

On average, the CI minutes have increased by 39sec because we introduced another clippy check for WASM.

```
Before:
    6s + 3m54s + max(5m50s, 1m34s) + max(16s, 5m50s) + max(9m27s, 9m24s, 6m2s, 6m10s) = 1507s (25.12 min)

After:
    max(18s, 15s) + max(5m24s, 39s, 6m3s, 1m25s, 5m53s) + max(9m43s, 9m21s, 6m7s, 6m36s) = 954s (15.9 min)
```

### Before

Taken from https://github.com/paritytech/subxt/actions/runs/7710633300.
before

### After

Taken from https://github.com/paritytech/subxt/actions/runs/7714660920. 

after


.github/workflows/rust.yml
```diff
@@ -45,11 +45,69 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
+      - if: failure()
```
niklasad1, line 48: can you explain what this does?

niklasad1 (approved):
any numbers how much it improves things?

.github/workflows/rust.yml
lexnv: The intent behind this is to terminate any jobs that are running in parallel if one of them fails.

For example, if clippy fails first then we should terminate the check, wasm_check and docs steps. Unfortunately, it seems that this can only be achieved by making an API call. Ive replaced this code with [cancel-action](https://github.com/andymckay/cancel-action) script.

jsdw: Looks much better with the small 2 liner step (and the linked script looks ok)! We may want to use a hash version for the script to lock us to the specific code (as has been done in some other places) just in case!

jsdw (approved):
Looks good to me now modulo the modified code to break clippy :)

sergejparity (approved):

.github/workflows/rust.yml
```diff
@@ -108,11 +108,29 @@ jobs:
...
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: -p subxt-lightclient --no-default-features --features web,unstable-light-client -- -D warnings
+
+      - name: Cargo clippy for web subxt with light-client and jsonrpsee
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: -p subxt --no-default-features --features web,unstable-light-client,jsonrpsee --target wasm32-unknown-unknown   -- -D warnings
```
jsdw: Potentially this could also be trimmed down by doing osmething like on https://github.com/paritytech/subxt/pull/1409/files/c83e0fcd133d27996a6b91ef3c9e7b46e91670e4..7320c27a0ad2f5c2f67fc499c8e14db40d20a145#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dL163

Issue #1394, opened 2024-01-23
Create `subxt-core` crate; a no_std subset of subxt functionality?


jsdw:
We'd like to make Subxt `no_std` compatible where possible. However, making parts of subxt support `no_std` as it stands is a bit tricky, and would lead to a bunch of `#[cfg(not(feature = "std"))]` type annotations in a bunch of places and a bunch of refactoring needed to break apart the bits that can and can't run on `no_std`.

As an alternative approach, how about we take the `no_std` compatible functionality out of Subxt and put it into a `subxt-core` crate which is no_std compatible and which `subxt` would then depend on. This would move Subxt more towards being a crate that just pulls together the core logic with the RPC stuff, generated code and higher level interface, allowing users to depend only on the lower level crates that they need instead if needed/preferred (this is in line also with our `desub` work, which will take out the logic for decoding storage and extrinsics and have Subxt depend on it from elsewhere).

I think we'll need to iterate on what the interface looks like a bit, trying out and refining something as we go.

Some thoughts on what we could put in this crate:

- Functionality for constructing and signing transactions (inc traits so that eg subxt-signer could depend on `subxt-core` and not `subxt` when it uses the `subxt` feature flag).
- Functionality for decoding blocks, extrinsics* and signed extensions.
- Decoding constants
- Decoding storage entries*, and constructing them
- Encoding runtime calls
- (essentially, almost anything that the `OfflineClient` can do?)
- All of the traits that enable us to encode/decode these things from eg `scale_value::Value` or codegen or whatever else (subxt can re-export them).

Other thoughts:

- Could rely on `subxt-metadata` directly to provide the metadata it needs to decode/encode things (assuming that can be made no-std too, which I think should be ok).
- `subxt-signer` could depend on `subxt-core` rather than `subxt` when using the "subxt" feature flag.
- On its own, the interfaces we provide here should be well documented and easy to use; just as important as `subxt` interfaces.
- The code generated by `subxt-codegen` could be made `no-std` and be used in conjunction with `subxt-core` to give a static interface for doing these things.
- We could re-export `subxt-core` from `subxt` as `subxt::core`, and then eg `subxt-macro` could possibly end up pointing to `subxt::core` by default but could also be redirected to point directly at `subxt_core`?
[...]

PR #1411, opened 2024-02-02
[WIP] Subxt-Core Crate

tadeohepperle:
relates to #1394

This PR does a few things:
- make `subxt-signer` no-std compatible
- create a new `subxt-core` crate that is no-std compatible by default and offers a subset of `subxt`'s functionality.
- use this `subxt-core` crate in `subxt-signer` and in `subxt`.

What is not working yet:
- The `critical_section` feature of the `once_cell` crate used in subxt-signer results in linker errors when run in `no-std`.
- Subxt compiles, but subxt examples are not all adjusted yet
- Interface and imports need to be discucced and made nicer, some messiness has to be removed
- Docs, especially book needs to be adjusted


tadeohepperle:
I am having trouble getting `no_std` testing working properly.
I setup a crate `testing/no-std-tests` that pulls in subxt-metadata, subxt-signer and `subxt-core`. You can try to build it with either of these:
```
cargo run 
cargo build --target thumbv7em-none-eabi 
cargo build --target aarch64-unknown-none
```
With `cargo run` (the native default target), we get linker errors. The curious thing is that the linker errors on my linux machine are different from the erros James is getting on a macos target.

With the `thumbv7em-none-eabi` and `aarch64-unknown-none` targets the problem is, that parts of the rust core prelude (e.g. `Some` and `None` variants, `Iterator` trait, etc. are not found by the compiler).

signer/src/ecdsa.rs
```diff
@@ -6,10 +6,22 @@
...
 use secrecy::ExposeSecret;
 
+struct GlobalSecp256K1Context;
+impl core::ops::Deref for GlobalSecp256K1Context {
+    type Target = Secp256k1<All>;
+
+    fn deref(&self) -> &Self::Target {
+        static ONCE: once_cell::sync::OnceCell<Secp256k1<All>> = once_cell::sync::OnceCell::new();
+        ONCE.get_or_init(|| Secp256k1::new())
+    }
```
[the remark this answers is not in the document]
jsdw: I'd copy what Substrate does here to be on the safe side, which is to do:

```rust
#[cfg(feature = "std")]
let context = SECP256K1;
#[cfg(not(feature = "std"))]
let context = Secp256k1::signing_only();
```

In the sign/from_seed fns to select which context to use ie https://github.com/paritytech/polkadot-sdk/blob/d54412ce8bef113c7a515b3a08851e7d77553df5/substrate/primitives/core/src/ecdsa.rs#L390 and https://github.com/paritytech/polkadot-sdk/blob/d54412ce8bef113c7a515b3a08851e7d77553df5/substrate/primitives/core/src/ecdsa.rs#L460

Since only two places this context is used anyway, I think that's fine :)

signer/src/ecdsa.rs
```diff
@@ -91,8 +103,9 @@ impl Keypair {
     /// keypair.sign(b"Hello world!");
     /// ```
     pub fn from_phrase(mnemonic: &bip39::Mnemonic, password: Option<&str>) -> Result<Self, Error> {
-        let big_seed = seed_from_entropy(&mnemonic.to_entropy(), password.unwrap_or(""))
-            .ok_or(Error::InvalidSeed)?;
+        let (arr, len) = mnemonic.to_entropy_array();
```
jsdw, line 106: Nice catch to avoid an allocation anyway :)

signer/src/ecdsa.rs
```diff
@@ -192,31 +206,35 @@ pub fn verify<M: AsRef<[u8]>>(sig: &Signature, message: M, pubkey: &PublicKey) -
     };
     let message_hash = sp_core_hashing::blake2_256(message.as_ref());
     let wrapped = Message::from_digest_slice(&message_hash).expect("Message is 32 bytes; qed");
-    signature.verify(&wrapped, &public).is_ok()
+    GlobalSecp256K1Context
```
jsdw, line 209: Ah and here you can use https://docs.rs/secp256k1/latest/secp256k1/struct.Secp256k1.html#method.verification_only when in no-std rather than signing_only, I think!

signer/Cargo.toml
```diff
@@ -26,34 +27,36 @@ ecdsa = ["secp256k1"]
...
 # Make the keypair algorithms here compatible with Subxt's Signer trait,
 # so that they can be used to sign transactions for compatible chains.
-subxt = ["dep:subxt"]
+subxt = ["dep:subxt-core"]
 
 # The getrandom package is used via schnorrkel. We need to enable the JS
 # feature on it if compiling for the web.
-web = ["getrandom/js", "subxt?/web"]
-native = ["subxt?/native"]
+web = ["getrandom/js"]
```
jsdw, line 34: Nice to be able to ditch the native feature flag!

Does this work when "web" is enabled and "std" isn't, I wonder?

signer/Cargo.toml
```diff
@@ -26,34 +27,36 @@ ecdsa = ["secp256k1"]
...
-bip39 = { workspace = true }
-schnorrkel = { workspace = true, optional = true }
-secp256k1 = { workspace = true, features = ["recovery", "global-context"], optional = true }
+bip39 = { workspace = true, default-features = false, features = ["unicode-normalization"] }
+schnorrkel = { workspace = true, optional = true, default-features = false }
+secp256k1 = { workspace = true, optional = true, default-features = false, features = ["alloc", "recovery"] } # features = ["recovery", "global-context"], 
 secrecy = { workspace = true }
 
+# We only pull this in because `std::sync::OnceLock` is not available in no-std contexts.
+once_cell = { workspace = true, default-features = false, features = ["alloc", "critical-section"] }
```
jsdw, line 53: `critical-section` uses this crate to provide the only-one-thread-at-a-time behaviour:

https://crates.io/crates/critical-section

This crate requires that you enable different features to support different no-std targets (perhaps this explains why you hit a compile error in no-std).

I'm not sure whether this is a good idea to use or not. 

If we do use it, we should def enable `critical-section/std` when `std` feature is enabled though, so that it uses a standard mutex instead of anything strange in std cases. We'd also need to document to users to go read the `critical-section` docs (like once_cell does) because they would need to depend on `critical-section` themselves and add the relevant features for the targets they are trying to support.

But, all of this is just to enable the provided dev accounts at the end of the day, so perhaps the best route to take is revert to using `std::sync::OnceLock`, remove `once_cell` dep and just don't globally instantiate the dev accounts when in no-std mode. So we could alter the `once_static` macro thing to just use the OnceLock global when in std and recreate the account (run the `$expr`) every time when in no-std? (and then document on the macro that in no-std we run the expr every time)

subxt/src/lib.rs
```diff
@@ -65,11 +62,14 @@ mod macros;
...
     client::{OfflineClient, OnlineClient},
-    config::{Config, PolkadotConfig, SubstrateConfig},
     error::Error,
-    metadata::Metadata,
 };
 
+// pub use subxt_core::config;
+// pub use subxt_core::config::{Config, PolkadotConfig, SubstrateConfig};
+// pub use subxt_core::dynamic;
+pub use subxt_core::*;
```
jsdw, line 71: I'd prefer to be really explicit here so that it's obvious exactly what we're exporting :)

core/src/tx/mod.rs
```diff
@@ -168,7 +170,8 @@ impl<CallData: EncodeAsFields> TxPayload for Payload<CallData> {
             .map(|f| scale_encode::Field::new(f.ty.id, f.name.as_deref()));
 
         self.call_data
-            .encode_as_fields_to(&mut fields, metadata.types(), out)?;
+            .encode_as_fields_to(&mut fields, metadata.types(), out)
+            .expect("The fields are valid types from the metadata, qed;");
```
jsdw, line 174: There are various other reasons why the call data might not properly encode though; better to keep the `?`?

core/src/tx/mod.rs
jsdw, line 1: I guess in follow up PRs we can add more functionality; there's a bunch of TX stuff we could have in core too I think; most of the stuff around signing and encoding calls ready to send (just not the "online" bits eg actually sending them)

core/README.md
```diff
@@ -0,0 +1,3 @@
+# Subxt-Core
+
+This library provides core functionality using in `subxt` and `subxt-signer`. It should be no-std compatible.
```
jsdw, line 3: I think for now we should add a warning here and in the lib docs that this is not stable may still change significantly between releases, so better to rely on `subxt` where possible: there's a lot to look at and think about with this new crate, but at the same time it'd be good to feel like we can release it now and iterate on it without worrying too much about breaking things :)

core/src/client/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+        Self {
+            genesis_hash,
+            runtime_version,
+            metadata,
+            marker: core::marker::PhantomData,
+        }
+    }
+
+    pub fn metadata(&self) -> Metadata {
+        self.metadata.clone()
```
jsdw, line 30: This is super expensive to clone (unlike the subxt version of the Metadata it's not wrapped in an Arc, so we are cloning the entire metadata structure and everything in it every time).

As such, cloning `ClientBase` is also super expensive (whic his currently necessary because SignedExtensions are all asking for an owned version of it, for instance).

I'd be tempted to remove `Clone` from `Metadata` and `ClientBase` as a test to see how many places we are relying on it, and then see what we can do to pass it by reference instead in this no-std layer (since things shouldn't be needing to mutate the metadata or client base stuff anyways)

jsdw: Ooh, no I was wrong; you are pointing at a metadata which is stilkl wrapped in Arc, so ignore this :)

jsdw:
> Interface and imports need to be discucced and made nicer, some messiness has to be removed

I'd be curious what you are currently thinking is messy and could be improved; perhaps we should have a walk through of this stuff at some point to get a better feeling for it all as there is a lot to think about in this PR :)

tadeohepperle:
I started a new branch, because it was easier than trying to solve all conflicts, correctly. To be continued here: https://github.com/paritytech/subxt/pull/1466

PR #1419, opened 2024-02-09
Typed Storage Keys

### LINKED ISSUE #1201: Can we be more helpful re decoding storage keys?
Storage keys are often formed from `_concat` style hashers, which are formed from `concat(hash(scale_encoded_key), scale_encoded_key)`.

In theory, we know what the types of the different keys and hashers are when iterating over storage values, and so we might be able to provide back properly typed keys (when `_concat` hashers are used at least) as well as properly typed values.

I think that the ideal interface to aim for would be that we could do:

```rust
// Build a storage query to iterate
let storage_query = polkadot::storage().system().account_iter();

// Get back an iterator of results

while let Some(Ok((key, value))) = results.next().await {
    // Here, key is ideally of a type like `StorageKey<AccountId>` and value is `AccountInfo`.
    // currently, key is `Vec<u8>`
}

Ok(())
```

When there are multiple keys, could we return a tuple of eg `(StorageKey<Foo>, StorageKey<Bar>)`?

`StorageKey` would be a struct that has a method on it like:

```rust
impl <KeyType: DecodeAsType> StorageKey {
    // Try to decode key into the type given,
    // where the type might be Value or something for dynamic queries,
    // and None if the hasher wasn't a `_concat` one.
    pub fn as_key(&self) -> Result<Option<KeyType>, Error> { ... }

    // Just return all of the key bytes.
    pub fn bytes(&self) -> &[u8] { ... }
}
```

When the storage query is dynamic, we could use `StorageKey<Value>` maybe to dynamically decode the key instead where possible.

There may be quirks and caveats to this idea that haven't crossed my mind yet!

tadeohepperle:
Fixes #1201

## Background Info

Previously, static storage keys were represented as a `Static<Encoded>` (wrapper around a `Vec<u8>`) that is created from a concrete type `T: Encode` by scale encoding it.
Previously the ` storage::Address` struct was generic over the type of key: `K`, and held a `Vec<K>` as a value. This `K` was `Static<Encoded>` for static storage keys and `scale_value::Value` for dynamic storage keys.

## What changed

For static storage keys, used in the codegen, we add a new type `StorageKey<K>` that is just a value of K scale encoded into bytes, but keeps the type information of K in a `PhantomData` field. 

A new trait `StorageMultiKey` is introduced, that can be implemented by anything that represents one or multiple storage keys:
```rs
pub trait StorageMultiKey {
    fn keys_iter(&self) -> impl ExactSizeIterator<Item = &dyn EncodeAsType>;

    // Skips over the first 16 bytes of the storage address (pallet hash + entry hash) and then tries to
    // reconstruct the StorageMultiKey from the sequence of hashes (only possible if all hashes are concat hashes)
    fn decode_from_address_bytes(
        address_bytes: &[u8],
        hashers_and_ty_ids: &[(StorageHasher, u32)],
        metadata: &Metadata,
    ) -> Option<Result<Self, Error>>;
}
```
This trait is then implemented for:
- `StorageKey<K>`,
- tuples of storage keys: `(StorageKey<A>, StorageKey<B>,  ...)`
- `Vec<scale_value::Value>`

The `Address` struct representing a storage address is now generic over a `Keys: StorageMultiKey` type, instead of holding a Vec<K>.

This allows us to construct storage addresses that use tuples of storage keys `(StorageKey<A>, StorageKey<B>,  ...)` in the generated code.

Also when getting `StorageResponse`'s back from the Backend, we can now attempt to reconstruct the respective `StorageMultiKey` from the address bytes.


subxt/src/storage/storage_address.rs
```diff
@@ -54,64 +58,314 @@ pub struct Yes;
...
+    pub fn bytes(&self) -> &[u8] {
+        &self.bytes.0 .0
+    }
+}
+
+/// This trait should be implemented by anything that can be used as one or multiple storage keys.
+pub trait StorageMultiKey {
+    /// Iterator over the storage keys, each key implements EncodeAsType to
+    /// give the corresponding bytes to a `StorageHasher`.
+    fn keys_iter(&self) -> impl ExactSizeIterator<Item = &dyn EncodeAsType>;
```
jsdw, line 104: Ah ok, so while `&dyn EncodeAsType` won't work any more when I make `EncodeAsType` non-object-safe (ie generic over TypeResolvers), since Subxt will probably continue to use concrete types like `PortableRegistry` instead of relying too much on `TypeResolver`, it'll be easy enough to make a wrapper trait that is object safe in the same way I had to in `scale-encode` :)

subxt/src/storage/storage_address.rs
```diff
@@ -18,6 +20,8 @@ use subxt_metadata::{StorageEntryType, StorageHasher};
 pub trait StorageAddress {
     /// The target type of the value that lives at this address.
     type Target: DecodeWithMetadata;
+    /// The keys type used to construc this address.
```
lexnv, line 23: s/`construc`/`construct`

subxt/src/storage/storage_address.rs
```diff
@@ -54,64 +58,314 @@ pub struct Yes;
...
     pallet_name: Cow<'static, str>,
     entry_name: Cow<'static, str>,
-    storage_entry_keys: Vec<StorageKey>,
+    keys: Keys,
     validation_hash: Option<[u8; 32]>,
     _marker: std::marker::PhantomData<(ReturnTy, Fetchable, Defaultable, Iterable)>,
 }
 
+/// A storage key, mostly used for static encoded values.
+/// The original value is only given during construction, but can be
```
lexnv, line 77: dq: Unfinished doc

subxt/src/storage/storage_address.rs
```diff
@@ -54,64 +58,314 @@ pub struct Yes;
...
+        // It gives the same result as if you were to use `vec![]` as a `StorageMultiKey`.
+        std::iter::empty()
+    }
+
+    fn decode_from_address_bytes(
+        address_bytes: &[u8],
+        _hashers_and_ty_ids: &[(StorageHasher, u32)],
+        _metadata: &Metadata,
+    ) -> Option<Result<Self, Error>> {
+        if address_bytes.len() < 16 {
```
lexnv, line 136: Nit: I would add a comment that for a valid storage key, we'd need at least 8 bytes for the pallet hash and 8 for the entry

subxt/src/error/mod.rs
```diff
@@ -206,6 +206,9 @@ pub enum StorageAddressError {
         /// The number of fields in the metadata for this storage entry.
         fields: usize,
     },
+    /// The bytes of a storage address are not the expected address for decoding the storage keys of the address.
+    #[error("Storage address bytes are not the expected format")]
```
lexnv, line 210: I would add a small section about keys needing at least 16 bytes (pallet ++ entry)

subxt/src/storage/utils.rs
```diff
@@ -37,3 +42,48 @@ pub(crate) fn storage_address_root_bytes<Address: StorageAddress>(addr: &Address
...
+) -> Result<(), StorageAddressError> {
+    if address_bytes.len() >= 16 {
+        *address_bytes = &mut &address_bytes[16..];
+        Ok(())
+    } else {
+        Err(StorageAddressError::UnexpectedAddressBytes)
+    }
+}
+
+/// Strips the first few bytes off a hash produced byt a concat hasher.
```
lexnv, line 58: `byt a concat`?

subxt/src/storage/storage_address.rs
```diff
@@ -54,64 +58,314 @@ pub struct Yes;
...
+                }
+                Err(err) => return Some(Err(err)),
+            }
+        }
+        Some(Ok(result))
+    }
+}
+
+/// Generates StorageMultiKey implementations for tuples, e.g.
+/// ```rs,norun
```
lexnv, line 248: nit: probably this need changing to avoid _unknown attribute `norun`_

subxt/examples/storage_iterating_partial.rs
```diff
@@ -38,11 +38,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Get back an iterator of results.
     let mut results = api.storage().at_latest().await?.iter(storage_query).await?;
 
-    while let Some(Ok((key, value))) = results.next().await {
-        println!("Key: 0x{}", hex::encode(&key));
-        println!("Value: {:?}", value);
+    while let Some(Ok(kv)) = results.next().await {
+        println!("Keys decoded: {:?}", kv.keys);
+        println!("Key: 0x{}", hex::encode(&kv.key_bytes));
+        println!("Value: {:?}", kv.value);
```
jsdw, line 44: The examples look great!

subxt/src/storage/utils.rs
```diff
@@ -37,3 +42,48 @@ pub(crate) fn storage_address_root_bytes<Address: StorageAddress>(addr: &Address
     write_storage_address_root_bytes(addr, &mut bytes);
     bytes
 }
+
+/// Strips the first 16 bytes (8 for the pallet hash, 8 for the entry hash) off some storage address bytes.
+pub fn strip_storage_addess_root_bytes(
```
jsdw, line 47: Do we want to publically expose this and below?

We expose this `utils` module publically at the moment by the looks of it, which was pointless because the fns were all `pub(crate)` so probably a relic from before!

I'd propose just no longer exposing the `utils` mod at all and then we can turn the `pub(crate)`s into `pub`s (unless there is some reason to expose these functions publically)

subxt/src/storage/storage_address.rs
```diff
@@ -54,64 +58,314 @@ pub struct Yes;
...
+    }
+
+    /// Returns the scale-encoded bytes that make up this key
+    pub fn bytes(&self) -> &[u8] {
+        &self.bytes.0 .0
+    }
+}
+
+/// This trait should be implemented by anything that can be used as one or multiple storage keys.
+pub trait StorageMultiKey {
```
jsdw, line 101: There's quite a lot of stuff related to this `StorageMultiKey`; I think it should all be moved to a new file and then just re-exported from `storage::address` like some other stuff is in the `mod.rs`.

Personally I think I'd just call it `StorageKey` since it's a trait representing storage keys; the Multi bit is sortof irrelevant. And then the file could be eg `storage_key.rs`

jsdw: Ah I see; there is a `StorageKey` concrete type above! I can't recall but if that's just for the static codegen then we could call it `StaticStorageKey` and then this trait can still be `StorageKey`

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn keys_iter(&self) -> impl ExactSizeIterator<Item = &dyn EncodeAsType>;
+
+    /// Attempts to decode the StorageKey from a whole storage address.
+    /// The key/keys can only be recovered if all hashers are concat-style hashers.
+    /// Example: Imagine The `StorageKey` is a tuple (A,B) and the hashers are: [Blake2_128Concat, Twox64Concat].
+    /// Then the memory layout of the storage key is:
+    /// ```txt
+    /// | 8 bytes pallet hash | 8 bytes entry hash | 16 bytes hash of A | ... bytes of A | 8 bytes hash of B | ... bytes of B |
+    /// ```
+    /// Returns None, if any of the hashers is not a concat-style hasher.
```
jsdw, line 27: What if one of the keys is a concat hasher and one isn't?

I'd assume that the codegen for that _could_ allow eg `((), AccountId32)` to be decoded from storage keys like `blake2_256, blake2_256_concat`? I wonder how much complexity supporting something like this would add, if not already possible?

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,232 @@
...
+        address_bytes: &[u8],
+        _hashers_and_ty_ids: &[(StorageHasher, u32)],
+        _metadata: &Metadata,
+    ) -> Option<Result<Self, Error>> {
+        // We need at least 16 bytes, becauase there are 8 bytes for the pallet hash and
+        // another 8 bytes for the entry hash at the beginning of each storage address.
+        if address_bytes.len() < 16 {
+            return Some(Err(StorageAddressError::UnexpectedAddressBytes.into()));
+        }
+        Some(Ok(()))
```
jsdw, line 55: The tuple impls below use `if let Err(err) = strip_storage_addess_root_bytes(cursor) {` to do this check; should this do the same?

Would supporting this empty tuple impl in the macro below make the macro lots more complicated?

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,232 @@
...
+/// impl<A: EncodeAsType, B: EncodeAsType> StorageKey for (StorageKey<A>, StorageKey<B>) {
+///     fn keys_iter(&self) -> impl ExactSizeIterator<Item = &dyn EncodeAsType> {
+///         let arr = [&self.0 as &dyn EncodeAsType, &self.1 as &dyn EncodeAsType];
+///         arr.into_iter()
+///     }
+/// }
+/// ```
+macro_rules! impl_tuples {
+    ($($ty:ident $n:tt),+) => {{
+        impl<$($ty: EncodeAsType + ?Sized),+> StorageKey for ($( StaticStorageKey<$ty >),+) {
```
jsdw, line 174: Should this be more general and apply to any tuple whose members also impl `StorageKey`, rather than being restricted to tuples of `StaticStorageKey`?

subxt/src/storage/storage_key.rs
jsdw: Perhaps more to the point: since this is a trait, it offers the possibility to return `None` if decoding can't succeed, but the implementations get to decide when to use this, so the comment should probably be more like "this is expected to return None when X".

And this gets me thinking; is `None` just an error; ie if you try to decode storage keys into something of the wrong shape, the decoding will fail basically (which in most places in Subxt leadds to a `DecodeError` or something). So is there a point in returning `Option<Result<>>` and not just returning `Result` and having a suitable error for when decoding fails if none already exists? Option makes sense for streams etc that can return `None` under a normal circumstance, but I can't think what it would mean here.

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn decode_from_address_bytes(
+        address_bytes: &[u8],
+        _hashers_and_ty_ids: &[(StorageHasher, u32)],
+        _metadata: &Metadata,
+    ) -> Option<Result<Self, Error>> {
+        // We need at least 16 bytes, becauase there are 8 bytes for the pallet hash and
+        // another 8 bytes for the entry hash at the beginning of each storage address.
+        if address_bytes.len() < 16 {
+            return Some(Err(StorageAddressError::UnexpectedAddressBytes.into()));
+        }
```
jsdw, line 54: I wonder whether the bytes given to a `StorageKey` should only ever be for the subset of things that this `StorageKey` impl is expected to handle?

Then, a more generic tuple impl for `StorageKey` would do nothing except look at the hashers and delegate the relevant bytes and hasher impl to each entry in the tuple (or error if mismatch in count of entries).

Each impl would also then have it easy and doesn't need to trim any bytes from the start; it would be given exactly the bytes and hashers that it is expected to work with.

Then the general storage code that calls into this trait would just trim the relevant root bytes from the front once.

subxt/src/storage/utils.rs
```diff
@@ -37,3 +42,32 @@ pub(crate) fn storage_address_root_bytes<Address: StorageAddress>(addr: &Address
...
+        Err(StorageAddressError::UnexpectedAddressBytes)
+    }
+}
+
+/// Strips the first few bytes off a hash to possibly skip to the plan key value,
+/// if [`StorageHasher::hash_contains_unhashed_key()`] for this StorageHasher.
+///
+/// Returns `Err(..)` if there are not enough bytes.
+/// Returns `Ok(())` otherwise
+pub fn strip_storage_hash_bytes(
```
jsdw, line 63: I think I said this before but maybe lost the comment; does this need to be `pub`? And if not, let's just not `pub use utils` at all from `subxt::storage`, and then we can just have everythign in here be `pub` rather than `pub(crate)`

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+    ///
+    /// Example: Imagine The `StorageKey` is a tuple (A,B) and the hashers are: [Blake2_128Concat, Twox64Concat].
+    /// Then the memory layout of the storage address is:
+    /// ```txt
+    /// | 8 bytes pallet hash | 8 bytes entry hash | 16 bytes hash of A | ... bytes of A | 8 bytes hash of B | ... bytes of B |
+    /// ```
+    /// `cursor` should point into a region after those first 16 bytes, at the start of a new hash.
+    /// `hashers_and_ty_ids` should consume all the hashers that have been used for decoding, such that there are less hashers coming to the next key.
+    fn decode_from_bytes(
+        cursor: &mut &[u8],
```
jsdw, line 34: My gut feeling is that `cursor` should be `&[u8]`, not `&mut &[u8]`; ie this function is just given some bytes to decoee, and some hashers that they represent, and there's no need for it to mutate any external state.

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Iterator over the storage keys, each key implements EncodeAsType to
+    /// give the corresponding bytes to a `StorageHasher`.
+    fn keys_iter(&self) -> impl Iterator<Item = &dyn EncodeAsType>;
+
+    /// How many keys are there in total? Each key is an element that needs to be individually hashed.
+    // Note: Ideally we would use `impl ExactSizeIterator<Item = &dyn EncodeAsType>` for `keys_iter` above,
+    // But that plays poorly with the `Flatten` and `Chain` structs.
+    fn keys_len(&self) -> usize;
+
+    /// Attempts to decode the StorageKey from a storage address that has been stripped of its root bytes.
```
jsdw, line 24: Just to be more precise:

```suggestion
    /// Attempts to decode some part of a storage key, given some bytes and a slice of hashers and type 
    /// IDs that describe the expected format of them.
```

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+/// A storage key for static encoded values.
+/// The original value is only present at construction, but can be decoded from the contained bytes.
+#[derive(Derivative)]
+#[derivative(Clone(bound = ""), Debug(bound = ""))]
+pub struct StaticStorageKey<K: ?Sized> {
+    pub(super) bytes: Static<Encoded>,
+    pub(super) _marker: std::marker::PhantomData<K>,
```
jsdw, line 75: Why `pub(super)` here? (Maybe it's because we construct this somewhere, and perhaps adding a constructor or `From` impl would avoid it? I'll probably spot why later in the review..)

subxt/src/storage/utils.rs
jsdw: But moreover, I think this is only ever used in `storage_key.rs` si let's just have it in that file :) (I think the others are here because they are used in more than one place but would need to confirm)

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+                fields: 1,
+            }
+            .into());
+        };
+        *hashers_and_ty_ids = &hashers_and_ty_ids[1..]; // Advance cursor by 1
+        decode_storage_key_from_hash(cursor, hasher, *ty_id, metadata)
+    }
+}
+
+pub fn decode_storage_key_from_hash<K: ?Sized>(
```
jsdw, line 133: Does this need to be pub?

subxt/src/storage/storage_key.rs
jsdw: I need to think abotu this a bit more because `hashers_and_ty_ids` is in a similar place for me actually.

One advantage of taking `&mut` things is that each `StorageKey` impl can consume the hasher (_and bytes_) that it actually uses up, and hand the rest on to the next `StorageKey` impl (assuming we have eg a tuple of impls). This would let eg `(A, (B, C), D)` to work, because `(B, C)` would consume 2 lots of bytes/hashers when it's called.

However, looking at the code below it doesn't look like we increment the bytes cursor in the same way that we implement the hashers cursor.

This leads to the downside of `&mut` bytes/hashers: to benefit, each implementation _must_ then consume the correct bytes/hasher amount (or error), which is trickier to ensure.

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+            result.push(value.remove_context());
+            n += 1;
+        }
+        *hashers_and_ty_ids = &hashers_and_ty_ids[n..]; // Advance cursor by n
+        Ok(result)
+    }
+}
+
+/// Generates StorageKey implementations for tuples, e.g.
+/// ```rs,norun
```
lexnv, line 216: I think this needs to be `no_run` or `ignore` to avoid the clippy warning

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,288 @@
...
+        _metadata: &Metadata,
+    ) -> Result<Self, Error> {
+        if hashers_and_ty_ids.is_empty() {
+            return Err(StorageAddressError::WrongNumberOfHashers {
+                hashers: 0,
+                fields: 1,
+            }
+            .into());
+        }
+        *hashers_and_ty_ids = &hashers_and_ty_ids[1..]; // Advance cursor by 1
```
jsdw, line 64: Here is one place where we increment the `hashers_and_ty_ids` cursor, byt _dont_ increment the bytes cursor in the same way.

jsdw: Aah, but I see now that we don't know how many bytes to advance if the hasher is a concat one here, so we can't advance them.

Maybe we need to return an error here if the next hasher is a concat hasher, since if we receive a concat hasher here, we're expected to decode the remaining bytes into some type, but aren't?

subxt/src/storage/storage_key.rs
tadeohepperle: I think I had `&[u8]` at first, but if there is a slice of bytes that represents multiple levels of storage keys in a row, we don't know which bytes to hand to the second  storage key, because the offset of where the bytes of the second key starts is not clear until the first key is already decoded. Either we return this offset as a `usize` or use `&mut &[u8]` to chop the bytes off during decoding. But it can be hard to gurantee that the `&mut &[u8]` and the `&mut &[(StorageHasher, u32)]` are in sync.

subxt/src/storage/storage_key.rs
tadeohepperle: Oh this is not needed anymore at all, sorry.

subxt/src/storage/storage_key.rs
tadeohepperle: Oh good catch, we need to strip the hash bytes here as well.

subxt/src/storage/storage_key.rs
jsdw: I'm happy sticking with the `&mut` approach for now, having spent more time looking at things :)

jsdw (commented):
This is generally looking good to me now!

One thing that might be nice to have a play with is having the `StorageKey` trait methods be a bit more similar, eg have`encode_storage_key` and `decode_storage_key`. Having `encode_storage_key` instead of `keys_iter` and `key_len` would hide the internal details better and not require an `&dyn EncodeAsType` type, which won't be possible later).

Another thing I think would be good is to add some tests in the `storage_key.rs` file which generate various hashed bytes and then try to decode them via the trait impls we have here, checking that things fail and succeed as expected. It's very easy to make a mistake and eg forget to advance cursors properly, so tests would add a lot of reassurance :)

I still need to look properly at the codegen side. And we have a bunch of logic around getting the hasher info from the metadata and wonder if anything can be simplified a touch in that area too.

codegen/src/error.rs
```diff
@@ -39,15 +39,21 @@ pub enum CodegenError {
...
+    #[error("{0} type should be an variant/enum type. Make sure you are providing a valid substrate-based metadata")]
     InvalidType(String),
     /// Extrinsic call type could not be found.
-    #[error(
-        "Extrinsic call type could not be found. Make sure you are providing a valid substrate-based metadata"
-    )]
+    #[error("Extrinsic call type could not be found. Make sure you are providing a valid substrate-based metadata")]
     MissingCallType,
+    /// There are too many or too few hashers.
+    #[error("Could not Generate functions for storage entry {storage_entry_name}. There are {key_count} keys, but only {hasher_count} hashers. The number of hashers must equal the number of keys or be exactly 1.")]
```
niklasad1, line 48:
```suggestion
    #[error("Could not cenerate functions for storage entry {storage_entry_name}. There are {key_count} keys, but only {hasher_count} hashers. The number of hashers must equal the number of keys or be exactly 1.")]
```

subxt/src/storage/storage_address.rs
```diff
@@ -156,101 +166,76 @@ where
...
-                    // Provided more keys than fields.
-                    return Err(StorageAddressError::WrongNumberOfKeys {
-                        expected: type_ids.len(),
-                        actual: self.storage_entry_keys.len(),
-                    }
-                    .into());
-                }
+        // If the key is a tuple, we encode each value to the corresponding tuple type.
+        // If the key is not a tuple, encode a single value to the key type.
+        let type_ids = match &ty.type_def {
```
niklasad1, line 195: could be if let TypeDef::Tuple(tuple) { ... } else { key_ty }

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn decode_from_bytes(
+        bytes: &mut &[u8],
+        hashers_and_ty_ids: &mut &[(StorageHasher, u32)],
+        types: &PortableRegistry,
+    ) -> Result<Self, Error>
+    where
+        Self: Sized + 'static,
+    {
+        let Some((hasher, ty_id)) = hashers_and_ty_ids.first() else {
+            return Err(StorageAddressError::WrongNumberOfHashers {
```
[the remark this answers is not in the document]
niklasad1: if so you could use `hasher_and_ty_ids.split_first()` instead

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,433 @@
...
+            return Err(StorageAddressError::TooManyBytes.into());
+        }
+
+        Ok(result)
+    }
+}
+
+// Skip over the hash bytes (including any key at the end), returning bytes
+// representing the key if one exists, or None if the hasher has no key appended.
+fn consume_hash_returning_key_bytes<'a>(
```
niklasad1, line 206: read_key_from_hash

subxt/src/storage/storage_type.rs
```diff
@@ -290,6 +310,75 @@ where
...
+                }
+            }
+        }
+    }
+}
+
+/// Strips the first 16 bytes (8 for the pallet hash, 8 for the entry hash) off some storage address bytes.
+fn strip_storage_addess_root_bytes(address_bytes: &mut &[u8]) -> Result<(), StorageAddressError> {
+    if address_bytes.len() >= 16 {
+        *address_bytes = &address_bytes[16..];
```
niklasad1, line 363: this may end up with an empty slice if address_bytes.len == 16 ok?

codegen/src/api/storage.rs
```diff
@@ -75,43 +75,83 @@ fn generate_storage_entry_fns(
...
             {
                 // An N-map; return each of the keys separately.
-                TypeDef::Tuple(tuple) => tuple
-                    .fields
-                    .iter()
-                    .enumerate()
-                    .map(|(idx, f)| storage_entry_map(idx, f.id))
-                    .collect::<Vec<_>>(),
+                TypeDef::Tuple(tuple) => {
+                    let key_count = tuple.fields.len();
```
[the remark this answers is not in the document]
tadeohepperle: I think in reality, this is unlikely, but if they were, we just return `vec![]` below.

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,433 @@
...
+    pub fn decoded(&self) -> Result<K, Error> {
+        let decoded = K::decode(&mut self.bytes())?;
+        Ok(decoded)
+    }
+}
+
+impl<K: ?Sized> StaticStorageKey<K> {
+    /// Returns the scale-encoded bytes that make up this key
+    pub fn bytes(&self) -> &[u8] {
+        &self.bytes.0 .0
```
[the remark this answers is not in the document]
tadeohepperle: rustfmt makes me do this :( I hate it too, looks so weird, but apparently this is the official formatting for nested tuples.

subxt/src/storage/storage_key.rs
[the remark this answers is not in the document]
tadeohepperle: We completely break the coding on error and throw away the cursors afterwards. So if we encounter an error, we just say the entire storage key could not be decoded and do not proceed.

subxt/src/storage/storage_key.rs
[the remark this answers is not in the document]
jsdw: I saw this too and assumed it must be rustfmt being strange; arggh oh well :)

subxt/src/storage/storage_type.rs
tadeohepperle: Yes, that is true, but it is okay I think.

jsdw: Yeah I think that's ok; if it matters (ie we expect a non-empty hash next) then the next thing will error instead

metadata/src/lib.rs
```diff
@@ -475,6 +475,35 @@ pub enum StorageHasher {
...
+            StorageHasher::Blake2_128 => 16,
+            StorageHasher::Blake2_256 => 32,
+            StorageHasher::Twox128 => 16,
+            StorageHasher::Twox256 => 32,
+            StorageHasher::Identity => 0,
+        }
+    }
+
+    /// Returns true if the key used to produce the hash is appended to the hash itself.
+    pub fn ends_with_key(&self) -> bool {
```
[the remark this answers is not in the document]
jsdw: I chanegd it from `contains_key` or something I think! I went with `ends_` just to fit in more nicely with the above method which is giving the length of the hash apart from the key bit.

`contains_key` would work too though for me!

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,433 @@
...
+use scale_value::Value;
+use subxt_metadata::StorageHasher;
+
+use derivative::Derivative;
+
+/// This trait should be implemented by anything that can be used as one or multiple storage keys.
+pub trait StorageKey {
+    /// Iterator over the storage keys, each key implements EncodeAsType to
+    /// give the corresponding bytes to a `StorageHasher`.
+    fn keys_iter(&self) -> impl Iterator<Item = &dyn EncodeAsType>;
```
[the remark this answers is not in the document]
jsdw: I had a more general comment that it'd be nice to "hide away" the internal details (ie the liust of keys contained within) and instead have a method like `encode_to_bytes` which more closely mirrors `decode_from_bytes` below. (Also, when the update comes we won't be able to return `&dyn EncodeAsType`).

If that worked out, and then I were also thinking about method naming next, I'd then suggest `encode_storage_key` and `decode_storage_key` to more closely matc hthe `StorageKey` trait name.

subxt/src/storage/storage_key.rs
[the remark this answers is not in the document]
niklasad1: cool

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    /// Attempts to decode the StorageKey given some bytes and a set of hashers and type IDs that they are meant to represent.
+    /// The bytes passed to `decode` should start with:
+    /// - 1. some fixed size hash (for all hashers except `Identity`)
+    /// - 2. the plain key value itself (for `Identity`, `Blake2_128Concat` and `Twox64Concat` hashers)
+    fn decode(
+        bytes: &mut &[u8],
+        hashers: &mut StorageHashersIter,
+        types: &PortableRegistry,
+    ) -> Result<Self, Error>
```
jsdw, line 113: This interface looks really clean now, nice one!

Small naming nit: I'd suggest something like `encode_storage_key` and `decode_storage_key` so there's a nice link between the method names and the trait involved :)

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,445 @@
...
+// An iterator over all type ids of the key and the respective hashers.
+pub struct StorageHashersIter {
+    hashers_and_ty_ids: Vec<(StorageHasher, u32)>,
+    idx: usize,
+}
+
+impl StorageHashersIter {
+    /// Creates a new [`StorageHashersIter`]. Looks at the [`StorageEntryType`] and
+    /// assigns a hasher to each type id that makes up the key.
+    pub fn new(storage_entry: &StorageEntryType, types: &PortableRegistry) -> Result<Self, Error> {
```
jsdw, line 25: Can this be `pub(crate)` so that constructing this is an internal-to-subxt thing only? :)

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    pub fn next_or_err(&mut self) -> Result<(StorageHasher, u32), Error> {
+        self.next().ok_or_else(|| {
+            StorageAddressError::TooManyKeys {
+                expected: self.hashers_and_ty_ids.len(),
+            }
+            .into()
+        })
+    }
```
jsdw, line 72: Should this be a part of the public API? Maybe it could just be a small internal helper fn instead :)

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    pub fn reset(&mut self) {
+        self.idx = 0;
+    }
```
jsdw, line 76: `reset()` type functions on iterators feel a bit unidiomatic to me (and mutating an iterator scares me a little). Not really a big deal but we could be more idiomatic perhaps by just making the iteration be a separate thing like so:

```rust
// rename `StorageHAshersIter` to this but keep methods the same:
pub struct StorageHashers { 
  // same methods.. 
 
    pub fn iter(&'_ self) -> StorageHashersIter<'_> {
        StorageHashersIter { hashers: self, idx: 0 }
    }
}

// impl `Iterator` on a separate type instead (then we can iterate over them 
// as many times as we want and no iterators are being mutated. Maybe a bit like:
pub struct StorageHashersIter<'a> {
    idx: usize,
    hashers: &'a StorageHashers
}

impl <'a> Iterator for StorageHashersIter<'a> {
    fn next(&mut self) {
        let item = self.hashers.hashers_and_ty_ids.get(self.idx).copied()?;
        self.idx += 1;
        Some(item)
    }
}
```

jsdw: Maybe this is a bit of a pain though, and if so I'm happy to stick with the current approach!

jsdw:
This looks great to me now; just a couple of very small suggestions! Great job @tadeohepperle!

subxt/src/storage/storage_key.rs
tadeohepperle: Oh yeah, the `reset` is a bit ugly, I implemented a seperate `StorageHashersIter<'a>` struct now.

subxt/src/storage/storage_key.rs
```diff
@@ -182,6 +202,7 @@ impl<K: ?Sized> StorageKey for StaticStorageKey<K> {
         hashers: &mut StorageHashersIter,
         types: &PortableRegistry,
     ) -> Result<(), Error> {
+        println!("{:?}", hashers);
```
jsdw, line 205: Intentional? :)

subxt/src/storage/storage_key.rs
```diff
@@ -12,19 +12,18 @@ use derivative::Derivative;
...
     hashers_and_ty_ids: Vec<(StorageHasher, u32)>,
-    idx: usize,
 }
 
-impl StorageHashersIter {
-    /// Creates a new [`StorageHashersIter`]. Looks at the [`StorageEntryType`] and
+impl StorageHashers {
+    /// Creates new [`StorageHashers`] from a storage entry. Looks at the [`StorageEntryType`] and
     /// assigns a hasher to each type id that makes up the key.
     pub fn new(storage_entry: &StorageEntryType, types: &PortableRegistry) -> Result<Self, Error> {
```
jsdw, line 25: pub(crate) perhaps so we don't need to care about this being in the public API? :)

jsdw: Ah I see; the whole thing is now!

jsdw (approved):
One `println!()` that I think is unintended but otherwise looks awesome :)

metadata/src/lib.rs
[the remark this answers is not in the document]
lexnv: Either looks good to me :D

subxt/src/storage/storage_key.rs
```diff
@@ -0,0 +1,466 @@
...
+        type T4C = (
+            ((), StaticStorageKey<u32>),
+            (StaticStorageKey<String>, StaticStorageKey<Era>),
+        );
+
+        let era = Era::Immortal;
+        for h0 in hashers {
+            for h1 in key_preserving_hashers {
+                for h2 in key_preserving_hashers {
+                    for h3 in key_preserving_hashers {
```
lexnv, line 430: Nice! 👍

lexnv (approved):
Nice job here 👍

PR #1420, opened 2024-02-09
[ci] Run cronjob with app secrets

alvicsam:
PR adds app tokens so the job "Upload artifacts" has running CI in the PR it creates.

cc https://github.com/paritytech/ci_cd/issues/928
cc @tadeohepperle @pavelsupr

### DIFF
--- a/.github/workflows/actions/use-nodes/action.yml
+++ b/.github/workflows/actions/use-nodes/action.yml
@@ -18,6 +18,7 @@ runs:
         name: nightly-polkadot-binary
 
     - name: decompress polkadot binary
+      shell: bash
       run: |
         tar -xzvf ./polkadot.tar.gz
 
--- a/.github/workflows/actions/use-nodes/action.yml
+++ b/.github/workflows/actions/use-nodes/action.yml
@@ -17,6 +17,10 @@ runs:
         workflow: build-nodes.yml
         name: nightly-polkadot-binary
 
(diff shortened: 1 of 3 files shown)


jsdw:
I turned this into a draft since it looks like it's in progress; please turn back to proper PR once it's ready for us to look at (I'm not sure the exact goals/reason for this change atm!) :)

alvicsam:
@jsdw thanks! I updated PR description

@tadeohepperle the credentials work but now the action [fails](https://github.com/paritytech/subxt/actions/runs/7843391726/job/21403872613) because of the error:

```
remote: error: Trace: 7feebb3939405d32c7ba980df6692009a70fa25f7beea09036fe2722b667b968        
remote: error: See https://gh.io/lfs for more information.        
remote: error: File polkadot is 123.89 MB; this exceeds GitHub's file size limit of 100.00 MB        
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com./
```

lexnv (approved):
This looks good to me! Thanks for the help here!
I would probably also strip the binary, similarly to what Niklas suggested.

What would be the next step to get this merged? 🤔

jsdw (approved):
Looks good, nice one!

I guess we can compress the other two polkadot binaries if we feel the need as well, but if it works at the mo then all good :)

PR #1439, opened 2024-02-22
Extrinsic Params Refinement

tadeohepperle:
Follow-Up PR to #1432, see discussion between James and me there to get some background information.

This PR introduces a new trait `RefineParams` that the `OtherParams` (renamed to `Params`) of signed extensions must implement:
```
pub trait RefineParams<T: Config> {
    fn refine(&mut self, _account_nonce: u64, _block_number: u64, _block_hash: T::Hash) {}
}
``` 

The 'refinement' of params is executed by an online client before creating an extrinsic and can automatically configure the newest account nonce for `CheckNonce` and the latest block hash for `CheckMortality`.


subxt/src/tx/tx_client.rs
```diff
@@ -103,11 +103,14 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
     }
 
     /// Create a partial extrinsic.
-    pub fn create_partial_signed_with_nonce<Call>(
+    ///
+    /// Note: This function will **not** refine the params to use the latest account nonce and the latest block hash/number.
+    /// Instead your call will get default values of nonce = 0 and mortality = Immortal for the signed extensions, if the params
+    /// have not been 'refined' before.
+    pub fn create_partial_signed_offline<Call>(
```
tadeohepperle, line 110: I am not sure about the name here, maybe `create_partial_signed_unrefined`? Ideally, we would have different implementations of `create_partial_signed` for offline and online clients, but without [`specialization`](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) I don't see how that would be possible.

lexnv: An alternative to specialization with stable rust is the autoref workaround. (James pointed this out a while ago and we ended up using it in substrate for metadata v15)

The `default` implementation should be implemented on the `&T`. Whenever users what to override that, they must implement it on `T`. Then the compiler will add as many references as needed to T to find the method call. Not entirely sure if this is applicable here

subxt/src/config/refine_params.rs
```diff
@@ -0,0 +1,53 @@
...
+
+//! Refining params with values fetched from the chain
+
+use crate::Config;
+
+/// Types implementing [`RefineParams`] can be modified to reflect live information from the chain.
+pub trait RefineParams<T: Config> {
+    /// Refine params to an extrinsic. There is usually some notion of 'the param is already set/unset' in types implementing this trait.
+    /// The refinement should most likely not affect cases where a param is in a 'is already set by the user' state.
+    fn refine(&mut self, _account_nonce: u64, _block_number: u64, _block_hash: T::Hash) {}
```
lexnv, line 13: Are there currently other types that substrate/polkadot extensions take for constructing their state?
I'm thinking more about how we could easily extend this in the future without causing a breaking change

subxt/src/book/usage/transactions.rs
```diff
@@ -137,11 +137,10 @@
 //!     Value::from_bytes("Hello there")
 //! ]);
 //!
-//! // Construct the tx but don't sign it. You need to provide the nonce
-//! // here, or can use `create_partial_signed` to fetch the correct nonce.
-//! let partial_tx = client.tx().create_partial_signed_with_nonce(
+//! // Construct the tx but don't sign it. The account nonce here defaults to 0.
+//! // You Can use `create_partial_signed` to fetch the correct nonce.
```
lexnv, line 141: nit: `You can`

subxt/src/tx/tx_client.rs
jsdw: Hehe it might be possible but we probably shouldn't go down that route for fear of over-complexifying :D

subxt/src/book/usage/transactions.rs
jsdw, line 141:
```suggestion
//! // You can use `create_partial_signed` to fetch the correct nonce.
```

subxt/src/config/extrinsic_params.rs
```diff
@@ -53,13 +55,12 @@ pub trait ExtrinsicParams<T: Config>: ExtrinsicParamsEncoder + Sized + 'static {
...
     /// help construct your [`ExtrinsicParams`] object.
-    type OtherParams;
+    type Params: RefineParams<T>;
 
     /// Construct a new instance of our [`ExtrinsicParams`].
     fn new<Client: OfflineClientT<T>>(
-        nonce: u64,
         client: Client,
-        other_params: Self::OtherParams,
+        other_params: Self::Params,
```
jsdw, line 63:
```suggestion
        params: Self::Params,
```

subxt/src/config/refine_params.rs
jsdw: One thing that we could do to _maybe_ minimise the risk of breaking changes a little is to wrap these into a struct like:

```rust
pub struct RefineData<T: Config> {
    account_nonce: u64,
    block_number: u64,
    block_hash: T::Hash
}

impl <T: Config> RefineData {
    pub fn account_nonce(&self) -> u64 { ... }
    pub fn block_number(&self) -> u64 { ... }
    pub fn block_hash(&self) -> u64 { ... }
}
```

That way, we can add params without changing the existing interface :)

subxt/src/config/refine_params.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! Refining params with values fetched from the chain
+
+use crate::Config;
+
+/// Types implementing [`RefineParams`] can be modified to reflect live information from the chain.
+pub trait RefineParams<T: Config> {
```
jsdw, line 10: Nit: I'd call this `RefinableParams` (because then you read `Params: RefinableParams<T>` which says.. these params are refinable!) and then call method `refine_params` I reckon so it's more obvious it relates to the trait

subxt/src/config/mod.rs
```diff
@@ -12,6 +12,7 @@ mod default_extrinsic_params;
 mod extrinsic_params;
 
 pub mod polkadot;
+pub mod refine_params;
```
jsdw, line 15: I wouldn't export this mod I reckon; there's no real need since the only item it exports is below anyway :)

subxt/src/tx/tx_client.rs
```diff
@@ -165,6 +167,30 @@ where
     T: Config,
     C: OnlineClientT<T>,
 {
+    /// Fetch the latest block header and account nonce from the backend and use them to refine [`ExtrinsicParams::Params`].
+    pub async fn refine_params(
```
jsdw, line 171: I wouldn't publically expose this; I think in general that I don't want users to really know about "refining params". If they are offline, they can obtain the relevant data and configure the params exactly as they want, and if they are online then we will be smart and apply some sensible defaults for them is all :)

subxt/src/tx/tx_client.rs
```diff
@@ -103,11 +103,14 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
     }
 
     /// Create a partial extrinsic.
-    pub fn create_partial_signed_with_nonce<Call>(
+    ///
+    /// Note: This function will **not** refine the params to use the latest account nonce and the latest block hash/number.
+    /// Instead your call will get default values of nonce = 0 and mortality = Immortal for the signed extensions, if the params
+    /// have not been 'refined' before.
```
jsdw, line 109:
```suggestion
    /// Note: if not provided, the default account nonce will be set to 0 and the default mortality will be _immortal_.
    /// This is because this method runs offline, and so is unable to fetch the data needed for more appropriate values.
```

Just to make it clear that because offline, we have to use worse defaults, and to avoid mentioning any new concept like "refining params"

subxt/src/tx/tx_client.rs
```diff
@@ -135,12 +135,15 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
     }
 
     /// Creates a signed extrinsic without submitting it.
-    pub fn create_signed_with_nonce<Call, Signer>(
+    ///
+    /// Note: This function will **not** refine the params to use the latest account nonce and the latest block hash/number.
+    /// Instead your call will get default values `nonce = 0` and `mortality = Immortal` for the signed extensions, if the
+    /// params have not been manually set or 'refined' before.
```
jsdw, line 141:
```suggestion
    /// Note: if not provided, the default account nonce will be set to 0 and the default mortality will be _immortal_.
    /// This is because this method runs offline, and so is unable to fetch the data needed for more appropriate values.
```

As above

subxt/src/tx/tx_client.rs
jsdw: I like the `_offline` suffix too FWIW, and also this suffix plays nicely with my suggested comment tweak so that it is consistent that if we call an `_offline` function we get worse defaults because it runs offline.

jsdw:
Some small suggestions and such but overall this looks really good, and doesn't feel like it adds too much complexity in the end while giving us the ability to set some great defaults!

It's sortof a shame that we have to ask for the block hash and account nonce etc now every time, but I'm ok with it for now!

(if perf is an issue here then users can provide their own params, and _potentially_ as an optimisation we could expand the `RefinableParams` trait with a `can_be_refined(&self) -> bool { true }` method, and then signed extensions can return false from that when they don't need refining, and if nothing needs refining we can skip obtaining the details at all :))

subxt/src/tx/tx_client.rs
tadeohepperle: Ah I remember reading dtolnays article about autoref specialization on stable rust a while ago, but it seemed like a weird workaround to me.

subxt/src/config/refine_params.rs
tadeohepperle: Hmm, I also like the name with the _able_, but I thought in rust it is generally a convention to about that, e.g. `Clone` instead of `Clonable`, `Serialize` instead of `Serializable`,[ etc.](https://github.com/rust-lang/api-guidelines/discussions/28)

jsdw: Hmm true! Maybe it's better to leave it as is then, just for consistency :)

jsdw (approved):
This looks great to me; good job Tadeo!

lexnv (approved):
Nice job here!

PR #1442, opened 2024-02-23
[audit] Update DispatchError to match latest in polkadot-sdk

jsdw:
The auditors spotted that our DispatchError type had fallen a bit out of sync with the latest one found here https://github.com/paritytech/polkadot-sdk/blob/master/substrate/primitives/runtime/src/lib.rs#L564-L598.

### DIFF
--- a/subxt/src/error/dispatch_error.rs
+++ b/subxt/src/error/dispatch_error.rs
@@ -59,6 +59,9 @@ pub enum DispatchError {
         "Some resource (e.g. a preimage) is unavailable right now. This might fix itself later."
     )]
     Unavailable,
+    /// Root origin is not allowed.
+    #[error("Root origin is not allowed.")]
+    RootNotAllowed,
 }
 
 /// An error relating to tokens when dispatching a transaction.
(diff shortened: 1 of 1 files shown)


subxt/src/error/dispatch_error.rs
```diff
@@ -59,6 +59,9 @@ pub enum DispatchError {
         "Some resource (e.g. a preimage) is unavailable right now. This might fix itself later."
     )]
     Unavailable,
+    /// Root origin is not allowed.
+    #[error("Root origin is not allowed.")]
+    RootNotAllowed,
```
lexnv, line 64: Would be interesting if we could automatically check that all types we have from substrate are in sync.

I'm not aware of a clean way to do that.
We could probably explore in the future a custom crate that compares the tokens from this type to a github link, or crate version.

```rust

#[sync_test(
   git = "https://github.com/paritytech/polkadot-sdk/blob/master/substrate",
   path  = "substrate/primitives/runtime/src/lib.rs"
)
enum SubstrateType {
  ...
}

lexnv (approved):

subxt/src/error/dispatch_error.rs
jsdw: That's an interesting idea!

niklasad1 (approved):

PR #1444, opened 2024-02-23
[audit] Avoid a panic in case we try decoding naff DryRunResult bytes

jsdw:


subxt/src/backend/legacy/rpc_methods.rs
```diff
@@ -535,6 +535,13 @@ impl DryRunResultBytes {
         // dryRun returns an ApplyExtrinsicResult, which is basically a
         // `Result<Result<(), DispatchError>, TransactionValidityError>`.
         let bytes = self.0;
+
+        // We expect at least 2 bytes. In case we got a naff response back (or
```
niklasad1, line 539: naff response back? 

some cool british expression? :)

lexnv (approved):

subxt/src/backend/legacy/rpc_methods.rs
jsdw: :D Yup, cool, exactly! "dodgy" or "malformed" are also applicable!

niklasad1 (approved):

PR #1445, opened 2024-02-23
[audit] Fix error mapping to wrong transaction status

jsdw:
Probably a copypaste error when the code was written :)


niklasad1 (approved):
nice catch by the auditors :)

lexnv (approved):

PR #1452, opened 2024-02-26
[lightclient] lookup chain ID just once at beginning of each message

jsdw:
Just a small change to attempt to improve the isolation of data by chain IDs a little better; require that we fetch all data once by chain ID rather than using chain ID to lookup data in multiple places


lexnv (approved):
Looks good to me! This simplifies the code quite a bit, making it easier to follow!

tadeohepperle (approved):
:+1:

lightclient/src/background.rs
```diff
@@ -114,19 +129,20 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
...
-        *next = next.wrapping_add(1);
-        id
+    fn for_chain_id(
+        &mut self,
+        chain_id: smoldot_light::ChainId,
+    ) -> (
+        &mut ChainData,
+        &mut smoldot_light::Client<TPlatform, TChain>,
+    ) {
+        let chain_data = self.chain_data.entry(chain_id).or_default();
```
niklasad1, line 143: is chain_data just a `Vec`?

lightclient/src/background.rs
```diff
@@ -138,26 +154,28 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
                 sender,
                 chain_id,
             } => {
-                let id = self.next_id(chain_id);
+                let (data, client) = self.for_chain_id(chain_id);
```
niklasad1, line 157: I would probably rename data to `chain_state` or something.

data is too generic

lightclient/src/background.rs
```diff
@@ -138,26 +154,28 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
...
+                let id = data.next_id();
```
niklasad1, line 158: Ok, we got chain_id -> ChainState now right should be good

I was afraid that we may re-use this globally but no worries.

niklasad1 (approved):
Looks clean.

Small style thingy I would like to rename `data` to `chain_state/per_chain_data` or something similar because I got confused about several times :brain:

jsdw:
Good idea; I renamed it to `chain_data` since that's the name of the struct; def feels nicer :)

PR #1455, opened 2024-02-27
Allow either backend to be used in light client code

jsdw:
I don't really like that this needs a new trait etc, and so I think this hints that there's a cleaner way to go about this whole thing.

The light client interface is really just an RpcClient at the end of the day, so maybe it was wrong to go about having a `LightClient` struct, which is essentially just a wrapper around `OnlineClient` anyway (this was my idea, so my bad :)). 

Maybe the light client APIs should be just focused around creating an RpcClient to talk to whatever chain, and then we add some of the light client functions as high level helpers to `OnlineClient` (and then people can alternately create a custom backend with custom RPC client etc to make one instead).


jsdw:
Closed in favoru of #1475

PR #1457, opened 2024-02-28
Suggestions for storage value decoding

jsdw:
This PR was quite complicated, so here I just offer suggestions for things to improve instead of tryign to write comments for it all!

One thing that is missing I think is some tests in `storage_key.rs` to check that we can form various tuple keys etc and they will correctly decode some bytes. I spotted one or two things I thought might fail, so would be great to have some test coverage here :)

### DIFF
--- a/metadata/src/lib.rs
+++ b/metadata/src/lib.rs
@@ -476,14 +476,14 @@ pub enum StorageHasher {
 }
 
 impl StorageHasher {
-    /// The hash produced by a [`StorageHasher`] has 1 or 2 of the following components:
-    /// 1. A fixed size hash. (not present for [`StorageHasher::Identity`])
-    /// 2. The key that was used as an input to the hasher. (only present for
+    /// The hash produced by a [`StorageHasher`] can have these two components, in order:
+    ///
+    /// 1. A fixed size hash. (not present for [`StorageHasher::Identity`]).
+    /// 2. The SCALE encoded key that was used as an input to the hasher (only present for
     /// [`StorageHasher::Twox64Concat`], [`StorageHasher::Blake2_128Concat`] or [`StorageHasher::Identity`]).
     ///
-    /// This function returns the number of hash bytes that must be skipped to get to the 2. component, the unhashed key.
-    /// To check whether or not an unhashed key is contained, see [`StorageHasher::hash_bytes_before_unhashed_key`].
-    pub fn hash_bytes_before_unhashed_key(&self) -> usize {
+    /// This function returns the number of bytes used to represent the first of these.
+    pub fn len_excluding_key(&self) -> usize {
         match self {
             StorageHasher::Blake2_128Concat => 16,
             StorageHasher::Twox64Concat => 8,
@@ -495,10 +495,8 @@ impl StorageHasher {
         }
     }
 
-    /// Returns if the key a hash was produces for, is contained in the hash itself.
-    /// If this is `true`, [`StorageHasher::hash_bytes_before_unhashed_key`] can be used,
-    /// to find the offset of the start byte of the key from the start of the hash bytes.
-    pub fn hash_contains_unhashed_key(&self) -> bool {
+    /// Returns true if the key used to produce the hash is appended to the hash itself.
+    pub fn ends_with_key(&self) -> bool {
         matches!(
             self,
             StorageHasher::Blake2_128Concat | StorageHasher::Twox64Concat | StorageHasher::Identity
--- a/subxt/src/error/mod.rs
+++ b/subxt/src/error/mod.rs
@@ -207,8 +207,14 @@ pub enum StorageAddressError {
         /// The number of fields in the metadata for this storage entry.
         fields: usize,
     },
+    /// We weren't given enough bytes to decode the storage address/key.
+    #[error("Not enough remaining bytes to decode the storage address/key")]
+    NotEnoughBytes,
+    /// We have leftover bytes after decoding the storage address.
+    #[error("We have leftover bytes after decoding the storage address")]
+    TooManyBytes,
     /// The bytes of a storage address are not the expected address for decoding the storage keys of the address.
-    #[error("Storage address bytes are not the expected format. Addresses need to be at least 16 bytes (pallet ++ entry) and follow a structure given by the hashers defined in the metadata.")]
+    #[error("Storage address bytes are not the expected format. Addresses need to be at least 16 bytes (pallet ++ entry) and follow a structure given by the hashers defined in the metadata")]
     UnexpectedAddressBytes,
     /// An invalid hasher was used to reconstruct a value from a chunk of bytes that is part of a storage address. Hashers where the hash does not contain the original value are invalid for this purpose.
     #[error("An invalid hasher was used to reconstruct a value with type ID {ty_id} from a hash formed by a {hasher:?} hasher. This is only possible for concat-style hashers or the identity hasher")]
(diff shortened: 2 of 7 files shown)


metadata/src/lib.rs
```diff
@@ -476,14 +476,14 @@ pub enum StorageHasher {
 }
 
 impl StorageHasher {
-    /// The hash produced by a [`StorageHasher`] has 1 or 2 of the following components:
-    /// 1. A fixed size hash. (not present for [`StorageHasher::Identity`])
-    /// 2. The key that was used as an input to the hasher. (only present for
+    /// The hash produced by a [`StorageHasher`] can have these two components, in order:
```
jsdw, line 479: Since the functions are a part of our public API here, I tried to give them slightly nicer names

subxt/src/error/mod.rs
```diff
@@ -211,12 +211,10 @@ pub enum StorageAddressError {
     #[error("Storage address bytes are not the expected format. Addresses need to be at least 16 bytes (pallet ++ entry) and follow a structure given by the hashers defined in the metadata.")]
     UnexpectedAddressBytes,
     /// An invalid hasher was used to reconstruct a value from a chunk of bytes that is part of a storage address. Hashers where the hash does not contain the original value are invalid for this purpose.
-    #[error("An invalid hasher was used to reconstruct a value of type {ty_name} (id={ty_id}) from a hash formed by a {hasher:?} hasher. This is only possible for concat-style hashers or the identity hasher")]
+    #[error("An invalid hasher was used to reconstruct a value with type ID {ty_id} from a hash formed by a {hasher:?} hasher. This is only possible for concat-style hashers or the identity hasher")]
```
jsdw, line 214: I removed the `type_name`, really jsut to simplify the code a touch since in most other places we only return the type ID anyway for this sort of error.

subxt/src/storage/storage_key.rs
```diff
@@ -21,25 +22,29 @@ pub trait StorageKey {
...
     /// ```
-    /// `cursor` should point into a region after those first 16 bytes, at the start of a new hash.
-    /// `hashers_and_ty_ids` should consume all the hashers that have been used for decoding, such that there are less hashers coming to the next key.
+    ///
+    /// Implementations of this must advance the `bytes` and `hashers_and_ty_ids` cursors to consume any that they are using, or
+    /// return an error if they cannot appropriately do so. When a tuple of such implementations is given, each implementation
+    /// in the tuple receives the remaining un-consumed bytes and hashers from the previous ones.
     fn decode_from_bytes(
-        cursor: &mut &[u8],
+        bytes: &mut &[u8],
```
jsdw, line 38: both this and the below are "cursors", so I renamed this one

subxt/src/storage/storage_key.rs
```diff
@@ -50,18 +55,24 @@ impl StorageKey for () {
...
         hashers_and_ty_ids: &mut &[(StorageHasher, u32)],
-        _metadata: &Metadata,
+        metadata: &Metadata,
     ) -> Result<Self, Error> {
-        if hashers_and_ty_ids.is_empty() {
-            return Err(StorageAddressError::WrongNumberOfHashers {
-                hashers: 0,
-                fields: 1,
+        // If no hashers, we just do nothing (erroring if ).
```
jsdw, line 62: When there _is_ a hasher, this also needs to skip over any key bytes if they exist (which is what `consume_hash_returning_key_bytes` does.

When there is not a hasher, we just exit happily unless there are leftover bytes, which would be unexpected.

subxt/src/storage/storage_key.rs
```diff
@@ -125,53 +136,28 @@ impl<K: ?Sized> StorageKey for StaticStorageKey<K> {
...
-
-    let bytes = *cursor;
-    if let Err(err) = scale_decode::visitor::decode_with_visitor(
-        cursor,
-        ty_id,
-        metadata.types(),
-        scale_decode::visitor::IgnoreVisitor,
-    ) {
-        return Err(scale_decode::Error::from(err).into());
+        // Advance the bytes cursor, returning any key bytes.
```
jsdw, line 140: With the `consume_hash_returning_bytes` function I thought it would be cleaner then to inline the other code for this impl

subxt/src/storage/storage_key.rs
```diff
@@ -125,53 +136,28 @@ impl<K: ?Sized> StorageKey for StaticStorageKey<K> {
             }
             .into());
         };
-        *hashers_and_ty_ids = &hashers_and_ty_ids[1..]; // Advance cursor by 1
-        decode_storage_key_from_hash(cursor, hasher, *ty_id, metadata)
-    }
-}
 
-pub fn decode_storage_key_from_hash<K: ?Sized>(
```
jsdw, line 133: I removed this specific-to-StaticStorageKey function and made a slightly more general `consume_hash_returning_bytes` one that could then be used in a couple of other places too, to cut down the number of util functions about

subxt/src/storage/storage_key.rs
```diff
@@ -186,32 +172,67 @@ impl StorageKey for Vec<scale_value::Value> {
...
-            let Some((hasher, ty_id)) = hashers_and_ty_ids_iter.next() else {
-                // Still bytes left, but no hashers and type ids anymore to pull from: this is an unexpected error.
-                return Err(StorageAddressError::UnexpectedAddressBytes.into());
-            };
-            strip_storage_hash_bytes(cursor, hasher)?;
-            let decoded = DecodedValueThunk::decode_with_metadata(cursor, *ty_id, metadata)?;
-            let value = decoded.to_value()?;
-            result.push(value.remove_context());
-            n += 1;
+        for (hasher, ty_id) in hashers_and_ty_ids.iter() {
```
jsdw, line 183: I had a look and `DecodedValueThunk::decode_with_metadata` looks like it consumes _all_ input bytes (not sure why tbh!), so the value decoding would have failed when more than one hasher I think.

I just decoded directly into a Value and figured it may as well go into a for loop where we check for remaining bytes at the end.

subxt/src/storage/storage_key.rs
```diff
@@ -186,32 +172,67 @@ impl StorageKey for Vec<scale_value::Value> {
...
         }
-        *hashers_and_ty_ids = &hashers_and_ty_ids[n..]; // Advance cursor by n
+
         Ok(result)
     }
 }
 
+// Skip over the hash bytes (including any key at the end), returning bytes
+// representing the key if one exists, or None if the hasher has no key appended.
+fn consume_hash_returning_key_bytes<'a>(
```
jsdw, line 208: This just captures a thing we want to do in a few places, which is to skip over any bytes a StorageHasher represents, and then if there are bytes for the key, then return them.

Using this means you don't have to worry about consuming the right number of bytes anywhere else.

subxt/src/storage/storage_type.rs
```diff
@@ -311,7 +310,8 @@ where
     }
 }
 
-pub(crate) fn storage_hasher_type_id_pairs(
+/// Return a vec of hashers and the associated type IDs for the keys that are hashed.
+fn storage_hasher_type_id_pairs(
```
jsdw, line 314: making it not pub didn't seem to break anything offhand :D

subxt/src/storage/utils.rs
```diff
@@ -6,17 +6,12 @@
...
-use crate::{
-    error::{Error, StorageAddressError},
-    metadata::Metadata,
-};
+use crate::{error::Error, metadata::Metadata};
 
 /// Return the root of a given [`StorageAddress`]: hash the pallet name and entry name
 /// and append those bytes to the output.
-pub(crate) fn write_storage_address_root_bytes<Address: StorageAddress>(
+pub fn write_storage_address_root_bytes<Address: StorageAddress>(
```
jsdw, line 14: utils wasn't exporting anything before, so now the module is not exported at all so these fns don't need to be pub(crate) to hide themselves any more.

I moved the added fns that were only used in one place into the place they were used.

tadeohepperle (approved):
Looks all good to me!

PR #1458, opened 2024-02-29
Make storage_page_size for the LegacyBackend configurable

### LINKED ISSUE #1454: Speed-up account iteration
Dear subxt team,

I'm currently using latest `subxt v0.34.0` and I would like to iterate through all existing accounts of the Aleph Zero network. I'm following this example: https://github.com/paritytech/subxt/blob/master/subxt/examples/storage_iterating.rs

Unfortunately, it seems like iterating the accounts is extremely slow.
I know from earlier subxt versions, there was the possibility to specify an additional `page_size` param to the `iter()`. Is there another option I'm missing that allows me to do a fast iteration through all the accounts?

Thanks for all the help and efforts.

jsdw:
Closes #1454 by allowing the user to configure the number of storage entries returned when iterating using the `LegacyBackend`. To make use of this you'll have to manually instantiate a backend etc, ie something like this:

```rust
// Connect to node:
let rpc_client = subxt::backend::rpc::RpcClient::from_insecure_url("ws://localhost:9944").await?;

// Use this connection to make calls using our legacy backend:
let backend = LegacyBackend::builder()
    .storage_page_size(1024)
    .build(rpc_client);

// Use the legacy backend to drive our Subxt client:
let client = OnlineClient::<PolkadotConfig>::from_backend(Arc::new(backend)).await?;
```

I added a builder just to mirror the `UnstableBackend` interface more closely, and to be a place where potentially we can expose more configuration if need be. It's definitely not ideal that you only have one global page size setting per client, but it's also a bit of a stop gap until we migrate to the new backend which streams storage entries and thus has no such configuration.

### DIFF
--- a/subxt/src/backend/legacy/mod.rs
+++ b/subxt/src/backend/legacy/mod.rs
@@ -22,18 +22,56 @@ use std::task::{Context, Poll};
 // Expose the RPC methods.
 pub use rpc_methods::LegacyRpcMethods;
 
+/// Configure and build an [`LegacyBackend`].
+pub struct LegacyBackendBuilder<T> {
+    storage_page_size: u32,
+    _marker: std::marker::PhantomData<T>,
+}
+
+impl<T: Config> Default for LegacyBackendBuilder<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T: Config> LegacyBackendBuilder<T> {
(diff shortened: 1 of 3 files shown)


tadeohepperle (approved):
Looks good to me! :)

lexnv (approved):

PR #1459, opened 2024-02-29
testing: Test light client with substrate binary

lexnv:
This PR adds a `long-running` feature flag for the light-client testing.
All subxt tests will use the light client to connect to a substrate binary when the feature flag is enabled.
This means that every test will spawn a substrate binary and a smoldot client will warp sync to the head of the chain.

We currently have a minimal subset of tests for the lightclient.
This PR prepares the testing for validating the light client as the default subxt client.

A few tests were disabled because they were taking too long to complete, mostly due to the storage iterations.


testing/integration-tests/proc-macro/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "integration-tests-proc-macro"
```
jsdw, line 2: Nit: I'd call this crate something like `subxt-test-proc-macro` I think, so it's super easy to relate it to the `#[subxt_test]` attr it exports

subxt/Cargo.toml
```diff
@@ -61,7 +61,7 @@ unstable-metadata = []
 
 # Activate this to expose the Light Client functionality.
 # Note that this feature is experimental and things may break or not work as expected.
-unstable-light-client = ["subxt-lightclient", "tokio-stream"]
+unstable-light-client = ["subxt-lightclient", "tokio-stream", "tokio"]
```
jsdw, line 64: This means that `unstable-light-client` pulls in tokio unconditionally when enabled, but what about when the `web` feature is enabled and we want to then not pull in tokio for wasm compilation?

Maybe since none of it is used, we don't hit any compile issues anyway. I can't think of a way offhand to say "when we have `native` and `unstable-light-client`, pull in tokio too, otherwise dont' (except maybe some build step which would make understanding everything complicated).

lexnv: Yep, sadly I had to bring this in and enable my hackish patch to run the unstable backend with the light-client. I think this is going to be handled by https://github.com/paritytech/subxt/pull/1455.

Ideally, I wanted to test both the legacy-backwith the light-client and the unstable-backend. I think in the end we can just use the unstable-backend for light-client testing in our CI. I've used `cfg-aliases` for our testing module, looks good at a first glance, that may simplify a bit our feature-flags conditions

jsdw: I agree that we only need to test with the unstable backend!

Once I've done some refactoring of the light client stuff (ie the follow up to #1455) then we'll be able to configure it with an unstable backend rather than requiring tokio etc right away, but I think it's ok to do this here for now anyway, as long as the tokio dep being bought in doesn't break wasm compilation for the light client :)

testing/integration-tests/proc-macro/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    Error,
+};
+
+#[proc_macro_attribute]
+pub fn subxt_test(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let subxt_attr = match syn::parse::<SubxtTestAttr>(attr) {
+        Ok(subxt_attr) => subxt_attr,
+        Err(err) => return err.into_compile_error().into(),
+    };
+    let timeout_duration = subxt_attr.timeout.unwrap_or(60 * 5);
```
jsdw, line 20: I wonder how many tests actually require longer than this? Since its our macro, we might as well set default in it that we are happy using for all of our tests. I can see many (all?) tests have 800s timeouts set!

testing/integration-tests/build.rs
```diff
@@ -0,0 +1,9 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        lightclient: { any(feature = "unstable-light-client", feature = "unstable-light-client-long-running") },
+        fullclient: { all(not(feature = "unstable-light-client"), not(feature = "unstable-light-client-long-running")) },
```
tadeohepperle, line 7: dq: Why are these needed?

tadeohepperle (approved):
Looks good to me, nice to have this in subxt.

subxt/src/client/light_client/mod.rs
```diff
@@ -93,7 +93,29 @@ impl RawLightClient {
...
+        use crate::backend::unstable::UnstableBackend;
+        use std::sync::Arc;
+        let (backend, mut driver) = UnstableBackend::builder().build(rpc_client);
+        let future = async move {
+            use futures::StreamExt;
+            while let Some(val) = driver.next().await {
+                if let Err(e) = val {
+                    // This is a test; bail if something does wrong and try to
+                    // ensure that the message makes it to some logs.
+                    eprintln!("Error driving unstable backend in tests (will panic): {e}");
```
niklasad1, line 106: I think the panic is sufficient here :)

testing/integration-tests/proc-macro/src/lib.rs
niklasad1, line 1: I wonder whether you can extract this to a separate PR, because I reckon that we could merge it right away.
Because you added the to tokio feature to the light client which is probably a blocker get this merged quite soon

testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -76,7 +89,7 @@ async fn block_subscriptions_are_consistent_with_eachother() -> Result<(), subxt
     Ok(())
 }
 
-#[tokio::test]
+#[subxt_test(timeout = 800)]
```
niklasad1, line 92: looks great

subxt/Cargo.toml
niklasad1: This will break the light client WASM unless you make remove the default features `"time", "rt-multi-thread"` below

testing/integration-tests/build.rs
niklasad1: these are just way an easy way to create an alias for `#[cfg(fullclient)]` and `#[cfg(lightclient)` such that we don't have repeat that anymore in the code.

such as:

```rust
// test only for the lightclient
#[cfg(lightclient)]
#[test] 
fn foo () {}
```

testing/integration-tests/build.rs
```diff
@@ -0,0 +1,9 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        lightclient: { any(feature = "unstable-light-client", feature = "unstable-light-client-long-running") },
```
niklasad1, line 6: I would prefer `light-client` here if that's possible

testing/integration-tests/build.rs
niklasad1, line 7: full-client

testing/integration-tests/build.rs
lexnv: Ahh this doesn't work with `cfg_aliases`, it hits recursion limit for whatever reason.

From https://github.com/katharostech/cfg_aliases?tab=readme-ov-file#syntax-and-error-messages:
> The aliase names are restricted to the same rules as rust identifiers which, for one, means that they cannot have dashes ( - ) in them

testing/integration-tests/build.rs
lexnv: It allows us to say `cfg(lightclient)` instead of `cfg(any(feature = "unstable-light-client", feature = "unstable-light-client-long-running"))`, to not repeat the longer format

lexnv:
CI not running anymore closing and reopening

PR #1460, opened 2024-02-29
`scale-type-resolver` integration

tadeohepperle:
This PR updates subxt to the latest versions of `scale-value`, `scale-decode`, `scale-encode` and `scale-typgen` which rely on the `scale-type-resolver` crate James wrote.


subxt/src/utils/wrapper_opaque.rs
```diff
@@ -74,57 +74,45 @@ impl<T> WrapperKeepOpaque<T> {
...
+        let visitor = visitor::new(out, |_, _| {
+            // Check that the target shape lines up: any other shape but composite it wrong.
+            Err(Error::new(ErrorKind::WrongShape {
                 actual: Kind::Struct,
-                expected: type_id,
-            }));
-        };
-
-        // Check that the name also lines up.
-        if ty.path.ident().as_deref() != Some("WrapperKeepOpaque") {
```
tadeohepperle, line 98: Dq: We had this check here before, but with the new type resolvers, it does not seem possible anymore with the new resolving strategy. What should we do here?

subxt/src/tx/tx_payload.rs
```diff
@@ -65,10 +65,6 @@ pub struct Payload<CallData> {
     validation_hash: Option<[u8; 32]>,
 }
 
-/// A boxed transaction payload.
-// Dev Note: Arc used to enable easy cloning (given that we can't have dyn Clone).
-pub type BoxedPayload = Payload<Arc<dyn EncodeAsFields + Send + Sync + 'static>>;
```
tadeohepperle, line 70: Note: This was removed, because `EncodeAsFields` is not object safe anymore and in the public element room for subxt noone seemed to care about it not being supported anymore.

subxt/src/blocks/extrinsic_types.rs
```diff
@@ -420,18 +420,16 @@ where
 
     /// Decode and provide the extrinsic fields back in the form of a [`scale_value::Composite`]
     /// type which represents the named or unnamed fields that were present in the extrinsic.
-    pub fn field_values(
-        &self,
-    ) -> Result<scale_value::Composite<scale_value::scale::TypeId>, Error> {
```
tadeohepperle, line 425: Note: previously we were able to return `scale_value::Composite<u32>`, but that would require changes to `scale_value`, e.g. creating a new fn `decode_composite_as_fields`, that maps the type id correctly.

jsdw: I don't think that this is a deal breaker, but I also think that exposing a fewwstanding `decode_as_fields` in `scale-value` alongside `decode_as_type` would be good to have (and should be a patch bump release too :)). That way, `scale-value` provides all of the same capabilityies as before (and internally its `DecodeAsFields` impl will just use this freestanding function too)

subxt/src/utils/wrapper_opaque.rs
jsdw: I can see a couple of small arguments for exposing the enum name and type name (perhaps optionally only) in `scale-type-resolver`. For now, I think the point of this is that the type has slightly different encoding, and so we don't want to use it when the actual input wasn't also using it.

I think we can remove the check for now, but let's add a TODO in the code and/or issue to add support for optional names in `scale-type-resolver` :)

jsdw: Side note: this is good feedback too; if you do find other details that are missing like this, then please do make note of them, perhaps in issues in `scale-type-resolver` pointing to where a thing was needed, and then we can consider adding some bits in one go!

jsdw (approved):
Looks great to me, good job getting this done!

subxt/src/utils/wrapper_opaque.rs
```diff
@@ -74,57 +74,45 @@ impl<T> WrapperKeepOpaque<T> {
...
 
-        let Some(ty) = types.resolve(type_id) else {
-            return Err(Error::new(ErrorKind::TypeNotFound(type_id)));
-        };
-
-        // Do a basic check that the target shape lines up.
-        let scale_info::TypeDef::Composite(_) = &ty.type_def else {
-            return Err(Error::new(ErrorKind::WrongShape {
+        let visitor = visitor::new(out, |_, _| {
+            // Check that the target shape lines up: any other shape but composite it wrong.
```
niklasad1, line 86:
```suggestion
            // Check that the target shape lines up: any other shape but the composite is wrong.
```

niklasad1 (approved):

PR #1466, opened 2024-03-07
`subxt-core` crate

tadeohepperle:
Introduces a `subxt-core` crate that provides a `no_std` compatible subset of `subxt`'s functionality.


core/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+edition.workspace = true
+rust-version.workspace = true
+publish = true
+
+license.workspace = true
+readme = "README.md"
+repository.workspace = true
+documentation.workspace = true
+homepage.workspace = true
+description = "Sign extrinsics to be submitted by Subxt"
```
niklasad1, line 14: probably a copy-paste nit :)

core/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+keywords = ["parity", "subxt", "extrinsic", "no-std"]
+
+[features]
+default = ["std"]                                                                                                                                                # "std"
+std = [
+    "scale-info/std",
```
niklasad1, line 20: why are not all crates that has `default-features = false` added here?

For instance codec/std is not enabled and many more....

core/src/client/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Each client should be able to provide access to the following fields
+/// - runtime version
+/// - genesis hash
+/// - metadata
+
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""), Clone(bound = ""))]
+pub struct ClientMetadata<C: Config> {
+    pub genesis_hash: C::Hash,
```
niklasad1, line 13: Make private? 

I see no reason to make these pub when you have added getters from them

core/src/client/mod.rs
```diff
@@ -0,0 +1,58 @@
+use derivative::Derivative;
+
+use crate::{config::Config, metadata::Metadata};
+
+/// Each client should be able to provide access to the following fields
+/// - runtime version
+/// - genesis hash
+/// - metadata
+
```
niklasad1, line 9:
```suggestion
```

core/src/client/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+/// Runtime version information needed to submit transactions.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RuntimeVersion {
+    /// Version of the runtime specification. A full-node will not attempt to use its native
+    /// runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
+    /// `spec_version` and `authoring_version` are the same between Wasm and native.
+    pub spec_version: u32,
```
niklasad1, line 46: I think it's better to add "getter methods" to expose these, in case we would want to change internal implementation/representation at some point without breaking changes.

core/src/client/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::{config::Config, metadata::Metadata};
+
+/// Each client should be able to provide access to the following fields
+/// - runtime version
+/// - genesis hash
+/// - metadata
+
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""), Clone(bound = ""))]
+pub struct ClientMetadata<C: Config> {
```
niklasad1, line 12: not sure about the name `ClientMetadata` could be `ChainState`, `ClientState` or something.

Because it's not only metadata but runtime version and genesis hash....

tadeohepperle: Yes, I am also not sure about the name. I named it `ClientBase` before. `ClientState` sounds good, any opinions on this @jsdw @lexnv ?

core/src/config/substrate.rs
```diff
@@ -225,12 +227,69 @@ where
...
+/// A number type that can be serialized both as a number or a string that encodes a number in a
+/// string.
+///
+/// We allow two representations of the block number as input. Either we deserialize to the type
+/// that is specified in the block type or we attempt to parse given hex value.
+///
+/// The primary motivation for having this type is to avoid overflows when using big integers in
+/// JavaScript (which we consider as an important RPC API consumer).
+#[derive(Copy, Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(untagged)]
```
niklasad1, line 244: This type is only needed for the RPC stuff and I wonder whether we could move to the RPC stuff instead? 

IIRC, the RPC stuff doesn't compile for no-std and that's the reason why to separate it

core/README.md
```diff
@@ -0,0 +1,3 @@
+# Subxt-Core
+
+This library provides core functionality using in `subxt` and `subxt-signer`. It should be no-std compatible.
```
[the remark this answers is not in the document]
tadeohepperle: Oh yeah true, we currently just make sure that it compiles to `thumbv7em-none-eabi` which should be enough of a proof though.

core/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! # Subxt-core
+//!
+//! `#[no_std]` compatible core crate for subxt.
+
```
niklasad1, line 8: we probably want some more lints here such as 

```
#![cfg_attr(docsrs, feature(doc_cfg))]
```

to work with the macros you added in core/src/macros.rs

core/src/events/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Return the topics associated with this event.
+    pub fn topics(&self) -> &[T::Hash] {
+        &self.topics
+    }
+}
+
+/// Details for the given event plucked from the metadata.
+pub struct EventMetadataDetails<'a> {
+    pub pallet: PalletMetadata<'a>,
```
niklasad1, line 245: sorry for being annoying as hell but again maybe add getter for these instead of public fields?

core/src/config/substrate.rs
tadeohepperle: Yeah, but the problem I have is that the `SubstrateConfig` should be defined in `subxt-core` and this has the associated type `Header = SubstrateHeader<32, _>` which has a field, `number: u32,` that is tagged with 
```
[serde(
        serialize_with = "serialize_number",
        deserialize_with = "deserialize_number"
)]
```
where `deserialize_number` uses `NumberOrHex`. I am not sure if I can easily move all of these Serialize implmentations to the RPC stuff, especially because the `SubstrateHeader` is now in a different crate.

core/src/metadata/metadata_type.rs
```diff
@@ -72,3 +73,64 @@ impl codec::Decode for Metadata {
         subxt_metadata::Metadata::decode(input).map(Metadata::new)
     }
 }
+
+pub trait MetadatExt {
```
niklasad1, line 77:
```suggestion
pub trait MetadataExt {
```

core/src/metadata/metadata_type.rs
```diff
@@ -72,3 +73,64 @@ impl codec::Decode for Metadata {
...
+    ) -> Result<subxt_metadata::RuntimeApiMetadata, MetadataError>;
+
+    fn custom_value_by_name_err(
+        &self,
+        name: &str,
+    ) -> Result<subxt_metadata::CustomValueMetadata, MetadataError>;
+}
+
+impl MetadatExt for subxt_metadata::Metadata {
+    /// Identical to `metadata.pallet_by_name()`, but returns an error if the pallet is not found.
```
niklasad1, line 100: move the documentation to trait instead :)

core/src/metadata/metadata_type.rs
niklasad1, line 77: would be nice with some docs on the trait

core/src/runtime_api/mod.rs
```diff
@@ -69,11 +74,7 @@ pub trait RuntimeApiPayload {
 #[derive(Derivative)]
 #[derivative(
     Clone(bound = "ArgsData: Clone"),
-    Debug(bound = "ArgsData: std::fmt::Debug"),
```
niklasad1, line 72: why are the trait bounds removed?

core/src/tx/mod.rs
```diff
@@ -49,26 +49,14 @@ pub struct ValidationDetails<'a> {
...
 pub struct Payload<CallData> {
     pallet_name: Cow<'static, str>,
     call_name: Cow<'static, str>,
     call_data: CallData,
     validation_hash: Option<[u8; 32]>,
 }
 
-/// A boxed transaction payload.
-// Dev Note: Arc used to enable easy cloning (given that we can't have dyn Clone).
-pub type BoxedPayload = Payload<Arc<dyn EncodeAsFields + Send + Sync + 'static>>;
```
niklasad1, line 70: this doesn't work anymore?

I recall James talking about this as well at some point

core/src/utils/bits.rs
```diff
@@ -144,45 +146,43 @@ impl<Store: BitStore, Order: BitOrder> codec::Encode for DecodedBits<Store, Orde
...
     type Value<'scale, 'info> = DecodedBits<Store, Order>;
     type Error = scale_decode::Error;
+    type TypeResolver = R;
 
     fn unchecked_decode_as_type<'scale, 'info>(
         self,
         input: &mut &'scale [u8],
-        type_id: scale_decode::visitor::TypeId,
-        types: &'info scale_info::PortableRegistry,
+        type_id: &R::TypeId,
```
niklasad1, line 159: hmm, what are these changes? 

I thought the TypeResolver stuff wasn't integrated yet and done another PR?

signer/Cargo.toml
```diff
@@ -26,34 +27,35 @@ ecdsa = ["secp256k1"]
...
+pbkdf2 = { workspace = true, default-features = false }
+sha2 = { workspace = true, default-features = false }
+hmac = { workspace = true, default-features = false }
 zeroize = { workspace = true }
-bip39 = { workspace = true }
-schnorrkel = { workspace = true, optional = true }
-secp256k1 = { workspace = true, features = ["recovery", "global-context"], optional = true }
+bip39 = { workspace = true, default-features = false }
+schnorrkel = { workspace = true, optional = true, default-features = false }
+secp256k1 = { workspace = true, optional = true, default-features = false, features = ["alloc", "recovery"] } # features = ["recovery", "global-context"],
```
niklasad1, line 50: why is "global-context" removed?

core/src/tx/mod.rs
tadeohepperle: Yeah, We removed this because `EncodeAsFields` is no longer object-safe with the new `scale-type-resolver` integration. So we could add an `EncodeAsFieldsWithResolver<R>` trait to `scale_encode` to make it work. But we decided against it for now, because we asked in the public element room if anyone is actually using this and noone was apposed to it being removed.

core/src/utils/bits.rs
tadeohepperle: I built the subxt-core pr on top of the type-resolver PR because I did not want to resolve the conflicts in the end, where files in subxt are modified but deleted in the `subxt-core` PR.

signer/src/ecdsa.rs
```diff
@@ -161,9 +165,9 @@ impl Keypair {
         // From sp_core::ecdsa::sign_prehashed:
         let wrapped = Message::from_digest_slice(&message_hash).expect("Message is 32 bytes; qed");
         let recsig: RecoverableSignature =
-            SECP256K1.sign_ecdsa_recoverable(&wrapped, &self.0.secret_key());
+            Secp256k1::signing_only().sign_ecdsa_recoverable(&wrapped, &self.0.secret_key());
```
niklasad1, line 168: yeah ok, this probably because they some `lazy_static` stuff that doesn't compile for no-std?

then removing global-context makes sense

signer/src/ecdsa.rs
```diff
@@ -109,7 +112,8 @@ impl Keypair {
     pub fn from_seed(seed: Seed) -> Result<Self, Error> {
         let secret = SecretKey::from_slice(&seed).map_err(|_| Error::InvalidSeed)?;
         Ok(Self(secp256k1::Keypair::from_secret_key(
-            SECP256K1, &secret,
+            &Secp256k1::signing_only(),
```
niklasad1, line 115: yeah ok, this probably because they some lazy_static stuff that doesn't compile for no-std?

then removing global-context makes sense

signer/Cargo.toml
tadeohepperle: IIRC "global-context" is not compatible with `no-std` and we use `Secp256k1::signing_only()` and `Secp256k1::verification_only()` to get an appropriate context where needed.

signer/src/ecdsa.rs
```diff
@@ -192,31 +196,36 @@ pub fn verify<M: AsRef<[u8]>>(sig: &Signature, message: M, pubkey: &PublicKey) -
...
-    signature.verify(&wrapped, &public).is_ok()
+
+    Secp256k1::verification_only()
+        .verify_ecdsa(&wrapped, &signature, &public)
+        .is_ok()
 }
 
 /// An error handed back if creating a keypair fails.
-#[derive(Debug, PartialEq, thiserror::Error)]
+#[derive(Debug, PartialEq, Display, From)]
```
niklasad1, line 206: impl StdError for std?

```rust
#[cfg(feature = std)]
impl std::error::Error  for Error {}
```

tadeohepperle: What do you mean by this?

signer/src/lib.rs
```diff
@@ -40,9 +43,3 @@ pub use secrecy::{ExposeSecret, SecretString};
 // SecretUri's can be parsed from strings and used to generate key pairs.
 // DeriveJunctions are the "path" part of these SecretUris.
 pub use crypto::{DeriveJunction, SecretUri, SecretUriError, DEV_PHRASE};
-
-#[cfg(any(
```
niklasad1, line 44: nice

jsdw:
Could we split out making `subxt-signer` no-std compatible into a separate PR? I do remember already reviewing and looking at such a thing, so hopefully it would already encompass those past comments etc also if not already :)

core/src/lib.rs
jsdw: IMO we should have `#![deny(missing_docs)]` everywhere :)

signer/src/ecdsa.rs
niklasad1: I mean that the error type doesn't implement std::error::Error anymore.

So, do we want implement for std::error::Error if --feature std is enabled on the crate?

signer/src/sr25519.rs
```diff
@@ -187,24 +192,25 @@ pub fn verify<M: AsRef<[u8]>>(sig: &Signature, message: M, pubkey: &PublicKey) -
 }
 
 /// An error handed back if creating a keypair fails.
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, Display, From)]
```
niklasad1, line 195: same here, do we want to implement std::error::Error for the error type here?

subxt/src/backend/unstable/mod.rs
```diff
@@ -407,17 +406,15 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
                     Some(ev) => ev,
                 };
 
-                let runtime_details = match runtime_event {
+                      let runtime_details = match runtime_event {
```
niklasad1, line 409: the indentation looks strange, is mixing spaces/tabs perhaps?

subxt/src/client/offline_client.rs
```diff
@@ -21,6 +21,8 @@ pub trait OfflineClientT<T: Config>: Clone + Send + Sync + 'static {
     fn genesis_hash(&self) -> T::Hash;
     /// Return the provided [`RuntimeVersion`].
     fn runtime_version(&self) -> RuntimeVersion;
+    /// Return the Client Metadata.
+    fn client_metadata(&self) -> ClientState<T>;
```
niklasad1, line 25: I don't get why we need the client_metadata here since it contains metadata, genesis_hash and runtime_version? 
To it's redundant to the old APIs?

subxt/src/macros.rs
```diff
@@ -56,7 +56,7 @@ macro_rules! cfg_jsonrpsee_web {
 macro_rules! cfg_reconnecting_rpc_client {
 	($($item:item)*) => {
 		$(
-			#[cfg(all(feature = "unstable-reconnecting-rpc-client"))]
```
niklasad1, line 59: nice catch

niklasad1 (commented):
I like it and nice work but have a ponder of my remaining comments nothing big for my-side.

I'll take another go at after you have looked at my comments and as James suggested it may be nice to split the core crate and signer stuff in different PRs if possible. The signer stuff looks solid to me but may be slow without the OnceCell/lazy stuff... for no-std

subxt/src/backend/unstable/mod.rs
tadeohepperle: Oh, this is rustfmt giving up formatting [(without warning! see this discussion)](https://users.rust-lang.org/t/rustfmt-silently-fails-to-work/75485/4), if a single expression cannot be formatted to fit in the line length limit.

subxt/src/client/offline_client.rs
tadeohepperle: It is a bit redundant, I agree. But we need some bundled form of the client data to use it in the creation of signed extensions. They we previously generic over the type of client, but now the signed extensions live in `subxt-core` but the client traits should stay in `subxt`. I did not want to remove the old methods either because if you just want to get e.g. the `runtime_version` there is no need to clone the `Arc` the metadata is wrapped up in.

signer/src/ecdsa.rs
tadeohepperle: Oh true! This is a really good catch.

subxt/src/events/mod.rs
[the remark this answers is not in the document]
tadeohepperle: I think  `subxt::events` has the submodules `events_client` and `events_type`, so I prefer to keep it in a seperate directory.

core/src/config/extrinsic_params.rs
```diff
@@ -7,47 +7,10 @@
 //! [`crate::config::DefaultExtrinsicParams`] provides a general-purpose
 //! implementation of this that will work in many cases.
 
-use crate::{client::OfflineClientT, Config};
-use core::fmt::Debug;
+use crate::{client::ClientState, Config, ExtrinsicParamsError};
+use alloc::vec::Vec;
 
 use super::refine_params::RefineParams;
```
jsdw, line 13:
```suggestion
use super::refine_params::RefineParams;

```

core/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+};
+pub use error::{Error, ExtrinsicParamsError, MetadataError, StorageAddressError};
+pub use events::{EventDetails, Phase, StaticEvent};
+pub use metadata::Metadata;
+pub use signer::Signer;
+
+pub use utils::{
+    to_hex, AccountId32, Encoded, Era, MultiAddress, MultiSignature, Static, UncheckedExtrinsic,
+    WrapperKeepOpaque, Yes, H160, H256, H512,
+};
```
jsdw, line 40: My gut feeling here is that we are exposing too much at the top level. I think a good palce to start would be to expose very little or nothing at all here, so that the import paths make it clearer what the different types are eg `subxt_core::config::extrinsic_params::RefineParams` is long but makes it much clearer what `RefineParams` relates to than `subxt_core::RefineParams` :)

subxt/examples/runtime_apis_dynamic.rs
```diff
@@ -1,5 +1,5 @@
 #![allow(missing_docs)]
-use subxt::dynamic::Value;
+use subxt::ext::scale_value::Value;
```
jsdw, line 2: Nit: it's still exposed as `subxt::dynamic::Value` which is a little cleaner :)

subxt/Cargo.toml
```diff
@@ -53,7 +53,7 @@ jsonrpsee = [
 # Enable this to pull in extra Substrate dependencies which make it possible to
 # use the `sp_core::crypto::Pair` Signer implementation, as well as adding some
 # `From` impls for types like `AccountId32`. Cannot be used with "web".
-substrate-compat = ["sp-core", "sp-runtime"]
+substrate-compat = ["sp-core", "sp-runtime", "subxt-core/substrate-compat"]
```
jsdw, line 56: Do we need `sp-core` and `sp-runtime` here at all now?

core/src/constants/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::{
+    metadata::{DecodeWithMetadata, MetadataExt},
+    Error, Metadata, MetadataError,
+};
+
+/// Run validation logic against some constant address you'd like to access. Returns `Ok(())`
+/// if the address is valid (or if it's not possible to check since the address has no validation hash).
+/// Return an error if the address was not valid or something went wrong trying to validate it (ie
+/// the pallet or constant in question do not exist at all).
+pub fn validate_constant<Address: ConstantAddress>(
```
jsdw, line 22: Use in `subxt::constants::COnstantsClient::validate()`?

core/src/constants/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+                MetadataError::ConstantNameNotFound(address.constant_name().to_owned())
+            })?;
+        if actual_hash != expected_hash {
+            return Err(MetadataError::IncompatibleCodegen.into());
+        }
+    }
+    Ok(())
+}
+
+pub fn get_constant<Address: ConstantAddress>(
```
jsdw, line 40: Use in `subxt::constants::COnstantsClient::at()`?

core/src/events/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+    ApplyExtrinsic(u32),
+    /// Finalizing the block.
+    Finalization,
+    /// Initializing the block.
+    Initialization,
+}
+
+/// The event details.
+#[derive(Debug, Clone)]
+pub struct EventDetails<T: Config> {
```
jsdw, line 41: How hard would it be to port over `Events` too? 

Right now, the "entry point" for working with events in `subxt_core` is to get hold of a single event from some events bytes and then call `Events::decode_from` to decode them.

It'd be much nicer I think if the entry point was to get bytes representing all of the events (which is how they come back from a node) and then calling `Events::decode_from` on the bytes. To make this work I reckon we could:

- Move `Events` over to `subxt_core`, too
- Rename `Events::new` to `Events::decode_from` and make it pub.
- Get rid of `Events::block_hash()` and `ExtrinsicEvents::block_hash()`: it looks like this is just here to be helpful but isn't actually needed internally, and if people want to know the block hash then they can fetch the block first and then ask for events from it etc. This allows us to get rid of `block_hash` from `Events::decode_from`, making it cleaner.
- Make `Events::new_from_client` a freestanding function exposed in `subxt::events::new_from_client`; now we don't have anything mentioning `OnlineClient` in Events I think :)

jsdw:
I still need to have another pass or two over this, but generally I think that this is really good stuff!

One thing that I think we're missing right now is a coherent "story" behind the things exposed in `subxt_core`, ie it feels more like a "collection of things that we could make no-std" than a useful library in its own right right now.

I think we'll just need to iterate more to get there though! Things that would help are examples and documentation showing off the different workflows. 

One step towards this I think is for each module like `constants`/`events` etc could have module level docs showing example(s) of how you can use it to decode some bytes you've obtained or pull a constant out of the metadata or encode some storage query or runtime query or whatever (and then decode the bytes you'd get back). 

In the act of doing this we might also find that some bits can move back into Subxt or some more bits to `subxt-core` or whatever in order to give us this coherent interface. Overall, the goal isn't to put _everything possible_ in `subxt-core` but more to expose a useful subset of the Subxt functionality in a no-stf compatible crate :)

We'll also want to work more on the `transaction` module and add functionality to create and sign transactions to really make that module feel more complete; this is an important part of the story afterall and probably one of the main use cases of such a no-std crate.

core/src/lib.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! # Subxt-core
+//!
+//! `#[no_std]` compatible core crate for subxt.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+pub extern crate alloc;
```
jsdw, line 10: Don't need feature flag? :)

core/src/error.rs
```diff
@@ -0,0 +1,177 @@
...
+    },
+    /// A signed extension in use on some chain was not provided.
+    #[display(
+        fmt = "The chain expects a signed extension with the name {_0}, but we did not provide one"
+    )]
+    UnknownSignedExtension(String),
+    /// Some custom error.
+    #[display(fmt = "Error constructing extrinsic parameters: {_0}")]
+    #[cfg(feature = "std")]
+    Custom(CustomExtrinsicParamsError),
```
jsdw, line 156: Rather than removing custom errors entirely, I'd maybe do something like `scale-decode` does and say "if std, expect Display+Debug+std::error::Error impl, else just expect Display+Debug". relevant code snippet from `scale-decode`:

```rust
enum Error {
    // ...

    /// A custom error.
    #[from]
    #[display(fmt = "Custom error: {_0}")]
    Custom(Box<dyn CustomError>),
}

/// Anything implementing this trait can be used in [`ErrorKind::Custom`].
#[cfg(feature = "std")]
pub trait CustomError: std::error::Error + Send + Sync + 'static {}
#[cfg(feature = "std")]
impl<T: std::error::Error + Send + Sync + 'static> CustomError for T {}

/// Anything implementing this trait can be used in [`ErrorKind::Custom`].
#[cfg(not(feature = "std"))]
pub trait CustomError: core::fmt::Debug + core::fmt::Display + Send + Sync + 'static {}
#[cfg(not(feature = "std"))]
impl<T: core::fmt::Debug + core::fmt::Display + Send + Sync + 'static> CustomError for T {}
```

(Note: `std::error::Error` requires Debug and Display already, so we only mention those when we don't use that)

core/src/config/refine_params.rs
```diff
@@ -14,7 +14,7 @@ pub struct RefineParamsData<T: Config> {
 }
 
 impl<T: Config> RefineParamsData<T> {
-    pub(crate) fn new(account_nonce: u64, block_number: u64, block_hash: T::Hash) -> Self {
+    pub fn new(account_nonce: u64, block_number: u64, block_hash: T::Hash) -> Self {
```
jsdw, line 17: I guess this needs to be called in Subxt now, so can't be `(crate)`. 

Perhaps we should add `#[doc(hidden)]` instead and a comment that it is not a part of the public API? The reason for this would be so that we can change it without breaking anything. A reason against is that maybe users want to test custom signed exts by using this sort of thing though, so maybe just leaving it pub is ok :)

core/src/config/substrate.rs
jsdw: I'm ok with it being here; it _is_ an RPC type but it's also just a specific concrete type we decide to use in our `SubstrateConfig` impl, so it should live next to it I reckon :)

core/src/custom_values/custom_value_address.rs
```diff
@@ -33,28 +33,13 @@ impl CustomValueAddress for str {
...
 /// A static address to a custom value.
 #[derive(Derivative)]
-#[derivative(
-    Clone(bound = ""),
-    Debug(bound = ""),
-    Eq(bound = ""),
-    Ord(bound = ""),
-    PartialEq(bound = "")
-)]
+#[derivative(Clone(bound = ""), Debug(bound = ""))]
```
jsdw, line 38: IIRC Niklas picked up on this already but I thought the bounds were all back, so: any reason for no more Eq/Ord/PartialEq here? :)

core/src/events/mod.rs
```diff
@@ -194,8 +187,8 @@ pub struct EventDetails<T: Config> {
 }
 
 impl<T: Config> EventDetails<T> {
-    // Attempt to dynamically decode a single event from our events input.
-    fn decode_from(
+    /// Attempt to dynamically decode a single event from our events input.
+    pub fn decode_from(
```
jsdw, line 191: Perhaps no need for this to be `pub` now that `Events` is our logical "entry point" into this stuff? :)

core/src/events/mod.rs
```diff
@@ -45,7 +52,8 @@ impl<T: Config> std::fmt::Debug for Events<T> {
 }
 
 impl<T: Config> Events<T> {
-    pub(crate) fn new(metadata: Metadata, block_hash: T::Hash, event_bytes: Vec<u8>) -> Self {
+    /// Create a new [`Events`] instance from the given bytes.
+    pub fn decode_from(metadata: Metadata, event_bytes: Vec<u8>) -> Self {
```
jsdw, line 56: Woop, looks great, nice to see Events moved into here :)

core/src/metadata/metadata_type.rs
jsdw: Still needs doing by the looks of it @tadeohepperle :)

core/src/tx/mod.rs
```diff
@@ -49,15 +49,7 @@ pub struct ValidationDetails<'a> {
...
-#[derive(Derivative)]
-#[derivative(
-    Clone(bound = "CallData: Clone"),
-    Debug(bound = "CallData: std::fmt::Debug"),
-    Eq(bound = "CallData: std::cmp::Eq"),
-    Ord(bound = "CallData: std::cmp::Ord"),
-    PartialEq(bound = "CallData: std::cmp::PartialEq"),
-    PartialOrd(bound = "CallData: std::cmp::PartialOrd")
-)]
+#[derive(Clone, Debug)]
```
jsdw, line 52: More removed bounds that we could put back? :)

core/src/utils/account_id.rs
```diff
@@ -100,16 +105,16 @@ impl AccountId32 {
...
+    #[display(fmt = "Base 58 requirement is violated")]
     BadBase58,
-    #[error("Length is bad")]
+    #[display(fmt = "Length is bad")]
     BadLength,
-    #[error("Invalid checksum")]
+    #[display(fmt = "Invalid checksum")]
     InvalidChecksum,
-    #[error("Invalid SS58 prefix byte.")]
+    #[display(fmt = "Invalid SS58 prefix byte.")]
```
jsdw, line 117: Perhaps worth an `if std, impl std::error::Error` just to be consistent with how we ususally treat errors?

core/src/utils/wrapper_opaque.rs
```diff
@@ -83,7 +86,7 @@ impl<T> EncodeAsType for WrapperKeepOpaque<T> {
         use scale_encode::error::{Error, ErrorKind, Kind};
 
         let visitor = visitor::new(out, |_, _| {
-            // Check that the target shape lines up: any other shape but the composite is wrong.
+            // Check that the target shape lines up: any other shape but composite it wrong.
```
jsdw, line 89: Fix not quite right hehe :P

```suggestion
            // Check that the target shape lines up: any other shape but composite is wrong.
```

core/src/utils/wrapper_opaque.rs
```diff
@@ -113,8 +116,6 @@ impl<T, R: TypeResolver> Visitor for WrapperKeepOpaqueVisitor<T, R> {
     ) -> Result<Self::Value<'scale, 'info>, Self::Error> {
         use scale_decode::error::{Error, ErrorKind};
```
jsdw, line 118: I'd like to keep that TODO unless there's an issue for it; just a remindfer for future us to add the extra check here in the future :)

core/src/dynamic.rs
```diff
@@ -19,16 +16,16 @@ pub use scale_value::{At, Value};
...
-// Lookup constants dynamically.
+/// Lookup constants dynamically.
 pub use crate::constants::dynamic as constant;
 
-// Lookup storage values dynamically.
+/// Lookup storage values dynamically.
 pub use crate::storage::dynamic as storage;
 
-// Execute runtime API function call dynamically.
+/// Execute runtime API function call dynamically.
```
jsdw, line 28: Nit: IIRC adding comments to types like this that have more comments leads to a concatenation of this comment + the one on the actual type (but I'd have to actually check that!). That may be why they weren't doc comments before

core/src/dynamic.rs
```diff
@@ -66,8 +63,8 @@ impl DecodedValueThunk {
     pub fn encoded(&self) -> &[u8] {
         &self.scale_bytes
     }
-    /// Decode the SCALE encoded storage entry into a dynamic [`DecodedValue`] type.
-    pub fn to_value(&self) -> Result<DecodedValue, Error> {
+    /// Decode the SCALE encoded storage Value into a dynamic [`DecodedValue`] type.
```
jsdw, line 66: I think "storage entry" makes more sense than "storage Value" (the Value is what it would be decoded into possibly, but not here)


```suggestion
    /// Decode the SCALE encoded storage entry into a dynamic [`DecodedValue`] type.
```

core/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub mod custom_values;
+pub mod dynamic;
+pub mod error;
+pub mod events;
+pub mod metadata;
+pub mod runtime_api;
+pub mod signer;
+pub mod storage;
+pub mod tx;
+pub mod utils;
```
jsdw, line 25: Nice; much cleaner :)

core/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub use config::Config;
+pub use error::Error;
+pub use metadata::Metadata;
+pub use signer::Signer;
```
jsdw, line 30: I'd move `signer.rs` into the tx folder and export it from here and expose as `subxt_core::tx::Signer`. Why? Because signing is specific to txs, so keeping it in the same place (rather than exposing a top level `subxt_core::Signer` as now feels more right :)

signer/Cargo.toml
```diff
@@ -15,7 +15,7 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa", "subxt", "std", "native"]
+default = ["sr25519", "ecdsa", "subxt", "std"]
```
jsdw, line 18: Lovely to get rid of a feature flag!

subxt/src/backend/legacy/mod.rs
```diff
@@ -181,20 +181,16 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
 
     async fn current_runtime_version(&self) -> Result<RuntimeVersion, Error> {
         let details = self.methods.state_get_runtime_version(None).await?;
-        Ok(RuntimeVersion {
-            spec_version: details.spec_version,
-            transaction_version: details.transaction_version,
-        })
+        Ok(RuntimeVersion::new(
+            details.spec_version,
```
jsdw, line 185: The only thing I don't reallt like about `RuntimeVersion` becoming a private struct with accessors etc is that because `spec_version` and `transaction_version` are actually the same type, it's now a little easier to accidentally mix them up here.

subxt/src/client/offline_client.rs
```diff
@@ -21,6 +21,8 @@ pub trait OfflineClientT<T: Config>: Clone + Send + Sync + 'static {
     fn genesis_hash(&self) -> T::Hash;
     /// Return the provided [`RuntimeVersion`].
     fn runtime_version(&self) -> RuntimeVersion;
+    /// Return the [subxt_core::client::ClientState] (metadata, runtime version and genesis hash).
+    fn client_state(&self) -> ClientState<T>;
```
jsdw, line 25: If we have this method, then let's give it a default impl too since it's just a concat of a few of the above :)

subxt/src/client/online_client.rs
```diff
@@ -286,7 +285,17 @@ impl<T: Config> OnlineClient<T> {
     /// Return the runtime version.
     pub fn runtime_version(&self) -> RuntimeVersion {
         let inner = self.inner.read().expect("shouldn't be poisoned");
-        inner.runtime_version.clone()
+        inner.runtime_version
+    }
+
+    /// Return the [subxt_core::client::ClientState] (metadata, runtime version and genesis hash).
+    pub fn client_state(&self) -> ClientState<T> {
+        let inner = self.inner.read().expect("shouldn't be poisoned");
```
jsdw, line 293: I guess even with a default impl we'll want this bit anyay though to avoid multiple locks, but it could be seen as an optimisation anyway, so it's all good :)

subxt/src/macros.rs
jsdw: I guess now we'll only use this if _some_ of the feature flag exists :'(

jsdw:
This is great stuff; left another round of comments but all pretty minor things, and we can thing more about docs etc later :)

@tadeohepperle if you have time to address the small comments and get things compiling then I'm happy to see this merge now!

subxt/src/backend/legacy/mod.rs
niklasad1: yeah agree, maybe we could make them pub and #[doc(hidden)] instead.

Because it's in subxt-core it won't work unless these are fields public

core/src/constants/constant_address.rs
```diff
@@ -31,22 +33,15 @@ pub trait ConstantAddress {
 #[derivative(
     Clone(bound = ""),
     Debug(bound = ""),
-    Eq(bound = ""),
-    Ord(bound = ""),
```
niklasad1, line 35: no Ord impl anymore?

core/src/constants/mod.rs
niklasad1: yes, perhaps it's better to move this to the constant client stuff

core/src/custom_values/custom_value_address.rs
```diff
@@ -33,28 +33,19 @@ impl CustomValueAddress for str {
...
-/// Used to signal whether a [`CustomValueAddress`] can be decoded.
-pub struct Yes;
-
 /// A static address to a custom value.
 #[derive(Derivative)]
 #[derivative(
     Clone(bound = ""),
     Debug(bound = ""),
-    Eq(bound = ""),
-    Ord(bound = ""),
```
niklasad1, line 45: same here no Ord?

testing/integration-tests/src/full_client/codegen/polkadot.rs
niklasad1, line 1: Is these changes related to this refactoring? Lots of changes at least

testing/no-std-tests/Cargo.toml
```diff
@@ -7,7 +7,9 @@ resolver = "2"
 
 [dependencies]
 subxt-metadata = { path = "../../metadata", default-features = false }
-subxt-signer = { path = "../../signer", default-features = false, features = ["sr25519"] }
+subxt-core = { path = "../../core", default-features = false }
+subxt-signer = { path = "../../signer", default-features = false, features = ["subxt", "sr25519"] }
+subxt-macro = { path = "../../macro" }                                                                         # Not used yet, but will be used here very soon, to verify the generated code is applicable.
```
niklasad1, line 12: I'm pretty slow and that follow the comment :package: 

I suppose you mean that we don't test that the generated code is no-std?

niklasad1 (approved):
awesome, great job.

just some small comments

testing/integration-tests/src/full_client/codegen/polkadot.rs
tadeohepperle: Yeah, there should be a lot of changes, because we use a lot of types from `subxt_core` now instead of pulling them from `subxt`. Additionally we use rust prelude types from `alloc` instead of from `std`.

jsdw (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/parity-tech-update-for-march/7226/1

PR #1475, opened 2024-03-11
Rework light client

jsdw:
The main changes:

- Move almost all light client code into `subxt_lightclient`; only a little glue in Subxt to integrate it.
- `LightClient` struct to create light client instance given relay chain spec, and to create parachain connections on top of this.
- `LightClientRpc` returned from each of the above which is the means to interact with the chain.
- Subxt implements `RpcClientT` for this, so that it can be used to construct an `OnlineClient`.
- Function provided for low level control but not exported yet (maybe best to hide Smoldot away unless somebody has a good reason to need to configure it directly.
- Each light client RPC connection has an associated `BackgroundTask`. This isn't exposed yet, but can eventually be in order that users can drive it and exert backpressure if need be (I'm unsure what effect that would have on Smoldot though!). 

Other changes along the way:
- light client background task is now per-chain-connection so no need to combine and then separate messages from different chains, and in the future we can opt to drive each one manually if we want.
- Handle error messages on subscriptions, too.
- Generally split errors up so that we get quite specific erros back when constructing, calling or getting responses back from the light client stuff.

Things not done:
- Exposing something like `OnlineClient::from_chain_spec()`; need a relay + parachain version etc, prob need a builder or something to make it nice. Instead I made it easier to instantiate an OnlineClient from an RPC client (`impl Into`) so it's actually pretty easy anyway now.
- I didn't add a nice `from_url` to the `ChainSpecConfig`. I staretd doing this but it involved pulling in extra deps and feature flags etc, and the code was just easier to keep in `subxt`. So I exposed a utility function to fetch chain spec from url instead. I'm ok with this because it's actually only really a dev use case anyway; we prob shouldn't encourage it too much!


lightclient/src/background.rs
```diff
@@ -1,43 +1,47 @@
...
+pub type SubscriptionResponse = Result<
+    (
+        SubscriptionId,
+        mpsc::UnboundedReceiver<Result<Box<RawValue>, JsonRpcError>>,
+    ),
+    LightClientRpcError,
+>;
+
+/// Type of subscription IDs we can get back.
+pub type SubscriptionId = String;
```
jsdw, line 31: It was previously an unsigned int, but even if Smoldot _currently_ does that, I didn't think we should rely on it since it probably also expects us to treat the valeus as opaque.

lightclient/src/background.rs
```diff
@@ -46,29 +50,160 @@ pub enum FromSubxt {
...
+
+        let bg_handle = BackgroundTaskHandle { to_backend: tx };
+
+        (bg_task, bg_handle)
+    }
+
+    /// Run the background task, which:
+    /// - Forwards messages/subscription requests to Smoldot from the front end.
+    /// - Forwards responses back from Smoldot to the front end.
+    pub async fn run(self) {
```
jsdw, line 147: A simplified version of what was called `start_task` before. No longer takes in any args, so the API for `BackgroundTask` is super simple; just `.run().await` it to start it running (and in the future we can return something users can poll which will just wrap this).

lightclient/src/background.rs
```diff
@@ -46,29 +50,160 @@ pub enum FromSubxt {
...
+        let mut channels = self.channels;
+        let mut data = self.data;
+
+        loop {
+            tokio::pin! {
+                let from_front_fut = channels.from_front.next().fuse();
+                let from_back_fut = channels.from_back.next().fuse();
+            }
+
+            futures::select! {
```
jsdw, line 158: I moved to this I think to avoid allocating somewhere or annoying lifetime things; can't remember exactly now. It just takes two channels as inputs though and so I think should have no cancellation concerns (this is the usual footgun with `select!`

lightclient/src/background.rs
```diff
@@ -100,87 +227,67 @@ struct PendingSubscription {
...
+    /// The unsubscribe method to call when the user drops the receiver
+    /// part of the channel.
+    unsubscribe_method: String,
 }
 
 /// The state of the subscription.
 struct ActiveSubscription {
     /// Channel to send the subscription notifications back to frontend.
-    sender: mpsc::UnboundedSender<Box<RawValue>>,
+    notification_sender: mpsc::UnboundedSender<Result<Box<RawValue>, JsonRpcError>>,
```
jsdw, line 239: I tweaked subscriptions a bit to:

a) return any subscription notification errors that come back (previously these were being ignored I think) and
b) send back the ID and repsonse channel rather than one channel for ID and one for responses, just to simplify usage a little.

lightclient/src/chain_config.rs
```diff
@@ -0,0 +1,71 @@
...
+#[non_exhaustive]
+#[derive(thiserror::Error, Debug)]
+pub enum ChainConfigError {
+    /// The provided chain spec is the wrong shape.
+    #[error("Invalid chain spec format")]
+    InvalidSpecFormat,
+}
+
+/// Configuration to connect to a chain.
+pub struct ChainConfig<'a> {
```
jsdw, line 18: This can grow to have more options like about max json RPC subs etc if we want to expose them. A user can then either build this config and configure it, or pass a chain spec `&str` directly, which will convert into this easily.

lightclient/src/rpc.rs
```diff
@@ -0,0 +1,132 @@
...
+        result: Box<RawValue>,
+    },
+    NotificationError {
+        /// RPC method that generated the notification.
+        method: String,
+        /// Subscription ID.
+        subscription_id: String,
+        /// Result.
+        error: Box<RawValue>,
+    },
```
jsdw, line 39: This file is copied from the previous impl in `background.rs`, but I added this `NotificationError` to handle the case when we get errors back in subscriptions. Worth double checking that this is right :)

subxt/src/utils/fetch_chain_spec.rs
```diff
@@ -0,0 +1,114 @@
...
+    InvalidUrl,
+    #[error("Cannot fetch chain spec: Invalid URL scheme.")]
+    InvalidScheme,
+    #[error("Cannot fetch chain spec: Handshake error establishing WS connection.")]
+    HandshakeError,
+}
+
+cfg_jsonrpsee! {
+    /// Fetch a chain spec from an RPC node at the given URL.
+    pub async fn fetch_chainspec_from_rpc_node(url: impl AsRef<str>) -> Result<Box<RawValue>, FetchChainspecError> {
```
jsdw, line 24: I tried putting a fn like this (`from_url`) on `ChainConfig` but it required pulling lots of deps in etc, and since it's a dev use case mostly I was sortof ok with just exposing this util function instead in Subxt, since it's easier to do it here. Not the cleanest though!

subxt/examples/light_client_parachains.rs
```diff
@@ -16,85 +12,34 @@ const ASSET_HUB_SPEC: &str =
...
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    // The smoldot logs are informative:
+    // The lightclient logs are informative:
     tracing_subscriber::fmt::init();
 
-    // Connecting to a parachain is a multi step process.
+    // Instantiate a light client with the Polkadot relay chain,
+    // and connect it to Asset Hub, too.
+    let (lightclient, polkadot_rpc) = LightClient::relay_chain(POLKADOT_SPEC)?;
```
lexnv, line 20: Looks amazing!

lexnv (approved):
LGTM! I like the idea of having a background task per chain ID, this should isolate things quite nicely 👍

jsdw:
I added back a compile error and docsrs thing that went missing, and updated the book, swapping the examples around so now connecting to parachains is the basic example, and connecting to a local node is more advanced :)

lightclient/src/background.rs
tadeohepperle: Looks good, btw you asked me exactly this question "What is the usual usual footgun with select!" when interviewing me as a job applicant about a year ago.

lightclient/src/background.rs
```diff
@@ -100,87 +227,67 @@ struct PendingSubscription {
     /// Send the method response ID back to the user.
```
tadeohepperle, line 227:
```suggestion
    /// the method response ID back to the user.
```

lightclient/src/background.rs
```diff
@@ -253,129 +357,133 @@ impl<TPlatform: PlatformRef, TChain> BackgroundTask<TPlatform, TChain> {
...
                     tracing::trace!(target: LOG_TARGET, "Received subscription id={sub_id} chain={chain_id:?}");
 
-                    let (sub_id_sender, active_subscription) = pending_subscription.into_parts();
-                    if sub_id_sender.send(Ok(result)).is_err() {
+                    let (sub_tx, sub_rx) = mpsc::unbounded_channel();
+
+                    // Send the method response and a channel to receive notifications back.
+                    if pending_subscription
+                        .response_sender
+                        .send(Ok((sub_id.clone(), sub_rx)))
```
tadeohepperle, line 395: Could we avoid allocation here by using `Arc<str>` for subscription ids? Probably irrelevant though, feel free to ignore.

lightclient/src/lib.rs
```diff
@@ -2,52 +2,259 @@
...
+    ///
+    /// Panics when called outside of a `tokio` runtime context.
+    ///
+    /// ## Web
+    ///
+    /// If smoldot panics, then the promise created will be leaked. For more details, see
+    /// https://docs.rs/wasm-bindgen-futures/latest/wasm_bindgen_futures/fn.future_to_promise.html.
+    pub fn parachain<'a>(
+        &self,
+        chain_config: impl Into<ChainConfig<'a>>,
```
tadeohepperle, line 141: This interface with the `relay_chain` and `parachain` functions looks really clean.

tadeohepperle (approved):
I think the interface is well designed, good job!

lightclient/src/background.rs
jsdw: I need to make a note somewhere because if I move to doing a code review style task then `future::select!` is def a candidate to put somewhere :D

lightclient/src/background.rs
jsdw: I'd guess it's not worth while because we (technically, but not really in reality) need to allocate a `String` anyway for the deserializing bit, and then the `Arc` is another alloc, and we onlyclone it once anyways :)

lightclient/src/background.rs
niklasad1: you are good, streams are cancel-safe.

I like tokio::select better because it doesn't require pinning the futures

jsdw: I normally prefer the function call version to avoid some footguns but iirc I had an issue that made it hard to use it nicely; I'll have to try again at some point!

Polkadot-Forum:

https://forum.polkadot.network/t/parity-tech-update-for-march/7226/1

PR #1476, opened 2024-03-11
rpc v2: chainhead support multiple finalized block hashes in `FollowEvent::Initialized`

niklasad1:
Companion for https://github.com/paritytech/polkadot-sdk/pull/3445 where it's now possible that the initialized event
can contain multiple finalized block hashes.

### DIFF
--- a/subxt/src/backend/unstable/follow_stream_unpin.rs
+++ b/subxt/src/backend/unstable/follow_stream_unpin.rs
@@ -35,9 +35,11 @@ pub struct FollowStreamUnpin<Hash: BlockHash> {
     // Futures for sending unpin events that we'll poll to completion as
     // part of polling the stream as a whole.
     unpin_futs: FuturesUnordered<UnpinFut>,
-    // Each new finalized block increments this. Allows us to track
-    // the age of blocks so that we can unpin old ones.
-    rel_block_num: usize,
+    // Each time a new finalized block is seen, we give it an age of `next_rel_block_age`,
+    // and then increment this ready for the next finalized block. So, the first finalized
+    // block will have an age of 0, the next 1, 2, 3 and so on. We can then use `max_block_life`
+    // to say "unpin all blocks with an age < (next_rel_block_age-1) - max_block_life".
+    next_rel_block_age: usize,
     // The latest ID of the FollowStream subscription, which we can use
     // to unpin blocks.
     subscription_id: Option<Arc<str>>,
@@ -117,17 +119,16 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                     let mut finalized_block_hashes =
                         Vec::with_capacity(details.finalized_block_hashes.len());
 
+                    // Pin each of the finalized blocks. None of them will show up again (except as a
+                    // parent block), and so they can all be unpinned immediately at any time. Increment
+                    // the block age for each one, so that older finalized blocks are pruned first.
                     for finalized_block in &details.finalized_block_hashes {
-                        let rel_block_num = this.rel_block_num;
-
-                        // Pin this block, but note that it can be unpinned any time since it won't show up again (except
-                        // as a parent block, which we are ignoring at the moment).
-                        // The first finalized block gets the starting block_num.
+                        let rel_block_age = this.next_rel_block_age;
                         let block_ref =
-                            this.pin_unpinnable_block_at(rel_block_num, *finalized_block);
+                            this.pin_unpinnable_block_at(rel_block_age, *finalized_block);
 
                         finalized_block_hashes.push(block_ref);
-                        this.rel_block_num += 1;
+                        this.next_rel_block_age += 1;
                     }
 
                     FollowStreamMsg::Event(FollowEvent::Initialized(Initialized {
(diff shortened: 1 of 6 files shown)


subxt/src/backend/unstable/mod.rs
```diff
@@ -321,7 +321,9 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
             .events()
             .filter_map(|ev| {
                 let out = match ev {
-                    FollowEvent::Initialized(init) => Some(init.finalized_block_hash.into()),
+                    FollowEvent::Initialized(init) => {
+                        init.finalized_block_hashes.last().map(|b| b.clone().into())
```
niklasad1, line 325: we could refactor latest_finalized_block_ref to return all block_hashes but this should be equivalent to the old code.

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -117,11 +118,15 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
                     let rel_block_num = this.rel_block_num;
                     // Pin this block, but note that it can be unpinned any time since it won't show up again (except
                     // as a parent block, which we are ignoring at the moment).
-                    let block_ref =
-                        this.pin_unpinnable_block_at(rel_block_num, details.finalized_block_hash);
+
```
jsdw, line 121: Nit: remove this newline and update the comment above now?

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -117,11 +118,15 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
...
+                    let finalized_block_hashes = details
+                        .finalized_block_hashes
+                        .iter()
+                        .map(|h| this.pin_unpinnable_block_at(rel_block_num, *h))
```
jsdw, line 125: I believe that `rel_block_num` should be incremented for each finalized block, so that the oldest one is unpinned first etc.

subxt/src/backend/unstable/mod.rs
jsdw: Yeah, I don't think there's any reason to report the older ones anyway; we just want to know the most recent one so that we can submit txs etc with it in mind :)

testing/integration-tests/src/full_client/client/unstable_rpcs.rs
```diff
@@ -122,11 +126,12 @@ async fn chainhead_unstable_storage() {
 
     let mut blocks = rpc.chainhead_unstable_follow(false).await.unwrap();
     let event = blocks.next().await.unwrap().unwrap();
-    let hash = match event {
-        FollowEvent::Initialized(init) => init.finalized_block_hash,
+    let hashes = match event {
+        FollowEvent::Initialized(init) => init.finalized_block_hashes,
```
jsdw, line 130: Nit: Personally for these tests I'd just do something like:

```rust
let hash = match event {
    FollowEvent::Initialized(init) => init.finalized_block_hashes.last().unwrap(),
    _ => panic!("Unexpected event"),
};
```

The spec guarantees we'll get at least 1 anyway

subxt/src/backend/unstable/mod.rs
```diff
@@ -444,9 +451,7 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
         &self,
     ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
         self.stream_headers(|ev| match ev {
-            FollowEvent::Initialized(ev) => {
-                vec![ev.finalized_block_hash]
-            }
+            FollowEvent::Initialized(init) => init.finalized_block_hashes,
```
jsdw, line 454: When we add reconnection stuff in, I guess we might tweak eg this stream_finalized function to try and pick up from where they left off (ie keep track of the latest finalized block hash and if see an init event, ignore any blocks we get back before this latest one).

But for now it's all good!

jsdw (approved):
LGTM, Just a couple of things!

jsdw (approved):
LGTM!

subxt/src/backend/unstable/follow_stream_unpin.rs
```diff
@@ -114,14 +115,23 @@ impl<Hash: BlockHash> Stream for FollowStreamUnpin<Hash> {
...
+                        Vec::with_capacity(details.finalized_block_hashes.len());
+
+                    for finalized_block in &details.finalized_block_hashes {
+                        // Pin this block, but note that it can be unpinned any time since it won't show up again (except
+                        // as a parent block, which we are ignoring at the moment).
+                        let block_ref =
+                            this.pin_unpinnable_block_at(rel_block_num, *finalized_block);
+
+                        finalized_block_hashes.push(block_ref);
+                        rel_block_num += 1;
```
lexnv, line 130: nit: would `this.rel_block += 1` work here?

lexnv (approved):
LGTM! Thanks for fixing this 🙏

subxt/src/backend/unstable/follow_stream_unpin.rs
jsdw: Good spot; now there are multiple we should increment `this.rel_block_num` here!

niklasad1: the old code didn't increment that but sure lemme fix that..

PR #1477, opened 2024-03-13
`no_std` compatibility for `subxt-signer`

tadeohepperle:
This PR makes subxt-signer no-std compatible. This was factored out into a seperate PR because the[`subxt-core` PR ](https://github.com/paritytech/subxt/pull/1466) got a bit big. For this reason, this PR might already make more changes than necessary to the dependency structure, but it is all to make the integration with the main `subxt-core` PR a smooth experience.


signer/src/crypto/secret_uri.rs
```diff
@@ -115,10 +117,10 @@ impl std::str::FromStr for SecretUri {
 }
 
 /// This is returned if `FromStr` cannot parse a string into a `SecretUri`.
-#[derive(Debug, Copy, Clone, PartialEq, thiserror::Error)]
+#[derive(Debug, Copy, Clone, PartialEq, Display)]
 pub enum SecretUriError {
     /// Parsing the secret URI from a string failed; wrong format.
-    #[error("Invalid secret phrase format")]
+    #[display(fmt = "Invalid secret phrase format")]
```
jsdw, line 123: I agree with Niklas in the subxt-core PR; let's add:

```rust
#[cfg(feature = "std")]
impl std::error::Error for SecretUriError {}
```

To this and any other Error type. This means that on "std" the errors will impl `Error`

signer/src/ecdsa.rs
```diff
@@ -192,31 +196,36 @@ pub fn verify<M: AsRef<[u8]>>(sig: &Signature, message: M, pubkey: &PublicKey) -
...
+
+    Secp256k1::verification_only()
+        .verify_ecdsa(&wrapped, &signature, &public)
+        .is_ok()
 }
 
 /// An error handed back if creating a keypair fails.
-#[derive(Debug, PartialEq, thiserror::Error)]
+#[derive(Debug, PartialEq, Display, From)]
 pub enum Error {
```
jsdw, line 207: impl `std::error::Error` when std :)

signer/src/utils.rs
```diff
@@ -17,10 +17,19 @@
...
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "std")] {
+                    $(#[$attr])*
+                    $vis fn $name() -> &'static $ty {
+                        static VAR: std::sync::OnceLock<$ty> = std::sync::OnceLock::new();
+                        VAR.get_or_init(|| { $expr })
+                    }
+                } else {
+                    $(#[$attr])*
+                    $vis fn $name() -> $ty {
```
jsdw, line 29: I wonder whether there is a way to have this also be `&'static $ty` (ideally enabling "std" wouldn't change the type signature and thus risk breaking anything).

Offhand I can't think of a good solution to this (we went with this approach in the first place to avoid needing `OnceCell` and `critical-sections` which was annoying per-platform feature flags etc).

Let's just emit nothing here in no-std, ie just don't provide dev keys. This would mean that "std" feature was purely additive which is nice.

Later, we could have this expose two versions of each function; one that is like here and just computes the thing the whole time (eg `$name_new()`), and one that returns `&'static $ty` (eg `$name()`) and is only available on "std". But for that we'd need to be able to concat tokens to build the names which requires another dep like `paste`. We can leave this for another time I think.

jsdw: Oh actually ignore this; `once_static` it only used for acouple of internal funcs, so I think we are ok actually!

I still feel uneasy about the different type signatures being hidden in the macro though so perhaps we remove the magic here, and in the one place its used we just duplicate the fns or something?

signer/src/crypto/secret_uri.rs
```diff
@@ -115,10 +117,10 @@ impl std::str::FromStr for SecretUri {
...
     InvalidFormat,
 }
```
jsdw, line 126: I left another comment about this but resolved it to start fresh:

The `once_static!` usage below returns `&'static regex::Regex` in "std" and `regex::Regex` in no-std. This is hidden in the `once_static!` macro. Being internal code, the type difference doesn't matter asl ong as it compiles in both cases anyway. What I do think is scary is that the `once_static!` macro hides this change, and in the future we might use it on something else (something public maybe) and forget that it emits different type signs in std and no-std.

I looked into `regex` a bit, and it already has had effort put into optimising its own Clone impl with internal ref counters etc.

So, let's just use `once_static_cloned` here, which has a nice identical-type-signature "no-std" fallback, and delete the `once_static` macro entirely!

testing/no-std-tests/src/main.rs
```diff
@@ -34,14 +34,14 @@ extern crate alloc;
...
+    // Subxt Metadata compiles:
     use codec::Decode;
     let bytes: alloc::vec::Vec<u8> = alloc::vec![0, 1, 2, 3, 4];
     subxt_metadata::Metadata::decode(&mut &bytes[..]).expect_err("invalid byte sequence");
 
     const METADATA: &[u8] = include_bytes!("../../../artifacts/polkadot_metadata_small.scale");
     subxt_metadata::Metadata::decode(&mut &METADATA[..]).expect("should be valid metadata");
+
+    // Subxt Signer compiles:
+    let _alice = subxt_signer::sr25519::dev::alice();
```
jsdw, line 46: Can we extend this a little to:

- create, sign and verify some random payload
- do that with each of sr25519 and ecdsa

That'd give me confidence that we fully support no-std at runtime when using this crate (because there are differences to how we sign etc in std vs no-std) :)

signer/Cargo.toml
```diff
@@ -15,7 +15,8 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa", "subxt", "native"]
+default = ["sr25519", "ecdsa", "subxt", "std", "native"]
+std = []
```
jsdw, line 19: When `std` is enabled we should also enable std on the other crates I think in order to potentially benefit from any std specific things they add

jsdw (commented):
Looks good; just some minor comments :)

testing/no-std-tests/src/main.rs
tadeohepperle: Okay I tried this and it works fine for `sr25519` but for `ecdsa` we get a build error from `arm-none-eabi-gcc`:

```

error: failed to run custom build command for `secp256k1-sys v0.9.2`

Caused by:
  process didn't exit successfully: `/home/tadeo/code/subxt/testing/no-std-tests/target/debug/build/secp256k1-sys-6ae4a38afd356447/build-script-build` (exit status: 1)
  --- stdout
  TARGET = Some("thumbv7em-none-eabi")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-unknown-linux-gnu")
  cargo:rerun-if-env-changed=CC_thumbv7em-none-eabi
  CC_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CC_thumbv7em_none_eabi
  CC_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  RUSTC_LINKER = None
  cargo:rerun-if-env-changed=CROSS_COMPILE
  CROSS_COMPILE = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  cargo:warning=Compiler version doesn't include clang or GCC: "arm-none-eabi-gcc" "--version"
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("true")
  CARGO_CFG_TARGET_FEATURE = Some("dsp,mclass,thumb-mode,thumb2,v5te,v6,v6k,v6t2,v7")
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em-none-eabi
  CFLAGS_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em_none_eabi
  CFLAGS_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=CC_thumbv7em-none-eabi
  CC_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CC_thumbv7em_none_eabi
  CC_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  RUSTC_LINKER = None
  cargo:rerun-if-env-changed=CROSS_COMPILE
  CROSS_COMPILE = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  cargo:warning=Compiler version doesn't include clang or GCC: "arm-none-eabi-gcc" "--version"
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("dsp,mclass,thumb-mode,thumb2,v5te,v6,v6k,v6t2,v7")
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em-none-eabi
  CFLAGS_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em_none_eabi
  CFLAGS_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=CC_thumbv7em-none-eabi
  CC_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CC_thumbv7em_none_eabi
  CC_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CC
  TARGET_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  RUSTC_LINKER = None
  cargo:rerun-if-env-changed=CROSS_COMPILE
  CROSS_COMPILE = None
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
  cargo:warning=Compiler version doesn't include clang or GCC: "arm-none-eabi-gcc" "--version"
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("dsp,mclass,thumb-mode,thumb2,v5te,v6,v6k,v6t2,v7")
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em-none-eabi
  CFLAGS_thumbv7em-none-eabi = None
  cargo:rerun-if-env-changed=CFLAGS_thumbv7em_none_eabi
  CFLAGS_thumbv7em_none_eabi = None
  cargo:rerun-if-env-changed=TARGET_CFLAGS
  TARGET_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:warning=depend/secp256k1/contrib/lax_der_parsing.c:7:10: fatal error: string.h: No such file or directory
  cargo:warning=    7 | #include <string.h>
  cargo:warning=      |          ^~~~~~~~~~
  cargo:warning=compilation terminated.
  cargo:warning=In file included from depend/secp256k1/src/util.h:13,
  cargo:warning=                 from depend/secp256k1/src/field.h:10,
  cargo:warning=                 from depend/secp256k1/src/group.h:10,
  cargo:warning=                 from depend/secp256k1/src/precomputed_ecmult_gen.c:3:
  cargo:warning=/usr/lib/gcc/arm-none-eabi/13.2.0/include/stdint.h:9:16: fatal error: stdint.h: No such file or directory
  cargo:warning=    9 | # include_next <stdint.h>
  cargo:warning=      |                ^~~~~~~~~~
  cargo:warning=compilation terminated.

  --- stderr


  error occurred: Command "arm-none-eabi-gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-g" "-fno-omit-frame-pointer" "-mthumb" "-march=armv7e-m" "-I" "depend/secp256k1/" "-I" "depend/secp256k1/include" "-I" "depend/secp256k1/src" "-I" "wasm/wasm-sysroot" "-Wall" "-Wextra" "-Wno-unused-function" "-Wno-unused-parameter" "-DSECP256K1_API=" "-DENABLE_MODULE_ECDH=1" "-DENABLE_MODULE_SCHNORRSIG=1" "-DENABLE_MODULE_EXTRAKEYS=1" "-DENABLE_MODULE_ELLSWIFT=1" "-Dprintf(...)=" "-DECMULT_GEN_PREC_BITS=4" "-DECMULT_WINDOW_SIZE=15" "-DUSE_EXTERNAL_DEFAULT_CALLBACKS=1" "-DENABLE_MODULE_RECOVERY=1" "-o" "/home/tadeo/code/subxt/testing/no-std-tests/target/thumbv7em-none-eabi/debug/build/secp256k1-sys-74715f18bcbdb71a/out/c928380082b86401-precomputed_ecmult_gen.o" "-c" "depend/secp256k1/src/precomputed_ecmult_gen.c" with args "arm-none-eabi-gcc" did not execute successfully (status code exit status: 1).


```

jsdw: Yup let's just comment out any such tests that don't work for now with a note :)

jsdw (approved):
LGTM, nice one!

testing/no-std-tests/src/main.rs
jsdw: Ah I see you already did that!

signer/src/crypto/secret_uri.rs
```diff
@@ -115,14 +117,14 @@ impl std::str::FromStr for SecretUri {
...
 
 /// This is returned if `FromStr` cannot parse a string into a `SecretUri`.
-#[derive(Debug, Copy, Clone, PartialEq, thiserror::Error)]
+#[derive(Debug, Copy, Clone, PartialEq, Display)]
```
niklasad1, line 120: impl std::error::Error for this one? I saw that James commented on it but I can't find the impl :)

testing/no-std-tests/Cargo.toml
```diff
@@ -7,6 +7,7 @@ resolver = "2"
 
 [dependencies]
 subxt-metadata = { path = "../../metadata", default-features = false }
+subxt-signer = { path = "../../signer", default-features = false, features = ["sr25519"] }                     # Note: feature "ecdsa" does not compile to `no-std` right now
```
niklasad1, line 10: It's probably a good idea to document it somewhere or raise an issue for it

testing/no-std-tests/src/main.rs
```diff
@@ -34,14 +34,27 @@ extern crate alloc;
...
     subxt_metadata::Metadata::decode(&mut &METADATA[..]).expect("should be valid metadata");
+
+    // Subxt Signer compiles:
+    use subxt_signer::sr25519;
+    let keypair = sr25519::dev::alice();
+    let message = b"Hello!";
+    let _signature = keypair.sign(message);
+    let _public_key = keypair.public_key();
+
+    // Note: `ecdsa` is not compiling for the `thumbv7em-none-eabi` target.
```
niklasad1, line 52: same here would be nice to have an issue for the ecdsa no-std stuff

niklasad1 (approved):

PR #1491, opened 2024-03-21
chore: bump sp-core, sp-runtime and sp-keyring

niklasad1:


jsdw (approved):

niklasad1:
it's seems like these bump leaks the `--feature std` in the workspace which breaks the no-std stuff

lexnv (approved):

jsdw:
Oh strange; I wonder whether their default features changed or something then; it'd be great to find out anyway. There's no urgent rush in getting the release out if we can fix this one first :)

niklasad1:
I think it's should be fine after you moved sp-crypto-hashing, tested it locally

jsdw:
Oh interesting; not sure why that would have made a difference but hey ho :)

PR #1497, opened 2024-03-21
unstable_rpc: Add transactionBroadcast and transactionStop

lexnv:
This PR adds the `transaction_unstable_broadcast` and `transaction_unstable_stop` APIs.

While at it, added a test to check that the broadcasted transaction gets into a finalized block.


subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -288,6 +288,28 @@ impl<T: Config> UnstableRpcMethods<T> {
 
         Ok(TransactionSubscription { sub, done: false })
     }
+
+    /// Broadcast the transaction on the p2p network until the
+    ///[`Self::transaction_unstable_stop`] is called.
```
niklasad1, line 293:
```suggestion
    /// [`Self::transaction_unstable_stop`] is called.
```

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -288,6 +288,28 @@ impl<T: Config> UnstableRpcMethods<T> {
...
+            .request("transaction_unstable_broadcast", rpc_params![to_hex(tx)])
+            .await
+    }
+
+    /// Stop the broadcasting process of the transaction.
+    ///
+    /// The operation ID is obtained from the [`Self::transaction_unstable_broadcast`] method.
+    ///
+    /// Returns an error if the operation ID does not correspond to any active transaction for this connection.
+    pub async fn transaction_unstable_stop(&self, operation_id: &str) -> Result<(), Error> {
```
niklasad1, line 308: maybe you could add a simple test just to make sure that node the refuses an invalid operation id

niklasad1 (approved):

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -288,6 +288,28 @@ impl<T: Config> UnstableRpcMethods<T> {
 
         Ok(TransactionSubscription { sub, done: false })
     }
+
+    /// Broadcast the transaction on the p2p network until the
+    ///[`Self::transaction_unstable_stop`] is called.
+    ///
+    /// Returns an operation ID that can be used to stop the broadcasting process.
+    /// Returns `None` if the server cannot handle the request at the moment.
+    pub async fn transaction_unstable_broadcast(&self, tx: &[u8]) -> Result<Option<String>, Error> {
```
jsdw, line 297: Has the PR to rename to `v1` merged in Substrate already? :)

jsdw (approved):
A couple of minor things but yay for more RPCs :)

When this merges we can also move our TX submitting to use this method and thus be "reconnect" safe at some point which will be nice (though much more complicated, because aside from searching blocks for the TX we need to periodically validate it to know whether to stop checking for it etc).

subxt/src/backend/unstable/rpc_methods.rs
niklasad1: yupp, https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/rpc-spec-v2/src/transaction/api.rs#L50

testing/integration-tests/src/full_client/client/unstable_rpcs.rs
```diff
@@ -309,3 +309,65 @@ async fn next_operation_event<
...
+        let extrinsics = finalized.extrinsics().await.unwrap();
+        let block_extrinsics = extrinsics
+            .iter()
+            .map(|res| res.unwrap())
+            .collect::<Vec<_>>();
+
+        // All blocks must contain the timestamp, we expect the next extrinsics to be our own.
+        let Some(tx) = block_extrinsics.get(1) else {
+            continue;
+        };
```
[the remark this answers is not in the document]
lexnv: There is just a small inconsistency with this:
- create_signed_offline returns the extrinsic bytes including the prefix of `compact encoding length`
- `ExtrinsicDetails` will strip the compact prefix:
https://github.com/paritytech/subxt/blob/a211e98c64b9fb42fafd4cd867ad439502527767/subxt/src/blocks/extrinsic_types.rs#L224

I've adjusted the code to reflect this, not sure if we should do anything else to align for ex with polkadot-api 🤔

PR #1501, opened 2024-03-25
feat(signer): ethereum implementation

### LINKED ISSUE #1467: Add ethereum compatible signer to subxt_signer?
Anybody can implement a custom signer for Subxt to sign things as they need.

An example which I put together just now which _seems_ to work for creating ethereum-compatible signatures and using `AccountId20`s is here:

https://gist.github.com/jsdw/13240f9341e433ea639b89d0d4235c8b

I'd be interested to check with other eth-compatible chains whether this solution works for all of them, and if so, it might be a good idea to add this in a more complete form to `subxt_signer`.

For now though, if anybody needs eth style signing, please check out the Gist above and let me know whether it works for you!

(this has come up previously in https://github.com/paritytech/subxt/issues/1037)

ryanleecode:
### Description

Fixes: #1467

Adds an ethereum signer implementation to the signer crate. 

### Notes
- The signer shares the ecdsa signing method since the only thing that is different is how the message is hashed
- Instead of using `for _ in 0..20` to add fuzzing to the tests, I'm using [proptest](https://proptest-rs.github.io/proptest/)

### Testing

You can test this against moonbeam using chopsticks with the gist provided in the original issue:

https://gist.github.com/jsdw/13240f9341e433ea639b89d0d4235c8b

`npx @acala-network/chopsticks@latest --config=chopsticks.yml`

```rs
const ALITH: (&str, &str) = (
    "b6c8ddd1748f5c360e7efaa9a2f45a7872cc67a79599a889d07f40aa68b6de75",
    "adcaa4dabe8d19461eea3380dee40d4ca334f89a099b269c1ce66afb8de69461",
);
const BALTHASAR: (&str, &str) = (
    "f61ca7bfb5ee9e710cd259e7bcd0723fbb0dfb204442701c0ccb82e9c2f03a2b",
    "4d691af46d07a9b5d72e3bd89fc5f4c949708a249a0386bab32552789b8258fe",
);
```

chopsticks.yml
```yaml
endpoint: wss://wss.api.moonbeam.network/
mock-signature-host: true
block: 4784047
db: ./db.sqlite

import-storage:
  System:
    Account:
      -
        -
          - "0x1ba2efe6ea21ee83e8925cd00fa6f562893919aa"
        - providers: 1
          data:
            free: "100000000000000000000000"
        -
          - "0x2ea13d9f33c8bf0c6dcdba10c739fe92857058af"
        - providers: 1
          data:
            free: "100000000000000000000000"
        -
          - "0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac"
        - providers: 1
          data:
            free: "100000000000000000000000"
  TechCommitteeCollective:
    Members: ["0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac"]
  CouncilCollective:
    Members: ["0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac"]
  TreasuryCouncilCollective:
    Members: ["0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac"]
  AuthorFilter:
    EligibleRatio: 100
    EligibleCount: 100

```


signer/Cargo.toml
```diff
@@ -15,7 +15,7 @@ description = "Sign extrinsics to be submitted by Subxt"
 keywords = ["parity", "subxt", "extrinsic", "signer"]
 
 [features]
-default = ["sr25519", "ecdsa", "subxt", "std", "native"]
+default = ["sr25519", "ecdsa", "eth", "subxt", "std", "native"]
```
niklasad1, line 18: I wonder whether this should be a default feature because I reckon "eth accounts/signer" is rather niche use case and most users won't need it.

Doesn't matter that much it's easy to opt-out anyway, @jsdw thoughts?

signer/src/eth.rs
```diff
@@ -0,0 +1,269 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! An ethereum signer implementation.
+use derive_more::{Display, From};
```
niklasad1, line 6:
```suggestion

use derive_more::{Display, From};
```

signer/src/eth.rs
```diff
@@ -0,0 +1,269 @@
+// Copyright 2019-2023 Parity Technologies (UK) Ltd.
```
niklasad1, line 1:
```suggestion
// Copyright 2019-2024 Parity Technologies (UK) Ltd.
```

signer/src/eth.rs
```diff
@@ -0,0 +1,269 @@
...
+
+//! An ethereum signer implementation.
+use derive_more::{Display, From};
+use hex::FromHex;
+use keccak_hash::keccak;
+use secp256k1::{Keypair, Message, Secp256k1, SecretKey};
+
+/// An ethereum signer implementation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EthereumSigner(Keypair);
```
niklasad1, line 13: I would prefer to call this `Signer` or `KeyPair` because it's already in module named `eth` and the `Ethereum` prefix can be omitted.

signer/src/eth.rs
```diff
@@ -0,0 +1,269 @@
...
+
+impl From<crate::ecdsa::Keypair> for EthereumSigner {
+    fn from(kp: crate::ecdsa::Keypair) -> Self {
+        EthereumSigner(kp.0)
+    }
+}
+
+impl EthereumSigner {
+    /// Construct an ethereum signer from a hex-encoded private key.
+    pub fn from_private_key_hex(hex: &str) -> Result<EthereumSigner, Error> {
```
niklasad1, line 23: Can we unify this API with existing [from_seed, from_uri, from_phrase](https://github.com/paritytech/subxt/blob/master/signer/src/sr25519.rs#L65-#L124) or is it some other reason?

signer/src/eth.rs
niklasad1: can we wrap the `ecdsa::KeyPair` here instead?

signer/src/eth.rs
ryanleecode: i added it. i think its ok not to add test for these since its just a passthrough to the ecdsa module which already have tests.

ryanleecode: I kind of feel like it would be better if they just did `let keypair: eth::Keypair = ecdsa::Keypair::from_phrase(...).into()` though

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+    };
+    let message_hash = keccak(message.as_ref());
+    let wrapped =
+        Message::from_digest_slice(message_hash.as_bytes()).expect("Message is 32 bytes; qed");
+    let pub_key = secp256k1::PublicKey::from_slice(&pub_key.0).expect("valid public key");
+
+    Secp256k1::verification_only()
+        .verify_ecdsa(&wrapped, &signature, &pub_key)
+        .is_ok()
+}
```
jsdw, line 179: Could you do the same thing you did with `sign`, ie does this work out to be identical to `ecdsa::verify` except for the hashing?

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+
+        fn sign(&self, signer_payload: &[u8]) -> T::Signature {
+            self.sign(signer_payload).into()
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
jsdw, line 266: I reckon it's worth also copying the moonbeam tests here to do a sanity chekc that our impl lines up with theirs:

https://github.com/moonbeam-foundation/moonbeam/blob/e70ee0d427dfee8987d5a5671a66416ee6ec38aa/primitives/account/src/lib.rs#L217

(For the other impls we can sanity check against substrate, but here we want to sanity check that we align with eth chains)

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+
+impl From<ecdsa::Keypair> for Keypair {
+    fn from(kp: ecdsa::Keypair) -> Self {
+        Self(kp)
+    }
+}
+
+impl Keypair {
+    /// Construct a keypair from a hex-encoded private key.
+    pub fn from_private_key_hex(hex: &str) -> Result<Self, Error> {
```
niklasad1, line 32: Technically this API supports `"<64 hex digits>"` but not `"0x<64 hex digits>`.

I think we should remove this because it's better to have a uniform API because [SecretUri](https://docs.rs/subxt-signer/latest/subxt_signer/struct.SecretUri.html) already supports hex strings with the format "0x<64 hex digits>".

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+    #[display(fmt = "{_0}")]
+    Inner(ecdsa::Error),
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
+/// Dev accounts, helpful for testing but not to be used in production,
+/// since the secret keys are known.
+pub mod dev {
```
jsdw, line 201: Can you confirm whether moonbeam uses these dev keys at all? Mythos doesn't; they have their own well known private keys for dev accounts.

If moonbeam uses these dev accounts then let's keep them, else if the well known keys are shared between eth impls let's use those instead, else let's ditch them.

The readme gives account details but also says that they come from substrate keys, so we can generate the test accounts and check that they match here if not done already: https://github.com/moonbeam-foundation/moonbeam/blob/e70ee0d427dfee8987d5a5671a66416ee6ec38aa/README.md?plain=1#L56

signer/src/eth.rs
jsdw: Yeah, let's remove this; the right way should be to do any hex stuff yourself and then call `from_seed` on the 32 bytes.

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+    let pub_key = secp256k1::PublicKey::from_slice(&pub_key.0).expect("valid public key");
+
+    Secp256k1::verification_only()
+        .verify_ecdsa(&wrapped, &signature, &pub_key)
+        .is_ok()
+}
+
+/// An error handed back if creating the keypair fails.
+#[derive(Debug, PartialEq, Display, From)]
+pub enum Error {
```
jsdw, line 183: I'd remove the `Inner` variant (remember that this is a public interface, so `Inner` is a bit weird to expose to users).

For now I'd lean to just doing `type Error = escda::Error` if identical errors from this and ecdsa (less future proof but I'm not too worried about that). If there are unique errors from this module (doesn't look like it offhand) then have an `enum Error` which looks a lot like the ecdsa one but with whatever changes needed.

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+            assert_eq!(Signer::sign(&keypair, msg_as_bytes), keypair.sign(msg_as_bytes));
+        }
+
+        #[test]
+        fn check_account_id(keypair in keypair()) {
+            let account_id = {
+                let uncompressed = keypair.0.0.public_key().serialize_uncompressed();
+                let hash = keccak(&uncompressed[1..]).0;
+                let hash20 = hash[12..].try_into().expect("should be 20 bytes");
+                AccountId20(hash20)
```
jsdw, line 314: By having this code here, the assumption is that it is a known good way to create account IDs. Perhaps link to the moonbeam source or something where they do the same to show why this is the correct way?

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+        type Address = super::AccountId20;
+        type Signature = super::Signature;
+        type Hasher = subxt::config::substrate::BlakeTwo256;
+        type Header =
+            subxt::config::substrate::SubstrateHeader<u32, subxt::config::substrate::BlakeTwo256>;
+        type ExtrinsicParams = subxt::config::SubstrateExtrinsicParams<Self>;
+        type AssetId = u32;
+    }
+
+    type Signer = dyn subxt::tx::Signer<StubEthRuntimeConfig>;
```
jsdw, line 285: Nit; perhaps call this `SubxtSigner` to make it clear below what this is?

signer/src/eth.rs
```diff
@@ -0,0 +1,346 @@
...
+            let inner = secp256k1::Keypair::from_secret_key(
+                &Secp256k1::new(),
+                &secret,
+            );
+
+            Keypair(ecdsa::Keypair(inner))
+        }
+    }
+
+    proptest! {
```
jsdw, line 299: Cool! I haven't used proptest before so this is neat to see! I wonder whether it's worth opening another issue to add proptest tests for the other signers too?

signer/Cargo.toml
jsdw: The only reason I tend to default to having all feature flags is for nicer IDE/docs support (otherwise loads of stuff is greyed out). So this is consistent with what I did with the other signers, but I'd be very open to figuring out an approach which has IDEs/docs working nicely even without feature flags enabled here; I just haven't bothered :)

jsdw:
Looking good so far @ryanleecode! Just left a few thoughts and comments :)

signer/src/eth.rs
ryanleecode: By use you mean its prefunded on their devnet?

ryanleecode:
So there is a difference in the `derive` implementation as well. For ethereum the derivation path uses soft junctions and it looks like this `m/44'/60'/0'/0/0`, where `m` is the mnemonic .. i.e. `format!("{}/44'/60'/0'/0/0", DEV_PHRASE)`. 

But the ecdsa implementation does not allow soft junctions.  

```rs
pub fn derive<Js: IntoIterator<Item = DeriveJunction>>(
    &self,
    junctions: Js,
) -> Result<Self, Error> {
    let mut acc = self.0.secret_key().clone().secret_bytes();
    for junction in junctions {
        match junction {
            DeriveJunction::Soft(_) => return Err(Error::SoftJunction),
            DeriveJunction::Hard(junction_bytes) => {
                acc = ("Secp256k1HDKD", acc, junction_bytes)
                    .using_encoded(sp_crypto_hashing::blake2_256)
            }
        }
    }
    Self::from_seed(acc)
}
```

signer/src/eth.rs
ryanleecode: https://github.com/paritytech/subxt/issues/1506

signer/src/eth.rs
ryanleecode: https://github.com/paritytech/subxt/issues/1506

ryanleecode:
I got rid of `from_uri` and `derive` because they don't make sense in the eth context. for an eth account to be valid it has to be derivation path `m/44'/60'/0'/0/{index}`. Their notation is also different from substrate's because it uses apostrophe to denote hard derivations.

Instead the `from_phrase` function now has a third parameter `index`, which denotes the "account number" in the eth specific derivation path. 

Maybe we should also bring back the private key constructor since I imagine it would be handy.

Cargo.toml
```diff
@@ -159,7 +159,7 @@ keccak-hash = { version = "0.10.0", default-features = false }
 secrecy = "0.8.0"
 sha2 = { version = "0.10.8", default-features = false }
 zeroize = { version = "1", default-features = false }
-eth_checksum = { version = "0.1.2", default-features = false }
+ethaddr = { version = "0.2.2", default-features = false }
```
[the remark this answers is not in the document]
ryanleecode: this library is just to compute the correct checksum for implementing `Display`

https://github.com/nlordell/ethprim-rs/blob/main/ethaddr/src/checksum.rs#L10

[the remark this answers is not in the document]
niklasad1: why do we need the public key "checksum" a.k.a EIP55 in the Display impl?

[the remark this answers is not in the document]
ryanleecode: its the defacto standard for getting the string representation of an address. Both web3.js and ethers always return the address as a checksum address.

web3 js: https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#note-on-checksum-addresses
ethers: https://docs.ethers.org/v6/api/address/#getAddress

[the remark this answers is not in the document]
niklasad1: Gotcha, I'm still not super happy to add this dependency and I'm wondering if we could implement this checksum ourselves?

The only concern is whether this library is maintained properly or not...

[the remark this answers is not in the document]
ryanleecode: yeah we can just impl ourselves. fixed :)

jsdw (approved):
as it stands now I'm happy to merge this once CI is happy; Great job @ryanleecode!

lexnv (approved):
LGTM! Nice once @ryanleecode ! 🚀

Polkadot-Forum:

https://forum.polkadot.network/t/paritytech-update-for-april/7646/1

PR #1505, opened 2024-03-27
rpc: add full support reconnecting rpc client

### LINKED ISSUE #1423: reconnecting rpc client
## What's the current behavior

Currently, subxt has recently added experimental support for a reconnecting rpc client, which is primarily
for users of the legacy RPC API

The main benefits of an "automatic reconnecting client" are as follows:
- Automatically reconnect on spurious connection hickups
- Avoid re-create all active subscriptions; for instance, it is possible that plenty of storage subscriptions are lost and need to be re-subscribed.
- The API for reconnecting/understand when a connection is closed is complicated in subxt
- The subscription API in subxt may return None when the connection is closed or that the subxt can't keep up with the server (the server is producing items faster than subxt can consume).
- Rotate between a list of endpoints (not supported yet).

Then the drawbacks are:
- Subscription messages may be lost when reconnecting.
- Subscription can have side effects, such as `submitAndWatch`
- Automatic re-connecting doesn't work well for the RPC spec v2 chain API, the "chain_head_unstable_follow subscription" returns the subscriptionID which will be used to query further states from the chain_head API. Thus, when reconnecting, the subscription ID must be "updated" to use the new one, and the old subscription ID from the old connection may not work.

## What subxt needs

subxt needs two different behaviors as long as it's compatible with both the RPC APIs.

### Legacy RPC

Provide the notion that users can decide whether they want to re-subscribe on certain subscriptions or not.

```rust
let rpc = Client::builder()
// This is the global default policy, unless specified elsewhere this is used.
.retry_policy(ExponentialBackoff::from_millis(100).max_delay(Duration::from_secs(10)))
// For the `tx_submitAndWatch` we don't attempt at resubmitting the tx again; and let the user handle this if wanted
.local_policy(subxt::rpc::Method::TransactionSubmitAndWatch, Policy::DoNotRetry)
```

It's basically not possible to get the state of transaction if connection is closed and up to the user if re-submit or not for such cases.

Use this as the default behavior.

### RPC v2 spec

Because the API is very different it should be quite straight forward to implement auto-reconnect directly in subxt such as andlet users decide whether
the want to create some subscriptions, calls on startup... which is "initial"

```rust
async fn client_auto_reconnect(
    initial: Vec<RpcCalls>,
    retry_policy: RetryPolicy,
) -> Result<(), ()> {
    let driver = create_rpc_backend(initial_state.clone(), retry_policy).await?;

    loop {
        // Reconnect if err
        if let Err(RpcDisconnect(r)) = driver.next() {
            driver = create_rpc_backend(initial.clone(), retry_policy).await?;
        }
    }
}
```

Also implement custom re-connecting logic should much easier to implement for users. It just a stream that returns an Err(Disconnect) then handle it. 

### Both

Improve subscription API such that indicates clearly whether the connection was closed or that subscription couldn't keep with the server. See https://github.com/paritytech/jsonrpsee/pull/1255 which is needed for this.

//cc @jsdw thoughts?

niklasad1:
This PR improves the reconnecting rpc client so that it works with the unstable RPC backend, which requires re-starting `chainhead_follow` along with other stuff that is required to update the state when a new connection is established.

Close [#1423](https://github.com/paritytech/subxt/issues/1423)

Some notes about the PR:

- RPC calls that are not part of "Backend API" such as "LegacyRpcMethods", are not "re-tried," and it's up to the user to handle such things. subxt now exposes two util methods: `retry, retry_stream` to retry calls, and subscriptions if one wants to retry such calls if the connection was lost.
- subxt is responsible for retrying RPC calls, and the rpc-reconnecting-client "only reconnects"
- Resubscribing is handled by subxt on re-connections, but subscriptions with side effects such as `author_submitAndWatch` are not retried, and it's up to the caller to handle such things.
- When re-subscribing to blocks during a reconnection, it's possible that a few blocks may be missed. In such scenarios, subxt will emit an error on the stream itself.


subxt/src/backend/unstable/mod.rs
```diff
@@ -145,20 +146,30 @@ impl<T: Config> UnstableBackend<T> {
...
+            let follow_handle = follow_handle.clone();
             let block_refs = f(ev).into_iter();
 
             futures::stream::iter(block_refs).filter_map(move |block_ref| {
-                let sub_id = sub_id.clone();
                 let methods = methods.clone();
+                let follow_handle = follow_handle.clone();
 
                 async move {
+                    // TODO(niklasad1): naive impl which will fetch the subscription ID for every block
```
niklasad1, line 162: //cc @jsdw suggestions here?

subxt/examples/legacy_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,71 @@
...
+    let backend: LegacyBackend<PolkadotConfig> =
+        LegacyBackend::builder().build(RpcClient::new(rpc.clone()));
+
+    let api: OnlineClient<PolkadotConfig> = OnlineClient::from_backend(Arc::new(backend)).await?;
+
+    // Print when the RPC client reconnects.
+    tokio::spawn(async move {
+        loop {
+            rpc.on_reconnect().await;
+            println!("RPC client reconnect initiated");
```
jsdw, line 49: Nit: when we get to this line I guess it's reconnected already? :)

subxt/examples/unstable_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,158 @@
...
+    let api: OnlineClient<PolkadotConfig> = OnlineClient::from_backend(Arc::new(backend)).await?;
+
+    let a = api.clone();
+    tokio::spawn(async move { subscribe_runtime_versions(a).await.unwrap() });
+    let a = api.clone();
+    tokio::spawn(async move { subscribe_to_a_few_blocks(a).await.unwrap() });
+
+    submit_retry_transaction(&api).await?;
+
+    futures::future::pending::<()>().await;
```
jsdw, line 57: Nit: we'll never get to the line below; I guess deliberate for testing atm :)

subxt/src/blocks/blocks_client.rs
```diff
@@ -127,7 +127,7 @@ where
         let client = self.client.clone();
         header_sub_fut_to_block_sub(self.clone(), async move {
             let sub = client.backend().stream_finalized_block_headers().await?;
-            BlockStreamRes::Ok(sub)
+            BlockStreamRes::Ok(sub.stream.expect("TODO use retry subscription here???"))
```
jsdw, line 130: Why would we need to do anything special here? Doesn't the backend handle any retry stuff internally anyways (assuming a retry RPC client is used)?

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,161 @@
...
+        |err: &Error| err.is_disconnected_will_reconnect(),
+    )
+    .await
+}
+
+/// Retry policy
+///
+/// Does nothing if a non-reconnecting rpc client hasn't been enabled.
+#[derive(Debug, Clone)]
+pub struct RetryPolicy(tokio_retry::strategy::ExponentialBackoff);
```
jsdw, line 140: Why do we actually need any sort of retry policy in Subxt? ie wouldn't the reconnecting RPC client handle this all? It would use a retry policy internally, and we would just get back Pending from it until it reconnects and is able to start processing whatever is waiting

niklasad1: It's used for retrying individual futures in the backend impls

So, the reconnecting is manged by the reconnecting client but not individual method calls.
For the unstable backend it's not used for instance subscribing to blocks, runtime upgrades etc because it's managed by the driver itself.

For other methods and the legacy ones we want to retry calls until they are successful or returns an error which isn't Rpc::DisconnectWillReconnect

subxt/src/blocks/blocks_client.rs
niklasad1: yes correct

subxt/src/backend/legacy/rpc_methods.rs
```diff
@@ -36,9 +40,13 @@ impl<T: Config> LegacyRpcMethods<T> {
         key: &[u8],
         hash: Option<T::Hash>,
     ) -> Result<Option<StorageData>, Error> {
-        let params = rpc_params![to_hex(key), hash];
-        let data: Option<Bytes> = self.client.request("state_getStorage", params).await?;
-        Ok(data.map(|b| b.0))
+        self.retry
```
jsdw, line 43: I'm not sure how I feel about having the retry logic in the `rpc_methods` files. I sortof think that it would be best to keep all of the retry stuff in the Backend impls so that it's really obvious what will be retried and how, and keeping the raw rpc method impls super simple so that they are just the lowest level way to make a basic api call to a node.

niklasad1: so I'm quite scared the folks will use legacy stuff through the rpc methods and then miss the retry logic but I agree it would be nicer to have it just in the backend impls

subxt/src/backend/unstable/mod.rs
```diff
@@ -294,7 +330,13 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
 
         // Subscribe to the body response and get our operationId back.
         let follow_events = self.follow_handle.subscribe().events();
+
+        // `methods` already retries the call.
+        //
+        // NOTE: the sub_id could be "out-dated" if a reconnection occurs
```
jsdw, line 336: This is prob another reason to keep retry logic all in one place ie in these method impls. We can retry this whole call including getting a new subscription ID if theres any disconnection :)

subxt/src/backend/legacy/rpc_methods.rs
niklasad1: if you have a look at the legacy stuff then lots of stuff will not have retry logic if we don't add to the rpc methods

subxt/src/backend/unstable/mod.rs
jsdw: The subscription ID is emitted in a Ready message each time the subscription is renewed for whatever reason, but the `.events()` fn ignores it.

So perhaps we could remove the `.events()` above and then keep an eye out for new Ready messages to update the subscription ID here as needed, while manually filtering out the events we don't care about?

We have a similar issue in `unstable/storage_items.rs` where we get a subscription ID once and then subscribe to stuff and prob should do a similar sort of thing there too?

subxt/src/backend/utils.rs
jsdw: > So, the reconnecting is manged by the reconnecting client but not individual method calls.

Would it be possible for the individual calls, upon seeing a "DisconnectedWillReconnect" error, to immediatelytry reconnecting by resubscribing or whatever? And then the RPC client will return `Poll::Pending` until it's ready (ie it actually reconnects). And it could also be up to the RPC client how many times to try reconnecting etc; if it gives up then it emits some `Error::ConnecitonClosed` or whatever it is and then the backend gives up and errors back to the user :)

In that way, the backend/subxt doesn't need to think at all about retry policies etc and it's entirely up to the RPC client itself to manage.

subxt/examples/legacy_reconnecting_rpc_client.rs
niklasad1: well this future returns when client starts to "reconnect" but I can change the APIs to more clean

subxt/src/backend/legacy/rpc_methods.rs
niklasad1: I have removed the retry logic for the rpc methods stuff for now and the idea is to expose a helper function `retry_with_policy` that should be easy to use...

subxt/examples/legacy_reconnecting_rpc_client.rs
jsdw: Aah ok; I assumed that `on_reconnect.await` would mean reconnection has finished by the time the await ends :)

subxt/src/backend/utils.rs
niklasad1: I'm a little confused by this.

So, the reconnecting client supporst to resubscribe/retry method calls and reconnecting but only reconnecting is enabled now in subxt because that we don't want to resubscribe to stuff like `authorSubmitAndWatch` because it has side-effects.

Instead it's up to backend impl to decide whether retry or not but the reconnecting rpc client reconnects "automatically".

Indeed, we could return Poll::Pending in the reconnecting client instead of having this code: https://github.com/paritytech/subxt/pull/1505/files#diff-fc19ff73c9c406c6d7e3da3c3dda26c8cc6f2727ae358037d1be63b6cfdfdfb2R96-R110. Yepp, that would be neat

subxt/Cargo.toml
```diff
@@ -99,7 +101,8 @@ subxt-metadata = { workspace = true, features = ["std"] }
 subxt-lightclient = { workspace = true, optional = true, default-features = false }
 
 # Reconnecting jsonrpc ws client
-reconnecting-jsonrpsee-ws-client = { version = "0.3", optional = true }
+reconnecting-jsonrpsee-ws-client = { git = "https://github.com/niklasad1/reconnecting-jsonrpsee-ws-client", branch = "na-experimental-branch", optional = true }
+finito = { git = "https://github.com/niklasad1/finito/", optional = true }
```
niklasad1, line 105:
```suggestion
finito = { version = "0.1.0", optional = true }
```

subxt/src/backend/legacy/mod.rs
```diff
@@ -111,7 +143,7 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
     ) -> Result<StreamOfResults<Vec<u8>>, Error> {
         let keys = StorageFetchDescendantKeysStream {
             at,
-            key,
+            key: key.clone(),
```
niklasad1, line 146:
```suggestion
            key,
```

subxt/src/backend/legacy/mod.rs
```diff
@@ -142,7 +174,7 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
     ) -> Result<StreamOfResults<StorageResponse>, Error> {
         let keys_stream = StorageFetchDescendantKeysStream {
             at,
-            key,
+            key: key.clone(),
```
niklasad1, line 177:
```suggestion
            key,
```

subxt/src/backend/utils.rs
niklasad1: I had a look at this and it's a quite hard to implement such a thing but I realized we already have this functionality in the reconnecting-rpc-client.

So I enabled "retry" for rpc calls/subscriptions from the reconnecting rpc client but turned off "resubscribing" which achieves the same thing.

The downside is that it's quite hard to implement custom retry behavior rather than the global policy but I fine with that for now.

We could enable that with a HashMap and just turn it off for certain methods....

niklasad1: I'm closing this discussion my last comment isn't accurate anymore.

TLDR; subxt is handling retry logic for subscriptions, method calls and the reconnecting client just reconnects.

subxt/examples/legacy_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,150 @@
...
+use subxt::tx::TxStatus;
+use subxt::{OnlineClient, PolkadotConfig};
+use subxt_signer::sr25519::dev;
+
+// Generate an interface that we can use from the node's metadata.
+#[subxt::subxt(runtime_metadata_path = "../artifacts/polkadot_metadata_small.scale")]
+pub mod polkadot {}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
jsdw, line 25: I like this example in that everything is "the same" except it'll retry in the background and you can monitor the reconnects if you like (I assume that bit is entirely optional and I'd probably tweak the comment above that `tokio::spawn` to prefix with "Optional: " but that's just a nit :))

subxt/examples/unstable_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,161 @@
...
+            .await?,
+    );
+
+    let r = RpcClient::new(rpc.clone());
+    let (backend, driver) = UnstableBackend::builder().build(r.clone());
+
+    // The unstable backend needs driving:
+    tokio::spawn(drive_rpc_backend(driver));
+
+    // Print when the RPC client starts to reconnect.
```
jsdw, line 44:
```suggestion
    // Optional: print when the RPC client starts to reconnect.
```

subxt/examples/unstable_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,161 @@
...
+use subxt::tx::TxStatus;
+use subxt::{OnlineClient, PolkadotConfig};
+use subxt_signer::sr25519::dev;
+
+// Generate an interface that we can use from the node's metadata.
+#[subxt::subxt(runtime_metadata_path = "../artifacts/polkadot_metadata_small.scale")]
+pub mod polkadot {}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
jsdw, line 26: Nit: I wonder whether it's worth making one of these exmaples shorter, because most of the code is identical in both cases anyways.

Ideally we'd reference both of these somewhere in the book, and then we can show the "long" one first and the "short" one which just shgows the alternate backend being instantiated, second

subxt/src/backend/legacy/mod.rs
```diff
@@ -77,190 +79,299 @@ impl<T: Config> LegacyBackend<T> {
...
+        let sub = retry(|| subscribe_runtime_version(&self.methods)).await?;
+        let methods = self.methods.clone();
+
+        let retry_sub = RetrySubscription::new(
+            sub,
+            Box::new(move || {
+                let methods = methods.clone();
+                Box::pin(async move { subscribe_runtime_version(&methods).await })
+            }),
+        );
```
jsdw, line 270: I'm currently a bit confused by the `RetrySubscription` API, in that we:

1. create a stream wrapped in `retry`, and then
2. pass that to the `RetrySubscription`, along with _another_ future that will be ran to restart the thing that's already going to be restarted via `retry`?

I wonder whether you could instead have two methods:

- `retry(|| fut)`; the one you have already which calls the closure to retry a fut whenever disconnected, and
- `retry_stream(|| stream)`; return items from the stream, and call the closure to create a new stream whenever disconnected (could accept a closure that returns a future to create the stream to be more useful).

So then the utils exposed are nice and simple (just two functions) and the code could instead be more like:

```rust
let retry_sub = retry_stream({
    let methods = methods.clone();
    move || subscribe_runtime_version(&methods).await
});
```

Maybe a crappy idea but I'm just wanting to remove the duplication where we are creating the same stream in two places each time :)

niklasad1: I get your point it's horrible to read :)

niklasad1:
@CodiumAI-Agent /describe

subxt/examples/legacy_reconnecting_rpc_client.rs
niklasad1: I have merged these into a single example with some comments have to use the unstable backend

niklasad1:
@CodiumAI-Agent /review

niklasad1:
@CodiumAI-Agent /describe

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
         keys: Vec<Vec<u8>>,
         at: T::Hash,
     ) -> Result<StreamOfResults<StorageResponse>, Error> {
-        let methods = self.methods.clone();
+        async fn fut<T: Config>(
+            keys: Vec<Vec<u8>>,
+            at: T::Hash,
+            methods: LegacyRpcMethods<T>,
+        ) -> Result<StreamOfResults<StorageResponse>, Error> {
```
jsdw, line 92: Nit: do we need this separate fut function when all of the ones below are inline in the `retry`? :)

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
-            last_finalized_block_num,
-        );
-        let sub = sub.map(|r| {
-            r.map(|h| {
-                let hash = h.hash();
-                (h, BlockRef::from_hash(hash))
+        let this = LegacyBackend {
+            methods: self.methods.clone(),
+            storage_page_size: self.storage_page_size,
+        };
```
jsdw, line 286: Is there any reason not to just do `let this = self.clone();` here and below?

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
-        let keys = StorageFetchDescendantKeysStream {
-            at,
-            key,
-            storage_page_size: self.storage_page_size,
-            methods: self.methods.clone(),
-            done: Default::default(),
-            keys_fut: Default::default(),
-            pagination_start_key: None,
-        };
+        retry(|| async {
```
jsdw, line 119: Is this retry doing anything? there is no `.await` inside it so AFAIU it's just returning the storage stream and will never retry?

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
-        let keys_stream = StorageFetchDescendantKeysStream {
-            at,
-            key,
-            storage_page_size: self.storage_page_size,
-            methods: self.methods.clone(),
-            done: Default::default(),
-            keys_fut: Default::default(),
-            pagination_start_key: None,
-        };
+        retry(|| async {
```
jsdw, line 153: Same; is this `retry` doing anything? No `.await`s so no chance of a `DisconnectedWillReconnect` error during the async block?

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
         &self,
     ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
-        let sub = self.methods.chain_subscribe_all_heads().await?;
-        let sub = sub.map(|r| {
-            r.map(|h| {
-                let hash = h.hash();
-                (h, BlockRef::from_hash(hash))
+        let methods = self.methods.clone();
+
+        let retry_sub = retry_stream(move || {
```
jsdw, line 239: `stream_all_block_headers`, `stream_best_block_headers` and `stream_finalized_block_headers` are a bit more complicated I think.

At the moment, you're silently restarting the stream on reconnection. This possibly misses blocks as a result but the user would have no idea that they might have missed anything.

I think there are two options that prob make more sense:
1. Restart the stream on reconnection, but tell the user about the disconnection (ie pass on the `Err`) so that they know blocks may have been missed.
   - Means that the user has to explicitly accept that a disconnect happened (ie handle the error without bailing), else they will bail.
2. "catch up" and download any blocks missed, so that for the user it appears mostly seamless, and there are no important gaps.
   - In reality, we'll miss some best/new blocks (and blocks may get pruned etc while we are disconnected), but we can fill in any gaps by giving the finalized blocks.
   - `stream_finalized_block_headers` already uses `subscribe_to_block_headers_filling_in_gaps` which does something like this, so maybe we can make use of that, or similar,
     if need be.

It's probably easiest to do 1 to begin with, but being able to be seamless by doing 2 would be cool (and similar to what we'd want to do in the unstable backend).

Both here and in the unstable backend, if we aim to do 2 and can't be seamless (ie the blocks we need to fill in the gaps are no longer available to download) then we should emit an error and end.

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
 
     async fn submit_transaction(
         &self,
         extrinsic: &[u8],
     ) -> Result<StreamOfResults<TransactionStatus<T::Hash>>, Error> {
-        let sub = self
-            .methods
-            .author_submit_and_watch_extrinsic(extrinsic)
-            .await?;
+        let sub = retry(|| self.methods.author_submit_and_watch_extrinsic(extrinsic)).await?;
```
jsdw, line 329: Ok, so here we retry the initial future that submits the transaciton, but then don't retry the stream.

It might be safer to start by not retrying anything at all here to avoid this case:

1. User submits transaction
2. Disconnection happens while we are waiting for RPC response (but server got the TX)
3. On reconnect, we try submitting again and get an error like `TransactionStatus::Invalid`. User assumes that transaction failed, but actually the original TX did in fact get submitted and included.

So let's just return a disconnected error or whatever straight away, because then the user knows that their TX may or may not have made it to the server and can act accordingly?

subxt/src/backend/legacy/mod.rs
```diff
@@ -84,183 +85,249 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
...
-            at,
-            key,
-            storage_page_size: self.storage_page_size,
-            methods: self.methods.clone(),
-            done: Default::default(),
-            keys_fut: Default::default(),
-            pagination_start_key: None,
-        };
+        retry(|| async {
+            let keys_stream = StorageFetchDescendantKeysStream {
```
jsdw, line 154: `StorageFetchDescendantKeysStream`, further down this file, should also handle the `DisconnectedWillReconnect` error if possible :)

subxt/src/backend/unstable/mod.rs
```diff
@@ -145,34 +148,51 @@ impl<T: Config> UnstableBackend<T> {
...
+            })
+            .filter_map(|(sub_id, v)| async {
+                match v {
+                    FollowStreamMsg::Event(ev) => Some((sub_id, ev)),
+                    FollowStreamMsg::Ready(id) => {
+                        *sub_id.write().unwrap() = id;
+                        None
+                    }
+                }
+            })
```
jsdw, line 171: If we wanted to avoid modifying the underlying stuff and get rid of the lock, maybe something like this would work:

```rust
self.follow_handle.subscribe().scan(None, |sub_id, ev| {
    if let FollowStreamMsg::Ready(id) = &ev {
        *sub_id = Some(id);
        Some((sub_id.expect("sub_id should always be present"), ev))
    } else {
        Some((sub_id.expect("sub_id should always be present"), ev))
    }
}).filter_map(|(sub_id, ev)| {
    if let FollowStreamMsg::Event(ev) = ev {
        Some((sub_id, ev))
    } else {
        None
    }
})
```

`follow_handle.subscribe()` should always return a `Ready` event with a sub_id in it before anything else.

Doing the above also means you can ditch the `let sub_id = get_subscription_id(&follow_handle).await?;` line, which is basically doing another subscribe etc under the hood to get this Ready event.

subxt/src/backend/unstable/mod.rs
```diff
@@ -194,130 +214,146 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
...
+            let queries = keys.iter().map(|key| StorageQuery {
+                key: &**key,
+                query_type: StorageQueryType::Value,
+            });
 
-        let storage_items =
-            StorageItems::from_methods(queries, at, &self.follow_handle, self.methods.clone())
-                .await?;
+            let storage_items =
+                StorageItems::from_methods(queries, at, &self.follow_handle, self.methods.clone())
```
jsdw, line 224: This doesn't need addressing in this PR, since we currently just emit a Disconnected error to users if the storage stream is interrupted, but a nice improvement would be to tweak `StorageItems` so that if it sees the `DisconnectedWillReconnect` error, it will pick up from where it left off w.r.t streaming storage items if possible (and _not_ restart the whole stream of items again).

This is the same sort of thing I said in an above comment re legacyBackend storage streams :)

subxt/src/backend/unstable/mod.rs
```diff
@@ -475,10 +524,8 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
         let mut seen_blocks_sub = self.follow_handle.subscribe().events();
 
         // Then, submit the transaction.
-        let mut tx_progress = self
-            .methods
-            .transactionwatch_v1_submit_and_watch(extrinsic)
-            .await?;
+        let mut tx_progress =
+            retry(|| self.methods.transactionwatch_v1_submit_and_watch(extrinsic)).await?;
```
jsdw, line 528: As with the unstable backend, it's probably safer to not retry this for now, since retrying could lead to the user being returned `Invalid` while in fact the TX was actually submitted.

subxt/src/backend/unstable/follow_stream.rs
```diff
@@ -148,6 +148,14 @@ impl<Hash> Stream for FollowStream<Hash> {
                             continue;
                         }
                         Poll::Ready(Err(e)) => {
+                            // Re-start if a reconnecting backend was enabled.
+                            if e.is_disconnected_will_reconnect() {
+                                this.stream = InnerStreamState::New;
+                                return Poll::Ready(Some(Ok(FollowStreamMsg::Event(
+                                    FollowEvent::Stop,
+                                ))));
```
jsdw, line 156: Nit; you could do this instead here and below:

```rust
if e.disconnected_will_reconnect() {
    // Restart if a reconnecting RPC client is used, and gets disconnected.
    this.stream = InnerStreamState::Stopped;
    continue;
}
```

We do that in a couple of other places already so it'd be a bit more consistent :)

subxt/src/backend/unstable/mod.rs
```diff
@@ -350,6 +388,8 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
     async fn stream_runtime_version(&self) -> Result<StreamOfResults<RuntimeVersion>, Error> {
         // Keep track of runtime details announced in new blocks, and then when blocks
         // are finalized, find the latest of these that has runtime details, and clear the rest.
+        //
+        // The backend driver already re-starts the subscription if gets re-connected.
```
jsdw, line 392: If we are disconnected long enough, can we miss a runtime version entirely? I'm not sure whether I care too much about this though!

subxt/src/backend/unstable/mod.rs
```diff
@@ -145,34 +148,51 @@ impl<T: Config> UnstableBackend<T> {
...
+            .filter_map(|(sub_id, v)| async {
+                match v {
+                    FollowStreamMsg::Event(ev) => Some((sub_id, ev)),
+                    FollowStreamMsg::Ready(id) => {
+                        *sub_id.write().unwrap() = id;
+                        None
+                    }
+                }
+            })
+            .flat_map(move |(sub_id, ev)| {
```
jsdw, line 172: This `stream_headers` is used to stream new/best/finalized blocks. One thing we want to be sure of is that no important blocks are missed (much like in the legacy backend) orrr we give an error to the user if they might be, so that they know it's possible.

It's a bit easier to pick up where we left off in the unstable backend; we can keep track of the latest finalized block we send to the user, and then on reconnect, if that block is in the `Initialized` event then we pick up from there and stream all the blocks after it, else we return the error to the user so that it's up to them if they want to resume (and miss blocks) or stop.

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,244 @@
...
+    resubscribe: ResubscribeGetter<T>,
+    state: PendingOrStream<T>,
+}
+
+impl<T> std::marker::Unpin for RetrySubscription<T> {}
+
+impl<T> Stream for RetrySubscription<T> {
+    type Item = Result<T, Error>;
+
+    fn poll_next(
```
jsdw, line 31: I wonder whether we could code golf this to somehting more like:

```rust
fn poll_next(
    mut self: Pin<&mut Self>,
    cx: &mut std::task::Context<'_>,
) -> Poll<Option<Self::Item>> {
    loop {
        match self.state {
            PendingOrStream::Pending(ref mut fut) => match ready!(fut.poll_unpin(cx)) {
                Err(e) => {
                    if e.is_disconnected_will_reconnect() {
                        self.state = PendingOrStream::Pending((self.resubscribe)());
                    } else {
                        return Poll::Ready(Some(Err(e)));
                    }
                }
                Ok(stream) => {
                    self.state = PendingOrStream::Stream(stream);
                },
            },
            PendingOrStream::Stream(ref mut s) => match s.poll_next_unpin(cx) {
                Poll::Ready(Some(Err(e))) if e.is_disconnected_will_reconnect() => {
                    self.state = PendingOrStream::Pending((self.resubscribe)());
                }
                other => return other,
            },
        };
    }
}
```

I also made the future failing return the err instead of `Ok(None)`. Maybe we should have a `done` flag that we set on such an error so that it returns `None` afterwards instead of polling again after errors? Or maybe just let the underlying thing react however it wants and we `fuse()` this is we prefer.

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,244 @@
...
+                }
+
+                // This applies only to the rpc v2 method calls/unstable backend.
+                //
+                // Once the backend sees `Error::DisconnectedWillReconnect` it will
+                // update the state, and subsequent `subscribe` will not proceed until
+                // a new subscription ID is obtained.
+                //
+                // It's still possible that subscription ID could be read before
+                // the backend has updated the state, but it is quite unlikely.
```
jsdw, line 100: I guess this depends on the ordering that futures/streams are polled in.

Eg take this backend function:

```rust
async fn block_header(&self, at: T::Hash) -> Result<Option<T::Header>, Error> {
    retry(|| async {
        let sub_id = get_subscription_id(&self.follow_handle).await?;
        self.methods.chainhead_v1_header(&sub_id, at).await
    })
    .await
}
```

Where `get_subscription_id` is basically `follow_handle.subscribe().subscription_id().await`.

If `self.methods.chainhead_v1_header(&sub_id, at).await` returns `DisconnectedWillReconnect`, we immediately re-run this async block which calls `get_subscription_id(&self.follow_handle).await?`, but if the `FollowStreamDriver` hasn't noticed that there's an error yet because it hasn't been polled enough to see a `Stop` event, then you'll get back a stale subscription ID from it (pretty much synchronously, because it's cached).

One solution might be to keep track of the subscription ID we are using. If we get given the same ID again after seeing a `DisconnectedWillReconnect` error, then we know it's wrong now and we need to try again. That could be done hackily in eg the code above, but I think there's probably a nicer way to guarantee that we get back a valid subscription ID while keeping the above clean.

I think it's worth raising an issue to think about how best to fix this, but for now I'm cool with this hack :)

subxt/examples/reconnecting_rpc_client.rs
jsdw, line 1: Because we should strive to add this to the book eventually, could we follow the naming convention and call this `setup_client_reconnecting_rpc.rs` or something like that? (ie, it'll be in the `setup -> client` section probably, alongside the one on configuring a custom RPC client)

subxt/src/backend/legacy/mod.rs
niklasad1: I had some issues to get it to compile IIRC but lemme check again.

subxt/src/backend/legacy/mod.rs
niklasad1: It's a workaround/hack because T: Clone is not satisfied here and it won't work

niklasad1: I could implement manually though if you want

subxt/src/backend/legacy/mod.rs
niklasad1: good catch, done

subxt/src/backend/legacy/mod.rs
niklasad1: should we retry this in StorageFetchDescendantValuesStream: https://github.com/paritytech/subxt/pull/1505/files/6261cb8f057e2be38d3b82c564b34ccb8dfe50b5..0a837a4b75b6a499cc3e68762c0f18656437cf03#diff-17bc33d8a1d22f8fcb082c5eff935dfa21b562c9d9133b520c357e9975e48bdaR584

subxt/src/backend/unstable/mod.rs
niklasad1: yes we can

subxt/src/backend/legacy/mod.rs
```diff
@@ -142,7 +149,7 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
     ) -> Result<StreamOfResults<StorageResponse>, Error> {
         let keys_stream = StorageFetchDescendantKeysStream {
             at,
-            key,
+            key: key.clone(),
```
niklasad1, line 152:
```suggestion
            key,
```

subxt/src/backend/legacy/mod.rs
niklasad1: The same applies for runtime upgrades as well but maybe not as likely to occur.

2) that seems quite complicated and prefer not implement such a thing in this PR.

So I guess you are suggesting that we should emit a message or error message "MissedMessage(Vec<BlockRef>)" on the stream and it's to the user to decide whether to fetch missed the blocks?

subxt/src/backend/legacy/mod.rs
jsdw: Yeah I wondered if that was the case :)

subxt/src/backend/legacy/mod.rs
jsdw: Offhand, yup I think adding a retry around that should make `StorageFetchDescendantValuesStream` work :)

subxt/src/backend/legacy/mod.rs
jsdw: Ooh yeah lets manually impl `Clone` then because it's a bit useless with that default impl as it stands. whoops!

subxt/src/client/online_client.rs
```diff
@@ -3,9 +3,10 @@
 // see LICENSE for license details.
 
 use super::{OfflineClient, OfflineClientT};
+use crate::backend::StreamOfResults;
```
lexnv, line 6: nit: I think we can move these imports back to `crate::backend` below. I do like the grouping of rustfmt nightly for these

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    // Create a new client with with a reconnecting RPC client.
+    let rpc = Client::builder()
+        // Reconnect with exponential backoff
+        //
+        // This API is "iterator-like" and we use `take` to limit the number of retries.
+        .retry_policy(
+            ExponentialBackoff::from_millis(100)
+                .max_delay(Duration::from_secs(60))
+                .take(100),
```
lexnv, line 33: nit: Maybe 3 for the example? I'd be afraid users copy thsese configs and retry 100 in production env

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    // Create a new client with with a reconnecting RPC client.
+    let rpc = Client::builder()
+        // Reconnect with exponential backoff
+        //
+        // This API is "iterator-like" and we use `take` to limit the number of retries.
+        .retry_policy(
+            ExponentialBackoff::from_millis(100)
+                .max_delay(Duration::from_secs(10))
+                .take(100),
```
niklasad1, line 33:
```suggestion
                .take(3),
```

subxt/src/backend/legacy/mod.rs
niklasad1: I have modified the RetryStream impl to emit `DisconnectWillReconnect` now, so should fine now

subxt/src/backend/unstable/mod.rs
```diff
@@ -145,34 +147,54 @@ impl<T: Config> UnstableBackend<T> {
...
         let methods = self.methods.clone();
-        let headers = self.follow_handle.subscribe().events().flat_map(move |ev| {
-            let sub_id = sub_id.clone();
-            let methods = methods.clone();
 
-            let block_refs = f(ev).into_iter();
+        let headers = self
+            .follow_handle
+            .subscribe()
+            .scan(None, |sub_id, ev| {
```
niklasad1, line 155: annoying cloning but should be cheaper than RwLock :)

subxt/src/client/online_client.rs
```diff
@@ -3,9 +3,10 @@
...
 use crate::custom_values::CustomValuesClient;
 use crate::{
-    backend::{legacy::LegacyBackend, rpc::RpcClient, Backend, BackendExt, StreamOfResults},
+    backend::{legacy::LegacyBackend, rpc::RpcClient, Backend, BackendExt},
```
niklasad1, line 9:
```suggestion
    backend::{legacy::LegacyBackend, rpc::RpcClient, Backend, BackendExt, StreamOfResults},
```

subxt/src/client/online_client.rs
```diff
@@ -3,6 +3,7 @@
 // see LICENSE for license details.
 
 use super::{OfflineClient, OfflineClientT};
+use crate::backend::StreamOfResults;
```
niklasad1, line 6:
```suggestion
```

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+/// Retry a future until it doesn't return a disconnected error.
+pub async fn retry<T, F, R>(mut retry_future: F) -> Result<R, Error>
+where
+    F: FnMut() -> T,
+    T: Future<Output = Result<R, Error>>,
+{
+    const REJECTED_MAX_RETRIES: usize = 10;
```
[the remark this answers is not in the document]
niklasad1: It's hard to reproduce now and I think I managed to produce once so far after the changes to clean the state after `DisconnectWillReconnect` occured.

One retry was sufficient then and 10 is probably not needed.

subxt/src/backend/utils.rs
niklasad1: Good idea, I added this but my code is slightly more verbose because I can't mutate it when have reference to in the match

subxt/src/backend/utils.rs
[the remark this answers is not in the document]
jsdw: I'm not sure that this is an issue with the RPC APIs themselves. We can raise a separate issue I think to tweak the code such that we are guaranteed to get a valid subscription ID from the follow subscription once we detect a disconnect (which could come from a separate RPC call and not be known yet to the follow subscription task).

subxt/src/backend/utils.rs
lexnv: Could it be possible to add `PendingOrStream::EmptyState` variant and do a `std::mem::replace` to circumvent that?

niklasad1: I'll resolve this and have look at the code in utils.

I added another enum variant PendingOrStream::Closed

subxt/src/backend/unstable/mod.rs
lexnv: IIUC, we could probably guard around the following edge-cases:

```
// Finalized blocks T0:
// We always keep track of the last one, in this case D.
chain: A -> B -> C -> D

// We get disconnected at T1 and reconnected at T1 + K
// Reconnect to chainHead will give us a new subID and the Initialized event
FollowEvent::Initialized { finalized_blocks: [C, D, E, F]}

// We search for our last finalized block D in the event Initialized
// We find it and continue normally. However if it is not present, we return FollowEvent::Stop
// because we cannot uphold the chainHead guarantee

```

jsdw: > Yes that makes sense but what I'm trying to explain is that the follow_stream stream impl doesn't keep state it just passes the events to next layer and it's currently no way to pass that DisconnectWillReconnect to the next layer except returning an error on the stream which in turn will emit an error on the backend driver.

I think it will pass a `stop` event though on disconnect, and then a new `initialized` event etc on reconnect. So as Alex said, you can do something like:

- Keep track of the last finalized block
- Wait for new `initialized` event.
    - If the last finalized block is in the list of finalized blocks in the `initialized` event then we can emit the remaining finalized blocks from that event and continue as normal.
    - If the last finalized block is **not** in the `initialized` list, then we emit some error and end the stream, because otherwise the user would miss blocks and not know about it.

subxt/src/backend/utils.rs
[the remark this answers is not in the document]
niklasad1: Hmm, to me it's not really issue in the spec because the server should really return a valid subscription ID if it's made on a new connection. This is just relevant to "poll order" of the pending futures when reconnect occurs which we could fix i.e, OrderedFutures or something.

I think we could just do the following:


```rust
    let mut sub_id = get_subscription_id(&self.follow_handle).await?;
    
    loop {
    
        tokio::select!( {
            // probably need to add this as internal thing to backend
            _ = client.reconnected() => {
                sub_id = get_subscription_id(&self.follow_handle).await?;
            }
            res = retry(|| self.methods.chainhead_v1_header(&sub_id, at)) => {
                break res;
            }
        }
    }
```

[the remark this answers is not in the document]
jsdw: > This is just relevant to "poll order" of the pending futures when reconnect occurs

Yeah; basically:

- The follow stream subscription that you pull events from is driven in some task by some driver somewhere.
- RPC requests like `chainhead_v1_header().await` are driven (probably in a separate task).

So if the latter returns `DisconnectedWillReconnect`, it's no guarantee that the former has been polled and also has seen this error and handled it.

> I think we could just do the following

You _could_ add something like `.reconnected()` to the backend, but that doesn't necessarily mean that everything has been polled and synced?

One way to sync the two things might be to share some state between them, but I have a feeling that would become messy.

A way I have in mind to try out at some point is:

- maintain a counter in the follow subscription thing which starts at 0.
- every time we get a new `initialized` event, increment this counter.
- now, if some call relying on the follow subscription eg `chainhead_v1_header` returns `DisconnectedWillReconnect`, we note the current counter value and keep trying the follow subscription again until we see that the counter increments (ie, it's been polled and has noticed that disconnect happens)
- when this happens, we know that it's reconnected and try the new subscription ID.  

(We could maybe also do this by just waiting for the subscription ID to change, rather than waiting for a counter to increment, but we don't have any guarantee about subscription Ids being random etc, so it wouldn't be very safe to do it this way)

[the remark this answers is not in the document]
jsdw: But yeah, as long as the current hack seems to work for now, I'm happy if we look into this as a followup because it'll need some thought and iteration to try it and, if it seems good, abstract over the idea nicely and avoid over complicated code :)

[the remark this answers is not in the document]
niklasad1: > You could add something like .reconnected() to the backend, but that doesn't necessarily mean that everything has been polled and synced?

Yes, you are correct but both are needed. It's not elegant to retry each future in loop and if we could abort the pending futures when we know that the connection is lost and retry them after the chainHead_v1_follow has been established again.

Yes, let's create an issue after this PR instead :)

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    let backend: LegacyBackend<PolkadotConfig> =
+        LegacyBackend::builder().build(RpcClient::new(rpc.clone()));
+
+    let api: OnlineClient<PolkadotConfig> = OnlineClient::from_backend(Arc::new(backend)).await?;
+
+    // Optionally print if the RPC client reconnects.
+    let rpc2 = rpc.clone();
+    tokio::spawn(async move {
+        loop {
+            rpc2.reconnect_started().await;
```
lexnv, line 65: nit: Would returning a different future from this function simplify things?

```rust
// Wait to see when the reconnection was initiated.
let reconnected_future = rpc.reconnect_initiated().await;

// Wait until reconnection finishes.
reconnected_future.await;
```

subxt/src/backend/unstable/mod.rs
```diff
@@ -194,127 +216,142 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
...
+            });
+
+            Ok(StreamOf(Box::pin(storage_result_stream)))
+        })
+        .await
     }
 
     async fn genesis_hash(&self) -> Result<T::Hash, Error> {
-        self.methods.chainspec_v1_genesis_hash().await
+        retry(|| self.methods.chainspec_v1_genesis_hash()).await
```
lexnv, line 315: nit: Should we retry on `.call` as well similar to `block_header`?

subxt/src/backend/unstable/mod.rs
```diff
@@ -496,15 +498,40 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for UnstableBackend<T> {
     async fn stream_finalized_block_headers(
         &self,
     ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
-        let stream = self
```
niklasad1: would be nice to unify with stream_headers such that closure could used in FollowStreamFinalizedHeads but can't use such thing since we need the state to decide how to deal with it.

subxt/src/backend/unstable/mod.rs
niklasad1: yes, thanks

subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -379,6 +380,77 @@ struct SubscriberDetails<Hash: BlockHash> {
     waker: Option<Waker>,
 }
 
+pin_project! {
```
jsdw, line 384: Do we need this? I don't think anything needs to be pinned really, so I suspect you can just write something like:

```rust
impl Unpin for FollowStreamFinalizedHeads {}
```

Then, you can access self and do anything with 100% safe rust in the stream impl :)

subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -379,6 +380,77 @@ struct SubscriberDetails<Hash: BlockHash> {
...
+    pub fn new(stream: FollowStreamDriverSubscription<Hash>) -> Self {
+        Self {
+            stream,
+            sub_id: None,
+            last_seen_block: None,
+        }
+    }
+}
+
+impl<Hash: BlockHash> Stream for FollowStreamFinalizedHeads<Hash> {
```
jsdw, line 406: This looks good to me!As a nit I'd be tempted to put it in a different file like the stream in `storage_items.rs` is (or put some of this stuff into a utils file or something) since it's not really a part of the follow stream driver code

subxt/examples/setup_reconnecting_rpc_client.rs
jsdw: Quite a neat idea :)

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    let mut last_seen_block = None;
+
+    while let Some(block) = blocks_sub.next().await {
+        let block = match block {
+            Ok(b) => b,
+            Err(e) => {
+                // This can only happen on the legacy backend and the unstable backend
+                // will handle this internally.
+                if e.is_disconnected_will_reconnect() {
+                    println!("The RPC connection was lost and we may loose a few blocks");
```
jsdw, line 90:
```suggestion
                    println!("The RPC connection was lost and we may lose a few blocks");
```

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    });
+
+    // Run for at most 100 blocks and print a bunch of information about it.
+    //
+    // The subscription is automatically re-started when the RPC client has reconnected.
+    // You can test that by stopping the polkadot node and restarting it.
+    let mut blocks_sub = api.blocks().subscribe_finalized().await?.take(100);
+    // Track the last block to determine how many blocks that were missed when reconnecting.
+    let mut last_seen_block = None;
```
jsdw, line 81: Being just an example, I wonder if it's worth this extra complexity vs just printing block details and continuing if disconnected_will_reconnect error :)

subxt/src/backend/unstable/follow_stream_driver.rs
niklasad1: yepp, the thingy was that the FollowStreamSubscription wasn't exported outside this and I placed here to avoid sharing it with `pub crate`.

I can move it if it ok to make it pub crate :)

subxt/examples/setup_reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,107 @@
...
+    //        }
+    //    }
+    // });
+    // let api: OnlineClient<PolkadotConfig> = OnlineClient::from_backend(Arc::new(backend)).await?;
+    // ```
+
+    let backend: LegacyBackend<PolkadotConfig> =
+        LegacyBackend::builder().build(RpcClient::new(rpc.clone()));
+
+    let api: OnlineClient<PolkadotConfig> = OnlineClient::from_backend(Arc::new(backend)).await?;
```
jsdw, line 59: Just as a note: in the "default" case, one can also just to something like `OnlineClient::from_rpc(rpc_client)`, which might be an easier thing to point them at in an example.

(And perhaps they can then figure out how to use custom backends in a separate example)

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Create a retry stream that will resubscribe on disconnect.
+///
+/// It's important to note that this function is intended to work only for stateless subscriptions.
+/// If the subscription takes input or modifies state, this function should not be used.
+pub(crate) async fn retry_stream<F, R>(mut sub_stream: F) -> Result<StreamOfResults<R>, Error>
+where
+    F: FnMut() -> ResubscribeFuture<R> + Send + 'static + Clone,
+    R: Send + 'static,
+{
+    loop {
```
jsdw, line 164: It's a shame that we are sortof duplicating the try logic in an outer and then inner loop here, I think just to return the initial error in the `Result` before we are inside a stream and errors come in the streams.

I wonder if it makes sense to change the signature of this fn to return `-> StreamOfResults<R>` (ie no outer `Result`) and then the body can be simpler eg:

```rust
// future that will resubscribe to the stream, internally retrying 
// if any DisconnectedWillReconnect (I wonder whether `retry` could be 
// used to make this bit even shorter too..)
let resubscribe = Box::new(move || {...});

// The extra Box is to encapsulate the retry subscription type
return Ok(StreamOf::new(Box::pin(RetrySubscription {
    state: PendingOrStream::Stream(v),
    resubscribe,
})));
```

jsdw: Maybe best to leave as is for this PR though because changing it above might hjave knock-on effects on the signatures of backend fns or whatever.

subxt/src/backend/legacy/mod.rs
```diff
@@ -431,6 +506,11 @@ impl<T: Config> Stream for StorageFetchDescendantKeysStream<T> {
                         return Poll::Ready(Some(Ok(keys)));
                     }
                     Err(e) => {
+                        if e.is_disconnected_will_reconnect() {
```
jsdw, line 509: Perhaps worth a comment, because this is doing something a bit weird: if a future returns an `Err(DisconnectedWillReconnect)`, we expect to _keep polling it_ and it will internally retry itself :)

subxt/src/backend/rpc/reconnecting_rpc_client.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// Default: 1024
+    pub fn set_max_logging_length(mut self, max: u32) -> Self {
+        self.max_log_len = max;
+        self
+    }
+
+    /// Configure which retry policy to use.
+    ///
+    /// Default: Exponential backoff 10ms
+    pub fn retry_policy<T>(self, retry_policy: T) -> Builder<T> {
```
jsdw, line 136: So to check, `T` here is the same as `P` as in it's expected to be `Iterator<Item = Duration> + Send + Sync + 'static + Clone`?

subxt/src/backend/unstable/follow_stream_driver.rs
```diff
@@ -380,6 +380,107 @@ struct SubscriberDetails<Hash: BlockHash> {
     waker: Option<Waker>,
 }
 
+/// A stream that subscribes to finalized blocks
+/// and indicates whether a block was missed if was restarted.
+#[derive(Debug)]
+pub struct FollowStreamFinalizedHeads<Hash: BlockHash, F> {
+    stream: FollowStreamDriverSubscription<Hash>,
```
niklasad1, line 387: This type @jsdw :point_up:

subxt/src/backend/utils.rs
```diff
@@ -0,0 +1,291 @@
...
+                                pending: (self.resubscribe)(),
+                                err,
+                            }
+                        }
+                    }
+                },
+                PendingOrStream::Closed => NextState::Closed(None),
+            };
+
+            match next_state {
```
jsdw, line 97: Still a shame that we need two enums; `PendingOrStream` + this `NextState` one; I was hoping we could compact it down to one enum :D

subxt/src/backend/unstable/follow_stream_driver.rs
jsdw: that struct is `pub` in this file though, so we could import it from a separate file OK I think? :D

jsdw (approved):
LGTM, nice job @niklasad1!

I left some nits and suggestions but nothing that would block it merging :)

subxt/src/backend/unstable/follow_stream_driver.rs
niklasad1: it's a pita because the test helpers are not accessible outside this module :(

subxt/src/backend/rpc/reconnecting_rpc_client.rs
niklasad1: yes perhaps best to add these bounds here

subxt/src/backend/legacy/mod.rs
niklasad1: true, changed it to continue the loop instead

subxt/src/backend/utils.rs
niklasad1: haha, I thought it was quite readable but fine less code is probably better

subxt/src/backend/utils.rs
niklasad1: I think you will be happier now after I removed the !DRY stuff

subxt/src/backend/unstable/follow_stream_driver.rs
jsdw: Ah I see! either way it's all good and we can always shuffle stuff about in the future or whatever anyways!

subxt/src/backend/utils.rs
```diff
@@ -125,7 +140,25 @@ where
...
 /// If the subscription takes input or modifies state, this function should not be used.
-pub(crate) async fn retry_stream<F, R>(sub_stream: F) -> Result<StreamOfResults<R>, Error>
+///
+/// # Example
+///
+/// ```no_run
+/// use subxt::backend::{utils::retry_stream, StreamOf};
+/// use futures::future::FutureExt;
+///
+///
```
niklasad1, line 150:
```suggestion
```

PR #1507, opened 2024-03-28
testing: Prepare light client testing with substrate binary and add subxt-test macro

lexnv:
In this PR
- `subxt-macro` is added. The macro runs the tests with a default timeout (6 minutes if not specified)
- added cfg aliases for our testing with a new `unstable-light-client-long-running` feature flag to reuse subxt testing with the lightclient
- some tests are disabled in the context of the light-client
- chainHead initialized event supports the legacy format (to make it work with our outdated lightclient version)
- small testing refactoring to decouple the construction of transaction from the submit operation (might still be needed in the future if we want to retry failed light-client operations)

Builds on top of:  https://github.com/paritytech/subxt/pull/1459 (because the CI was not running anymore)


testing/integration-tests/proc-macro/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "subxt-test-proc-macro"
```
jsdw, line 2: Nit: also rename the folder to `subxt-test-proc-macro`?

testing/integration-tests/src/utils/tx_retries.rs
```diff
@@ -0,0 +1,51 @@
...
+                tracing::info!("Transaction error: {}, retrying...", err.to_string());
+
+                // Retry if the client is not connected.
+                tokio::time::sleep(std::time::Duration::from_secs(RETRY_TIME)).await;
+            }
+            Err(other) => return Err(other),
+        }
+    }
+
+    submit().await
```
jsdw, line 50: So, if using the light client, we try submitting a couple of additional times to handle light client specific issues?

I wonder if there's anything we can do to avoid needing to do this; does needing it imply that using the light client to submit transactions now is not that robust and will fail sometimes? Is there anything we can do before we try submitting that might help things (ie wait a few blocks), so that by the time we submit we don't need anything special? Is this just because of the legacy backend and everything works ok with the unstable backend (and in that case I think I'd only try testing the unstable backend with the light client)?

lexnv: Yep that makes sense. I've removed the `tx_retries` for now.
Instead maybe we should report those `tx::Dropped` events to smoldot if they still happen after an upgrade.

jsdw (approved):
LGTM!

Is the effect of removing that function that some tests will fail sometimes now when using the light client?

lexnv:
When we run tests with the `unstable-light-client-long-running` we'll probably encounter more flaky failures to due `TransactionEvent::Dropped`.  However, that is not enabled now in the CI and I would generally wait to enable that until we have: https://github.com/paritytech/subxt/pull/1400.

I'll look into that issue shortly, hopefully we could handle that and create a PR in smoldot if the fix is not too involved

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -391,6 +390,30 @@ pub struct Initialized<Hash> {
...
+        #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct InitializedIR<Hash> {
+            finalized_block_hashes: Option<Vec<Hash>>,
+            finalized_block_hash: Option<Hash>,
+            finalized_block_runtime: Option<RuntimeEvent>,
+        }
+
+        let ir = InitializedIR::deserialize(deserializer)?;
+        let finalized_block_hashes = ir
```
niklasad1, line 405: slightly weird comment; we could actually throw an error if both `finalized_block_hash` and `finalized_block_hashes`.

now, we checking `finalized_block_hashes` first and if that doesn't exist we fallback to `finalized_block_hash` which is probably fine but if one inserts both by accident it lead to confusing scenarios but fine :)

testing/integration-tests/subxt-test-macro/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+use syn::{
+    parse::{Parse, ParseStream},
+    Error,
+};
+
+/// Environment variable for setting the timeout for the test.
+const SUBXT_TEST_TIMEOUT: &str = "SUBXT_TEST_TIMEOUT";
+
+/// Default timeout for the test.
+const DEFAULT_TIMEOUT: u64 = 60 * 6;
```
niklasad1, line 18: would be nicer to use Duration here IMO

niklasad1 (approved):
awesome, I really like this

testing/integration-tests/subxt-test-macro/src/lib.rs
lexnv: We would still need to pass std::time::Duration to the `quote!` below, which does not impl `toTokens` , I've renamed this to illustrate better that we are counting seconds, thanks!

PR #1508, opened 2024-03-28
Make subxt-core ready for publishing

jsdw:
Tadeo did some great work getting a bunch of no-std compatible bits moved out of subxt into this new `subxt-core` crate. The aim of this PR is to tidy up and polish this crate so that it's ready to be released and made use of directly.

Some things I did here:

- Move Extrinsic decoding stuff to `subxt_core`
- Move Transaction creation/signing stuff to `subxt_core`
- Expose methods in storage and runtime APIs to encode/decode things and use in Subxt
- Provide a top level example for each of the areas (tx, runtime_apis, events, constants, custom_values, storage, blocks) to show how to build/encode/decode the relevant bits, and overall docs to link to these main areas.
- Various other tweaks to consistify the structure and such.


cli/src/commands/explore/pallets/calls.rs
```diff
@@ -154,7 +154,10 @@ fn mocked_offline_client(metadata: Metadata) -> OfflineClient<SubstrateConfig> {
         H256::from_str("91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3")
             .expect("Valid hash; qed");
 
-    let runtime_version = subxt::client::RuntimeVersion::new(9370, 20);
+    let runtime_version = subxt::client::RuntimeVersion {
+        spec_version: 9370,
+        transaction_version: 20,
+    };
```
jsdw, line 160: I reverted to just a struct so that the args (of the same type) were named

codegen/src/lib.rs
```diff
@@ -224,7 +224,16 @@ impl CodegenBuilder {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the path provided is not an absolute path.
     pub fn set_subxt_crate_path(&mut self, crate_path: syn::Path) {
+        if absolute_path(crate_path.clone()).is_err() {
+            // Throw an error here, because otherwise we end up with a harder to comprehend error when
+            // substitute types don't begin with an absolute path.
+            panic!("The provided crate path must be an absolute path, ie prefixed with '::' or 'crate'");
+        }
```
jsdw, line 236: We shouldn't need to clone crate_path here but need to change scale-typegen to avoid.

core/src/constants/address.rs
```diff
@@ -10,7 +12,7 @@ use derive_where::derive_where;
 
 /// This represents a constant address. Anything implementing this trait
 /// can be used to fetch constants.
-pub trait ConstantAddress {
+pub trait AddressT {
```
jsdw, line 15: I standardised on all addresses/payloads being either `AddressT` (just looking something up; constants & storage) or `PayloadT` (submitting/calling something; TX's & runtime APIs)

core/src/constants/mod.rs
```diff
@@ -1,29 +1,56 @@
...
-pub fn validate_constant<Address: ConstantAddress>(
-    metadata: &subxt_metadata::Metadata,
-    address: &Address,
-) -> Result<(), Error> {
+/// When the provided `address` is statically generated via the `#[subxt]` macro, this validates
+/// that the shape of the constant value is the same as the shape expected by the static address.
+///
+/// When the provided `address` is dynamic (and thus does not come with any expectation of the
+/// shape of the constant value), this just returns `Ok(())`
+pub fn validate<Address: AddressT>(metadata: &Metadata, address: &Address) -> Result<(), Error> {
```
jsdw, line 53: instead of `validate_constant` etc, aim for `constants::validate` etc.

core/src/metadata/metadata_type.rs
```diff
@@ -70,68 +76,6 @@ impl TryFrom<frame_metadata::RuntimeMetadataPrefixed> for Metadata {
 
 impl codec::Decode for Metadata {
     fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
-        subxt_metadata::Metadata::decode(input).map(Metadata::new)
-    }
-}
-
-/// Some extension methods on [`subxt_metadata::Metadata`] that return Errors instead of Options.
-pub trait MetadataExt {
```
jsdw, line 78: I didn't see a need for this offhand and just added the one missing function above on the concrete type instead.

core/src/runtime_api/payload.rs
jsdw, line 1: This was just moved from mod.rs; I should have done a `git mv` because after thenchanging `mod.rs` it didn't pick up the move.

core/src/storage/utils.rs
```diff
@@ -1,50 +1,24 @@
...
-    addr.append_entry_bytes(metadata, &mut bytes)?;
-    Ok(bytes)
-}
-
-/// Outputs a vector containing the bytes written by [`write_storage_address_root_bytes`].
-pub fn storage_address_root_bytes<Address: StorageAddress>(addr: &Address) -> Vec<u8> {
-    let mut bytes = Vec::new();
-    write_storage_address_root_bytes(addr, &mut bytes);
-    bytes
-}
```
jsdw, line 46: moded to `mod.rs` as get_address_bytes and get_address_root_bytes

cli/src/commands/explore/pallets/calls.rs
lexnv: This is more specific, I like it

artifacts/demo_chain_specs/polkadot.json
niklasad1, line 1: why is this changed?

artifacts/polkadot_metadata_small.scale
niklasad1, line 1: why is this changed?

artifacts/demo_chain_specs/polkadot.json
lexnv: We need to update this periodically to speed up the light-client syncing. Indeed, we also have a CI step for this :D

core/src/blocks/extrinsics.rs
```diff
@@ -0,0 +1,717 @@
...
+
+        assert_matches!(
+            result.err(),
+            Some(crate::Error::Block(
+                crate::error::BlockError::UnsupportedVersion(3)
+            ))
+        );
+    }
+
+    // TODO add this back when we have tx encode logic in core.
```
niklasad1, line 632: Can you open an issue for this as well? Easy to forgot

What is stopping us to have tx encode logic in core?

core/src/client.rs
```diff
@@ -1,61 +1,31 @@
...
-    }
-
-    pub fn runtime_version(&self) -> RuntimeVersion {
-        self.runtime_version
-    }
-
-    pub fn genesis_hash(&self) -> C::Hash {
-        self.genesis_hash
-    }
+    /// Genesis hash.
```
niklasad1, line 14: why pub now?

core/src/constants/address.rs
niklasad1: perhaps modify the docs then above :point_up_2:

core/src/client.rs
lexnv: I think is more in-line with `RuntimeVersion`, but either way is good for me

core/src/tx/signer.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019-2023 Parity Technologies (UK) Ltd.
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
```
lexnv, line 1: nit: Should we rename `Singer` -> `SignerT` for consistency with `AddrT` / `PayloadT`?
I've also seen it imported as `SignerT` for testing

core/src/client.rs
niklasad1: It's probably because we use it outside subxt-core but ideally we wouldn't make it pub. It's hard to add stuff without breaking the API...

lexnv (approved):
LGTM! Nice one 🚀

core/src/client.rs
niklasad1: I would probably add doc(hidden) to it to be on the safe-side

artifacts/demo_chain_specs/polkadot.json
niklasad1: yeah I know but I don't understand why it was changed in this PR :thinking:

jsdw: Some bits of the codegen were tweaked to point to subxt-core, breaking the codegen in the tests, and so I ran our CI rust thing to update the artifacts to fix this :)

core/src/client.rs
jsdw: Yeah, I made runtime version things pub just because it felt better constructing it with named params (and it's basically just a named tuple). I think I followed the same logic here and just made it pub since it's just a tuple of 3 things you can fetch individually otherwise. I could go either way really, but at the time preferred the simplicity I guess of it just being a simple data struct :)

I don't think it should be `doc(hidden)` personally because it's a used as a param in a couple of places so people should be able to name it.

core/src/constants/address.rs
jsdw: What do you think needs modifying? I think they are still relevant? :)

artifacts/polkadot_metadata_small.scale
jsdw: See https://github.com/paritytech/subxt/pull/1508#discussion_r1565762611

core/src/blocks/extrinsics.rs
jsdw: Ooh good point; I added this TODO when I hadn't ported over the extrinsic creation stuff to subxt-core, but now I think I can re-enable that test or something :)

core/src/tx/signer.rs
jsdw: Good question! There are probably a bunch of other traits here that we could, if we wanted, add `T` on the end of! In this PR I just standardised the "address/payload" traits to have a `T` but maybe we should go further! What do you reckon? (miight be better to do a full standarisation as a separate PR but I'd have to check hoiw many traits we have left to add the T too :D)

core/src/blocks/extrinsics.rs
jsdw: Re-enabled it!

core/src/tx/signer.rs
jsdw: I had a look at the other traits we have and now I'm not sure which way is best! I liked the standardising here but also think that adding `T` onto everything feels a bit naff (and it's not what Rust does on any of the basic traits).

Lemme have a play; maybe I should undo the `T` thing and do something else..

niklasad1 (approved):

core/src/constants/address.rs
niklasad1: yepp, I misunderstood your comment

core/src/tx/signer.rs
jsdw: (I'll sort this out in a small separate PR anyways :))

Polkadot-Forum:

https://forum.polkadot.network/t/paritytech-update-for-april/7646/1

PR #1522, opened 2024-04-08
storage_type: Strip key proper hash and entry bytes (32 instead of 16)

### LINKED ISSUE #1517: Dynamic storage iter example no longer works
Always get `Error: StorageAddress(TooManyBytes)` with `wss://rpc.polkadot.io`.

```rs
use subxt::{config::polkadot::H256, OnlineClient, PolkadotConfig};

let api = <OnlineClient<PolkadotConfig>>::from_url("wss://rpc.polkadot.io").await?;
let storage_query = subxt::dynamic::storage("System", "Account", Vec::new());

while let Some(r) = results.next().await {
	match r {
		Ok(kv) => {
			println!("Keys decoded: {:?}", kv.keys);
			println!("Key: {}", array_bytes::bytes2hex("0x", &kv.key_bytes));
			println!("Value: {:?}", kv.value.to_value()?);
		},
		Err(e) => println!("Error: {e:?}"),
	}
}
```

lexnv:
This PR ensures that we can properly decode storage key addresses.

There was an issue confirmed by https://github.com/paritytech/subxt/issues/1517, where the decoding of keys for dynamic storage queries would always fail at:

https://github.com/paritytech/subxt/blob/75c905dcb7ce61832d4d6395ffd2cbd7f28a89a6/subxt/src/storage/storage_type.rs#L260-L264

After investigating the code path, the error is coming from the following lines: 
https://github.com/paritytech/subxt/blob/75c905dcb7ce61832d4d6395ffd2cbd7f28a89a6/core/src/storage/storage_key.rs#L272-L275

The check appeared to be correct, as well as consuming hasher keys via `consume_hash_returning_key_bytes`.

After inspecting the key address bytes, we wrongly assumed the storage prefix bytes length.

Our storage keys start with the following:

```
twox_128(storage prefix) ++ twox_128(storage entry name)
```

In the issue, it is `twox_128(System) ++ twox_128(Accounts)`.

The output of the twox is: [16 bytes](https://paritytech.github.io/polkadot-sdk/master/frame_support/struct.Twox128.html#associatedtype.Output), not 8 as stated and assumed in our code base.

### Testing Done
- reproduced the issue via storage_iter_dynamic example
- Compared addr bytes with expected prefix:
  - twox_128("System"): 26aa394eea5630e07c48ae0c9558cef7
  - twox_128("Account"): b99d880ec681799c0cf30e8886371da9

Closes: https://github.com/paritytech/subxt/issues/1517

### DIFF
--- a/testing/integration-tests/src/full_client/storage/mod.rs
+++ b/testing/integration-tests/src/full_client/storage/mod.rs
@@ -237,7 +237,7 @@ async fn storage_pallet_storage_version() -> Result<(), subxt::Error> {
 }
 
 #[subxt_test]
-async fn storage_decode_keys() -> Result<(), subxt::Error> {
+async fn storage_iter_decode_keys() -> Result<(), subxt::Error> {
     use futures::StreamExt;
 
     let ctx = test_context().await;
--- a/testing/integration-tests/src/full_client/storage/mod.rs
+++ b/testing/integration-tests/src/full_client/storage/mod.rs
@@ -235,3 +235,54 @@ async fn storage_pallet_storage_version() -> Result<(), subxt::Error> {
         .await?;
     Ok(())
 }
+
+#[subxt_test]
+async fn storage_decode_keys() -> Result<(), subxt::Error> {
+    use futures::StreamExt;
+
+    let ctx = test_context().await;
+    let api = ctx.client();
+
+    let storage_static = node_runtime::storage().system().account_iter();
+    let results_static = api
+        .storage()
+        .at_latest()
+        .await?
+        .iter(storage_static)
+        .await?;
+
+    let storage_dynamic = subxt::dynamic::storage("System", "Account", vec![]);
+    let results_dynamic = api
+        .storage()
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):
A test for iterating over storage keys would be nice :)

testing/integration-tests/src/full_client/storage/mod.rs
```diff
@@ -235,3 +235,54 @@ async fn storage_pallet_storage_version() -> Result<(), subxt::Error> {
         .await?;
     Ok(())
 }
+
+#[subxt_test]
+async fn storage_decode_keys() -> Result<(), subxt::Error> {
```
jsdw, line 240: Nit: add `iter` to the test name?

```suggestion
async fn storage_iter_decode_keys() -> Result<(), subxt::Error> {
```

jsdw (approved):

jsdw:
Awesome; thanks for digging into this!

PR #1526, opened 2024-04-12
Tweaks to eth signer

jsdw:
@ryanleecode I finally got to having a pass over #1501 and thought I'd just suggest tweaks in the form of a branch! Lemme know what you reckon, and if all good then we can merge this into your PR and I reckon it's good to go :)

The tweaks:
- Put behind unstable flag for now
- Test in CI
- Separate `DerivationPath` so that people can configure it more arbitrarily.
- A couple of misc tweaks


signer/Cargo.toml
```diff
@@ -15,16 +15,26 @@ description = "Sign extrinsics to be submitted by Subxt"
...
+]
 
 # Pick the signer implementation(s) you need by enabling the
 # corresponding features. Note: I had more difficulties getting
 # ecdsa compiling to WASM on my mac; following this comment helped:
 # https://github.com/rust-bitcoin/rust-bitcoin/issues/930#issuecomment-1215538699
 sr25519 = ["schnorrkel"]
 ecdsa = ["secp256k1"]
-eth = ["keccak-hash", "secp256k1", "bip32"]
+unstable-eth = ["keccak-hash", "ecdsa", "secp256k1", "bip32"]
```
jsdw, line 37: We pull in the ecdsa::Keypair/sign/verify for instance, so need to ensure this stuff exists too :)

signer/src/eth.rs
```diff
@@ -175,11 +204,8 @@ pub enum Error {
     #[from(ignore)]
     InvalidSeed,
     /// Invalid derivation index.
-    #[display(fmt = "Invalid derivation index: {_0}")]
-    InvalidDerivationIndex(bip32::Error),
-    /// Invalid phrase.
-    #[display(fmt = "Cannot parse phrase: {_0}")]
-    InvalidPhrase(bip39::Error),
```
jsdw, line 182: We weren't using this error anywhere

signer/src/lib.rs
```diff
@@ -33,8 +33,8 @@ pub mod sr25519;
 pub mod ecdsa;
 
 // An ethereum signer implementation.
-#[cfg(feature = "eth")]
-#[cfg_attr(docsrs, doc(cfg(feature = "eth")))]
+#[cfg(feature = "unstable-eth")]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable-eth")))]
```
jsdw, line 37: I thought it'd be better to play it a little safe and call this unstable for now, so we can merge it quicker and work on the interface or whatever later if we need to.

signer/src/eth.rs
```diff
@@ -98,15 +97,45 @@ impl Keypair {
...
+/// A derivation path.
+#[derive(Clone, Debug)]
+pub struct DerivationPath {
+    inner: bip32::DerivationPath,
 }
 
-impl Default for Derivation {
-    fn default() -> Self {
-        Self::Soft
+impl DerivationPath {
```
jsdw, line 106: Move the derivation path stuff to a separate struct so that we can accept arbitrary derivation strings and have a couple of nice helpers for "common" things

signer/src/eth.rs
```diff
@@ -98,15 +103,47 @@ impl Keypair {
...
+/// `"m/44'/60'/0'/0/0"`, or we can construct one using the helpers [`DerivationPath::empty()`]
+/// and [`DerivationPath::eth()`].
+#[derive(Clone, Debug)]
+pub struct DerivationPath {
+    inner: bip32::DerivationPath,
+}
+
+impl DerivationPath {
+    /// An empty derivation path (in other words, just use the master-key as is).
+    pub fn empty() -> Self {
```
ryanleecode, line 116: needs a test

ryanleecode (commented):
looks great. needs a test for the empty derivation path and then its ready to go

signer/src/eth.rs
```diff
@@ -31,35 +32,39 @@ impl Keypair {
...
-        .expect("valid private key");
+        // `to_seed` isn't available unless std is enabled in bip39.
+        #[cfg(feature = "std")]
+        let seed = mnemonic.to_seed(password.unwrap_or(""));
+        #[cfg(not(feature = "std"))]
+        let seed = mnemonic.to_seed_normalized(password.unwrap_or(""));
+
+        // TODO: Currently, we use bip32 to derive private keys which under the hood uses
+        // the Rust k256 crate. We _also_ use the secp256k1 crate (which is very similar).
+        // It'd be great if we could 100% use just one of the two crypto libs. bip32 has
```
niklasad1, line 63: let's open a PR to the bip32 crate to update the secp256k1 crate then :)

jsdw: I almost did this hehe, but yup definitely; I think when this merges we can open an issue in Subxt to avoid thedupe crates and then go from there :)

signer/src/eth.rs
jsdw: Good point!

jsdw:
Ok, I added a test using some of the vectors in BIP39; these compare mnemonic + password vs seed.

In the process of doing this, I realised that our `from_seed` functions were a lie; they were actually "from secret key bytes". So I renamed them to `from_secret_key_bytes` and added a proper `from_seed` function here, which is then used in the new tests.

CI isn't running but we'll spot any issues once it merges with the other branch I guess!

PR #1534, opened 2024-04-15
Remove dupe storage item if we get one back, to be compatible with Smoldot + legacy RPCs

### LINKED ISSUE #1453: LightClient storage iterator returns duplicates
I am using the light client feature to try and iterate over all members of the Fellowship.  
Code looks like this:  
```rust
let key = collectives::storage().fellowship_collective().members_iter();
let mut query = parachain_api.storage().at_latest().await.unwrap().iter(key).await?;

while let Some(Ok((id, fellow))) = query.next().await {
    // ..
}
```

But somehow it returns duplicate values. The full code is on this branch: https://github.com/paritytech/subxt/compare/master...oty-light-client-iter-duplicate  
Note that it does work when using an RPC node.

### LINKED ISSUE #1525: Light Client: Iteration on map storage will never stop.
The storage `devices` in my testnet has only one entry, but the iteration will never stop while querying via the light client.
Dependences:
```
subxt = { version = "0.35.2", features = ["unstable-light-client", "unstable-reconnecting-rpc-client"] }
subxt-signer = { version = "0.35.2", features = ["subxt"] }
subxt-lightclient = "0.35.2"
```

Screenshot 2024-04-12 at 15 18 34

jsdw:
Smoldot returns the provided `start_key` when fetching storage items (see https://github.com/smol-dot/smoldot/issues/1692), whereas we don't expect this back from Substrate (it's badly named really; `start_key` does imply it will be handed back to me).

In Subxt we remove such an entry if we find it when iterating over storage entries, to avoid duplicate entries being returned.

Closes #1453, closes #1525

### DIFF
--- a/subxt/src/backend/legacy/mod.rs
+++ b/subxt/src/backend/legacy/mod.rs
@@ -379,6 +379,11 @@ pub struct StorageFetchDescendantKeysStream<T: Config> {
     storage_page_size: u32,
     // What key do we start paginating from? None = from the beginning.
     pagination_start_key: Option<Vec<u8>>,
+    // // What key did we see last? In Smoldot, the legacy storage API will hand
+    // // back the `start_key` (ie the key we saw last time) as the first key,
+    // // whereas in Substrate it will not. Here we'll want to skip over it to avoid
+    // // returning it twice.
+    // last_seen_key: Option<Vec<u8>>,
     // Keys, future and cached:
     keys_fut: Option<Pin<Box<dyn Future<Output = Result<Vec<Vec<u8>>, Error>> + Send + 'static>>>,
     // Set to true when we're done:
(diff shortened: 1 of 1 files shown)


subxt/src/backend/legacy/mod.rs
```diff
@@ -379,6 +379,11 @@ pub struct StorageFetchDescendantKeysStream<T: Config> {
     storage_page_size: u32,
     // What key do we start paginating from? None = from the beginning.
     pagination_start_key: Option<Vec<u8>>,
+    // // What key did we see last? In Smoldot, the legacy storage API will hand
+    // // back the `start_key` (ie the key we saw last time) as the first key,
+    // // whereas in Substrate it will not. Here we'll want to skip over it to avoid
+    // // returning it twice.
+    // last_seen_key: Option<Vec<u8>>,
```
lexnv, line 386: nit: Do we need this anymore?

lexnv (approved):

subxt/src/backend/legacy/mod.rs
jsdw: Ah whoops this can go now!

PR #1535, opened 2024-04-15
Rename traits to remove T suffix

jsdw:
In the end I went for a pattern like:

```text
mod            trait    concrete type              static type        dynamic type
--------------------------------------------------------------------------------------
constants      Address  DefaultAddress<ReturnTy>   StaticAddress<..>  DynamicAddress   
custom_values  Address  -                          StaticAddress<..>  str
storage        Address  DefaultAddress<K,R,F,D,I>  StaticAddress<..>  DynamicAddress<K>
runtime_api    Payload  DefaultPayload<Args,Ret>   StaticPayload<..>  DynamicPayload     
tx             Payload  DefaultPayload<CallData>   StaticPayload<..>  DynamicPayload
```

Where the static and dynamic types are mostly just aliases to the `DefaultX` concrete type.

### DIFF
--- a/subxt/src/tx/tx_client.rs
+++ b/subxt/src/tx/tx_client.rs
@@ -7,7 +7,7 @@ use crate::{
     client::{OfflineClientT, OnlineClientT},
     config::{Config, ExtrinsicParams, Header, RefineParams, RefineParamsData},
     error::{BlockError, Error},
-    tx::{PayloadT, Signer as SignerT, TxProgress},
+    tx::{Payload, Signer as SignerT, TxProgress},
     utils::PhantomDataSendSync,
 };
 use codec::{Compact, Decode, Encode};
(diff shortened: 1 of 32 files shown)


subxt/src/tx/tx_client.rs
```diff
@@ -94,7 +94,7 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
         params: <T::ExtrinsicParams as ExtrinsicParams<T>>::Params,
     ) -> Result<SubmittableExtrinsic<T, C>, Error>
     where
-        Call: PayloadT,
+        Call: Payload,
         Signer: SignerT<T>,
```
lexnv, line 98: I think this is fine, just thinking out loud, how would the following look?
- import `crate::tx::Signer` (not as `Signer T`)
- Use `<Call, Sign> where Sing: Signer<T>`?

But yea, we don't have to do it since we already have an `OfflineClientT and OnlineClientT` to distinguish between concret impl and traits

lexnv (approved):
LGTM! Nice one! 👍

subxt/src/tx/tx_client.rs
jsdw: Ohyeah, we do use the T pattern for `OfflineClient` and `OnlineClient` too don't we; I guess we are still not super consistent hehe. At least those traits are generally not supposed to be imported/used I guess!

> Use <Call, Sign> where Sing: Signer<T>?

I think that makes sense!

niklasad1 (approved):

PR #1539, opened 2024-04-19
Stabilize transactionWatch methods

lexnv:
This PR stabilizes the transactionWatch methods in response to:
- https://github.com/paritytech/polkadot-sdk/pull/4171

cc @paritytech/subxt-team


niklasad1 (approved):
I wonder whether we should rename it `fn transaction_watch_v1_xx(..)` instead of `fn transactionwatch_v1_xx` but all good

PR #1553, opened 2024-04-22
Fixed import in `subxt-signer::eth`

### LINKED ISSUE #1552: Wrong import in `subxt-signer::eth`
I found a weird issue with this line
https://github.com/paritytech/subxt/blob/e1fb401619d80d38d3697752b07c0769a5d6121a/signer/src/eth.rs#L271

I started by importing the dependency
```sh
cargo init foo
cd foo
cargo add subxt-signer --git https://github.com/paritytech/subxt/ --rev e1fb401 --features unstable-eth
```
with this simple setup, `cargo check` does not complain, but `cargo build` gives me the following error:
image

muraca:
fixes #1552

### DIFF
--- a/signer/src/eth.rs
+++ b/signer/src/eth.rs
@@ -268,7 +268,7 @@ pub mod dev {
 #[cfg(feature = "subxt")]
 mod subxt_compat {
     use subxt_core::config::Config;
-    use subxt_core::tx::Signer as SignerT;
+    use subxt_core::tx::signer::Signer as SignerT;
 
     use super::*;
 


niklasad1:
hmm, most likely a merge conflict or something caused this.

Ideally we should check this in the CI but thank you :rainbow:

niklasad1 (approved):

muraca:
> Ideally we should check this in the CI 

I saw that the CI only does cargo check with the proper feature, it's weird that this problem is being shown when building only.

niklasad1:
It's re-producible with `cd signer && cargo check --features unstable-eth` as well.

It's probably that some features is leaked in our workspace that caused that to work in the CI or that the feature is not checked at all. Anyway thanks for the PR

PR #1565, opened 2024-04-29
Update deps to use `scale-type-resolver` 0.2

jsdw:
Mainly this just means converting `&type_id`s to `type_id`s as we pass them by value and not reference now.

Also add back a check in `WrapperKeepOpaque` to confirm that the struct we're pointing at is the right one, to resovle a small TODO.

### DIFF
--- a/core/src/utils/wrapper_opaque.rs
+++ b/core/src/utils/wrapper_opaque.rs
@@ -86,7 +86,7 @@ impl<T> EncodeAsType for WrapperKeepOpaque<T> {
                 expected_id: format!("{type_id:?}"),
             }))
         })
-        .visit_composite(|(_type_id,out), _path, _fields| {
+        .visit_composite(|(_type_id, out), _path, _fields| {
             self.data.encode_to(out);
             Ok(())
         });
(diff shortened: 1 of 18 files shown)


core/src/utils/wrapper_opaque.rs
```diff
@@ -105,11 +106,19 @@ impl<T, R: TypeResolver> Visitor for WrapperKeepOpaqueVisitor<T, R> {
...
         value: &mut scale_decode::visitor::types::Composite<'scale, 'info, R>,
-        _type_id: &R::TypeId,
+        _type_id: R::TypeId,
     ) -> Result<Self::Value<'scale, 'info>, Self::Error> {
         use scale_decode::error::{Error, ErrorKind};
-
-        // TODO: When `scale-type-resolver` [provides struct names](https://github.com/paritytech/scale-type-resolver/issues/4), check that this struct name is `WrapperKeepOpaque`
+        use scale_decode::visitor::DecodeError;
+
+        if value.name() != Some("WrapperKeepOpaque") {
```
jsdw, line 114: We can get the struct name now for a little additional sanity check that our custom logic here makes sense given the target type.

lexnv (approved):

niklasad1 (approved):

PR #1598, opened 2024-05-20
Implemented `sign_prehashed` for `ecdsa::Keypair` and `eth::Keypair`

muraca:
Some people might find this useful 😄 

cc @niklasad1 @jsdw thanks ❤️

### DIFF
--- a/signer/src/eth.rs
+++ b/signer/src/eth.rs
@@ -106,11 +106,15 @@ impl Keypair {
 
     /// Signs an arbitrary message payload.
     pub fn sign(&self, signer_payload: &[u8]) -> Signature {
-        let message_hash = keccak(signer_payload);
+        self.sign_prehashed(&keccak(signer_payload))
+    }
+
+    /// Signs a pre-hashed message.
+    pub fn sign_prehashed(&self, message_hash: &H256) -> Signature {
         let wrapped =
             Message::from_digest_slice(message_hash.as_bytes()).expect("Message is 32 bytes; qed");
         Signature(ecdsa::internal::sign(&self.0 .0.secret_key(), &wrapped))
-    }
+    } 
 }
 
 /// A derivation path. This can be parsed from a valid derivation path string like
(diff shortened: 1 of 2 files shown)


jsdw (commented):
Thank you! Out of interest, is this something you found yourself wanting?

I'm thinking it's probably worth adding the same to the `ecdsa` impl for consistency (internally both `sign_prehashed` fns would end up being identical but hey ho)

muraca:
> Thank you! Out of interest, is this something you found yourself wanting?

That is correct - and it's something that is often found in other similar libraries, including [ethers](https://docs.rs/ethers-signers/2.0.14/src/ethers_signers/wallet/mod.rs.html#149)

muraca:
> I'm thinking it's probably worth adding the same to the ecdsa impl for consistency (internally both sign_prehashed fns would end up being identical but hey ho)

You could call `self.0.sign_prehashed` instead of duplicating

jsdw:
> You could call self.0.sign_prehashed instead of duplicating

Yup, true! :)

muraca:
> > You could call self.0.sign_prehashed instead of duplicating
> 
> 
> 
> Yup, true! Are you up for adding that also?

Sure, I'll do it later today

signer/src/eth.rs
```diff
@@ -171,6 +169,12 @@ impl AsRef<[u8; 65]> for Signature {
     }
 }
 
+impl From<ecdsa::Signature> for Signature {
+    fn from(s: ecdsa::Signature) -> Self {
+        Self(s.0)
+    }
+}
```
jsdw, line 176: Super small nit, but perhaps we could ditch this and just do the conversion manually in `sign_prehashed`? Only reason being that having the signatures be distinct feels a little better :)

Otherwise, looks great!

muraca: The alternative would be something like 
```rust
Signature(self.0.sign_prehashed(&message).0)
```
I went this route because it felt more readable

muraca:
Sorry for this mess, I'm having a few issues with rust analyzer. I'll reboot my laptop and try again

signer/src/eth.rs
jsdw: That looks good to me! The `From` impl is a public thing so I was hesitant to want toe expose that sort of thing :)

jsdw:
No worries, and thank yoU! Looking good to me now!

jsdw (approved):

signer/src/eth.rs
muraca: I have already changed it, the comment was just to explain the reasoning behind the choice 😁

niklasad1 (approved):

Issue #1503, opened 2024-03-26
Replace `derive_more` (and remove `syn` 1.0 in favour of 2.0)?


jsdw:
https://crates.io/crates/derive_more hasn't had a release in a couple of years, and particularly it's still using `syn` 1.0. Is there a sensible alternative?

We mainly use it for `Display` impls and mainly on error types (which `derivative` and `derive_where` don't offer) to be no-std friendly (we previously used `thiserror` for this, but it's not no-std friendly). We have a couple of `From` impl uses too.

I recently replaced `derivative` with `derive-where` for a similar reason; it hadn't seen a release in 3 years, was triggering `non-canonical` compiler warnings with some of its impls, and was still using `syn` 1.0. Would be good to eliminate `syn` 1.0 from the dependency tree entirely or as much as possible :)

jsdw:
One thing to confirm here is to do a `cargo tree -i syn` or whatever to check that we only have 2.0 in the tree, or do we need to remove/update anything else to make it be the case?

PR #1600, opened 2024-05-21
Remove `derive_more`

pkhry:
### Description 
- I decided to define a macro for Error wrapping to use in place of `From` from `derive_more`
- Reimplemented Display by hand.
- also removed `derive_more` from dependencies
### Questions: 
 Given that we used `derive_more` for errors, maybe we can use [SNAFU](https://github.com/shepmaster/snafu) for errors with no-std constraint given that `thiserror` doesn't support no-std
see #1503


core/src/macros.rs
```diff
@@ -18,4 +18,21 @@ macro_rules! cfg_substrate_compat {
...
+            }
+        }
+    };
+    ($ty:ident $(<$( $param:ident ),+>)? as $delegate_ty:ty, $expr:expr) => {
+        impl From<$ty> for $delegate_ty {
+            fn from(val: $ty) -> Self {
+                $expr(val)
+            }
+        }
+    };
```
jsdw, line 35: This block doesn't appear to be made use of, and the `$param` that you capture isn't made use of anywhere anyways, so I'd just remove it to keep things simple :)

signer/src/utils.rs
```diff
@@ -34,3 +34,20 @@ macro_rules! once_static_cloned {
...
+            }
+        }
+    };
+    ($ty:ident $(<$( $param:ident ),+>)? as $delegate_ty:ty, $expr:expr) => {
+        impl From<$ty> for $delegate_ty {
+            fn from(val: $ty) -> Self {
+                $expr(val)
+            }
+        }
+    };
```
jsdw, line 52: Same here; just remove this block

core/src/error.rs
```diff
@@ -38,11 +38,14 @@ pub enum Error {
...
-    }
-}
+convert_error!(ExtrinsicParamsError as Error, Error::ExtrinsicParams);
+convert_error!(BlockError as Error, Error::Block);
+convert_error!(MetadataError as Error, Error::Metadata);
+convert_error!(scale_decode::Error as Error, Error::Decode);
+convert_error!(scale_decode::visitor::DecodeError as Error, Error::Decode);
+convert_error!(scale_encode::Error as Error, Error::Encode);
+convert_error!(StorageAddressError as Error, Error::StorageAddress);
+convert_error!(codec::Error as Error, Error::Codec);
```
jsdw, line 48: Just a small thought to simplify this; one could do:

```rust
macro_rules! convert_error {
    ($module_path:path as $delegate_ty:ident :: $variant:ident) => {
        impl From<$module_path> for $delegate_ty {
            fn from(val: $module_path) -> Self {
                $delegate_ty::$variant(val.into())
            }
        }
    };
}

// usage:
convert_error!(FooError as Error::Foo);
```

Which looks a little more natural :)

Either way, I'd turn the `ty` into `path` because that's really what you're capturing here

signer/src/utils.rs
pkhry: done

core/src/macros.rs
pkhry: done

pkhry:
> Feel free to merge the two sub-PRs into this one; IMO it would be easier to see it all in one place, and you'll have to merge the others into here before we can properly review this one anyway!
> 
> I think I'd only normally stick to separate PRs when it's actually two distinct things that can both be merged to master independently :)

I initially made separate pr's to make reviewing easier, but good to know, i won't be splitting mr's in the future if they are changing the same thing

core/src/error.rs
```diff
@@ -4,155 +4,181 @@
...
+
 #[cfg(feature = "std")]
 impl std::error::Error for Error {}
 
-impl From<scale_decode::visitor::DecodeError> for Error {
-    fn from(value: scale_decode::visitor::DecodeError) -> Self {
-        Error::Decode(value.into())
-    }
-}
+convert_error!(ExtrinsicParamsError as Error::ExtrinsicParams);
```
niklasad1, line 49: What do you think about:

```rust
impl_from!(ExtrinsicParamsError => Error::ExtrinsicParams);
```

I find "as" slightly misleading because in my head `as == typecast` such as "usize::MAX as u64".... 

Ideally we could find a similar crate as derive_more or some other crate for error handling where it supports opting of from std::error::Error which is the reason why we are not using thiserror in subxt but this should work as work around for now.

niklasad1 (approved):

lexnv (approved):
LGTM! I do like the suggestion from Niklas to have a `impl_from` instead of `convert_err`

jsdw (approved):
Nice, LGTM! (I'd also be up for that `impl_from` and `=>` thing but easy either way :)

core/src/error.rs
pkhry: changed `convert_from` to `impl_from` with the `as` -> `=>` swap

niklasad1 (approved):

PR #1601, opened 2024-05-21
Remove `derive_more`'s `Display` usages

pkhry:
### Description
Reimplemented `Display` by hand. 

see #1503


core/src/utils/account_id.rs
```diff
@@ -105,19 +106,29 @@ impl AccountId32 {
...
-    #[display(fmt = "Invalid checksum")]
     InvalidChecksum,
-    #[display(fmt = "Invalid SS58 prefix byte.")]
     InvalidPrefix,
 }
 
+impl Display for FromSs58Error {
+    fn fmt(
+        &self,
+        f: &mut scale_info::prelude::fmt::Formatter<'_>,
```
niklasad1, line 121: I don't understand why you use fmt from scale info?

```suggestion
        f: &mut core::fmt::Formatter<'_>,
```

core/src/error.rs
```diff
@@ -164,33 +180,83 @@ pub enum StorageAddressError {
...
     UnknownSignedExtension(String),
     /// Some custom error.
-    #[display(fmt = "Error constructing extrinsic parameters: {_0}")]
     Custom(Box<dyn CustomError>),
 }
 
+impl Display for ExtrinsicParamsError {
+    fn fmt(
+        &self,
+        f: &mut scale_info::prelude::fmt::Formatter<'_>,
```
niklasad1, line 242: same here why not `core::fmt::Formatter`?

niklasad1 (commented):
Looks good overall 

Can you just replace `scale_info::prelude::fmt` with `core::fmt`?

signer/src/crypto/secret_uri.rs
```diff
@@ -117,13 +118,20 @@ impl core::str::FromStr for SecretUri {
...
     InvalidFormat,
 }
 
+impl Display for SecretUriError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            SecretUriError::InvalidFormat => write!(f, "Invalid secret phrase format"),
+        }
+    }
+}
```
jsdw, line 133: I wonder if there is a small, reasonably maintained crate out there that can help us with this over manual impls, though happy doing it manually if not!

core/src/utils/account_id.rs
pkhry: done

signer/src/crypto/secret_uri.rs
pkhry: i guess we can use Strum(uses syn 2.0) or we can just implement display by hand for now and hope that thiserror would start working in no-std or maybe try Using SNAFU for errors as it works in no-std(don't know if it was discussed before)

PR #1602, opened 2024-05-21
Remove `derive_more` from dependencies

### LINKED ISSUE #1503: Replace `derive_more` (and remove `syn` 1.0 in favour of 2.0)?

pkhry:
closes #1503


niklasad1 (approved):
Would be easier to review if you could omit the TOML fmt changes, I guess you are using some fmt tool for that.

I for one preferred to old format with features on line but perhaps we could add some fmt rules in this repo. Substrate is using [taplo](https://taplo.tamasfe.dev/) for that I guess we could as well

jsdw (approved):

pkhry:
> Would be easier to review if you could omit the TOML fmt changes, I guess you are using some fmt tool for that.

dropped formatting changes until we setup the autoformatting tool

PR #1604, opened 2024-05-22
POC: use `snafu` instead of `derive_more` for errors

pkhry:
### Description
- this is an alternative to #1600 (see #1503)
- Replaces `derive_more` usage with `snafu`

Changing the library was somewhat easy, however: 
- `snafu` uses its own custom `snafu::Error` type in no-std contexts, therefore we needed to define a wrapper type that implements `snafu::AsErrorSource` for external error types and this led to a bit of boilerplate. 
- `snafu` doesn't support tuple variants yet, so its impossible to define an `Error` enum like this:
```rust
#[derive(Snafu)]
enum Error {
 #[snafu(display("example text {0}"))]
  MyEnum(AnotherError) 
}
```
it will be defined as: 
```rust
#[derive(Snafu)]
enum Error {
 #[snafu(display("example text {source}"), context(false))]
  MyEnum {
    source: AnotherError
  } 
}
- for external types it will look like example below, given we defined `DisplayError` wrapper for compatibility
```rust
#[derive(Snafu)]
enum Error {
 #[snafu(display("example text {source}"), context(false))]
  MyEnum {
    #[snafu(source(from(AnotherErrorExternal,DisplayError)))]
    source: DisplayError<AnotherError>
  } 
}
```

- This is a POC to gather some discussion whether it's better to go with `snafu` or use some other library or define `Display`/`From` by hand
- This will also incur dowstream breakage for users of subxt-metadata, subxt-signer, subxt-core

### DIFF
--- a/core/src/error_utils.rs
+++ b/core/src/error_utils.rs
@@ -1,0 +1,24 @@
+use core::fmt::{self, Debug, Display};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct DisplayError<T>(pub T);
+
+impl<T> Debug for DisplayError<T>
+where
+    T: Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl<T> Display for DisplayError<T>
+where
+    T: Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl<T> snafu::Error for DisplayError<T> where T: Display + Debug {}
--- a/signer/src/ecdsa.rs
+++ b/signer/src/ecdsa.rs
@@ -5,12 +5,15 @@
 //! An ecdsa keypair implementation.
 use codec::Encode;
 
-use crate::crypto::{seed_from_entropy, DeriveJunction, SecretUri};
+use crate::{
+    crypto::{seed_from_entropy, DeriveJunction, SecretUri},
+    utils::DisplayError,
+};
 use core::str::FromStr;
-use derive_more::{Display, From};
 use hex::FromHex;
 use secp256k1::{ecdsa::RecoverableSignature, Message, Secp256k1, SecretKey};
 use secrecy::ExposeSecret;
+use snafu::Snafu;
 
 const SECRET_KEY_LENGTH: usize = 32;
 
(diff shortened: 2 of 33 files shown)


niklasad1:
> This will also incur dowstream breakage for users of subxt-metadata, subxt-signer, subxt-core

It's because that snafu uses each variant as a struct instead of tuple?

```rust
// snafu
#[derive(Debug, snafu::Snafu)]
pub enum Error {
    #[snafu(display("Metadata Error: {source}"), context(false))]
    Metadata {
        /// Error source
        source: MetadataError,
    }
}

// dervive_more
#[derive(Debug, dervive_more::Display)]
enum Error {
    #[display(fmt = "Metadata Error: {_0}")]
    Metadata(MetadataError),
}
```

In downstream code then one has to do:


```rust
   // snafu
   if let Err(Error::Metadata { e }) = my_err {}

   // derive_more
   if let Err(Error::Metadata(e)) = my_err {}
```

Looks like a good solution to me :)
Would be nice to use `context(false)` globally on the Error type but that doesn't seem possible
because it depends on display attribute is used....

niklasad1 (approved):
Looks good to me

lexnv (approved):
LGTM! This approach also looks good, nice 👍

core/src/error_utils.rs
```diff
@@ -0,0 +1,24 @@
+use core::fmt::{self, Debug, Display};
+
+#[derive(Clone, PartialEq, Eq)]
+pub struct DisplayError<T>(pub T);
```
jsdw, line 4: Offhand I can't quite see what the reason for this is?

niklasad1: > snafu uses its own custom snafu::Error type in no-std contexts, therefore we needed to define a wrapper type that implements snafu::AsErrorSource for external error types and this led to a bit of boilerplate.snafu uses its own custom snafu::Error type in no-std contexts, therefore we needed to define a wrapper type that implements snafu::AsErrorSource for external error types and this led to a bit of boilerplate.

Most likely :point_up:

signer/src/ecdsa.rs
```diff
@@ -213,22 +216,26 @@ pub(crate) mod internal {
...
-    #[from(ignore)]
+    #[snafu(display("Invalid seed for ECDSA, contained soft junction"))]
     SoftJunction,
     /// Invalid phrase.
-    #[display(fmt = "Cannot parse phrase: {_0}")]
-    Phrase(bip39::Error),
+    #[snafu(display("Cannot parse phrase: {source}"), context(false))]
+    Phrase {
+        #[snafu(source(from(bip39::Error, DisplayError)))]
+        source: DisplayError<bip39::Error>,
```
jsdw, line 231: So eg here, what do you gain from wrapping in `DisplayError`?

core/src/error_utils.rs
jsdw: Ahh, I should have paid more attention reading :)

pkhry:
closing in favor of #1600

PR #1638, opened 2024-06-11
Add 20-byte account id to subxt_core

### LINKED ISSUE #1576: `AccountId20` does not implement `Serialize`
I wanted to use the following legacy method to retrieve nonces from transaction pool, however the `AccountId20` type does not implement Serialize. 

https://github.com/paritytech/subxt/blob/f034ac486cb04667934a5399e564b8e877e96c5d/subxt/src/backend/legacy/rpc_methods.rs#L140-L150

I noticed that the `subxt-signer` crate does not depend on `serde`, and I'm not sure if you want to ever support this, therefore I'm opening this issue rather than a PR. 

In the meantime, for anyone who might incur in the same issue, my solution was to use a custom `AccountId20` type and make it compatible with the `subxt_signer::eth::Signature` type by implementing `From<subxt_signer::eth::AccountId20>`.
```rust
/// A 20-byte cryptographic identifier.
// We need this because subxt's AccountId20 does not implement Serialize.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Encode, Decode, serde::Serialize)]
pub struct AccountId20([u8; 20]);

impl From<subxt_signer::eth::AccountId20> for AccountId20 {
    fn from(value: subxt_signer::eth::AccountId20) -> Self {
        AccountId20(value.0)
    }
}

impl Config for MyConfig {
    // ... 
    type AccountId = AccountId20;
    type Address = AccountId20;
    type Signature = subxt_signer::eth::Signature;
    // ...
}
```

pkhry:
### Description: 
see #1576. This mr just moves the `AccountId20` implementation to subxt_core and adds serde's `Serialize` and `Deserialize` implementations for it.

closes #1576

### DIFF
--- a/codegen/src/lib.rs
+++ b/codegen/src/lib.rs
@@ -358,7 +358,7 @@ fn default_derives(crate_path: &syn::Path) -> DerivesRegistry {
 fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
     let mut type_substitutes = TypeSubstitutes::new();
 
-    let defaults: [(syn::Path, syn::Path); 12] = [
+    let defaults: [(syn::Path, syn::Path); 13] = [
         (
             parse_quote!(bitvec::order::Lsb0),
             parse_quote!(#crate_path::utils::bits::Lsb0),
@@ -370,6 +370,10 @@ fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
         (
             parse_quote!(sp_core::crypto::AccountId32),
             parse_quote!(#crate_path::utils::AccountId32),
+        ),
+        (
+            parse_quote!(fp_account::AccountId20),
+            parse_quote!(#crate_path::utils::AccountId20),
         ),
         (
             parse_quote!(sp_runtime::multiaddress::MultiAddress),
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -64,6 +64,9 @@ sp-core = { workspace = true, optional = true }
 sp-runtime = { workspace = true, optional = true }
 tracing = { workspace = true, default-features = false }
 
+# AccountId20
+keccak-hash = { workspace = true}
+
 [dev-dependencies]
 assert_matches = { workspace = true }
 bitvec = { workspace = true }
--- a/core/src/utils/account_id20.rs
+++ b/core/src/utils/account_id20.rs
@@ -1,0 +1,163 @@
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! `AccountId20` is a repressentation of Ethereum address derived from hashing the public key.
+
+use core::fmt::Display;
+
+use alloc::format;
+use alloc::string::String;
+use codec::{Decode, Encode};
+use keccak_hash::keccak;
+use serde::{Deserialize, Serialize};
+
+#[derive(
+    Copy,
+    Clone,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Encode,
+    Decode,
+    Debug,
+    scale_encode::EncodeAsType,
+    scale_decode::DecodeAsType,
+    scale_info::TypeInfo,
+)]
+/// Ethereum-compatible `AccountId`.
+pub struct AccountId20(pub [u8; 20]);
+
+impl AsRef<[u8]> for AccountId20 {
+    fn as_ref(&self) -> &[u8] {
+        &self.0[..]
+    }
+}
+
+impl AsRef<[u8; 20]> for AccountId20 {
+    fn as_ref(&self) -> &[u8; 20] {
+        &self.0
+    }
+}
+
+impl From<[u8; 20]> for AccountId20 {
+    fn from(x: [u8; 20]) -> Self {
+        AccountId20(x)
+    }
+}
+
+impl AccountId20 {
+    /// Convert to a public key hash
+    pub fn checksum(&self) -> String {
+        let hex_address = hex::encode(self.0);
+        let hash = keccak(hex_address.as_bytes());
+
+        let mut checksum_address = String::with_capacity(42);
+        checksum_address.push_str("0x");
+
+        for (i, ch) in hex_address.chars().enumerate() {
+            // Get the corresponding nibble from the hash
+            let nibble = hash[i / 2] >> (if i % 2 == 0 { 4 } else { 0 }) & 0xf;
+
+            if nibble >= 8 {
+                checksum_address.push(ch.to_ascii_uppercase());
+            } else {
+                checksum_address.push(ch);
+            }
+        }
+
+        checksum_address
+    }
+}
+
+/// An error obtained from trying to interpret a hex encoded string into an AccountId20
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+#[allow(missing_docs)]
+pub enum FromChecksumError {
+    BadLength,
+    InvalidChecksum,
+    InvalidPrefix,
+}
+
+impl Display for FromChecksumError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            FromChecksumError::BadLength => write!(f, "Length is bad"),
+            FromChecksumError::InvalidChecksum => write!(f, "Invalid checksum"),
+            FromChecksumError::InvalidPrefix => write!(f, "Invalid checksum prefix byte."),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for FromChecksumError {}
+
+impl Serialize for AccountId20 {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&self.checksum())
+    }
+}
+
+impl<'de> Deserialize<'de> for AccountId20 {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        String::deserialize(deserializer)?
+            .parse::<AccountId20>()
+            .map_err(|e| serde::de::Error::custom(format!("{e:?}")))
+    }
+}
+
+impl core::fmt::Display for AccountId20 {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "{}", self.checksum())
+    }
+}
+
+impl core::str::FromStr for AccountId20 {
+    type Err = FromChecksumError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.len() != 42 {
+            return Err(FromChecksumError::BadLength);
+        }
+        if !s.starts_with("0x") {
+            return Err(FromChecksumError::InvalidPrefix);
+        }
+        hex::decode(&s.as_bytes()[2..])
+            .map_err(|_| FromChecksumError::InvalidChecksum)?
+            .try_into()
+            .map(AccountId20)
+            .map_err(|_| FromChecksumError::BadLength)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn deserialisation() {
+        let key_hashes = vec![
+            "0xf24FF3a9CF04c71Dbc94D0b566f7A27B94566cac",
+            "0x3Cd0A705a2DC65e5b1E1205896BaA2be8A07c6e0",
+            "0x798d4Ba9baf0064Ec19eB4F0a1a45785ae9D6DFc",
+            "0x773539d4Ac0e786233D90A233654ccEE26a613D9",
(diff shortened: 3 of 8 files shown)


core/src/utils/account_id20.rs
```diff
@@ -0,0 +1,162 @@
...
+    InvalidChecksum,
+    InvalidPrefix,
+}
+
+impl Display for FromChecksumError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            FromChecksumError::BadLength => write!(f, "Length is bad"),
+            FromChecksumError::InvalidChecksum => write!(f, "Invalid checksum"),
+            FromChecksumError::InvalidPrefix => write!(f, "Invalid SS58 prefix byte."),
```
pkhry, line 90:
```suggestion
            FromChecksumError::InvalidPrefix => write!(f, "Invalid checksum prefix byte."),
```

core/src/utils/account_id20.rs
```diff
@@ -0,0 +1,162 @@
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! The "default" Substrate/Polkadot AccountId. This is used in codegen, as well as signing related bits.
```
niklasad1, line 5: I have troubles to understand `The "default" Substrate/Polkadot AccountId` part, to my understanding AccountId32 is the default one and AccountId20 is used for chains that are using ethereum accounts.

I would rather say that it's used for chains that is using ethereum accounts which are different from the default AccountId32 that is used by substrate by default.

For substrate-based chains that exposes AccountId20 in the metadata, this type can used to generate code when the correct AccountId rather than converting back and forth between AccountId20 and AccountId32

That is my understanding of it...

core/Cargo.toml
```diff
@@ -65,6 +65,10 @@ sp-core = { workspace = true, optional = true }
 sp-runtime = { workspace = true, optional = true }
 tracing = { workspace = true, default-features = false }
 
+# AccountId20
+keccak-hash = { workspace = true}
+
+
```
niklasad1, line 71:
```suggestion
```

core/src/utils/account_id20.rs
```diff
@@ -0,0 +1,162 @@
...
+    Ord,
+    PartialOrd,
+    Encode,
+    Decode,
+    Debug,
+    scale_encode::EncodeAsType,
+    scale_decode::DecodeAsType,
+    scale_info::TypeInfo,
+)]
+/// Eth public_key_hash
```
niklasad1, line 31:
```suggestion
/// A Ethereum-compatible `AccountId`.
```

core/src/utils/account_id20.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl core::fmt::Display for AccountId20 {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "{}", self.checksum())
+    }
+}
+
+impl core::str::FromStr for AccountId20 {
+    type Err = FromChecksumError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
niklasad1, line 126: This assumes `hex string` as input correct?

I think you need to check that it starts with `s.starts_with("0x")`  or use s.get(2..) because it can may panic if someone enters a String with unicode characters that isn't 1 byte.

Such as

```rust
    let v = String::from("🗻∈🌏");
    
    // panics
    println!("{}", &v[2..]);
```

It's a rare edge-case but I think we should check that :)

core/src/utils/account_id20.rs
```diff
@@ -0,0 +1,162 @@
...
+impl From<[u8; 20]> for AccountId20 {
+    fn from(x: [u8; 20]) -> Self {
+        AccountId20(x)
+    }
+}
+
+impl AccountId20 {
+    /// Convert to a public key hash
+    pub fn checksum(&self) -> String {
+        let hex_address = hex::encode(self.0);
```
niklasad1, line 55: I think this code is just moved but we think we could get rid of this `hex::encode call` it should be possible to iterate over the bytes instead of allocating a String.

This probably doesn't matter in practice but just a thought :)

core/Cargo.toml
```diff
@@ -65,6 +65,10 @@ sp-core = { workspace = true, optional = true }
 sp-runtime = { workspace = true, optional = true }
 tracing = { workspace = true, default-features = false }
 
+# AccountId20
+keccak-hash = { workspace = true}
```
niklasad1, line 69: quite annoying to add this dependency to core but no way around it I guess without feature gating it...

core/src/utils/account_id20.rs
pkhry: good catch, i didnt read into the top part of the file when copying license for accountId32 and brought this with it. replaced this part

signer/Cargo.toml
```diff
@@ -33,7 +33,7 @@ std = [
 # https://github.com/rust-bitcoin/rust-bitcoin/issues/930#issuecomment-1215538699
 sr25519 = ["schnorrkel"]
 ecdsa = ["secp256k1"]
-unstable-eth = ["keccak-hash", "ecdsa", "secp256k1", "bip32"]
+unstable-eth = ["subxt", "keccak-hash", "ecdsa", "secp256k1", "bip32"]
```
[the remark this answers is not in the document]
pkhry: replaced with `dep:subxt-core`

core/src/utils/account_id20.rs
pkhry: added the check

core/src/utils/account_id20.rs
niklasad1: no,  I don't think it matters just an unnecessary allocation :)

signer/Cargo.toml
```diff
@@ -33,7 +33,7 @@ std = [
 # https://github.com/rust-bitcoin/rust-bitcoin/issues/930#issuecomment-1215538699
 sr25519 = ["schnorrkel"]
 ecdsa = ["secp256k1"]
-unstable-eth = ["keccak-hash", "ecdsa", "secp256k1", "bip32"]
+unstable-eth = ["dep:subxt-core", "keccak-hash", "ecdsa", "secp256k1", "bip32"]
```
niklasad1, line 36: do we need the keccak-hash now when it's dependency in subxt-core?

niklasad1 (approved):
LGTM, nice to have this for sure.

I guess when the new metadata v16 then the codegen can be smart enough to use AccountId20 instead of AccountId32 from ethereum-style metadata/chains?

pkhry:
> LGTM, nice to have this for sure.
> 
> I guess when the new metadata v16 then codegen can smart enough to use AccountId20 instead of AccountId32 from ethereum-style metadata/chains?

We need to know the path for the accountId20 to provide a substitution in this case, If anyone can tag in with it we can add it to the substitution rules inside codegen/lib

codegen/src/lib.rs
```diff
@@ -371,6 +371,10 @@ fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
             parse_quote!(sp_core::crypto::AccountId32),
             parse_quote!(#crate_path::utils::AccountId32),
         ),
+        (
+            parse_quote!(fp_account::AccountId20),
```
niklasad1, line 375: Maybe add a comment that type comes from frontier which probably works for frontier/moonbeam but probably not for all ethereum-based chains.

I have no strong opinions whether to support this override or not, doesn't hurt I guess...

niklasad1 (approved):

codegen/src/lib.rs
jsdw: Mmm it might be that we can add a few overrides if needed to cater for the common chains, but otherwise the user can do this themselves.

I was hesitent but don't see a reason not to add this since its super unlikely to ever break anything that doesn't care :)

signer/src/eth.rs
```diff
@@ -6,11 +6,11 @@
 
 use crate::ecdsa;
 use alloc::format;
-use alloc::string::String;
 use core::fmt::{Display, Formatter};
 use core::str::FromStr;
 use keccak_hash::keccak;
 use secp256k1::Message;
+use subxt_core::utils::AccountId20;
```
jsdw, line 13: Let's avoid importing `subxt-core` here, and instead copy the pattern from eg `sr25519.rs` where we have a standalong `PublicKey` struct and then in the `subxt_compat` section we pull in `subxt-core` and have the relevant `From` impls etc to make `PublicKey` play nice with `subxt`'s `AccountId20` :)

signer/Cargo.toml
jsdw: From my prev comment, let's remove the subxt-core dep here :)

jsdw:
Looks good to me! Only things to do are fixing the CI issues and removing the `subxt-core` dep from `subxt-signer`'s eth feature by following the same pattern as `sr25519.rs` etc :)

signer/src/eth.rs
pkhry: what about the tests for eth? they will be enable only if "subxt" feature is also enabled

pkhry:
issues with subxt-macro related, to clippy

signer/src/eth.rs
jsdw: Ah the tests are no problem; currently we just assume the default features are enabled for them, which allows subxt-core anyways. I just think that we can avoid needing subxt-core by default for "eth" signing to fall in line with the other algorithms :)

signer/src/eth.rs
```diff
@@ -278,10 +150,99 @@ pub mod dev {
...
 #[cfg(feature = "subxt")]
 mod subxt_compat {
+    use super::*;
+    use crate::ecdsa;
     use subxt_core::config::Config;
     use subxt_core::tx::signer::Signer as SignerT;
+    use subxt_core::utils::AccountId20;
+    use subxt_core::utils::MultiAddress;
+
+    impl Keypair {
```
jsdw, line 160: Ah this isn't quite what I had in mind; I was hitnking you could do something more like sr25519.rs and have eg:

```rust
/// The public key for an [`Keypair`] key pair. This is equivalent to an `AccountId20`.
pub struct PublicKey(pub [u8; 20]);

pub struct Keypair(..)

// Impl here; not gated by subxt-compat feature:
impl Keypair {
    // ...
}

#[cfg(feature = "subxt")]
mod subxt_compat {
    use super::*;

    // Only in this block do we now need to import subxt-core to
    // describe how to convert our PublicKey above into an AccountId20:
    use subxt_core::utils::AccountId20;

    impl From<PublicKey> for AccountId20 {
        fn from(value: PublicKey) -> Self {
            value.to_account_id()
        }
    }

    impl PublicKey {
        /// A shortcut to obtain an [`AccountId20`] from a [`PublicKey`].
        /// We often want this type, and using this method avoids any
        /// ambiguous type resolution issues.
        pub fn to_account_id(self) -> AccountId20 {
            AccountId20(self.0)
        }
    }

    impl<T: Config> SignerT<T> for Keypair
    where
        T::AccountId: From<PublicKey>,
        // looks like T::Address is also AcocuntId20 so the one From impl above works,
        // but this allows other address types to be used as long as there's a way to convert:
        T::Address: From<PublicKey>, 
        // sameish: we just use the Signature type above so I think this Just Works, but
        // also leaves door open for people to use other Signature types in their Config as 
        // long as they provide a From impl:
        T::Signature: From<Signature>,
    {
        fn account_id(&self) -> T::AccountId {
            self.public_key().into()
        }

        fn address(&self) -> T::Address {
            self.public_key().into()
        }

        fn sign(&self, signer_payload: &[u8]) -> T::Signature {
            self.sign(signer_payload).into()
        }
    }
}
```

pkhry: updated in line with the suggestion

core/Cargo.toml
```diff
@@ -64,6 +64,9 @@ sp-core = { workspace = true, optional = true }
...
```
niklasad1, line 68: Can we hide the dependency `keccak-hash` behind the subxt-compat feature flag now?

signer/src/eth.rs
```diff
@@ -89,18 +97,10 @@ impl Keypair {
...
-    /// Obtain the [`ecdsa::PublicKey`] of this keypair.
-    pub fn public_key(&self) -> ecdsa::PublicKey {
-        self.0.public_key()
-    }
-
-    /// Obtains the public address of the account by taking the last 20 bytes
-    /// of the Keccak-256 hash of the public key.
-    pub fn account_id(&self) -> AccountId20 {
+    /// Obtain the [`eth::PublicKey`] of this keypair.
+    pub fn public_key(&self) -> PublicKey {
```
jsdw, line 101: Ah yeah ok; it's a bit different from sr25519 etc because the public key isn't directly the bytes we want for the acocuntId. I like this wrapper type though; I suppose it could just contain the `ecdsa::PublicKey` at this point, but I'm not too bothered either way :)

jsdw (approved):
LGTM, nice job @pkhry!

muraca:
Lovely, thank you!

jsdw:
Next time we should make sure the CI is passing before merging (which in this case means fixing the failing test); just incase something else was broken in the other CI steps that we missed :)

PR #1655, opened 2024-06-24
fix(tests): don't assume order of balance transfer extrinsics

niklasad1:


testing/integration-tests/src/full_client/blocks/mod.rs
```diff
@@ -235,14 +235,22 @@ async fn fetch_block_and_decode_extrinsic_details() {
...
-    tx.as_root_extrinsic::<node_runtime::Call>().unwrap();
-    let ext = tx
-        .as_extrinsic::<node_runtime::balances::calls::types::TransferAllowDeath>()
-        .unwrap()
-        .unwrap();
-    assert_eq!(ext.value, 10_000);
-    assert!(tx.is_signed());
+    let mut balance_transfer = None;
+
+    for tx in block_extrinsics {
```
jsdw, line 240: You may also be able to replace the loop with something like:

```rust
if let Some(ext) = block_extrinsics.find_first::<node_runtime::balances::calls::types::TransferAllowDeath>().unwrap() {

}
```

But it's good to aleo be testing `as_root_extrinsic` anyways :)

jsdw (approved):

testing/integration-tests/src/full_client/blocks/mod.rs
niklasad1: yeah, it's already tested above I noticed now

PR #1658, opened 2024-06-26
Allow PartialExtrinsic to be held across await points

jsdw:
One of the main use cases of `PartialExtrinsic` is that you can build it, then go and obtain a signature from elsewhere (quite possibly asynchronously), and then hand the signature back to the `PartialExtrinsic` to build the final extrinsic to submit.

`PartialExtrinsic` didn't implement `Send` previously, but with this PR it does. An example is also added to prove/show this.


niklasad1 (approved):

pgherveou (approved):
Nice, thanks for adding that in 🙏

Issue #1631, opened 2024-06-09
Support signing with json exported from polkadot.js


ethernomad:
Accounts managed by https://polkadot.js.org/apps/ or the browser extension can be downloaded as an encrypted json file, for example:

```json
{
  "encoded": "",
  "encoding": {
    "content": [
      "pkcs8",
      "sr25519"
    ],
    "type": [
      "scrypt",
      "xsalsa20-poly1305"
    ],
    "version": "3"
  },
  "address": "5DV3QBwhs7djHop5WEcUQ1BMFSmsqvqnDhrYh343x495AL4v",
  "meta": {
    "genesisHash": "",
    "name": "",
    "whenCreated": 1639212245729
  }
}
```

I'm not sure what the name of this format is.

It would be useful for subxt_signer to be able to load a keypair from a json file and the correct password. 

I can implement a pull request for this.

jsdw:
I suspect it's a semi arbitrary format that the PJS extension invented and that one or two other wallets then decided to allow import of.

Is it worth being able to do this programmatically eg via some utility function in Subxt signer I wonder? Or would it make more sense to have a binary that can do this? (Can `subkey` do this already I wonder? https://docs.substrate.io/reference/command-line-tools/subkey/)

ethernomad:
I have a use case where a program needs to have an encrypted key on disk that can be decrypted interactively. This json file format seems ideal because an account can be created / exported with ease from the browser extension.

I've almost finished implementing a from_json() method in subxt_signer::sr25519::Keypair

jsdw:
> I've almost finished implementing a from_json() method in subxt_signer::sr25519::Keypair

Ok awesome, would be good to see! I imagine it needs additional dependencies so we may want to have this behind a feature flag. I'm also currently wondering whether it should be attached to a specific algorithm (eg sr25519) or be a standalone utility function which takes the JSon and gives back the private key or whatever, but let's see what you come up with!

niklasad1:
Closed by https://github.com/paritytech/subxt/pull/1661

PR #1661, opened 2024-06-29
Support loading keys from Polkadot-JS accounts.

ethernomad:
A common way to manage keys for Substrate accounts is to use the Polkadot JS browser extension. Accounts in the extension can be exported to an encrypted JSON file.

It would be very useful for subxt to support loading keys from these files. Both long running services and interactive programs built with subxt would be able to load accounts exported from the browser extension.

The JSON schema is described here:
https://github.com/polkadot-js/common/blob/37fa211fdb141d4f6eb32e8f377a4651ed2d9068/packages/keyring/src/types.ts#L67

The password is hashed with Scrypt and then the ciphertext is decrypted with XSalsa20Poly1305.

The implementation in the pull request is very rigid, inline with the JS implementation.

Here is the original issue: #1631


Cargo.toml
```diff
@@ -161,6 +161,9 @@ keccak-hash = { version = "0.10.0", default-features = false }
 secrecy = "0.8.0"
 sha2 = { version = "0.10.8", default-features = false }
 zeroize = { version = "1", default-features = false }
+base64 = { version="0.22.1", default-features = false }
```
niklasad1, line 164:
```suggestion
base64 = { version = "0.22.1", default-features = false }
```

signer/Cargo.toml
```diff
@@ -35,6 +40,9 @@ sr25519 = ["schnorrkel"]
 ecdsa = ["secp256k1"]
 unstable-eth = ["keccak-hash", "ecdsa", "secp256k1", "bip32"]
 
+# Enable support for loading key pairs from polkadot-js json.
+json = ["std", "subxt", "sr25519", "base64", "scrypt", "crypto_secretbox", "serde", "serde_json"]
```
niklasad1, line 44: Do we really need `std` enabled for this `json` feature?

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+        // Decode from Base64.
+        let decoded = base64::engine::general_purpose::STANDARD.decode(self.encoded)?;
+        if decoded.len() < 68 {
+            return Err(Error::UnsupportedEncoding);
+        }
+
+        // Extract scrypt parameters.
+        // https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/scrypt/fromU8a.ts
+        let salt = &decoded[0..32];
+        let n = u32::from_le_bytes(decoded[32..36].try_into().unwrap());
```
niklasad1, line 123: return err here instead unwrap please, perhaps add another error variant for these scrypt params.

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+        let p = u32::from_le_bytes(decoded[36..40].try_into().unwrap());
+        let r = u32::from_le_bytes(decoded[40..44].try_into().unwrap());
+
+        if n != 32768 || p != 1 || r != 8 {
+            return Err(Error::UnsupportedScryptParameters { n, p, r });
+        }
+
+        // Hash password.
+        let scrypt_params = scrypt::Params::new(15, 8, 1, 32).unwrap();
```
niklasad1, line 132: please use expect if these are infallible with a short explaination

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+        let mut key = Key::default();
+        scrypt::scrypt(password.as_bytes(), salt, &scrypt_params, &mut key).unwrap();
```
niklasad1, line 134: This like fail-able, perhaps return an error if this fails?!

Do you know what's the default size of the Key here? Is it always gonna be correct https://docs.rs/scrypt/latest/scrypt/fn.scrypt.html?

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+        let public_key = &plaintext[85..117];
+
+        if header != [48, 83, 2, 1, 1, 48, 5, 6, 3, 43, 101, 112, 4, 34, 4, 32]
+            || div != [161, 35, 3, 33, 0]
+        {
+            return Err(Error::InvalidKeys);
+        }
+
+        // Generate keypair.
+        let keypair = sr25519::Keypair::from_ed25519_bytes(secret_key).unwrap();
```
niklasad1, line 160: I would also want to return err here instead of unwrap

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+struct EncryptedJsonDescriptor {
+    /// Descriptor for the content
+    content: Vec<String>,
+    /// The encoding (in current/latest versions this is always an array)
+    r#type: Vec<String>,
+    /// The version of encoding applied
+    version: String,
+}
+
+#[derive(Deserialize)]
```
niklasad1, line 71: what's the reason for not support all fields https://github.com/polkadot-js/common/blob/37fa211fdb141d4f6eb32e8f377a4651ed2d9068/packages/keyring/src/types.ts#L52-#L65? 

Subxt doesn't need them but perhaps some user may find them useful...

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct EncryptedJsonDescriptor {
```
niklasad1, line 62: this name is also quite confusing I would just call it EncryptionMetadata

niklasad1 (commented):
great stuff, I mostly want to get rid of the unwraps and rename a few things.

overall nice PR

signer/src/json.rs
jsdw: given that we check `decoded.len() < 68` above, I don't think it's necessary to also handle these errors as they should be infallible right? But I'd make them `.expect("length checked above")` perhaps :)

signer/src/json.rs
jsdw: An explanation of the arbitrary values would def be nice

signer/src/lib.rs
```diff
@@ -37,6 +37,11 @@ pub mod ecdsa;
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable-eth")))]
 pub mod eth;
 
+/// A polkadot-js account json loader.
+#[cfg(feature = "json")]
```
jsdw, line 41: `json` feels a bit vague as a feature name to me; perhaps `"polkadot-js-json-compat"` or similar makes more sense, to be clear that this is about decoding the specific polkadot-js JSON format?

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+    encoding: EncryptedJsonDescriptor,
+    /// The ss58 encoded address or the hex-encoded version (the latter is for ETH-compat chains)
+    address: AccountId32,
+    /// The underlying metadata associated with the keypair
+    meta: Meta,
+}
+
+impl KeyringPairJson {
+    /// Decrypt JSON keypair.
+    pub fn decrypt(self, password: &str) -> Result<sr25519::Keypair, Error> {
```
jsdw, line 99: I reckon that the API in this module would be cleaner if the public API for it looked something like this:

```rust
mod polkadot_js_compat {
    /// Given a JSON keypair as exported from Polkadot-JS, this returns an [`sr25519::Keypair`]
    pub fn decrypt_json(json: &str, password: &str) -> Result<sr25519::Keypair, Error> { .. }

    /// Possible errors when calling the [`from_json`] function.
    pub enum Error { .. }
}
```

That way there is no visible struct and just one method that takes the JSON string and gives back a keypair or an error.

signer/Cargo.toml
```diff
@@ -65,6 +73,11 @@ secp256k1 = { workspace = true, optional = true, features = [
     "recovery",
 ] }
 keccak-hash = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
+base64 = { workspace = true, optional = true, features = ["alloc"] }
+scrypt = { workspace = true, default-features = false, optional = true }
+crypto_secretbox = { workspace = true, optional = true, features = ["alloc", "salsa20"] }
```
jsdw, line 80:
```suggestion

# These are used if the polkadot-js-compat feature is enabled
serde = { workspace = true, optional = true }
serde_json = { workspace = true, optional = true }
base64 = { workspace = true, optional = true, features = ["alloc"] }
scrypt = { workspace = true, default-features = false, optional = true }
crypto_secretbox = { workspace = true, optional = true, features = ["alloc", "salsa20"] }
```

signer/src/sr25519.rs
```diff
@@ -122,6 +122,17 @@ impl Keypair {
         Ok(Keypair(keypair))
     }
 
+    /// Construct a keypair from a slice of bytes, corresponding to
+    /// an Ed25519 expanded secret key.
+    pub fn from_ed25519_bytes(bytes: &[u8]) -> Result<Self, Error> {
```
jsdw, line 127: Maybe for now make this `pub(crate)`

jsdw (commented):
I'm happy with this in principle; just some suggesitons on the overall API etc to add to Niklas's ones! Nice one!

ethernomad:
> User @ethernomad, please sign the CLA [here](https://cla.parity.io/sign_cla/161566/3).

I get a 404 when I try to sign the CLA - anyone know where I can report this?

Screenshot 2024-07-18 at 10-19-20 Parity Contributor Signatures

signer/src/json.rs
ethernomad: I improved the error handling.

signer/src/json.rs
ethernomad: I added expect() and the explanation from the JS implementation.

signer/src/json.rs
ethernomad: Switched to expect() - crypto_secretbox::Key is 32 bytes.

signer/src/json.rs
```diff
@@ -0,0 +1,204 @@
...
+    when_created: u64,
+}
+
+/// https://github.com/polkadot-js/common/blob/37fa211fdb141d4f6eb32e8f377a4651ed2d9068/packages/keyring/src/types.ts#L67
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[allow(dead_code)]
+pub struct KeyringPairJson {
+    /// The encoded string
+    encoded: String,
```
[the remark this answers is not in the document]
ethernomad: We can't change this - it's part of the file format.

signer/src/sr25519.rs
ethernomad: Fixed.

signer/src/json.rs
ethernomad: Implemented.

signer/src/lib.rs
ethernomad: done

signer/src/json.rs
ethernomad: done

signer/src/json.rs
ethernomad: As the KeyringPairJson struct is no longer pub I removed all the unused fields.

signer/Cargo.toml
ethernomad: This is necessary for the following command to succeed:
```
cargo test --no-default-features -F polkadot-js-compat
```

ethernomad:
This pull request is ready to review again.

signer/Cargo.toml
jsdw: Do you know why this is necessary for that command to succeed? I had another look over and it's not a huge deal but would be nice not to need the std feature :)

jsdw (approved):
Looks good to me! Couple of tiny nits but happy to see this merge now :)

Ideally we'd avoid needing the std feature for the polkadot-js-compat, but it's not really a big deal!

signer/src/polkadot_js_compat.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        Ok(keypair)
+    }
+}
+
+/// Given a JSON keypair as exported from Polkadot-JS, this returns an [`sr25519::Keypair`]
+pub fn decrypt_json(json: &str, password: &str) -> Result<sr25519::Keypair, Error> {
+    let pair_json: KeyringPairJson = serde_json::from_str(json)?;
+    Ok(pair_json.decrypt(password)?)
+}
```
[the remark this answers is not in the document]
ethernomad: done!

ethernomad:
The cla bot still crashes when I try to sign it.

jsdw:
> The cla bot still crashes when I try to sign it.

It seems to have been accepted anyways by the looks of things :)

signer/Cargo.toml
```diff
@@ -25,6 +25,11 @@ std = [
     "bip39/std",
     "schnorrkel/std",
     "secp256k1/std",
```
jsdw, line 27:
```suggestion
    "schnorrkel?/std",
    "secp256k1?/std",
```

In fact, these should also be optional only I think, though not an issue with this PR in itself!

jsdw:
Modulo a couple of tiny dependency related nits I'm happy with this! Thankyou @ethernomad!

niklasad1 (approved):

mordamax:
@ethernomad re. CLA bot - you've already signed it, so all good

image

it's just bad user-experience in its logic, when you try to sign once-again it gives 404 O_o which was unexpected for me to see too :D

PR #1676, opened 2024-07-10
Add hash method to ExtrinsicDetails

jsdw:
And ensure that it is identical to the `hash()` method from the `Transaction` type.


pgherveou (approved):
🙏

lexnv (approved):

PR #1683, opened 2024-07-17
feat(utils): add MultiSigner helper

FlorianFranzen:
This add the `MultiSIgner` to the types of supported multi types.


jsdw:
Thanks for the PR! Do you have any example of where this would be useful? I'm generally a bit hesitant to add things that rely on `substrate-compat` since it would be ideal to remove that feature completely at some point (especially since Subxt rarely keeps in line with `sp_*` releases, and chains will also be using various different versions)

jsdw:
We've removed the `substrate-compat` feature flag entirely now, and so I'll close this PR. perhaps this is a place we could improve docs or add an example of how to do such a thing though, so i'd be open to thinking of something in that area that would make it clear how to solve this :)

Issue #1677, opened 2024-07-11
Stabilise the `unstable-reconnecting-rpc-client`


jsdw:
We'd like to stabilise the automatic reconnecting logic and make it the default behaviour.

I think that this requires two things:

1. Making sure that the reconnecting logic in Subxt is tested. This means testing the Backend impls, but also some of the structs returned that hande eg storage streams and making sure that they all work as expected. Let's unit test what we can, and if needbe we can create a mock RpcClient implementation and see how things respond to that returning an error or disconnecting.
2. Niklas had the idea to take what we need from https://crates.io/crates/reconnecting-jsonrpsee-ws-client and inline it into Subxt, since I think that crate exposes a bunch of features we don't need given that we handle all of the reconnect logic in Subxt.

Once these are done and we have some confidence that it works, we can make this the default and remove the feature flag entirely. Maybe we should continue exposing the "not reconnecting" jsonrpsee client too in case people want to fall back to the old behaviour or handle reconnecting in some other way.

niklasad1:
Yes, I can confirm that this is the roadmap

pkhry:
I've encountered some issues with testing The `Backend` trait and its implementations. 
### Explanation
The `Backend` impl requires a `RpcClient` implementation which would connect to the Server.
But the problem is that RpcClient is using `serde_json::from_str`in `request` and `RpcSubscription`. RpcClient also expects `RpcClientT` to return `serde_json::raw::RawValue`.
So to test the `Backend` implementation we need to either mock the `RpcClient` or the server implementations. 
### Mocking the RpcClient 
Using concrete implementation of `RpcClientT` and then passing it to the `RpcClient`. 
The problem is that we don't have `Serialize` instances defined for many data types inside `subxt`  that we expect to return inside the responses from `RpcClientT`.
Ways to circumvent this without defining `Serialize` for every type we need are: 
 - `cfg_attr` conditional attribute 
 - just use `serde_json::value::Value`/`json!()` and build data types by hand to be returned in each case by the `RpcClientT`. 
 - Try to decouple `RpcClient` from concrete wire format, to be able to pass identity values of things we expect to get back from the `RpcClientT` -> this seems completetely useless as we use JSON-RPC.
 ### Mocking the Server 
 Seems like an overkill to me as this will require us to start up and kill a test server for every test and any additional wins in testing would be for `RpcClientT` implementations, but this feels like testing internals of the libraries we use. 
 ### In short: 
[...]

PR #1711, opened 2024-08-15
Integration tests for `unstable-reconnecting-rpc-client`

pkhry:
### Description
[Issue link](https://github.com/paritytech/subxt/issues/1677)
>>1. Making sure that the reconnecting logic in Subxt is tested. This means testing the Backend impls, but also some of the structs returned that hande eg storage streams and making sure that they all work as expected. Let's unit test what we can, and if needbe we can create a mock RpcClient implementation and see how things respond to that returning an error or disconnecting.

I've plugged the new rpc client into integration tests behind a feature-flag.

### Reconnection testing
I've added a few functions to have the ability to restart the `substrate-node` when needed in `client` tests, but so far it seems like handwriting a server with predetermined responses might be a better idea as  node restarts can take upwards of 30-60s inside a test.


niklasad1 (commented):
Cool,

So for me the MVP is to add one extra test for the unstable and stable backend where we restart the node and see that it's works as intended i.e, running all existing tests with reconnecting rpc client is not needed.  I don't know what's the best to test for this but I suggest 1) block subscription,  2) kill/restart the node 3) check that block subscription works after reconnect

AFAIU, that may take 1-3 minutes to do as starting the node can be slow which should be acceptable since some other jobs are already quite slow because they run in parallel.

Bonus stuff as the first step would be to test that unstable backend emit an error when the block subscription misses blocks during "reconnect".

testing/integration-tests/src/full_client/client/mod.rs
```diff
@@ -121,6 +121,9 @@ async fn transaction_validation() {
 
     wait_for_blocks(&api).await;
 
+    #[cfg(feature = "unstable-reconnecting-rpc-client")]
+    let _ctx = ctx.restart().await;
```
niklasad1, line 125: I would rather create a separate test for unstable-reconnecting stuff than injecting it here.

Also this would remove all feature-gated stuff in this PR and just feature-gate the test itself

pkhry: removed all the feature gating stuff and moved this to a separate test

niklasad1 (approved):
nice, looks good

testing/integration-tests/Cargo.toml
```diff
@@ -25,6 +25,10 @@ unstable-light-client-long-running = ["subxt/unstable-light-client"]
 # the default one which relies on the "old" RPC methods.
 unstable-backend-client = []
 
+# Enable the unstable reconnecting rpc-client
+unstable-reconnecting-rpc-client = []
+
+
```
niklasad1, line 31:
```suggestion
```

testing/integration-tests/src/full_client/client/mod.rs
```diff
@@ -409,3 +409,40 @@ async fn partial_fee_estimate_correct() {
...
+    let signed_extrinsic = api
+        .tx()
+        .create_signed(&tx, &alice, Default::default())
+        .await
+        .unwrap();
+
+    signed_extrinsic
+        .validate()
+        .await
+        .expect("validation failed");
```
jsdw, line 439: `.validate()` obtains the latest block ref and then calls some runtime API, but I'm not sure how much that is testing the reconnecting logic, so curious what you think about this? I guess the point is that Subxt reconnects and doesn't start throwing out errors for every call, which is definitely a good start!

A couple of other tests that we might want to add:

- Subscribe to storage entries and restart the node while streaming them, checking that you don't miss any inbetween or start over or whatever. (might require some setup though to get some storage entries of interest, or maybe subscribe to accountIds and expect the dev ones or something).

- Subscribe to finalized blocks and restart while this is happening, checking that the block numbers of returned blocks follow eachother nicely regardless (though 6s block times means you'd probably not often screw up)

testing/integration-tests/src/utils/node_proc.rs
```diff
@@ -58,26 +62,29 @@ where
...
 
+    pub async fn restart(mut self) -> Self {
+        tokio::task::spawn_blocking(move || {
+            if let Some(ref mut proc) = &mut self.proc {
+                let _ = proc.restart().unwrap();
+            }
+            self
+        })
+        .await
+        .expect("to succeed")
```
jsdw, line 73: Maybe a DQ but why does this have to be in a spawn_blocking?

jsdw (approved):
Generally looks good to me, and a good foundation for adding some more integration tests around restarting! Mainly I just have a Q about whether we need the feature flag, and what test(s) are best to have (though happy to start with the test you added and then add some more)!

testing/integration-tests/src/full_client/client/mod.rs
```diff
@@ -409,3 +409,40 @@ async fn partial_fee_estimate_correct() {
     // Both methods should yield the same fee
     assert_eq!(partial_fee_1, partial_fee_2);
 }
+
+#[cfg(feature = "unstable-reconnecting-rpc-client")]
```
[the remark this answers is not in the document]
niklasad1: we shouldn't need it anymore

testing/integration-tests/Cargo.toml
```diff
@@ -25,6 +25,9 @@ unstable-light-client-long-running = ["subxt/unstable-light-client"]
 # the default one which relies on the "old" RPC methods.
 unstable-backend-client = []
 
+# Enable the unstable reconnecting rpc-client
+unstable-reconnecting-rpc-client = []
```
niklasad1, line 29: this isn't needed anymore as you already enabled the unstable reconnecting feature for tests, I'm happy to remove this feature flag :)

testing/integration-tests/src/full_client/client/mod.rs
pkhry: > Subscribe to finalized blocks and restart while this is happening, checking that the block numbers of returned blocks follow eachother nicely regardless (though 6s block times means you'd probably not often screw up) 

This will return a toplevel error instead of retrying in the stream currently.

> Subscribe to storage entries and restart the node while streaming them, checking that you don't miss any inbetween or start over or whatever. (might require some setup though to get some storage entries of interest, or maybe subscribe to accountIds and expect the dev ones or something).
I've rewritten the test to check that clients follow the same rules for streaming finalized blocks

testing/integration-tests/src/full_client/client/mod.rs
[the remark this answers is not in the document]
pkhry: removed

testing/integration-tests/src/utils/context.rs
```diff
@@ -28,5 +31,9 @@ pub type TestContext = TestNodeProcess<SubstrateConfig>;
 pub type TestClient = OnlineClient<SubstrateConfig>;
 
 pub async fn test_context() -> TestContext {
-    test_context_with("alice".to_string()).await
+    test_context_with("alice".to_string(), RpcClientKind::Legacy).await
+}
+
+pub async fn test_context_unstable_rpc_client() -> TestContext {
```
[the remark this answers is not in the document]
pkhry: renamed

testing/integration-tests/Cargo.toml
pkhry: removed

Issue #1660, opened 2024-06-27
Support fetching metadata from crate


bkchr:
The `subxt` macro currently supports to fetch the metadata from the filesystem or from some RPC node. However, for the future zombienet rust tests, it would be nice to also specify a crate.

So, something like this:
```
subxt(runtime_crate = "crate_name", wasm_file_name = "name", features = ["fast-runtime"])
```

The `crate_name` would be required to be path of the same workspace. `wasm_file_name` should be optional and if not given, it should be guessed based on the build output. If there is more than one wasm file, the macro should complain. `features` is also optional and just activates some feature at compile time.

https://github.com/paritytech/polkadot-sdk/tree/85f0edae1d6c461081f42fcabd13e547e5a5b683/substrate/.maintain/build-only-wasm.sh could be helpful here.

jsdw:
From the linked issue, I would prefer that you can point Subxt at the WASM runtime and have it extract and use the metadata from it. Then, I think we could have:

- `runtime_metadata_path` - get metadata from path
- `runtime_metadata_insecure_url` - get metadata from (possibly not secure) URL
- `runtime_path` - get metadata from runtime at path

I'd prefer to avoid the subxt macro having to know or care anything about the structure of the `polkadot-sdk` repo if possible :)

bkchr:
> I'd prefer to avoid the subxt macro having to know or care anything about the structure of the `polkadot-sdk` repo if possible :)

I mean what I proposed does not require to know the structure of the repo? Only need to know how `substrate-wasm-builder` works, but this is reasonable.

pepoviola:
Hi 👋,  I think we want to pass the `crate_name` as @bkchr mention in order to compile _on the fly_ to get the `wasm` to use. I currently add a similar logic to the `build` script in the [pr](https://github.com/paritytech/polkadot-sdk/pull/4883/files#diff-fdf350a016dacbd38b74e0636692268491d531a9e03bdde2f5861c5d38dfb17f) but would be great to have this implemented in subxt.

Thanks!

jsdw:
Since we need to extract metadata from the runtime WASM regardless, adding `runtime_path` as a first step seems sensible anyway, and then we can consider actually building the runtime and extracting the wasm from it. 

[...]

PR #1720, opened 2024-08-21
Add support for generating metadata from runtime wasm files

pkhry:
### Description 
See #1660 

Adds `wasm_file_path` keyword support to `subxt::subxt` macro to allow extracting metadata from the runtime WASM.
 
No support for compiling target crate to wasm and using it to generate the Metadata for now.

code taken from [polkadot-sdk PR](https://github.com/paritytech/polkadot-sdk/blob/7fc345d4d8450b92fda03866af8e75d263c8ae64/polkadot/zombienet-sdk-tests/build.rs#L1)


Cargo.toml
```diff
@@ -115,6 +115,7 @@ which = "5.0.0"
 strip-ansi-escapes = "0.2.0"
 proptest = "1.5.0"
 hex-literal = "0.4.1"
+subwasmlib = { git = "https://github.com/chevdor/subwasm", rev = "v0.21.3" }
```
lexnv, line 118: nit; There's no released crate on crates io?

macro/src/lib.rs
```diff
@@ -23,6 +23,26 @@ use subxt_codegen::{
...
+        CodegenError::Fetch(subxt_codegen::error::FetchMetadataError::Other(
+            err.to_string(),
+        ))
+    }
+    let source =
+        Source::from_options(Some(wasm_path.to_path_buf()), None, None, None).map_err(to_error)?;
+    let subwasm = Subwasm::new(&source.try_into().map_err(to_error)?).map_err(to_error)?;
+    let mut output = vec![];
+    subwasm
+        .write_metadata(OutputFormat::Scale, None, &mut output)
```
lexnv, line 41: Ideally, it would be interesting if we could:
- decode the metadata in a way that we don't depend on substrate crates (`sp-*` `sc_executor`)
- extract from the wasm blob just the `Metadata_metadata()` bytes

If that turns out to be too complicated, we can still look into obtaining that information via substrate-crates. Using subwasmlib to fetch the metadata brings quite a tree of dependencies.

For extracting the wasm blob, we'd need to rely on `sp_maybe_compressed_blob`:
 https://github.com/chevdor/subwasm/blob/fae3e13bd1f8680bf32cbf2ed223223583f8d713/libs/wasm-loader/src/lib.rs#L152

For calling into the wasm blob and fetching the metadata, we'd rely on`sc-executor`:
https://github.com/chevdor/subwasm/blob/fae3e13bd1f8680bf32cbf2ed223223583f8d713/libs/wasm-testbed/src/lib.rs#L131-L147

lexnv (commented):
Nice! Looks good as a first step! 

I would try to explore decoding the wasm blob internally, or use the substrate crates directly 

edit; I think the end-goal here is to fetch the wasm blob directly from a repo (ie https://github.com/paritytech/subxt/issues/1660#issuecomment-2303077118), which then justifies the `subwasmlib ` dependency.

Cargo.toml
pkhry: nope

niklasad1: https://github.com/chevdor/subwasm/issues/98

jsdw:
> edit; I think the end-goal here is to fetch the wasm blob directly from a repo (ie https://github.com/paritytech/subxt/issues/1660#issuecomment-2303077118), which then justifies the subwasmlib dependency.

The end goal for me here is to point the subxt macro to some WASM runtime file (just as you can also point to the metadata.scale currently) and have it extract the metadata from that and use the metadata to generate an interface.

> I would try to explore decoding the wasm blob internally, or use the substrate crates directly

I agree; I think it'd be good to look into how easily we can avoid this subwasm crate and the various dependencies it brings in and to the decoding internally (or using the relevant `sp-*` crate as long as it doesn't bring in too much itself). So I'm roughly in agreement with https://github.com/paritytech/subxt/pull/1720#discussion_r1726521326 :)

macro/src/lib.rs
pkhry: rewritten to use direct dependencies in last commit

pkhry:
> > edit; I think the end-goal here is to fetch the wasm blob directly from a repo (ie [#1660 (comment)](https://github.com/paritytech/subxt/issues/1660#issuecomment-2303077118)), which then justifies the subwasmlib dependency.
> 
> The end goal for me here is to point the subxt macro to some WASM runtime file (just as you can also point to the metadata.scale currently) and have it extract the metadata from that and use the metadata to generate an interface.
> 
> > I would try to explore decoding the wasm blob internally, or use the substrate crates directly
> 
> I agree; I think it'd be good to look into how easily we can avoid this subwasm crate and the various dependencies it brings in and to the decoding internally (or using the relevant `sp-*` crate as long as it doesn't bring in too much itself). So I'm roughly in agreement with [#1720 (comment)](https://github.com/paritytech/subxt/pull/1720#discussion_r1726521326) :)

I've redone things using relevant crates directly and PR unfortunately is still pulling quite a lot of deps.

codegen/src/error.rs
```diff
@@ -57,6 +57,9 @@ pub enum CodegenError {
     /// Cannot generate types.
     #[error("Type Generation failed: {0}")]
     TypeGeneration(#[from] TypegenError),
+    /// Error when generating metadata from Wasm-runtime
+    #[error("Failed to generate metadata from wasm file\nreason: {0}")]
```
jsdw, line 61: Nit: Does the reason warrant a newline here? We don't use one in any of the above errors

testing/ui-tests/src/correct/wasm_runtime_metadata.rs
```diff
@@ -0,0 +1,14 @@
+#[subxt::subxt(wasm_file_path = "../../../../artifacts/westend_runtime.wasm")]
```
jsdw, line 1: Nit: could we call it `runtime_path` perhaps? Then we have `runtime_metadata_path` to point to the path of some runtime metadata file, and `runtime_path` to point to the path of some runtime :)

niklasad1: or "runtime_wasm_path" but the native runtime is substrate is removed now? So, runtime_path is maybe sufficient :)

macro/Cargo.toml
```diff
@@ -27,6 +27,11 @@ syn = { workspace = true }
 quote = { workspace = true }
 subxt-codegen = { workspace = true, features = ["fetch-metadata"] }
 scale-typegen = { workspace = true }
+sc-executor = { workspace = true }
+sp-maybe-compressed-blob = { workspace = true }
+sp-state-machine = { workspace = true }
+sp-io = { workspace = true }
+sc-executor-common = { workspace = true }
```
jsdw, line 34: I wondfer how many dependencies we end up pulling in here; possibly worth putting behind a feature flag like "runtime_file_path" and then perhaps not enabling it by default (since I think it's probably a bit of an edge case that people would point straight to a runtime WASM blob).

jsdw (commented):
A couple of small things re name and putting behind a feature flag but otherwise looks very clean; good job!

testing/ui-tests/src/correct/wasm_runtime_metadata.rs
jsdw: I'd be happy with `runtime_wasm_path` too; either that or `runtime_path` feels consistent with the other attrs to me :)

macro/Cargo.toml
pkhry: moved them to optional and hidden behind feature flag

macro/src/lib.rs
```diff
@@ -206,6 +212,22 @@ fn validate_type_path(path: &syn::Path, metadata: &Metadata) {
...
     let unstable_metadata = args.unstable_metadata.is_present();
+
+    #[cfg(feature = "runtime-path")]
+    if let Some(path) = &args.runtime_path {
+        if args.runtime_metadata_insecure_url.is_some() || args.runtime_metadata_path.is_some() {
+            abort_call_site!(
+                "Exclusively one of 'runtime_metadata_path', 'runtime_metadata_insecure_url' or `runtime_path` must be provided"
+            );
+        };
+        let root = std::env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
```
niklasad1, line 223: "." -> current_dir?

pkhry: Think so, I've just taken the default value from 'runtime_metada_path' branch tbh

macro/src/lib.rs
```diff
@@ -239,14 +261,9 @@ fn fetch_metadata(args: &RuntimeMetadataArgs) -> Result<subxt_codegen::Metadata,
                 .and_then(|b| subxt_codegen::Metadata::decode(&mut &*b).map_err(Into::into))
                 .map_err(|e| e.into_compile_error())?
         }
-        (None, None) => {
-            abort_call_site!(
-                "One of 'runtime_metadata_path' or 'runtime_metadata_insecure_url' must be provided"
-            )
-        }
-        (Some(_), Some(_)) => {
+        _ => {
```
niklasad1, line 264: This warning message isn't accurate if `--feature runtime-path` is enabled, so it could be improved by "feature-gating" this such as

```rust
match path {
    #[cfg(feature = "runtime-path")
    _ => {
           abort_call_site!(
                "Only one of 'runtime_metadata_path', 'runtime_metadata_insecure_url'  or  'runtime_path` can be provided"
                "Exclusively one of 'runtime_metadata_path' or 'runtime_metadata_insecure_url' must be provided"
            )
    }
    #[cfg(not(feature = "runtime-path"))
    _ => {
            abort_call_site!(
                "Only one of 'runtime_metadata_path' or 'runtime_metadata_insecure_url' can be provided"
                "Exclusively one of 'runtime_metadata_path' or 'runtime_metadata_insecure_url' must be provided"
            )
    }
};
```

Cargo.lock
niklasad1, line 1: nice call, to add a feature-flag for these wasm deps in substrate

niklasad1 (approved):

macro/src/lib.rs
pkhry: Yeah, good catch! changed per suggestion.

Cargo.toml
```diff
@@ -116,6 +116,8 @@ which = "5.0.0"
 strip-ansi-escapes = "0.2.0"
 proptest = "1.5.0"
 hex-literal = "0.4.1"
+sc-executor = "0.40.0"
+sc-executor-common = "0.35.0"
```
jsdw, line 120: Very much a Nit, but these are substrate crates too so could be below with the rest :)

jsdw (approved):
Looks great; nice work!

One tiny thing to do might also be to document this new command in the subct macro docs; I think we have some examples of different attrs and would be good to add one for this too.

PR #1729, opened 2024-08-26
fix: enable dependabot for all crates

niklasad1:
We have some crates outside the workspace which aren't updated on dependabot updates such as `testing/wasm-rpc-tests` and some example which this fixes.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
-    directory: "/"
+    directory: "**/*"
     schedule:
       interval: weekly
     ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
-    directory: "/"
+    directory: "**/*"
```
niklasad1, line 4: Specify all directories from the current layer and below recursively, using globstar, for locations of manifest files

jsdw (approved):
Ah cool, so the glob pattern will enable it for non workspace crates too!

PR #1739, opened 2024-08-30
Update example with `Client` instead of `RpcClient`

leonardocustodio:


niklasad1 (changes requested):
Hey,

I don't understand what you want to achieve with this PR

`subxt::backend::rpc::reconnecting_rpc_client::Client` has been renamed to `RpcClient` and this code doesn't compile anymore.

If you think Client is more readable for this example you have to do:

```rust
use subxt::backend::rpc::reconnecting_rpc_client::RpcClient as Client;
```

jsdw:
Will close this for now given the above

leonardocustodio:
I'm sorry. I suppose I looked at the wrong version. On mine, `RpcClient` would not compile, just `Client`. Maybe I was looking at an older version if it was renamed later.

PR #1751, opened 2024-09-02
Add tests for legacy `Backend` impl

pkhry:
### Description 
As in the title. This PR adds tests for `LegacyBackend` implementation in subxt/backend. 
Pr uses a mocked `RpcClient` implementation for providing expected responses. 

Part of the work on #1677


subxt/src/backend/legacy/mod.rs
```diff
@@ -95,28 +95,36 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
         keys: Vec<Vec<u8>>,
         at: T::Hash,
     ) -> Result<StreamOfResults<StorageResponse>, Error> {
-        retry(|| async {
-            let keys = keys.clone();
-            let methods = self.methods.clone();
-
-            // For each key, return it + a future to get the result.
-            let iter = keys.into_iter().map(move |key| {
+        fn get_entry<T: Config>(
```
niklasad1, line 98: I don't follow what you fixed with this but I think the previous code would try to fetch all storage keys in one retry block i.e, all storage key calls needs to be successful or retry it.

Looks more efficient to try each storage entry individually as you do now but maybe some other issue with it that I don't follow?

pkhry: Let's say we tried to fetch 2 entries:
- "A" and "B".
- "A" returns an OK result
- "B" returns an error
The function would return Ok(streamOf(OK,Err)

So it just would not retry

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,395 @@ pub enum TransactionStatus<Hash> {
...
+
+        impl RpcClientT for MockRpcClientStorage {
+            fn request_raw<'a>(
+                &'a self,
+                method: &'a str,
+                params: Option<Box<serde_json::value::RawValue>>,
+            ) -> super::rpc::RawRpcFuture<'a, Box<serde_json::value::RawValue>> {
+                Box::pin(async move {
+                    match method {
+                        "state_getStorage" => {
```
niklasad1, line 413: It would be quite useful to collect of rpc methods from the Backend to know that we cover the all exposed in that API such as method names changes etc but this works too.

pkhry: I mean a lot of them follow the same structure for legacy backend, but I can re add the tests for other methods too, it's just that they would add more changes by themselves by requiring serializers for expected data

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,395 @@ pub enum TransactionStatus<Hash> {
...
+                            let mut data = self.data.lock().await;
+                            let key = match serde_json::to_value(params.unwrap()).unwrap() {
```
niklasad1, line 415: I think jsonpsee is already in the dependency tree and you could simplify this code to use.

```rust
   let rpc_params = jsonrpsee::types::Params::new(params.map(|p| p.get());
   let key: sp_core::Bytes = rpc_params.one().unwrap();
```

Perhaps move to helper function to make it easier to read and follow

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,395 @@ pub enum TransactionStatus<Hash> {
...
+        /// The test covers them because they follow the simple pattern of:
+        /// ```no_run
+        ///  async fn THE_THING(&self) -> Result<T::Hash, Error> {
+        ///    retry(|| <DO THE THING> ).await
+        ///  }
+        /// ```
+        async fn simple_fetch() {
+            let hash = crate::utils::H256::random();
+            // Setup
+            let mut entries = StorageEntry::new();
```
niklasad1, line 600: Picking hairs, it's bit strange with StorageEntry here to me it looks just mocked data for each rpc call/subscription. 

I would rather just to call it subscription and request data or something

niklasad1 (commented):
Nice work, looks promising.

I would like some more wrapper functions to make the code a little easier to read such as `subscription` instead of using mpsc::channel etc. It took my a while to understand why the channel was not used for pure method calls etc.

subxt/src/backend/mod.rs
niklasad1: yeah, it doesn't work nicely with subscriptions anyway so ignore this

subxt/src/backend/legacy/mod.rs
jsdw: Yup ok, so previously we wrapped code in `retry`, but that code didn't actually do any async stuff and just returned a `StreamOf`. Now, we are retrying each individual async request to fetch a storage value. Good catch!

niklasad1: Cool, you could also create a separate PR for this if you want get it in ASAP :)

subxt/src/backend/mod.rs
pkhry: renamed to MockDataTable

subxt/src/backend/mod.rs
pkhry: done as suggested

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,395 @@ pub enum TransactionStatus<Hash> {
...
+        struct Data<R: Storage> {
+            request: R,
+            subscription: mpsc::Receiver<RpcResult<Vec<Item>>>,
+        }
+
+        struct MockRpcClientStorage {
+            data: Arc<Mutex<Data<StorageEntry>>>,
+        }
+
+        impl RpcClientT for MockRpcClientStorage {
```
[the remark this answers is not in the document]
pkhry: i'm currently trying to see if it would work with unstable backend testing

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,399 @@ pub enum TransactionStatus<Hash> {
...
+            }
+        }
+
+        fn runtime_version(num: u32) -> RuntimeVersion {
+            RuntimeVersion {
+                spec_version: num,
+                transaction_version: num,
+                other: HashMap::new(),
+            }
+        }
```
jsdw, line 495: This is a good start! I think it would be good to try and make it super clear/simple if possible, eg something along the lines of:

```rust
#[tokio::test]
async fn storage_fetch_values() {
    // Setup
    let mock_data = [
        ("ID1", Ok("Data1".as_bytes())),
        ("ID2", Err(RpcError::DisconnectedWillReconnect("Reconnecting".to_string()))),
        ("ID2", Ok("Data2".as_bytes())),
        ("ID3", Err(RpcError::DisconnectedWillReconnect("Reconnecting".to_string()))),
        ("ID3", Ok("Data3".as_bytes())),
    ];
    let mock_client = mock_storage_client(mock_data.clone());

    // Test
    let backend: LegacyBackend<Conf> = LegacyBackend::builder().build(rpc_client);
    let response = backend
        .storage_fetch_values(
            mock_data.iter().map(|x| x.0.clone().into()).collect(),
            crate::utils::H256::random(),
        )
        .await
        .unwrap();

    let response = response
        .map(|x| x.unwrap())
        .collect::<Vec<StorageResponse>>()
        .await;
    let expected: Vec<_> = mock_data
        .iter()
        .map(|(key, value)| storage_response(key, value))
        .collect();
    assert_eq!(expected, response)
}
```

pkhry: factored some of the common code in the last commit, let me know if it looks good to you

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,416 @@ pub enum TransactionStatus<Hash> {
...
+                subscription: Subscription::from_vec(subscription_data).await,
+            };
+            RpcClient::new(MockRpcClientStorage {
+                data: Arc::new(Mutex::new(data)),
+            })
+        }
+
+        #[tokio::test]
+        async fn storage_fetch_values() {
+            // Setup
```
niklasad1, line 561: nice, much easier to read now

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,416 @@ pub enum TransactionStatus<Hash> {
...
+        }
+
+        impl MockDataTable {
+            fn new() -> Self {
+                MockDataTable {
+                    items: HashMap::new(),
+                }
+            }
+
+            fn from_vec<'a, T: Serialize, I: IntoIterator<Item = (&'a str, RpcResult<T>)>>(
```
niklasad1, line 375: from_iter is probably more accurate but doesn't matter that much

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,417 @@ pub enum TransactionStatus<Hash> {
...
+        /// - `stream_runtime_version`
+        /// - `stream_all_block_headers`
+        /// - `stream_best_block_headers`
+        /// The test covers them because they follow the simple pattern of:
+        /// ```no_run
+        /// async fn stream_the_thing(
+        ///     &self,
+        /// ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
+        ///     let methods = self.methods.clone();
+        ///    let retry_sub = retry_stream(move || {
```
niklasad1, line 678:
```suggestion
        ///     let retry_sub = retry_stream(move || {
```

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,417 @@ pub enum TransactionStatus<Hash> {
...
+                Ok::<crate::client::RuntimeVersion, crate::Error>(client_runtime_version(0)),
+                Ok(client_runtime_version(4)),
+                Ok(client_runtime_version(5)),
+            ]);
+
+            while let Some(res) = results.next().await {
+                if res.is_ok() {
+                    assert_eq!(expected.pop_front().unwrap().unwrap(), res.unwrap())
+                } else {
+                    assert!(matches!(
```
niklasad1, line 730: Ok, this special assert is because the Error doesn't impl PartialEq?

pkhry: Yup, deriving it all the way through is a bit painful

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,417 @@ pub enum TransactionStatus<Hash> {
...
+
+            // Test
+            let backend: LegacyBackend<Conf> = LegacyBackend::builder().build(rpc_client);
+            let response = backend.genesis_hash().await.unwrap();
+
+            assert_eq!(hash, response)
+        }
+
+        #[tokio::test]
+        /// This test should cover following methods:
```
niklasad1, line 668: Not sure whether we should be extra verbose say: ` This test should cover the logic in the following methods`

Your example below is explains it very well

niklasad1 (approved):
LGTM

subxt/src/backend/mod.rs
jsdw: That'd be more consistent yup :)

subxt/src/backend/mod.rs
```diff
@@ -325,9 +325,416 @@ pub enum TransactionStatus<Hash> {
...
+            type Hash = crate::utils::H256;
+            type AccountId = crate::utils::AccountId32;
+            type Address = crate::utils::MultiAddress<Self::AccountId, ()>;
+            type Signature = crate::utils::MultiSignature;
+            type Hasher = crate::config::substrate::BlakeTwo256;
+            type Header = crate::config::substrate::SubstrateHeader<u32, Self::Hasher>;
+            type ExtrinsicParams = DefaultExtrinsicParams<Self>;
+
+            type AssetId = u32;
+        }
```
jsdw, line 508: Prob worth a comment for why this is necessary instead of just importing SubstrateConfig or whatever :)

jsdw (approved):
The tests look much cleaner now; nice one!

I still would like to see a more generic MockRpcClient thing that I think might help to simplify some of the code around the tests, but can be tried out as part of a separate PR!

subxt/src/backend/mod.rs
pkhry: done

Issue #1567, opened 2024-05-02
reconnecting rpc: ensure poll order to avoid outdated subscription ID on restart


niklasad1:
If we fail because the subscription ID is invalid at the moment (ie not propagated before we submitted the call), then how likely it is to fail with the same error a second time?

This could also be a followup since rpc-v2 are hard to reason about wrt reconnections. Does the race stil reproduce with 3/4 retries?

_Originally posted by @lexnv in https://github.com/paritytech/subxt/pull/1505#discussion_r1584924860_

jsdw:
One idea offhand (but maybe not the best one!) is something like (I can't remember if I wrote this somewhere already): 

1. Keep a "connection count" in the follow subscription thing, which is returned along with the subscription ID in the ready event and/or is available with every event (like the subscription ID might be). 
2. When we get the data we want from the follow subscription, also note the current connection count at that time.
3. If some call that depends on data from follow subscription returns a `DisconnectedWillReconnect`, then go back to 2, but wait until the connection count increments before trying to do anything with the data.

So for a real example this might mean that we: 
- get a subscription ID and note the connection count when we do
- call eg `chainHead_v1_header(sub_id)` 
    - if the chainHead call return an error then loop around and try to get subscription Id again, waiting for an Id that has a connection count which is >=1 higher.

This should ensure that the follow subscription has reconnected at least once since we got the disconnected error, and isn't stale :)

PR #1765, opened 2024-09-09
add reconnecting tests for unstable_backend

pkhry:
### Description 
As in the title. This PR adds tests for `UnstableBackend` implementation in subxt/backend. 
The tests are mostly about testing `StorageResponses` in different cases of reconnections.
Pr uses a mocked `RpcClient` implementation for providing expected responses. 

### Additional changes
- refactor `MockRpcClient` to be usable across legacy backend and unstable_backend following [suggestion from last PR](https://github.com/paritytech/subxt/pull/1751#discussion_r1743372027)
- add test for stale subscription ID causing error when trying to fetch storage to have a test that should fail if the [relevant issue is fixed](https://github.com/paritytech/subxt/issues/1567) 

Part of the work on #1677


subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
-            request: MockDataTable,
-            subscription: Subscription,
+            data_table: MockDataTable,
+            subscription_channel: Option<Subscription>,
+            subscription_handlers: HashMap<String, SubscriptionHandler>,
+            method_handlers: HashMap<String, MethodHandler>,
+        }
+
+        impl Data {
+            async fn call_meth<'a>(
```
niklasad1, line 488:
```suggestion
            async fn call<'a>(
```

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+                params: Option<Box<serde_json::value::RawValue>>,
+            ) -> super::rpc::RawRpcFuture<'a, Box<serde_json::value::RawValue>> {
+                let method = self.method_handlers.get(meth).unwrap_or_else(|| {
+                    panic!("no method named {} registered. Params: {:?}", meth, params)
+                });
+
+                (*method)(&mut self.data_table, &mut self.subscription_channel, params)
+            }
+
+            async fn call_subscription<'a>(
```
niklasad1, line 500:
```suggestion
            async fn subscribe<'a>(
```

subxt/src/backend/mod.rs
```diff
@@ -336,54 +336,96 @@ pub struct StorageResponse {
...
-        };
-        use subxt_core::{config::DefaultExtrinsicParams, Config};
-        use tokio::sync::{mpsc, Mutex};
+        use std::time::Duration;
+
+        use super::*;
 
-        type RpcResult<T> = Result<T, RpcError>;
-        type Item = RpcResult<String>;
+        pub type SubscriptionHandler = Box<
```
niklasad1, line 366: quite complicated type and a bit hard to read but I'm fine to remove the lifetimes and cloning the data for each call if it simplifies the code...

it just tests anyway

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+                    };
+                });
             }
         }
 
         struct Data {
-            request: MockDataTable,
-            subscription: Subscription,
+            data_table: MockDataTable,
+            subscription_channel: Option<Subscription>,
```
niklasad1, line 482: can you move the subscription_channel to the subscription handler? or is this needed to keep the channel alive?

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+        pub struct MockRpcBuilder {
+            data: Data,
+        }
+
+        impl MockRpcBuilder {
+            pub fn new() -> Self {
+                Self {
+                    data: Data {
+                        data_table: MockDataTable::new(),
+                        subscription_channel: None,
```
niklasad1, line 524: is this always `None`? can we remove it?

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,615 @@ mod test {
...
+                            }
+                            response.map(|x| RawValue::from_string(x).unwrap())
+                        })
+                    }),
+                )
+                .add_mock_data_from_iter(mock_subscription_data)
+                .add_mock_data_from_iter(response_data)
+                .build();
+
+            let (backend, mut driver): (UnstableBackend<Conf>, _) =
```
niklasad1, line 1033: move this unstable backend stuff to a helper function? It's used for several tests

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,615 @@ mod test {
...
+                .collect::<Vec<Result<StorageResponse, Error>>>()
+                .await;
+
+            assert!(matches!(
+                response.as_slice(),
+                [
+                    Ok(resp1 @ StorageResponse { .. }),
+                    Ok(resp2 @ StorageResponse { .. }),
+                    Err(Error::Other(s))
+                ] if s == "errro"
```
niklasad1, line 1290: typo?

```suggestion
                ] if s == "error"
```

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,615 @@ mod test {
...
+
+            let binding = response
+                .unwrap()
+                .collect::<Vec<Result<StorageResponse, Error>>>()
+                .await;
+            let response = binding.last().unwrap();
+
+            assert!(matches!(
+                response,
+                Err(Error::Other(reason)) if reason == "errro"
```
niklasad1, line 1469: typo?

```suggestion
                Err(Error::Other(reason)) if reason == "error"
```

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+                subscription_name: &str,
+                subscription_handler: SubscriptionHandler,
+            ) -> Self {
+                self.data
+                    .subscription_handlers
+                    .insert(subscription_name.into(), subscription_handler);
+                self
+            }
+
+            pub fn add_mock_data_from_iter<
```
niklasad1, line 547: since this is the only way to add `mock_data`, why not removing the `from_iter`? 

```suggestion
            pub fn add_mock_data<
```

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -839,6 +874,22 @@ pub(crate) mod hashmap_as_tuple_list {
         deserializer.deserialize_any(HashMapVisitor(PhantomData))
     }
 
+    /// Serialize a number as string
```
niklasad1, line 877: I wonder rather move these helpers to the test module

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -778,6 +799,9 @@ pub(crate) mod unsigned_number_as_string {
     use serde::de::{Deserializer, Visitor};
     use std::fmt;
 
+    #[cfg(test)]
+    use serde::ser::Serializer;
```
niklasad1, line 803: move to test module?!

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -786,6 +810,14 @@ pub(crate) mod unsigned_number_as_string {
         deserializer.deserialize_any(NumberVisitor(std::marker::PhantomData))
     }
 
+    /// Serialize a number as string
```
niklasad1, line 813: move to test module?!

niklasad1 (commented):
nice work, I have mostly have comments to simplify the tests to make it a little easier to read.

I'll take a closer look at the test themselves in the next iteration

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: renamed "errro" into "error"

subxt/src/backend/mod.rs
pkhry: renamed "errro" into "error"

subxt/src/backend/mod.rs
pkhry: renamed

subxt/src/backend/unstable/rpc_methods.rs
pkhry: done

subxt/src/backend/unstable/rpc_methods.rs
pkhry: done

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+                    data: Data {
+                        data_table: MockDataTable::new(),
+                        subscription_channel: None,
+                        subscription_handlers: HashMap::new(),
+                        method_handlers: HashMap::new(),
+                    },
+                }
+            }
+
+            pub fn add_method(mut self, method_name: &str, meth: MethodHandler) -> Self {
```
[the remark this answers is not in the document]
pkhry: renamed to "method_handler"

subxt/src/backend/mod.rs
pkhry: we set it to `Some` when we get a subscription going via "chainHead_v1_follow"

subxt/src/backend/mod.rs
```diff
@@ -392,121 +434,212 @@ mod test {
...
+                                let _ = sender.send(i).await;
+                            }
+                        }
+                        Message::Single(item) => sender.send(item).await.unwrap(),
+                    };
+                });
             }
         }
 
         struct Data {
```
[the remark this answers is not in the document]
pkhry: done

subxt/src/backend/mod.rs
pkhry: i dont think cloning would help as we need mutable refs to mutate the data inside

subxt/src/backend/mod.rs
pkhry: yup, sometimes we reference it inside request handlers

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -745,11 +769,16 @@ pub enum TransactionStatus<Hash> {
...
+#[cfg_attr(test, derive(Serialize))]
 pub struct TransactionBlockDetails<Hash> {
     /// The block hash.
     pub hash: Hash,
     /// The index of the transaction in the block.
-    #[serde(with = "unsigned_number_as_string")]
+    #[serde(deserialize_with = "unsigned_number_as_string::deserialize")]
+    #[cfg_attr(
+        test,
+        serde(serialize_with = "unsigned_number_as_string::for_test::serialize")
```
niklasad1, line 780: aight, I see makes sense to have these in this module then

Missed the serialize_with

subxt/src/backend/mod.rs
```diff
@@ -336,54 +336,96 @@ pub struct StorageResponse {
...
-        use serde_json::value::RawValue;
-        use std::{
-            collections::{HashMap, VecDeque},
-            sync::Arc,
-        };
-        use subxt_core::{config::DefaultExtrinsicParams, Config};
-        use tokio::sync::{mpsc, Mutex};
+        use std::time::Duration;
+
+        use super::*;
```
jsdw, line 365:
```suggestion

    pub mod rpc_client {
        use std::time::Duration;
        use super::*;

```

Nit: for consistency with above

subxt/src/backend/mod.rs
```diff
@@ -336,54 +336,96 @@ pub struct StorageResponse {
...
 
-        impl MockDataTable {
-            fn new() -> Self {
-                MockDataTable {
-                    items: HashMap::new(),
+        impl<T> Message<T> {
+            pub fn single(self) -> T {
+                match self {
+                    Self::Single(s) => s,
+                    _ => todo!(),
```
jsdw, line 395: Does something need doing here still or should it be a `panic!`?

subxt/src/backend/mod.rs
```diff
@@ -336,54 +336,96 @@ pub struct StorageResponse {
...
+                }
+            }
+            pub fn many(self) -> RpcResult<Vec<T>> {
+                match self {
+                    Self::Many(s) => s,
+                    _ => todo!(),
```
jsdw, line 401: Does something need doing here still or should it be a `panic!`?

subxt/src/backend/mod.rs
jsdw: (if so I'd call this `unwrap_single` for consistency with eg unwrap_err etc and same for below)

subxt/src/backend/mod.rs
```diff
@@ -336,54 +336,96 @@ pub struct StorageResponse {
...
+                        items
+                            .into_iter()
+                            .map(|item| item.map(|x| serde_json::to_string(&x).unwrap()))
+                            .collect()
+                    })),
+                    Message::Single(item) => {
+                        Message::Single(item.map(|x| serde_json::to_string(&x).unwrap()))
+                    }
+                };
                 match self.items.entry(key) {
```
jsdw, line 429: Nit: you can maybe simplify this a touch with something like:

```rust
self.items.entry(key).or_default().push_back(item)
```

or_default to put empty VecDeque in if empty, and then push the item to it in either case

pkhry: Done

subxt/src/backend/mod.rs
pkhry: Replaced with panic

subxt/src/backend/mod.rs
pkhry: Done

subxt/src/backend/mod.rs
pkhry: Done

subxt/src/backend/mod.rs
niklasad1: It's fine to keep this, I thought it was possible to just clone the `Arc<Mutex<_>>` but it's one layer above this...

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
 #[cfg(test)]
 mod test {
     use super::*;
+    pub use crate::{backend::StorageResponse, error::RpcError};
+    pub use futures::StreamExt;
+    pub use serde::Serialize;
+    pub use std::collections::{HashMap, VecDeque};
+    pub use subxt_core::{config::DefaultExtrinsicParams, Config};
+    pub use tokio::sync::{mpsc, Mutex};
+    pub type RpcResult<T> = Result<T, RpcError>;
```
niklasad1, line 345: I would prefer to separate `use` and `type defs` here separated by a new line

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
+    pub type Item = RpcResult<String>;
+    pub use crate::backend::rpc::RawRpcSubscription;
```
niklasad1, line 347: I would re-export RawRpcFuture as well here and remove all the prefix `super::rpc::*` below.. see my other comments below

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
+        use super::*;
+        use std::time::Duration;
+
+        pub type SubscriptionHandler = Box<
+            dyn for<'a> Fn(
+                    &'a mut MockDataTable,
+                    &'a mut Option<Subscription>,
+                    Option<Box<serde_json::value::RawValue>>,
+                )
+                    -> super::rpc::RawRpcFuture<'a, super::rpc::RawRpcSubscription>
```
niklasad1, line 370:
```suggestion
                    -> RawRpcFuture<'a, RawRpcSubscription>
```

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
+                + Send,
+        >;
+
+        pub type MethodHandler = Box<
+            dyn for<'a> Fn(
+                    &'a mut MockDataTable,
+                    &'a mut Option<Subscription>,
+                    Option<Box<serde_json::value::RawValue>>,
+                )
+                    -> super::rpc::RawRpcFuture<'a, Box<serde_json::value::RawValue>>
```
niklasad1, line 380:
```suggestion
                    -> RawRpcFuture<'a, Box<serde_json::value::RawValue>>
```

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
-        use tokio::sync::{mpsc, Mutex};
+        pub enum Message<T> {
+            Many(RpcResult<Vec<T>>),
+            Single(T),
+        }
 
-        type RpcResult<T> = Result<T, RpcError>;
-        type Item = RpcResult<String>;
+        impl<T> Message<T> {
+            pub fn unwrap_single(self) -> T {
```
niklasad1, line 390: nice

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
+                    panic!(
+                        "no subscription named {} registered. Params: {:?}",
+                        sub, params
+                    )
+                });
+
+                (*sub)(&mut self.data_table, &mut self.subscription_channel, params)
             }
         }
+        pub struct MockRpcBuilder {
```
niklasad1, line 511: can we `#[dervive(Default)]` here and remove the new impl below to save some lines of code?

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,308 @@ pub struct StorageResponse {
...
+    enum Conf {}
+    impl Config for Conf {
+        type Hash = crate::utils::H256;
+        type AccountId = crate::utils::AccountId32;
+        type Address = crate::utils::MultiAddress<Self::AccountId, ()>;
+        type Signature = crate::utils::MultiSignature;
+        type Hasher = crate::config::substrate::BlakeTwo256;
+        type Header = crate::config::substrate::SubstrateHeader<u32, Self::Hasher>;
+        type ExtrinsicParams = DefaultExtrinsicParams<Self>;
+
```
niklasad1, line 607:
```suggestion
```

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -745,11 +769,16 @@ pub enum TransactionStatus<Hash> {
 
 /// Details of a block that a transaction is seen in.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[cfg_attr(test, derive(Serialize))]
 pub struct TransactionBlockDetails<Hash> {
     /// The block hash.
     pub hash: Hash,
     /// The index of the transaction in the block.
-    #[serde(with = "unsigned_number_as_string")]
+    #[serde(deserialize_with = "unsigned_number_as_string::deserialize")]
```
niklasad1, line 777: ok, deserialize is only implemented for this type and `with` is a shortcut for both serialize and deserialize impl

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,599 @@ mod test {
...
+            // Test
+            // This request should encounter an error on `request` and do a retry.
+            let response_hash = backend.genesis_hash().await.unwrap();
+
+            assert_eq!(hash, response_hash)
+        }
+
+        #[tokio::test]
+        // Failure as we do not wait for subscription id to be updated
```
niklasad1, line 1353: perhaps add a link https://github.com/paritytech/subxt/issues/1567

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,599 @@ mod test {
...
+            })
+        }
+
+        #[tokio::test]
+        async fn storage_fetch_values_returns_stream_with_single_error() {
+            let response_data = vec![(
+                "method_response",
+                Message::Single(Ok(response_started("Id1"))),
+            )];
+            let mock_subscription_data = vec![(
```
niklasad1, line 1028:
```suggestion
            // Events generated for the chainHead_v1_storage 
            let follow_event_data = vec![(
```

subxt/src/backend/mod.rs
```diff
@@ -734,4 +872,599 @@ mod test {
...
+                ("continue_response", Message::Single(Ok(()))),
+                ("continue_response", Message::Single(Ok(()))),
+                (
+                    "continue_response",
+                    Message::Single(Err(RpcError::DisconnectedWillReconnect("Error".into()))),
+                ),
+                ("continue_response", Message::Single(Ok(()))),
+                ("continue_response", Message::Single(Ok(()))),
+            ];
+            let mock_data = vec![
```
niklasad1, line 1184: yeah, ok I see why you separate these now but to follow how they relate to chainHead api is a bit tricky to follow.

It's not really easy for me to follow how the responses is linked the "follow_event/subscription data`,
So I wonder whether we provide something like:

```rust
(
     method_response(id1), 
     Message::Many(vec![Ok(storage_items("Id1", &[storage_result("ID1", "Data1")])), Ok(operation_continue("Id1"))])
)

```

where the "first item" is the method_response i.e, operationStarted and the "second item" is the subscription data i.e, follow_events on provided on that method response.

That way it simpler to understand the responses for each call but it may not worked that well with your test structure

niklasad1 (approved):
Great,

Minor thingy, whether we can improve the readability for the tests by linking responses and subscription messages together but it's not a blocker for this PR IMO.

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: hmm, given we know the exact ordering of the requests/responses/subscription messages i can try to refactor it to be one single stack of responses that will be popped one by one. But maybe dong this in a followup PR would be more appropriate?

niklasad1: Yeah it's fine to keep it out of this PR, let's hear what James and Alex thinks about this as well :)

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+        use rpc_methods::{
+            Bytes, Initialized, MethodResponse, MethodResponseStarted, OperationError, OperationId,
+            OperationStorageItems, RuntimeSpec, RuntimeVersionEvent,
+        };
+        use sp_core::H256;
+
+        use super::unstable::*;
+        use super::*;
+
+        fn run_backend(
```
lexnv, line 876: nit: maybe `build_backend`?

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+                            "0xcbca25e39f142387": 2,
+                            "0x687ad44ad37f03c2": 1,
+                            "0xab3c0572291feb8b": 1,
+                            "0xbc9d89904f5b923f": 1,
+                            "0x37c8bb1350a9a2a8": 1
+                    }
+            });
+            serde_json::from_value(spec).unwrap()
+        }
+        fn init_hash() -> crate::utils::H256 {
```
lexnv, line 922: nit: Maybe `random_hash()`?

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+        async fn storage_fetch_values_returns_stream_with_single_error() {
+            let response_data = vec![(
+                "method_response",
+                Message::Single(Ok(response_started("Id1"))),
+            )];
+            let mock_subscription_data = vec![(
+                "chainHead_v1_storage",
+                Message::Many(Ok(vec![Ok(operation_error("Id1")), Ok(FollowEvent::Stop)])),
+            )];
+            let rpc_client = setup_mock_rpc_client(false)
```
lexnv, line 1017: Nice! The API looks quite elegant!

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+                .build();
+
+            let backend = run_backend_spawn_background(rpc_client);
+
+            // Test
+            // This request should encounter an error on `request` and do a retry.
+            let response = backend
+                .storage_fetch_values(
+                    ["ID1".into(), "ID2".into(), "ID3".into()].into(),
+                    crate::utils::H256::random(),
```
lexnv, line 1045: nit: use the util function instead or even better we can remove it and import `utils::H256` so we can say `h256::random()` directly?

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+                vec![
+                    storage_response("ID1", "Data1"),
+                    storage_response("ID2", "Data2"),
+                    storage_response("ID3", "Data3"),
+                ],
+                response
+            )
+        }
+        #[tokio::test]
+        /// Tests the error behaviour in streams based on chainHead_v1_continue calls
```
lexnv, line 1134: nit: new line

subxt/src/backend/mod.rs
```diff
@@ -734,4 +857,600 @@ mod test {
...
+
+            // We try again and should fail mid way
+            let response = backend
+                .storage_fetch_values(
+                    ["ID1".into(), "ID2".into(), "ID3".into()].into(),
+                    crate::utils::H256::random(),
+                )
+                .await
+                .unwrap();
+            // operation returned FollowEvent::OperationError
```
lexnv, line 1257: dq: Does subxt retry on OperationError?

subxt/src/backend/mod.rs
lexnv: Yep it would be a bit easier to follow the test with Niklas suggestion. If it turns out too complicated to change it we can do a followup for sure :D

subxt/src/backend/unstable/rpc_methods.rs
```diff
@@ -745,11 +769,16 @@ pub enum TransactionStatus<Hash> {
 
 /// Details of a block that a transaction is seen in.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[cfg_attr(test, derive(Serialize))]
```
lexnv, line 772: Should we just expose a `derive(Serialize)` instead of placing it under a cfg(test)? That might turn out to be useful for some developers, wdyt?

lexnv (approved):
LGTM! Tiny comments around test styles. Maybe we can expose the `derive(Serialize)` implementation for some structures and we can think about grouping test data, but those can be a followup 👍

subxt/src/backend/mod.rs
niklasad1: No, all errors are except DisconnectWillReconnect and Rejected are handed back to the user to deal with.

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,293 @@ pub struct StorageResponse {
...
 
-        impl MockDataTable {
-            fn new() -> Self {
-                MockDataTable {
-                    items: HashMap::new(),
+        impl<T> Message<T> {
+            pub fn unwrap_single(self) -> T {
+                match self {
+                    Self::Single(s) => s,
+                    _ => panic!(),
```
jsdw, line 392: Nit:

```suggestion
                    _ => panic!("cannot unwrap_single on Message::Many"),
```

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,293 @@ pub struct StorageResponse {
...
-            fn from_iter<'a, T: Serialize, I: IntoIterator<Item = (&'a str, RpcResult<T>)>>(
-                item: I,
-            ) -> Self {
-                let mut data = Self::new();
-                for (key, item) in item.into_iter() {
-                    data.push(key.into(), item);
+            pub fn unwrap_many(self) -> RpcResult<Vec<T>> {
+                match self {
+                    Self::Many(s) => s,
+                    _ => panic!(),
```
jsdw, line 398: Nit:

```suggestion
                    _ => panic!("cannot unwrap_many on Message::Single"),
```

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,293 @@ pub struct StorageResponse {
...
+                        method_handler, params
+                    )
+                });
+
+                (*method)(&mut self.data_table, &mut self.subscription_channel, params)
             }
 
-            async fn write(&self, items: RpcResult<Vec<Item>>) {
-                self.sender.send(items).await.unwrap()
+            async fn subscribe<'a>(
```
jsdw, line 491: Is there a reason for this and `call`, above, to be async? it means that you have `.await.await` below. Can you have `fn call` and `fn subscribe` instead?

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,293 @@ pub struct StorageResponse {
...
+                    "state_getStorage",
+                    Box::new(|data, _sub, params| {
+                        Box::pin(async move {
+                            let params = params.map(|p| p.get().to_string());
+                            let rpc_params = jsonrpsee::types::Params::new(params.as_deref());
+                            let key: sp_core::Bytes = rpc_params.sequence().next().unwrap();
+                            let value = data.pop(key.0).unwrap_single();
+                            value.map(|v| serde_json::value::RawValue::from_string(v).unwrap())
+                        })
+                    }),
```
jsdw, line 615: I wonder how easily we could remove the Box's here to make the API cleaner.

Eg if `add_method` and `add_subscription` took in some generic like:

```rust
fn add_subscription(sub: &str, f: F) 
where
    Fn: Send + Fn(
        &'a mut MockDataTable,
        &'a mut Option<Subscription>,
        Option<Box<serde_json::value::RawValue>>,
    ) -> RawRpcFuture<'a, Box<serde_json::value::RawValue>>,
```

Then you could do the boxing internally and keep the API cleaner (maybe you also return some future that is then boxed or whatever instead of returning `RawRpcFuture<'a, Box<serde_json::value::RawValue>>` so that both boxes are removed?)

The downside is you probably end up having to copy these bounds in both the `Box` and in the `where` though could maybe avoid duplication with a tiny macro or trait like `trait SubscriptionFn: Fn(...)`.

Not a big deal really because this is just test code but might be a nice little ergonomic win :)

subxt/src/backend/mod.rs
niklasad1: I think the reason is because the mock data is behind an async mutex, that shouldn't be that hard to get rid of I guess because we don't hold it over other await points...

subxt/src/backend/mod.rs
pkhry: removed boxing form the callsite

subxt/src/backend/mod.rs
pkhry: removed `async`

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
```diff
@@ -336,177 +336,293 @@ pub struct StorageResponse {
...
+                params: Option<Box<serde_json::value::RawValue>>,
+            ) -> RawRpcFuture<'a, RawRpcSubscription> {
+                let sub = self.subscription_handlers.get(sub).unwrap_or_else(|| {
+                    panic!(
+                        "no subscription named {} registered. Params: {:?}",
+                        sub, params
+                    )
+                });
+
+                (*sub)(&mut self.data_table, &mut self.subscription_channel, params)
```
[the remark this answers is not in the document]
pkhry: Ok, so a few methods/subscriptions do in fact require access to the same data point, so capturing stuff in a closure ends up more cumbersome that the current approach in some cases

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: done

subxt/src/backend/mod.rs
pkhry: Ok, so after a bit of experimenting it looks like it's not possible without adding an external dependency or adding mock data item by item to the mock data set. using `dyn Serialize` is not possible(not object safe) and adding another dependency just to make tests seem a bit more pretty feels like an overkill.

PR #1766, opened 2024-09-11
Add eth_address

pgherveou:


signer/src/ecdsa.rs
```diff
@@ -150,6 +151,13 @@ impl Keypair {
         Self::from_secret_key(acc)
     }
 
+    /// Get the Ethereum address for this key pair.
+    /// See https://ethereum.org/en/developers/docs/accounts/#account-creation
+    pub fn eth_address(&self) -> H160 {
+        let pk = self.0.public_key().serialize_uncompressed();
+        H160::from_slice(&keccak_256(&pk[1..])[12..])
+    }
```
jsdw, line 159: Wouldn't you just use the `signer::eth` module in this crate if you want to work with eth addresses? I'd have to double check but think that will display them appropriately :)

pgherveou: ah did not realize there was another way to get the H160 ethereum address

jsdw: IIRC we added `eth` as an easy to use wrapper around the ecdsa stuff just to make working with eth addresses nice and easy!

PR #1768, opened 2024-09-11
fix flaky test with temporary solution

### LINKED ISSUE #1752: Subxt integration tests failed against latest Substrate build.

pkhry:
### Description
Temporary plug for flaky test in CI. 

When we encounter `e.disconnected_will_reconnect()` error in the test the blocks for new subscribers are getting pruned so we can't really test the fact that there's an intersection there. 
So for now this part is conditional.

Close #1752


jsdw (approved):
Makes sense to me at the mo!

lexnv (approved):
Looks clean! 🙏

PR #1785, opened 2024-09-26
Use frame-decode for core extrinsic decode logic

jsdw:
`frame-decode` is a crate to handle decoding historic or modern extrinsics and storage keys/values. (https://docs.rs/frame-decode/latest/frame_decode/)

- Here, we integrate it into Subxt for extrinsic decoding so that the decode logic lives in one place.
- A possible next step would be to see if it's worth integrating storage key decoding too!


core/src/blocks/extrinsics.rs
```diff
@@ -2,39 +2,60 @@
...
     _marker: core::marker::PhantomData<T>,
 }
 
 impl<T: Config> Extrinsics<T> {
     /// Instantiate a new [`Extrinsics`] object, given a vector containing
     /// each extrinsic hash (in the form of bytes) and some metadata that
     /// we'll use to decode them.
-    pub fn decode_from(extrinsics: Vec<Vec<u8>>, metadata: Metadata) -> Result<Self, BlockError> {
-        let ids = ExtrinsicPartTypeIds::new(&metadata)?;
+    pub fn decode_from(extrinsics: Vec<Vec<u8>>, metadata: Metadata) -> Result<Self, Error> {
```
jsdw, line 32: I moved the decoding work to this initial step.

This means that on the downside we:
- try to decode everything straight away, even if the user doesn't care (though I suspect they usually would)
- an error decoding an extrinsic will lead to a failure early on, preventing eg getting the extrinsic bytes

On the plus side, we:
- Can iterate over or `find` or whatever multiple times without re-doing the decoding work each time we do these things.

metadata/src/lib.rs
```diff
@@ -61,6 +65,63 @@ pub struct Metadata {
     custom: frame_metadata::v15::CustomMetadata<PortableForm>,
 }
 
+// Since we've abstracted away from frame-metadatas, we impl this on our custom Metadata
+// so that it can be used by `frame-decode` to obtain the relevant extrinsic info.
+impl frame_decode::extrinsics::ExtrinsicTypeInfo for Metadata {
```
jsdw, line 70: We wouldn't normally need this if we used frame_metadata's `RuntimeMetadataV15` or whatever directly, but since we abstracted it into our own Metadata crate so as to not care about the underlying version, we need to be able to get the info we need from this via the impl here.

lexnv (approved):
LGTM! 👍

pkhry (approved):
LGTM

Polkadot-Forum:

https://forum.polkadot.network/t/decoding-old-and-new-blocks-and-storage-entries-in-rust-october-2024-update/10276/1

PR #1786, opened 2024-09-26
deps: use polkadot-sdk umbrella crate

niklasad1:


core/Cargo.toml
```diff
@@ -31,8 +31,9 @@ std = [
     "tracing/std",
     "impl-serde/std",
     "primitive-types/std",
+    "polkadot-sdk/std",
 ]
-substrate-compat = ["sp-core", "sp-runtime"]
+substrate-compat = ["polkadot-sdk/sp-core", "polkadot-sdk/sp-runtime", "polkadot-sdk/std"]
```
jsdw, line 36: So sp-core wasn't std before and sp-runtime was I think.. I guess std is needed here for it to work?

jsdw (approved):
The only thing that's a little scary is that if we need to have std enabled for some but not all of the sp crates somewhere then I guess we'll run into issues having this umbrella crate, but on the other hand it makes it much easier to keep these versions in sync and uptodate so I think it's a net win!

core/Cargo.toml
jsdw: I wonder how many more deps it ends up bringing in; maybe worth checking?

niklasad1: It should only affect `--feature --substrate-compat` and not the std or no-std which I'm okay with...

Sure, lemme check

niklasad1:
> The only thing that's a little scary is that if we need to have std enabled for some but not all of the sp crates somewhere then

Yupp, good point I didn't think about it before :)

core/Cargo.toml
niklasad1: this branch brings 351 depends vs 350 on master, I think it's fine :)

metadata/Cargo.toml
```diff
@@ -21,7 +21,7 @@ std = ["scale-info/std", "frame-metadata/std"]
 scale-info = { workspace = true, default-features = false }
 frame-metadata = { workspace = true, default-features = false, features = ["current", "decode"] }
 codec = { package = "parity-scale-codec", workspace = true, default-features = false, features = ["derive"] }
-sp-crypto-hashing = { workspace = true }
+polkadot-sdk = { workspace = true, default-features = false, features = ["sp-crypto-hashing"] }
```
niklasad1, line 24:
```suggestion
polkadot-sdk = { workspace = true, features = ["sp-crypto-hashing"] }
```

core/Cargo.toml
jsdw: Yeah that's all good then!

Cargo.toml
```diff
@@ -138,13 +136,7 @@ instant = { version = "0.1.13", default-features = false }
...
 
 # Substrate crates:
-sp-core = { version = "34.0.0", default-features = false }
-sp-crypto-hashing = { version = "0.1.0", default-features = false }
-sp-runtime = "39.0.1"
-sp-keyring = "39.0.0"
-sp-maybe-compressed-blob = "11.0.0"
-sp-state-machine = "0.43.0"
-sp-io = "38.0.0"
+polkadot-sdk = { version = "0.5", default-features = false }
```
niklasad1, line 139:
```suggestion
polkadot-sdk = { version = "0.7", default-features = false }
```

lexnv (approved):

PR #1789, opened 2024-09-27
reconn-rpc-client: parse URL before connecting

niklasad1:
It was hard to figure whether one simply entered a faulty URL and it tried to reconnect according to the retry policy.

With this change we first parse url and then try to reconnect. This will detect invalid URLs etc directly instead of waiting for the retry to complete.


jsdw (approved):

lexnv (approved):
Nice!

PR #1802, opened 2024-10-01
rpc: stabilize chainhead backend

### LINKED ISSUE #1753: Stabilise the `UnstableBackend`
Nowadays, Substrate nodes support the `chainHead_v1` RPCs which Subxt's `UnstableBackend` relies on. Smoldot would need upgrading to 0.18 to support them, which has its own issues (eg #1749), but the lightclient feature is still unstable.

To stabilise the `UnstableBackend` we should:

1. Rename it to something like `ChainHeadBackend`
2. Add something like `ChainHeadBackendBuilder::build_with_background_driver()` which uses `tokio` or the WASM equivalent to spawn the background task which drives the driver.
3. Use this method instead of building a `LegacyBackend` in `OnlineClient::from_insecure_url` etc.
4. Think about how errors are propagated to the user if this background task is hidden away. Currently, errors are emitted there only. We may need to tweak the FollowStreamDriver etc to pass errors on to the current subscriptions too, so they make it out to users where appropriate.
5. Update Smoldot to 0.18 (if not already done) so that the light client feature will work with a default Subxt client, now that the default relies on chainHead_v1 APIs which we need 0.18 for.

All of the above should work both in WASM and native builds by default. This being true, we wouldn't need an additional feature flag or whatever I believe.

niklasad1:
Close https://github.com/paritytech/subxt/issues/1753


subxt/src/backend/chain_head/follow_stream_driver.rs
```diff
@@ -432,6 +432,10 @@ where
                     self.sub_id = Some(sub_id);
                     continue;
                 }
+                FollowStreamMsg::Event(FollowEvent::Stop) => {
```
niklasad1, line 435: This was missing to close down subscriptions when the backend was shutdown...

probably my own fault

niklasad1: aight, stopped was ignored because of the reconnecting stuff..

hmm, hard to decide that from the stream i.e, whether to close it on `Stop` or not...

subxt/src/backend/chain_head/mod.rs
```diff
@@ -92,26 +97,59 @@ impl<T: Config> UnstableBackendBuilder<T> {
...
         };
 
         (backend, driver)
     }
+
+    /// An API to build the backend and driver which will run in the background until completion
+    /// on the default runtime.
+    ///
+    /// On native this is spawned on the tokio runtime.
+    /// On web this is spawned on the wasm_bindgen.
```
jsdw, line 115:
```suggestion
    /// On native this task is spawned on the tokio runtime.
    /// On web this task is spawned via wasm_bindgen.
```

jsdw (approved):
Nice!

From our chat in the rust call I think we agree that it's worth adding eg a "runtime" feature flag which on native enables the relevant tokio thing and on web enables the web thing, but we could leave this for a separate PR and I don't think it's super urgent!

I guess raising an issue to look at the subscriptions halting is the other thing to do, but again not related to this PR :)

subxt/Cargo.toml
```diff
@@ -47,12 +47,21 @@ web = [
     "finito?/wasm-bindgen",
 ]
 
+# Feature flags for enabling default future executor runtimes.
```
niklasad1, line 50:
```suggestion
# Feature flag to enable the default future executor.
```

lexnv (approved):

subxt/src/backend/chain_head/mod.rs
```diff
@@ -111,9 +111,10 @@ impl<T: Config> ChainHeadBackendBuilder<T> {
     /// An API to build the backend and driver which will run in the background until completion
     /// on the default runtime.
     ///
-    /// - On non-wasm targets, this will spawn a tokio task to poll the driver.
-    /// - On wasm targets, this will spawn a wasm-bindgen task to poll the driver.
+    /// - On non-wasm targets, this will spawn a tokio task to run driver in the background until completion.
```
niklasad1, line 114:
```suggestion
    /// - On non-wasm targets, this will spawn the driver on tokio.
```

subxt/src/backend/chain_head/mod.rs
```diff
@@ -92,26 +97,61 @@ impl<T: Config> UnstableBackendBuilder<T> {
...
         };
 
         (backend, driver)
     }
+
+    /// An API to build the backend and driver which will run in the background until completion
+    /// on the default runtime.
+    ///
+    /// - On non-wasm targets, this will spawn the driver on tokio.
+    /// - On wasm targets, this will spawn a wasm-bindgen task to run driver in the background until completion.
```
niklasad1, line 115:
```suggestion
    /// - On wasm targets, this will spawn the driver on wasm-bindgen-futures.
```

subxt/src/backend/chain_head/mod.rs
```diff
@@ -92,26 +97,61 @@ impl<T: Config> UnstableBackendBuilder<T> {
...
             driver: follow_stream_driver,
         };
 
         (backend, driver)
     }
+
+    /// An API to build the backend and driver which will run in the background until completion
+    /// on the default runtime.
+    ///
+    /// - On non-wasm targets, this will spawn the driver on tokio.
```
niklasad1, line 114:
```suggestion
    /// - On non-wasm targets, this will spawn the driver on `tokio`.
```

PR #1817, opened 2024-10-08
chainhead backend: notify subscribers when the backend is closed

### LINKED ISSUE #1808: `FollowStreamFinalizedHeads` doesn't work properly for non-reconnecting backend
[FollowStreamFinalizedHeads](https://github.com/paritytech/subxt/blob/master/subxt/src/backend/chain_head/follow_stream_driver.rs#L383-#L392) is used both for reconnecting and non-reconnecting rpc clients currently which doesn't work properly for `non-reconnecting` because this stream assumes that after the `stop event` it will continue/be restarted.

This is not true for `non reconnecting backend` when the stop event is fired the backend is closed.  Thus, if one subscribes to finalized blocks and the backend is stopped then the stream will never return and wait forever.

So in order fix this we would need to inject a flag [here](https://github.com/paritytech/subxt/blob/master/subxt/src/backend/unstable/mod.rs#L155) whether the backend is reconnecting or not, or alternatively make `FollowStreamDriverSubscription` aware of this in some way.

####  Open terminal 1
```bash
substrate-node --dev
# Wait for the node to start
```

#### Open terminal 2
Run the following code

```rust
#[tokio::main]

    let rpc = RpcClient::from_url("ws://localhost:9944".to_string()).await?;
    let backend: ChainHeadBackend<PolkadotConfig> =
        ChainHeadBackendBuilder::default().build_with_background_driver(rpc.clone());
    let api = OnlineClient::from_backend(std::sync::Arc::new(backend)).await?;

    let mut blocks_sub = api.blocks().subscribe_finalized().await?.take(100);

    while let Some(block) = blocks_sub.next().await {
        let block = block?;

        let block_number = block.number();
        let block_hash = block.hash();

        println!("Block #{block_number} ({block_hash})");
    }

    Ok(())
}
```

#### Open terminal 3
pkill substrate-node

At this point (when substrate-node is stopped) then the finalized head stream should be terminated but it isn't....

niklasad1:
Close #1808 

This PR fixes that once [Shared::done](https://github.com/paritytech/subxt/blob/master/subxt/src/backend/chain_head/follow_stream_driver.rs#L196-#L199) then we also wake up the "waker" such that the FollowStreamSubscription poll impl gets notified about.

This issue was really that the FollowStreamSubscription wasn't woken up after that `Shared.done == true` was set.

Currently the backend only returns error once before it's shutdown and then return None but this forces users of the driver to run it completion and one doesn't run it to completion i.e, such as bailing on error then [Shared::done == true](https://github.com/paritytech/subxt/blob/master/subxt/src/backend/chain_head/follow_stream_driver.rs#L50-#L54) is never set.

Thus, I also had to change that we run the driver until completion.

Since we already now that backend is shutdown when it returns error we could already set `done == true`  [here](https://github.com/paritytech/subxt/blob/master/subxt/src/backend/chain_head/follow_stream_driver.rs#L55) but it's doesn't feel correct to me but would probably help users of the low-level API not run into such tricky issues ^^, thoughts?

### DIFF
--- a/subxt/src/backend/chain_head/follow_stream.rs
+++ b/subxt/src/backend/chain_head/follow_stream.rs
@@ -71,7 +71,9 @@ enum InnerStreamState<Hash> {
     /// We are polling for, and receiving events from the stream.
     ReceivingEvents(FollowEventStream<Hash>),
     /// We received a stop event. We'll send one on and restart the stream.
-    Stopped,
+    Stopped { restart: bool },
+    /// Report error.
+    Error(Option<Error>),
     /// The stream is finished and will not restart (likely due to an error).
     Finished,
 }
@@ -83,8 +85,9 @@ impl<Hash> std::fmt::Debug for InnerStreamState<Hash> {
             Self::Initializing(_) => write!(f, "Initializing(..)"),
             Self::Ready(_) => write!(f, "Ready(..)"),
             Self::ReceivingEvents(_) => write!(f, "ReceivingEvents(..)"),
-            Self::Stopped => write!(f, "Stopped"),
+            Self::Stopped { .. } => write!(f, "Stopped"),
             Self::Finished => write!(f, "Finished"),
+            Self::Error(_) => write!(f, "Error"),
         }
     }
 }
@@ -150,13 +153,20 @@ impl<Hash> Stream for FollowStream<Hash> {
                         Poll::Ready(Err(e)) => {
                             // Re-start if a reconnecting backend was enabled.
                             if e.is_disconnected_will_reconnect() {
-                                this.stream = InnerStreamState::Stopped;
+                                this.stream = InnerStreamState::Stopped { restart: true };
                                 continue;
                             }
 
-                            // Finish forever if there's an error, passing it on.
-                            this.stream = InnerStreamState::Finished;
-                            return Poll::Ready(Some(Err(e)));
+                            // Finish forever if there's an error which is done as follows:
+                            //
+                            // 1) Send the FollowEvent::Stop message which is propagated to the subscriber
+                            // 2) Send the error on the stream
+                            // 3) Finish the stream
+                            //
+                            this.stream = InnerStreamState::Error(Some(e));
+                            return Poll::Ready(Some(Ok(FollowStreamMsg::Event(
+                                FollowEvent::Stop { restart: false },
+                            ))));
                         }
                     }
                 }
@@ -175,14 +185,14 @@ impl<Hash> Stream for FollowStream<Hash> {
                         Poll::Ready(None) => {
                             // No error happened but the stream ended; restart and
                             // pass on a Stop message anyway.
-                            this.stream = InnerStreamState::Stopped;
+                            this.stream = InnerStreamState::Stopped { restart: true };
                             continue;
                         }
                         Poll::Ready(Some(Ok(ev))) => {
-                            if let FollowEvent::Stop = ev {
+                            if let FollowEvent::Stop { restart } = ev {
                                 // A stop event means the stream has ended, so start
                                 // over after passing on the stop message.
-                                this.stream = InnerStreamState::Stopped;
+                                this.stream = InnerStreamState::Stopped { restart };
                                 continue;
                             }
                             return Poll::Ready(Some(Ok(FollowStreamMsg::Event(ev))));
@@ -190,7 +200,7 @@ impl<Hash> Stream for FollowStream<Hash> {
                         Poll::Ready(Some(Err(e))) => {
                             // Re-start if a reconnecting backend was enabled.
                             if e.is_disconnected_will_reconnect() {
-                                this.stream = InnerStreamState::Stopped;
+                                this.stream = InnerStreamState::Stopped { restart: true };
                                 continue;
                             }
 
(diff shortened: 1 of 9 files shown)


jsdw:
Is the idea here that an `RpcClient` implementation would need to emit `FollowEvent::BackendClosed` (of the JSON of that at least) in order to tell subscriptions that they can actually completely stop? Whereas if the subscriptions just get a `FollowEvent::Stop` event then things might continue? (I guess the actual RpcClient impls would need tweaking too to make this PR useful, or maybe I'm missing something?)

subxt/src/backend/chain_head/follow_stream.rs
```diff
@@ -154,9 +160,16 @@ impl<Hash> Stream for FollowStream<Hash> {
...
+                            // Finish forever if there's an error which is done as follows:
+                            //
+                            // 1) Send the FollowEvent::BackendClosed message which tells to subscriber that subscription is closed.
+                            // 2) Send the error on the backend driver stream.
+                            // 3) Finish the stream.
+                            //
+                            this.stream = InnerStreamState::Error(Some(e));
+                            return Poll::Ready(Some(Ok(FollowStreamMsg::Event(
+                                FollowEvent::BackendClosed,
+                            ))));
```
jsdw, line 172: Ah, this is where we turn any error that's not a `DisconnectedWillReconnect` into something which closes the subscription, so RpcClients don't need to know anything special!

subxt/src/backend/chain_head/follow_stream.rs
```diff
@@ -72,6 +72,10 @@ enum InnerStreamState<Hash> {
     ReceivingEvents(FollowEventStream<Hash>),
     /// We received a stop event. We'll send one on and restart the stream.
     Stopped,
+    /// The backend was closed and tell everything to shutdown.
+    BackendClosed,
+    /// Report error.
+    Error(Option<Error>),
```
jsdw, line 78: Offhand I can't see why the Error is in an `Option` here? Can't see anywhere it's set to `None` :D

jsdw:
Looking at the PR, I wonder about a couple of things:

1. Would it make more sense to indeed forward the `Error`, since you can call `is_disconnected_will_reconnect()` on it to see if you need to handle that. In this PR we create a new state from the error that's a bit redundant perhaps because we already have the error? it feels most future proof to forward the error too, since then any other specific errors can be handled easily in subscriptions or whatever in the future.
2. If that doesn't work out, I also wonder whether to avoid adding `FollowEvent::BackendClosed` and instead have a wrapper enum like `enum FollowEventOrClosed { Event(FollowEvent), Closed }`. This means we don't have to handle for instance the possibility of the RpcClient returning an event like "backendClosed" (which should never happen), and then perhaps the `InnerStreamState` would need only one variant, `BackendClosed(Error)`,

I'd be up for seeing whether (1) was possible anyways or whether it adds a bunch of complexity or something that I'm not thinking about!

subxt/src/backend/chain_head/follow_stream.rs
niklasad1: It's just hack around that poll takes &mut and not possible to clone the error.

Thus, I had to use Option::take to take the error and then change the state.

jsdw: Aah, yeah ok gotcha!

Cloning the error is important if we wanted to take the alternate approach of forwarding the error to any subscribers, so hmm I can see why that could be a pain!

jsdw:
Yep; I like your idea of the `is_closed` flag as the quick fix, and then maybe the next step is to Arc the error so we can share it with the subscribers, which is especially useful if the driver task is hidden away behind a helper function :)

subxt/src/backend/chain_head/follow_stream_driver.rs
```diff
@@ -196,6 +196,13 @@ impl<Hash: BlockHash> Shared<Hash> {
     pub fn done(&self) {
         let mut shared = self.0.lock().unwrap();
         shared.done = true;
+
+        // Wake up all subscribers so they can see that we're done.
```
niklasad1, line 200:
```suggestion
        // Wake up all subscribers so they get notified that backend was closed
```

subxt/src/backend/chain_head/mod.rs
```diff
@@ -128,16 +128,16 @@ impl<T: Config> ChainHeadBackendBuilder<T> {
...
+            // NOTE: we need to poll the driver until it's done i.e returns None
+            // to ensure that the backend is shutdown properly.
             while let Some(res) = driver.next().await {
-                if let Err(e) = res {
-                    if !e.is_disconnected_will_reconnect() {
-                        tracing::debug!(target: "subxt", "chainHead driver was closed: {e}");
-                        break;
-                    }
+                if let Err(err) = res {
+                    tracing::debug!(target: "subxt", "ChainHeadBackendDriver got error={err}");
```
niklasad1, line 136:
```suggestion
                    tracing::debug!(target: "subxt", "chainHead backend error={err}");
```

subxt/src/backend/chain_head/mod.rs
```diff
@@ -128,16 +128,16 @@ impl<T: Config> ChainHeadBackendBuilder<T> {
...
-                    if !e.is_disconnected_will_reconnect() {
-                        tracing::debug!(target: "subxt", "chainHead driver was closed: {e}");
-                        break;
-                    }
+                if let Err(err) = res {
+                    tracing::debug!(target: "subxt", "chainHead backend error={err}");
                 }
             }
+
+            tracing::debug!(target: "subxt", "ChainHeadBackendDriver closed");
```
niklasad1, line 140:
```suggestion
            tracing::debug!(target: "subxt", "chainHead backend was closed");
```

jsdw (approved):
Nice!

subxt/src/backend/chain_head/follow_stream_driver.rs
```diff
@@ -196,6 +196,13 @@ impl<Hash: BlockHash> Shared<Hash> {
     pub fn done(&self) {
         let mut shared = self.0.lock().unwrap();
         shared.done = true;
+
+        // Wake up all subscribers so they get notified that backend was closed
```
niklasad1, line 200:
```suggestion
        // Wake up all subscribers so they get notified that the backend was closed
```

pkhry (approved):

PR #1822, opened 2024-10-11
fetch-metadata: remove tokio

niklasad1:
I don't think it's worth to use tokio in fetch-metadata for the blocking API and I removed in favor if futures::executor::block_on

### DIFF
--- a/macro/Cargo.toml
+++ b/macro/Cargo.toml
@@ -29,6 +29,7 @@ quote = { workspace = true }
 subxt-codegen = { workspace = true, features = ["fetch-metadata"] }
 scale-typegen = { workspace = true }
 polkadot-sdk = { workspace = true, optional = true, features = ["sp-io", "sc-executor-common", "sp-state-machine", "sp-maybe-compressed-blob", "sc-executor"] }
+futures = { workspace = true, features = ["executor"] }
 
 [lints]
 workspace = true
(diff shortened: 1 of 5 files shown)


lexnv (approved):

macro/Cargo.toml
```diff
@@ -29,6 +29,7 @@ quote = { workspace = true }
 subxt-codegen = { workspace = true, features = ["fetch-metadata"] }
 scale-typegen = { workspace = true }
 polkadot-sdk = { workspace = true, optional = true, features = ["sp-io", "sc-executor-common", "sp-state-machine", "sp-maybe-compressed-blob", "sc-executor"] }
+futures = { workspace = true, features = ["executor"] }
```
jsdw, line 32: I guess one reason for `tokio` was just that it was already likely to be included as a dep. I wonder if this adds any new deps to the tree or whether it was also in use somewhere anyways?

jsdw (approved):

macro/Cargo.toml
niklasad1: I don't think so because futures/executor is already enabled by default which compiles for wasm as wel

niklasad1, line 32:
```suggestion
futures = { workspace = true }
```

niklasad1: https://github.com/rust-lang/futures-rs/blob/0.3/futures/Cargo.toml#L35

macro/Cargo.toml
```diff
@@ -29,6 +29,7 @@ quote = { workspace = true }
 subxt-codegen = { workspace = true, features = ["fetch-metadata"] }
 scale-typegen = { workspace = true }
 polkadot-sdk = { workspace = true, optional = true, features = ["sp-io", "sc-executor-common", "sp-state-machine", "sp-maybe-compressed-blob", "sc-executor"] }
+futures = { workspace = true }
```
niklasad1, line 32:
```suggestion
futures = { workspace = true, features = ["executor"] }
```

niklasad1:
after thinking about this, if one doesn't use tokio then creating jsonrpsee::ws_client this will panic so I think it safer to keep tokio anyway

PR #1828, opened 2024-10-14
fix: `defalt-feature` -> `default-features` in Cargo.toml

niklasad1:
This is only used by docs.rs when publishing the crates which is why we are not notified about it.


jsdw (approved):
whoops :D

PR #1829, opened 2024-10-14
move `fetch metadata` to a separate crate `subxt_utils_fetchmetadata`

### LINKED ISSUE #1812: subxt: brings in `tls` dependencies when `default-features = false && features = ["native"]`
Hey hey,

I'm using the following config for subxt:

```toml
[package]
name = "..."
version = "0.1.0"
edition = "2021"

[dependencies]
subxt = { git = "https://github.com/paritytech/subxt", default-features = false, features = ["native"], branch = "master" }
tokio = { version = "1.40.0", features = ["rt-multi-thread", "macros", "time"] }
futures = "0.3"
serde_json = "1"
```

This config ends up leads to a bunch of TLS dependencies which I'm not expecting to see:

```
   Compiling rustls-pki-types v1.9.0
   Compiling rustls v0.23.14
   Compiling rustls-pemfile v2.2.0
   Compiling rustls-native-certs v0.7.3
   Compiling rustls-webpki v0.102.8
   Compiling tokio-rustls v0.26.0
   Compiling rustls-platform-verifier v0.3.4
   Compiling hyper-rustls v0.27.3
    Checking subxt-riscv-example v0.1.0 (/home/niklasad1/Github/temp/subxt-jsonrpsee-no-tls)
```

`subxt-macro` leaks jsonrpsee because of [this](https://github.com/paritytech/subxt/blob/master/macro/Cargo.toml#L29) and ideally `subxt-codegen` dependency should just be dev-dependency i.e, fetch the metadata at compile time and it shouldn't be needed as a dependency but maybe some reason that I don't follow :)

If that doesn't work we need to add way to feature-gate `fetch_metadata_from_url` from the macro.

//cc @jsdw @lexnv

niklasad1:
Close #1812 

This PR moves the `fetch metadata` to a separate crate and also makes it optional for the `subxt macro/codegen` use it behind a feature flag.

So, if one disables jsonrpsee in subxt then the following would end-up in a compile error:

```rust
#[subxt(runtime_metadata_insecure_url = "ws://localhost:9944")]
mod runtime {}

fn main() {}
```

```bash
error: 'runtime_metadata_insecure_url' requires the 'runtime-metadata-insecure-url' feature to be enabled
  --> src/main.rs:75:1
   |
75 | #[subxt::subxt(runtime_metadata_insecure_url = "ws://localhost:9944")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the attribute macro `subxt::subxt` (in Nightly builds, run with -Z macro-backtrace for more info)
```


codegen/src/error.rs
```diff
@@ -85,6 +85,7 @@ impl CodegenError {
 /// Error attempting to load metadata.
 #[derive(Debug, thiserror::Error)]
 #[non_exhaustive]
+
```
niklasad1, line 88:
```suggestion
```

macro/src/lib.rs
```diff
@@ -246,7 +243,12 @@ fn fetch_metadata(args: &RuntimeMetadataArgs) -> Result<subxt_codegen::Metadata,
                 .and_then(|b| subxt_codegen::Metadata::decode(&mut &*b).map_err(Into::into))
                 .map_err(|e| CodegenError::from(e).into_compile_error())?
         }
+        #[cfg(feature = "jsonrpsee")]
```
niklasad1, line 246: fetch metadata via `runtime_metadata_insecure_url` is only possible with the `feature = "jsonrpsee` now to enable opting out from jsonrpsee/tokio

codegen/Cargo.toml
```diff
@@ -13,8 +13,9 @@ description = "Generate an API for interacting with a substrate node from FRAME
 
 [features]
 default = []
-fetch-metadata = ["dep:jsonrpsee", "dep:tokio", "dep:frame-metadata"]
+fetch-metadata = []
```
jsdw, line 16: DQ: Looking at this now, what is the reason to add a new `jsonrpsee` feature flag here, when `fetch-metadata` always requires `jsonrpsee` anyway I think?

codegen/Cargo.toml
```diff
@@ -25,11 +26,12 @@ quote = { workspace = true }
 syn = { workspace = true }
 scale-info = { workspace = true }
 subxt-metadata = { workspace = true }
-jsonrpsee = { workspace = true, features = ["async-client", "client-ws-transport-tls", "http-client"], optional = true }
+jsonrpsee = { workspace = true, features = ["ws-client", "http-client"], optional = true }
```
jsdw, line 29: I guess this means that, without importing `jsonrpsee` myself to enable TLS features, that we won't be able to use fetch-metadata with httpS urls?

codegen/Cargo.toml
niklasad1: It's possible to fetch-metadata from a file without jsonrpsee that's why I kept it...

macro/Cargo.toml
```diff
@@ -16,6 +16,7 @@ description = "Generate types and helpers for interacting with Substrate runtime
 [features]
 web = ["subxt-codegen/web"]
 runtime-path = ["polkadot-sdk"]
+jsonrpsee = ["subxt-codegen/jsonrpsee"]
```
jsdw, line 19: I'd change this to `fetch-metadata = ["subxt-codegen/fetch-metadata"]` and ditch the `jsonrpsee` flag I reckon.

Then, whether you can fetch metadata from a URL or not depends on this feature being enabled, rather than us always bringing in jsonrpsee (at compile time) to allow that in the macro.

I think this is all that actually needs doing in this PR to remove jsonrpsee; no need for the etra `jsonrpsee` feature flag etc I think?

jsdw: (should also be noted that jsonrpsee only comes in at compile time for the macro to run, so I don't think it's awfully important here. The thing I would want to avoid is jsonrpsee being bought in as a runtime dependency evcen when we don't enable the `jsonrpsee` feature in subxt)

codegen/Cargo.toml
niklasad1: yes without jsonrpsee it will only work to fetch metadata from a file or runtime path...

macro/Cargo.toml
niklasad1: I think I rather breakout the fetch-metadata to a separate crate to simplify this

jsdw: Yeah we could do; it just always felt like overkill having a separate crate for this one thing!

I think at the mo I prefer just making `fetch-metadata` an optional feature on the macro crate and then that's all we really need to do here I think?!

jsdw: That said, we could consider adding a `utils` folder which contains some of the little crates we might end up wanting, and then one of those crates is `fetch-metadata` which contains the code to fetch metadata from URLs and is only imported in the macro if the `fetch-metadata` feature is enabled?

codegen/Cargo.toml
jsdw: Oohhh I missed that, hrmph! :D

codegen/Cargo.toml
niklasad1: ooooh, tls enabled by default by the ws-client and http-client :P

fetch-metadata/Cargo.toml
jsdw, line 1: Could we have it in `subxt/utils` perhaps, and then the crate could be `subxt-utils-fetchmetadata` or something like that, and this way it'd be clear it wasn't a particularly important crate at a glance (and then we can publish other small utils crates in the same format if there are other things to extract

niklasad1: sure good idea

utils/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! Subxt utils.
+
+// Internal helper macros
+#[macro_use]
+mod macros;
+
+macros::cfg_fetch_metadata! {
```
niklasad1, line 11: I want to add `fetch_chain_spec` here as well but I want to see whether James is happy with this and I'll do in another PR

macro/Cargo.toml
```diff
@@ -16,6 +16,7 @@ description = "Generate types and helpers for interacting with Substrate runtime
 [features]
 web = ["subxt-codegen/web"]
 runtime-path = ["polkadot-sdk"]
+runtime-metadata-insecure-url = ["subxt-utils/fetch-metadata-url"]
```
jsdw, line 19: Nit: Just to align with the above I'd clal it `runtime-url` orrr, rename both to match the attrs so we have:

`runtime-metadata-path`
`runtime-metadata-insecure-url`

utils/src/macros.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+macro_rules! cfg_fetch_metadata {
+    ($($item:item)*) => {
+        $(
+			#[cfg(any(feature = "fetch-metadata-url", feature = "fetch-metadata"))]
+			#[cfg_attr(docsrs, doc(cfg(any(feature = "fetch-metadata-url", feature = "fetch-metadata"))))]
+			$item
+		)*
+    };
```
jsdw, line 28: This appears to be a mixture of spaces and tabs; I'm surprised though that cargo fmt wouldn't have changed that if so!

jsdw (approved):
Looks good; A step in the right direction I think!

I'm still on the fence over having one utils crate versus several (because feature flags make everything messier) but this is a step up from the current state :)

utils/fetch-metadata/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019-2024 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! Subxt utils fetch metadata.
+
+// Internal helper macros
+#[macro_use]
+mod macros;
+
```
niklasad1, line 10:
```suggestion
```

lexnv (approved):
Nice! Had a brief view over the PR and looks solid 👍

PR #1832, opened 2024-10-21
chore(deps): bump `scale family crates`,  `primitive-types` and `impl-serde`

niklasad1:


jsdw (approved):

Cargo.toml
```diff
@@ -89,16 +89,16 @@ impl-serde = { version = "0.4.0", default-features = false }
...
 proc-macro2 = "1.0.86"
 quote = "1.0.37"
 regex = { version = "1.11.0", default-features = false }
 scale-info = { version = "2.11.3", default-features = false }
 scale-value = { version = "0.16.3", default-features = false }
 scale-bits = { version = "0.6.0", default-features = false }
-scale-decode = { version = "0.13.1", default-features = false }
-scale-encode = { version = "0.7.1", default-features = false }
+scale-decode = { version = "0.13.2", default-features = false }
+scale-encode = { version = "0.7.2", default-features = false }
```
niklasad1, line 101:
```suggestion
scale-encode = { version = "0.8.0", default-features = false }
```

jsdw:
The failure makes me wonder whether there is more than one version of scale-decode or scale-encode in the dependency tree now, hence "trait not implemented" type errors?

niklasad1:
> The failure makes me wonder whether there is more than one version of scale-decode or scale-encode in the dependency tree now, hence "trait not implemented" type errors?

We need to release scale-value as well :shrug: :upside_down_face:

Cargo.toml
```diff
@@ -89,16 +89,16 @@ impl-serde = { version = "0.4.0", default-features = false }
...
 jsonrpsee = { version = "0.24.5" }
 pretty_assertions = "1.4.1"
-primitive-types = { version = "0.12.2", default-features = false }
+primitive-types = { version = "0.13.1", default-features = false }
 proc-macro-error2 = "2.0.0"
 proc-macro2 = "1.0.86"
 quote = "1.0.37"
 regex = { version = "1.11.0", default-features = false }
 scale-info = { version = "2.11.3", default-features = false }
 scale-value = { version = "0.16.3", default-features = false }
```
niklasad1, line 98:
```suggestion
scale-value = { version = "0.17.0", default-features = false }
```

lexnv (approved):

jsdw (approved):

PR #1835, opened 2024-10-22
Make it clearer which extrinsic failed to decode

jsdw:
Add an index to errors relating to the extrinsics in a block failing to decode.


niklasad1 (approved):

niklasad1:
Looks like you need to update the light-client stuff as well but neat with some extra information to understand when decoding something fails.

lexnv (approved):

PR #1838, opened 2024-10-23
ci: run against older substrate/polkadot ref

niklasad1:


.github/workflows/build-nodes.yml
```diff
@@ -16,6 +16,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: paritytech/polkadot-sdk
+          # TODO: Temporary to pin a specific polkadot commit because of breakage.
```
niklasad1, line 19:
```suggestion
          # TODO: Temporary to pin a specific polkadot commit because of that substrate master doesn't work
```

PR #1843, opened 2024-10-25
Re-export jsonrpsee from `subxt::ext`

pgherveou:
Re-export so that downstream code can downcast RpcError to ErrorObject

### DIFF
--- a/subxt/src/lib.rs
+++ b/subxt/src/lib.rs
@@ -112,6 +112,10 @@ pub mod ext {
     pub use scale_value;
     pub use subxt_core;
 
+    cfg_jsonrpsee! {
+        pub use jsonrpsee;
+    }
+
     cfg_substrate_compat! {
         pub use subxt_core::ext::{sp_runtime, sp_core};
     }


subxt/src/error/mod.rs
```diff
@@ -19,6 +19,7 @@ pub use dispatch_error::{
 
 // Re-expose the errors we use from other crates here:
 pub use crate::metadata::Metadata;
+pub use jsonrpsee_types::error::ErrorObject;
```
niklasad1, line 22: you need to feature-gate this and I think this will work:

```suggestion
#[cfg(feature = "jsonrpsee")]
pub use jsonrpsee::types::error::ErrorObject;
```

jsdw:
I think you're referring to downcasting the `RpcError::Client` into the `jsonrpsee` error, which is `jsonrpsee_core::client::Error` by the looks of it.

I'm not against exposing that here as something like `JsonrpseeClientError`, but it would also need to be behind the "jsonrpsee" feature flag :)

pgherveou:
> I think you're referring to downcasting the `RpcError::Client` into the `jsonrpsee` error, which is `jsonrpsee_core::client::Error` by the looks of it.
> 
> I'm not against exposing that here as something like `JsonrpseeClientError`, but it would also need to be behind the "jsonrpsee" feature flag :)

I am trying to get this to work so I can unwrap the nested error message

https://github.com/paritytech/polkadot-sdk/blob/79294d4d2ffa4c4b9102836d1170f03b71bc1c24/substrate/frame/revive/rpc/src/client.rs?plain=1#L112-L119

subxt/src/error/mod.rs
```diff
@@ -19,7 +19,8 @@ pub use dispatch_error::{
 
 // Re-expose the errors we use from other crates here:
 pub use crate::metadata::Metadata;
-pub use jsonrpsee_types::error::ErrorObject;
+#[cfg(feature = "jsonrpsee")]
+pub use jsonrpsee::types::error::ErrorObject;
```
jsdw, line 23: I'd use the cfg macro we have like:

```
crate::macros::cfg_jsonrpsee! {
    // Things hidden by feature
}
```

But I also wonder whether the error you want is `jsonrpsee::core::client::Error`, because `ErrorObject`, which I think isn't very useful for downcasting etc?

pgherveou: maybe we can just re-export full crate here instead of jsonrpsee::types::...

subxt/src/error/mod.rs
```diff
@@ -19,6 +19,8 @@ pub use dispatch_error::{
 
 // Re-expose the errors we use from other crates here:
 pub use crate::metadata::Metadata;
+#[cfg(feature = "jsonrpsee")]
+pub use jsonrpsee::types::error::ErrorObject;
```
pgherveou, line 23:
```suggestion
crate::macros::cfg_jsonrpsee! {
    pub use jsonrpsee;
}
```

subxt/src/error/mod.rs
niklasad1: Yeah, you are correct it would be better to use `pub use jsonrpsee::core::client::Error as JsonrpseeClientError`. To support all errors that the jsonrpsee client can emit.

Maybe it works to downcast it to a ErrorObject if the error [is this variant](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/error.rs#L36-#L38)

jsdw: > Maybe it works to downcast it to a ErrorObject if the error [is this variant](https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/error.rs#L36-#L38)

downcasting basically just takes some `Box<dyn Error>` and tries to turn it into the concrete type, and for us that type is `jsonrpsee::core::client::Error` I think, so it would never work to downcast to `ErrorObject` unless I'm missing something, I think!

subxt/src/error/mod.rs
jsdw: If we wanted this, how about adding it in `subxt/src/lib.rs` to the `pub mod ext` (and then behind the cfg macro); that's how we generally expose external deps and would make sense I reckon :)

subxt/src/error/mod.rs
pgherveou: ah let me try maybe I was casting to the wrong type

> Maybe it works to downcast it to a ErrorObject if the error is this variant

I am probably hitting this nested variant

subxt/src/error/mod.rs
```diff
@@ -19,9 +19,6 @@ pub use dispatch_error::{
 
 // Re-expose the errors we use from other crates here:
 pub use crate::metadata::Metadata;
-crate::macros::cfg_jsonrpsee! {
-    pub use jsonrpsee;
-}
```
pgherveou: oups fixing..

niklasad1 (approved):

jsdw (approved):

PR #1845, opened 2024-10-28
Ignore unnecessary Extra and Call types in UncheckedExtrinsic and validation.

jsdw:
This isn't necessary for working with pallet_revive's `UncheckedExtrinsic` any more, but let's not expose/require types that we don't actually need any more to be a little more toelrant and prepare for V16 metadata which won't have an `extra` type anyway.

Fix number of account IDs from storage tests.

Old description:

1. Make extracting `Call` and `Extra` generic params from the extrinsic type optional. 

   Until now, the extrinsic type was basically always `UncheckedExtrinsic<Address, Call, Signature, Extra>`, and so the TypeInfo of that provided the address, call, signature and extra type IDs, which we'd extract in V14 metadata. (In V15 metadata these Ids are all explicit and we don't need to extract anything from the extrinsic type). 

   Now, we only _need_ `Address` and `Signature` types, making Subxt compatible with eg [`pallet_revive::evm::runtime::UncheckedExtrinsic<Address, Signature, E: EthExtra>`](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/revive/src/evm/runtime.rs#L57). Turns out that Subxt didn't really use the other two types anyway.

2. ~~When generating the Subxt interface from metadata, also find and substitute `pallet_revive::evm::runtime::UncheckedExtrinsic<Address, Signature>` for our `subxt::utile::UncheckedExtrinsic`, so that it's still possible to provide bytes as arguments that ask for that type in the generated APIs.~~

### DIFF
--- a/codegen/src/lib.rs
+++ b/codegen/src/lib.rs
@@ -351,7 +351,7 @@ fn default_derives(crate_path: &syn::Path) -> DerivesRegistry {
 fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
     let mut type_substitutes = TypeSubstitutes::new();
 
-    let defaults: [(syn::Path, syn::Path); 13] = [
+    let defaults: [(syn::Path, syn::Path); 14] = [
         (
             parse_quote!(bitvec::order::Lsb0),
             parse_quote!(#crate_path::utils::bits::Lsb0),
@@ -409,10 +409,17 @@ fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
         // `EncodeAsType` the bytes would be re-encoded. This leads to the bytes
         // being altered by adding the length prefix in front of them.
 
-        // Note: Not sure if this is appropriate or not. The most recent polkadot.rs file does not have these.
+        // This allows things calls that require an `UncheckedExtrinsic` type to be given our
+        // util one, which was be converted from a Vec<u8>:
         (
             parse_quote!(sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic),
             parse_quote!(#crate_path::utils::UncheckedExtrinsic),
+        ),
+        // This extends the above to also work with runtimes that set the UncheckedExtrinsic
+        // type to the one from pallet_revive (used for handling EVM transactions):
+        (
+            parse_quote!(pallet_revive::evm::runtime::UncheckedExtrinsic<Address, Call>),
+            parse_quote!(#crate_path::utils::UncheckedExtrinsic<Address, Call, u8, u8>),
         ),
     ];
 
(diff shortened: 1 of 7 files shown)


codegen/src/lib.rs
```diff
@@ -409,11 +409,18 @@ fn default_substitutes(crate_path: &syn::Path) -> TypeSubstitutes {
...
+        // This allows things calls that require an `UncheckedExtrinsic` type to be given our
+        // util one, which was be converted from a Vec<u8>:
         (
             parse_quote!(sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic),
             parse_quote!(#crate_path::utils::UncheckedExtrinsic),
         ),
+        // This extends the above to also work with runtimes that set the UncheckedExtrinsic
+        // type to the one from pallet_revive (used for handling EVM transactions):
+        (
+            parse_quote!(pallet_revive::evm::runtime::UncheckedExtrinsic<Address, Call>),
```
niklasad1, line 421: I wonder whether we need something special for [Acala uncheckedExtrinsic](https://github.com/AcalaNetwork/Acala/blob/master/primitives/src/unchecked_extrinsic.rs#L44-#L47) as well and double-check other chains as well?

jsdw: Yeah it looks like Subxt would run into issues with that type too when decoding from V14 metadata offhand!

That type doesn't expose _any_ of the generics we need though, so there's no obvious way to make it work for V14 metadata without somewhat custom hacks!

The second thing of being able to encode bytes into that type would be easy enough for them to create a type substitution for.

niklasad1: It's not big issue anyway since it only concern fetching metadata v14 and subxt by default is using v15

niklasad1 (approved):

jsdw:
Am holding off on this PR because there's a chance that the `TypeInfo` of the new extrinsic type will be updated anyway such that we don't need to actually make any changes here! If that's the case, we may still want to keep (1) above, to make it more likely that custom types will be supported in the future and not require a couple of types that we don't use, but it won't be super important!

metadata/src/utils/validation.rs
```diff
@@ -291,14 +291,8 @@ fn get_extrinsic_hash(
     let address_hash = get_type_hash(registry, extrinsic.address_ty, outer_enum_hashes);
     // The `RuntimeCall` type is intentionally omitted and hashed by the outer enums instead.
     let signature_hash = get_type_hash(registry, extrinsic.signature_ty, outer_enum_hashes);
-    let extra_hash = get_type_hash(registry, extrinsic.extra_ty, outer_enum_hashes);
```
jsdw, line 294: No need to validate this because we validate signed exts below. In V16 metadata this type will be removed anyway.

jsdw:
Closing in favour of a smaller fix: #1848

PR #1848, opened 2024-10-29
Fix substrate nightly

### LINKED ISSUE #1849: Subxt integration tests failed against latest Substrate build.

jsdw:
Now 16, not 14 account IDs by default on the node.

I also tweaked the ExtrincicParamTypeIds code to avoid allocating a hashmap and bailing if we find types we don't care about which don't have IDs.

Close #1849

### DIFF
--- a/metadata/src/from_into/v14.rs
+++ b/metadata/src/from_into/v14.rs
@@ -10,7 +10,6 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt::Write;
 use frame_metadata::{v14, v15};
-use hashbrown::HashMap;
 use scale_info::TypeDef;
 
 impl TryFrom<v14::RuntimeMetadataV14> for Metadata {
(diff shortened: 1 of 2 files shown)


niklasad1 (approved):

metadata/src/from_into/v14.rs
```diff
@@ -277,36 +276,33 @@ impl ExtrinsicPartTypeIds {
...
-            .iter()
-            .map(|ty_param| {
-                let Some(ty) = ty_param.ty else {
-                    return Err(TryFromError::TypeNameNotFound(ty_param.name.clone()));
-                };
-
-                Ok((ty_param.name.as_str(), ty.id))
-            })
-            .collect::<Result<_, _>>()?;
+        let find_param = |name: &'static str| -> Option<u32> {
```
niklasad1, line 279: Wo you know how big the extrinsic type params list is? 

We will do a linear search 4 times on instead 4 HashMap lookups which could be slower...

niklasad1: Ok, I checked it myself it's `extrinsic_ty.len() == 4`

lexnv (approved):

PR #1850, opened 2024-10-30
remove substrate compat

### LINKED ISSUE #1847: remove `substrate-compat` feature
subxt has a feature called `substrate-compat` which provides a few nice impls/features from polkadot-sdk such as:
- polkadot sdk signer
- Glue code for seamless conversions between `subxt types` and `polkadot-sdk types` such as AccountId32 and similar

This comes with the downside that subxt depends on polkadot-sdk which is brings in lots of dependencies and that subxt is dependent of the polkadot-sdk releases such as i) if there is no subxt release for a while it may use an outdated polkadot-sdk version ii) that a specific user wants to use a specific release iii) the code doesn't compile with different versions of some dependency.

In addition, subxt got a lot of issues where folks have issues with a certain polkadot-sdk release/branch and this will allow users to select which polkadot-sdk release/branch would makes this easier. Of course to some degree subxt is dependent to polkadot-sdk such new transaction extensions, signed extensions and similar things.

So the idea is that subxt will remove this feature, the user has to bring these dependencies themselves

It's important that subxt provides a nice and easy example how to do this without the `subxt substrate-compat`  feature.

niklasad1:
Close https://github.com/paritytech/subxt/issues/1847


subxt/examples/substrate_compat_signer.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    // Create a new API client, configured to talk to Polkadot nodes.
+    let api = OnlineClient::<PolkadotConfig>::new().await?;
+
+    let signer = {
+        let acc = sr25519::Pair::from_string("//Alice", None)?;
+        pair_signer::PairSigner::new(acc)
+    };
+
+    let dest = subxt_signer::sr25519::dev::bob().public_key().into();
```
jsdw, line 94:
```suggestion
    let dest = {
        let acc = sr25519::Pair::from_string("//Bob", None)?;
        MultiAddress::Address32(acc.public().0)
    };
```

Just because the aim is to show that subxt_signer isn't used at all and Subxt is playing nice with Substrate's signer stuff :)

subxt/examples/substrate_compat.rs
jsdw, line 1: Just repeating what I already said on the call: I'd either:

1. Remove this example, or
2. Show in the `fn main()` for it something that demonstrates when this approach is good to take (ie when would I copy this example versus copying the simpler `substrate_compat_signer` one which ultimately runs the same code)

subxt/examples/substrate_compat_signer.rs
jsdw, line 1: I like this example; for me it shows that it's not hard to use Substrate's signer stuff instead if people would rather do that!

jsdw (commented):
Based on the signer example I feel happy enough removing the substrate compat feature; it's not too hard plugging things like Substrate's signer stuff in, and this means we are independent and one can write SUbxt code against _any_ version of `polkadot-sdk` that they want :)

jsdw:
Whoops; my suggestion broke the cargo check; I forgot about the extra import needed!

subxt/src/book/usage/transactions.rs
```diff
@@ -77,7 +77,6 @@
 //! Let's see how to use each of these approaches:
```
jsdw, line 77: I think here I'd be inclined to only show how to do (1), and then link to your example for (2) since it's suddenly quite verbose

jsdw (approved):
This LGTM and I like the example!

Just a small doc change suggestion, but this can also be done as a followup if preferred!

subxt/src/book/usage/transactions.rs
```diff
@@ -66,8 +66,7 @@
 //! There are two main ways to create a compatible signer instance:
 //! 1. The `subxt_signer` crate provides a WASM compatible implementation of [`crate::tx::Signer`]
 //! for chains which require sr25519 or ecdsa signatures (requires the `subxt` feature to be enabled).
-//! 2. Alternately, Subxt can use instances of Substrate's `sp_core::Pair` to sign things by wrapping
-//! them in a new type pattern and implement `crate::tx::Signer`.
+//! 2. Alternately, implement your own [`crate::tx::Signer`] instance using the `polkadot-sdk` signer.
```
niklasad1, line 69:
```suggestion
//! 2. Alternately, implement your own [`crate::tx::Signer`] instance by wrapping it in a new type pattern.
     This can be used to use the polkadot-sdk signer instead of subxt signer.
```

subxt/src/book/usage/transactions.rs
```diff
@@ -94,76 +93,7 @@
...
-//! # is_subxt_signer(PairSigner::new(sp_keyring::AccountKeyring::Alice.pair()));
-//! # }
-//! ```
-//!
-//! See the `subxt_signer` crate or the `sp_core::Pair` docs for more ways to construct
-//! and work with key pairs.
-//!
-//! If this isn't suitable/available, you can either implement [`crate::tx::Signer`] yourself to use
-//! custom signing logic, or you can use some external signing logic, like so:
+//! //// 2.  See `subxt/examples/substrate_compat_signer.rs` how to use the polkadot-sdk signer with Subxt.
```
jsdw, line 96: Nit: Maybe have this below the code block instead, something like:

"For 2, see <...> to see how to you can intagrate things like `sp_core`'s signer in to Subxt."

subxt/src/book/usage/transactions.rs
```diff
@@ -66,8 +66,8 @@
 //! There are two main ways to create a compatible signer instance:
 //! 1. The `subxt_signer` crate provides a WASM compatible implementation of [`crate::tx::Signer`]
 //! for chains which require sr25519 or ecdsa signatures (requires the `subxt` feature to be enabled).
-//! 2. Alternately, Subxt can use instances of Substrate's `sp_core::Pair` to sign things by wrapping
-//! them in a `crate::tx::PairSigner` (requires the `substrate-compat` feature to be enabled).
+//! 2. Alternately, implement your own [`crate::tx::Signer`] instance by wrapping it in a new type pattern.
+     This can be used to use the polkadot-sdk signer instead of subxt signer.
```
niklasad1, line 70:
```suggestion
```

subxt/src/book/usage/transactions.rs
jsdw: On line 76 (sorry, can't comment there) I'd change "Let's see how to use each of these approaches:" into something like "Here's an example of the usual approach to signing:"

And then here, we could have:

```suggestion
//! 2. Alternately, implement your own [`crate::tx::Signer`] instance using something like the `polkadot-sdk` signer.
```

niklasad1: I changed this is slightly and add something similar at another place

lexnv (approved):

PR #1856, opened 2024-11-06
Migrate custom error trait impls to `thiserror`

pkhry:
### Description 

Migrates Error trait impls to `thiserror` as it now supports nostd


core/Cargo.toml
```diff
@@ -65,6 +65,9 @@ tracing = { workspace = true, default-features = false }
 # AccountId20
 keccak-hash = { workspace = true}
 
+# Error handling 
+thiserror = { workspace = true, default-features = false }
```
jsdw, line 69: I'd just lump this with the main blob of dependencies :)

pkhry: done

core/src/error.rs
```diff
@@ -4,203 +4,142 @@
...
-#[derive(Debug)]
+#[derive(Debug, DeriveError)]
 pub enum Error {
     /// Codec error.
-    Codec(codec::Error),
+    #[error("Scale codec error: {0}")]
+    Codec(#[from] codec::Error),
     /// Metadata error.
-    Metadata(MetadataError),
+    #[error("Metadata Error: {0}")]
```
niklasad1, line 19:
```suggestion
    #[error(transparent)]
```

core/src/error.rs
```diff
@@ -4,203 +4,142 @@
...
 use subxt_metadata::StorageHasher;
+use thiserror::Error as DeriveError;
 
 /// The error emitted when something goes wrong.
-#[derive(Debug)]
+#[derive(Debug, DeriveError)]
 pub enum Error {
     /// Codec error.
-    Codec(codec::Error),
+    #[error("Scale codec error: {0}")]
```
niklasad1, line 16:
```suggestion
    #[error(transparent)]
```

core/src/events.rs
```diff
@@ -376,7 +376,8 @@ impl<T: Config> EventDetails<T> {
             .map(|f| scale_decode::Field::new(f.ty.id, f.name.as_deref()));
 
         let decoded =
-            scale_value::scale::decode_as_fields(bytes, &mut fields, self.metadata.types())?;
+            scale_value::scale::decode_as_fields(bytes, &mut fields, self.metadata.types())
+                .map_err(Into::<scale_decode::Error>::into)?;
```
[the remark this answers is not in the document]
pkhry: decode_as_fields returns `DecodeError` and although `scale_decode::Error` implements `From<DecodeError>` compiler needs an explicit cast

niklasad1 (approved):
Looks good to me.

Would be good if you could double-check whether we should use `#[error(transparent)]` from every wrapped error type to avoid duplicate stuff in the display impl.

Such `Decode error: Decode error <some message>` but I'm not sure I could be wrong :P

core/src/events.rs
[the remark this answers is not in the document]
jsdw: I don't get why that would be the case though offhand, when it wasn't before? Also, out of interest, can you use `Into::into` or do you need the explicit type there too?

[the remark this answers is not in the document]
pkhry: No idea either.
Fails to infer the type without explicit annotation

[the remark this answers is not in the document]
jsdw: I had a look and there was a missing `From` impl for `scale_decode::visitor::DecodeError` (confusingly similar name!); adding that let the Into's go away :)

jsdw (approved):
Nice!

signer/src/polkadot_js_compat.rs
```diff
@@ -5,14 +5,15 @@
 //! A Polkadot-JS account loader.
 
 use base64::Engine;
-use core::fmt::Display;
 use crypto_secretbox::{
     aead::{Aead, KeyInit},
     Key, Nonce, XSalsa20Poly1305,
 };
 use serde::Deserialize;
 use subxt_core::utils::AccountId32;
```
niklasad1, line 14:
```suggestion
```

core/src/error.rs
```diff
@@ -41,6 +41,14 @@ impl From<scale_decode::visitor::DecodeError> for Error {
     }
 }
 
+// TODO: when `codec::Error` implements `core::Error`
+// remove this impl and replace it by thiserror #[from]
+impl From<codec::Error> for Error {
+    fn from(err: codec::Error) -> Error {
+        Error::Codec(err)
+    }
+}
```
jsdw, line 50: Why does not implementing `core::Error` stop you from using `#[from]` here?

signer/src/ecdsa.rs
```diff
@@ -222,33 +224,33 @@ pub(crate) mod internal {
...
-impl_from!(bip39::Error => Error::Phrase);
-impl_from!(hex::FromHexError => Error::Hex);
-
-#[cfg(feature = "std")]
-impl std::error::Error for Error {}
+impl From<bip39::Error> for Error {
+    fn from(err: bip39::Error) -> Self {
+        Error::Phrase(err)
+    }
+}
```
jsdw, line 253: User `#[from]` instead?

signer/src/sr25519.rs
```diff
@@ -203,35 +205,39 @@ pub fn verify<M: AsRef<[u8]>>(sig: &Signature, message: M, pubkey: &PublicKey) -
...
-
-impl Display for Error {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        match self {
-            Error::InvalidSeed => write!(f, "Invalid seed (was it the wrong length?)"),
-            Error::Phrase(e) => write!(f, "Cannot parse phrase: {e}"),
-            Error::Hex(e) => write!(f, "Cannot parse hex string: {e}"),
-            Error::Signature(e) => write!(f, "Signature error: {e}"),
-        }
+impl From<schnorrkel::SignatureError> for Error {
```
jsdw, line 224: Use `#[from]` instead of the below manual From impls?

jsdw (approved):
LGTM, just a few places it looks like you can use `#[from]` unless I'm not seeing the reason why not :)

core/src/error.rs
niklasad1:
```bash
error[E0599]: the method `as_dyn_error` exists for reference `&Error`, but its trait bounds were not satisfied
  --> core/src/error.rs:17:13
   |
17 |     Codec(#[from] codec::Error),
   |             ^^^^ method cannot be called on `&Error` due to unsatisfied trait bounds
   |
  ::: /home/niklasad1/.cargo/registry/src/index.crates.io-6f17d22bba15001f/parity-scale-codec-3.6.12/src/error.rs:26:1
   |
26 | pub struct Error {
   | ---------------- doesn't satisfy `parity_scale_codec::Error: AsDynError<'_>` or `parity_scale_codec::Error: StdError`
   |
   = note: the following trait bounds were not satisfied:
           `parity_scale_codec::Error: StdError`
           which is required by `parity_scale_codec::Error: AsDynError<'_>`
           `&parity_scale_codec::Error: StdError`
           which is required by `&parity_scale_codec::Error: AsDynError<'_>`
```

It's probably because thiserror implements core::error::Error and codec::Error implements std::error::Error but perhaps we could open issue in thiserror if there is some way around that

signer/src/ecdsa.rs
niklasad1: Same as :point_up:

PR #1859, opened 2024-11-08
Load latest metadata version from Wasm blobs.

### LINKED ISSUE #1858: `RUNTIME_APIS` is empty when expanded with wasm path
#1720 introduced a way to generate metadata from runtime wasm files, but when I try generate it like;

```rust
#[subxt_macro::subxt(
    runtime_path = "<my wasm path>"
)]
pub mod foo {}
```

The resulting code expanded with `cargo-expand` shows as

```rust
#[allow(dead_code, unused_imports, non_camel_case_types, unreachable_patterns)]
#[allow(clippy::all)]
#[allow(rustdoc::broken_intra_doc_links)]
pub mod foo {
    #[allow(unused_imports)]
    mod root_mod {
        pub use super::*;
    }
    pub static PALLETS: [&str; 63usize] = [
..
    ];
    // This is empty!
    pub static RUNTIME_APIS: [&str; 0usize] = [];
```

while expanding macros in previous ways like

```rust
#[subxt_macro::subxt(
    runtime_metadata_insecure_url = "wss://westend-rpc.polkadot.io")
]
pub mod foo {}
```

the `RUNTIME_APIS` is filled like

```rust
...
    pub static RUNTIME_APIS: [&str; 22usize] = [
        "Core",
        "Metadata",
        "BlockBuilder",
...
```

I've tried with wasm files from https://github.com/polkadot-fellows/runtimes/releases/ and https://github.com/paritytech/subxt/blob/master/artifacts/westend_runtime.wasm but all of them are expanded into empty `RUNTIME_APIS`

pkhry:
closes #1858

### DIFF
--- a/macro/src/wasm_loader.rs
+++ b/macro/src/wasm_loader.rs
@@ -14,6 +14,8 @@ use polkadot_sdk::{
 };
 use subxt_codegen::{CodegenError, Metadata};
 
+static SUPPORTED_METADATA_VERSIONS: [u32; 2] = [14, 15];
+
 /// Result type shorthand
 pub type WasmMetadataResult<A> = Result<A, CodegenError>;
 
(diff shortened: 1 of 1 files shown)


macro/src/wasm_loader.rs
```diff
@@ -39,17 +39,50 @@ fn call_and_decode(wasm_file: Vec<u8>) -> WasmMetadataResult<Metadata> {
...
+            CodegenError::Wasm("method \"Metadata_metadata_at_version\" doesnt exist".to_owned())
+        })?;
+
+    decode(encoded_metadata)
 }
 
 fn decode(encoded_metadata: Vec<u8>) -> WasmMetadataResult<Metadata> {
-    Metadata::decode(&mut encoded_metadata.as_ref()).map_err(Into::into)
+    // We slice the first byte from the metadata because it's wrapped inside an option and we know that its always `Some`
+    let metadata = <Vec<u8>>::decode(&mut &encoded_metadata[1..]).map_err(CodegenError::Decode)?;
```
lexnv, line 84: nit: Would this panic on empty `encoded_metadata`?

macro/src/wasm_loader.rs
```diff
@@ -39,17 +39,50 @@ fn call_and_decode(wasm_file: Vec<u8>) -> WasmMetadataResult<Metadata> {
...
+    let encoded_metadata = executor
+        .uncached_call(
+            runtime_blob,
+            &mut ext,
+            false,
+            "Metadata_metadata_at_version",
+            &version,
+        )
+        .map_err(|e| {
+            dbg!(e);
```
lexnv, line 75: Could we use tracing:: or log:: here instead of dbg?

pkhry: Oh, I forgot to remove this

macro/src/wasm_loader.rs
pkhry: Yeah, probably, will change

lexnv (approved):
Great job here! 

Thinking outloud here, maybe a followup: it would be useful at some point to extend the functionality of this module:
- have a `WasmBlobExecutor` wrapper that is handling just the executor building and runtime blob
- then we could expose `WasmBlobExecutor::versions` which calls into `Metadata_metadata_versions`
- similar for `WasmBlobExecutor::get_metadata(version: ..)` 

This way we can choose in our macros which version to use (most prob the lastest, but sometimes we may want to fallback to lastest stable instead of unstable)

macro/src/wasm_loader.rs
```diff
@@ -39,17 +39,50 @@ fn call_and_decode(wasm_file: Vec<u8>) -> WasmMetadataResult<Metadata> {
...
+            CodegenError::Wasm("method \"Metadata_metadata_versions\" doesnt exist".to_owned())
+        })?;
+    let mut versions = <Vec<u32>>::decode(&mut &version[..]).map_err(CodegenError::Decode)?;
 
-    decode(metadata)
+    // Highest version will always be the last one in the vec
+    versions.sort();
+
+    let version = versions
+        .last()
```
lexnv, line 60: I think iter::max could do the trick here as well

macro/src/wasm_loader.rs
ShoyuVanilla: I think that the following code would work 😄 

```rust
let metadata = <Option<Vec<u8>>>::decode(&mut &encoded_metadata[..]).map_err(CodegenError::Decode)?;
```

macro/src/wasm_loader.rs
```diff
@@ -39,17 +39,50 @@ fn call_and_decode(wasm_file: Vec<u8>) -> WasmMetadataResult<Metadata> {
...
+            "Metadata_metadata_versions",
+            &[],
+        )
+        .map_err(|_| {
+            CodegenError::Wasm("method \"Metadata_metadata_versions\" doesnt exist".to_owned())
+        })?;
+    let mut versions = <Vec<u32>>::decode(&mut &version[..]).map_err(CodegenError::Decode)?;
 
-    decode(metadata)
+    // Highest version will always be the last one in the vec
```
jsdw, line 56: I'd look for a specific version here that we know Subxt supports (ie 15, or 14 failing that).

The highest version is also sometimes the unstable metadata version which is `u32::MAX` here and this should be ignored.

macro/src/wasm_loader.rs
pkhry: refactored a bit to also fail on `None`

macro/src/wasm_loader.rs
```diff
@@ -39,17 +39,50 @@ fn call_and_decode(wasm_file: Vec<u8>) -> WasmMetadataResult<Metadata> {
...
+    let version = executor
+        .uncached_call(
+            runtime_blob.clone(),
+            &mut ext,
+            true,
+            "Metadata_metadata_versions",
+            &[],
+        )
+        .map_err(|_| {
+            CodegenError::Wasm("method \"Metadata_metadata_versions\" doesnt exist".to_owned())
```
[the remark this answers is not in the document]
pkhry: now the code will fallback to `Metadata_metadata`

macro/src/wasm_loader.rs
pkhry: Added a filter to only look for v14 and v15

niklasad1 (approved):

macro/src/wasm_loader.rs
```diff
@@ -57,3 +46,115 @@ fn maybe_decompress(file_contents: Vec<u8>) -> WasmMetadataResult<Vec<u8>> {
...
+        if versions.is_empty() {
+            return Err(CodegenError::Other(
+                "No supported metadata versions were returned".to_owned(),
+            ));
+        }
+
+        Ok(versions)
+    }
+
+    fn metadata(&mut self) -> WasmMetadataResult<Metadata> {
```
jsdw, line 109: Nit: I'd call this something like `load_legacy_metadata` or something to be consistent with below (and make clear it's the "old" call :))

macro/src/wasm_loader.rs
```diff
@@ -57,3 +46,115 @@ fn maybe_decompress(file_contents: Vec<u8>) -> WasmMetadataResult<Vec<u8>> {
...
+            .executor
+            .uncached_call(
+                self.runtime_blob.clone(),
+                &mut self.externalities,
+                false,
+                "Metadata_metadata_at_version",
+                &version.encode(),
+            )
+            .map_err(|e| {
+                dbg!(e);
```
jsdw, line 146: Remove this? :)

macro/src/wasm_loader.rs
pkhry: renamed to `load_legacy_metadata`

macro/src/wasm_loader.rs
pkhry: finally done

jsdw (approved):
LGTM, nice one!

Not needed in this PR but I wonder how easily one could make a coupel of small-as-possible test runtimes exposing the new and old APIs to test that we get the expected V14 or V15 metadatas from them. May not be worth the effort though..

Issue #1862, opened 2024-11-12
Signing fails when using lightclient feature for Wasm Apps


paulormart:
Hi Team, when using the `unstable-light-client` feature to build Wasm Apps, the transactions are not successfully submitted to the network, raising the error below:

```
SubxtError(Serialization(Error("missing field `num_peers`", line: 0, column: 0)))
```

niklasad1:
I think the broadcast event was removed from transactionWatch API, https://github.com/paritytech/json-rpc-interface-spec/pull/134

So, I guess we could make it an `num_peers: Option<u32`>` to work with older versions but quite weird that this event is generated without a count.

What do you guys reckon @jsdw @lexnv?

jsdw:
Agreed that we should just remove the `Broadcasted` event entirely since it's not a part of the stable spec (and the `broadcasted: bool` from the `Dropped` event). 

But yeah, that wouldn't resolve this issue. In the spec, it doesn't appear that there was ever a version with the "broadcasted" event but no number, so I'm not sure how this would happen. @paulormart do you have an easy way to reproduce this with a specific chain, and which chain did the error crop up on?

paulormart:
The easiest way to reproduce that I found, was to run the `wasm-example` with the `unstable-light-client` enabled and replace the default `OnlineClient` with a backend using a `LightClient` loaded from a network chain spec. 

I currently have this implemented on my own branch [here](https://github.com/turboflakes/subxt/tree/tmp-fix-tx-broadcasted/examples/wasm-example) using the Paseo network chainspec. And also pushed a PR with these changes here https://github.com/paritytech/subxt/pull/1863/commits/d19dd5a218c12eb31c7e42c92d7e5495f7718bd6

jsdw:
Thanks!

I opened #1865 to address this; it was a little like your PR but more aggressive in removing fields that were no logner a part of the spec :)

I didn't commit to using Paseo in the lightclient WASM example but instead reprocuded using a native light client example.

jsdw:
Closed by #1865

PR #1863, opened 2024-11-12
Temporary fix for tx broadcasted issue

paulormart:
Follow up of issue #1862 

This PR is a workaround I am currently using to fix the current mapping in TransactionStatus so that transactions are successfully submitted.

I have also changed the `wasm-example` to use the light client backend so that this could be tested on Paseo network.


paulormart:
Fixed on #1865

PR #1877, opened 2024-12-02
Re-export more `scale_value` functionality

cmichi:
We need access to some more functionality here in ink!.

### DIFF
--- a/core/src/dynamic.rs
+++ b/core/src/dynamic.rs
@@ -8,7 +8,7 @@
 use crate::metadata::{DecodeWithMetadata, Metadata};
 use alloc::vec::Vec;
 use scale_decode::DecodeAsType;
-pub use scale_value::{At, Value};
+pub use scale_value::{scale, serde, stringify, At, Value};
 
 /// A [`scale_value::Value`] type endowed with contextual information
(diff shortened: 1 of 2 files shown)


core/src/dynamic.rs
```diff
@@ -8,7 +8,7 @@
 use crate::metadata::{DecodeWithMetadata, Metadata};
 use alloc::vec::Vec;
 use scale_decode::DecodeAsType;
-pub use scale_value::{At, Value};
+pub use scale_value::{scale, serde, stringify, At, Value};
```
niklasad1, line 11: can't you guys use [subxt::ext::scale_value](https://github.com/paritytech/subxt/blob/master/subxt/src/lib.rs#L109) instead of relying of this re-export from core?

jsdw: I agree that it's better to use the ext imports (or depend on the same version of scale-value in Cargo.toml) instead of subxt exporting everything; lessens the burden on subxt if we don't have to re-export everything from dependent libraries :)

cmichi: Thanks, I wasn't aware of `subxt::ext`. That solves it.

PR #1884, opened 2024-12-12
docs: Improve substitute_type rules and alternatives

### LINKED ISSUE #1464: Unclear path substition rules
It is not clear how to specify the substitute `syn::Path`s required as argument.

A few things remain puzzling:

1. what is the relative base module needed to get a substitution match?
2. does it have to be absolute paths?
3. do bare types (so i.e. just an `Ident`) themselves suffice for matching?

```rust
let path = &ty.ty.path;
            // Don't generate a type if it was substituted - the target type might
            // not be in the type registry + our resolution already performs the substitution.
            if self.settings.substitutes.contains(dbg!(&path.segments)) {
                continue;
            }
```

is the relevant line, but it's not clear what `path.segments` contains and relative to what base.

mittal-parth:
- Clarifies path substitution rules
- Adds the recommend alternative of type conversion

Closes #1464


subxt/src/lib.rs
```diff
@@ -168,46 +168,29 @@ pub mod ext {
 ///
 /// ## `substitute_type(path = "...", with = "...")`
 ///
-/// This attribute replaces any reference to the generated type at the path given by `path` with a
-/// reference to the path given by `with`.
+/// Replaces references to a specific type in the generated code with a custom type.
+///
+/// ### Important Notes
+/// - The path to replace must be an absolute path (fully qualified)
+/// - The replacement type must implement the necessary traits for encoding and decoding
```
lexnv, line 175: nit: Maybe we can mention the traits here?

mittal-parth: Woudln't that be dynamic?

mittal-parth: @lexnv

subxt/src/lib.rs
```diff
@@ -168,46 +168,29 @@ pub mod ext {
...
-/// #     fn from(v: codec::Compact<Foo>) -> Foo {
-/// #         v.0
-/// #     }
-/// # }
-/// # fn main() {}
+/// mod polkadot {
+///     // Custom type to replace Perbill
+///     pub struct Foo(u32);
+///     // Note: Foo must implement required traits
+/// }
```
jsdw, line 189: This looks wrong to me offhand; the substititue path is "crate::Foo" but then we move the type into `crate::polkadot::Foo`? I'd just keep the type defined at the top level to avoid anybody thinking that there is a requirement for the extra mod or whatever.

PR #1892, opened 2025-01-02
fix: substrate runner, support new libp2p addr log

niklasad1:


testing/substrate-runner/src/lib.rs
```diff
@@ -291,7 +291,10 @@ fn try_find_substrate_port_from_output(r: impl Read + Send + 'static) -> Substra
 
         // Parse the p2p port line (present in debug logs)
         let p2p_port_line = line
+            // oldest message:
             .rsplit_once("New listen address: /ip4/127.0.0.1/tcp/")
+            // slightly newer message:
+            .or_else(|| line.rsplit_once("New listen address address=/ip4/127.0.0.1/tcp/"))
```
niklasad1, line 297: /cc @lexnv this log with duplicate "address" seems like a nit?

pkhry (approved):
LGTM. 

But a small nit: while we are at it can we make function `try_find_substrate_port_from_output` in `subxt-runner` return `Result<SubstrateNodeInfo, Error>` and fail early in the function if some of the values are not present?(we use the function only once on line 105 and we try to unconditionally extract  optinoal values)

niklasad1:
> But a small nit: while we are at it can we make function try_find_substrate_port_from_output in subxt-runner return Result<SubstrateNodeInfo, Error> and fail early in the function if some of the values are not present?(we use the function only once on line 105 and we try to unconditionally extract optinoal values)

We can't really just return early if some line couldn't be parsed because these are logs from the substrate node which we iterator line by line (i.e. std::io::Read) that contains a bunch of unrelated stuff which we are not interested in. So the assumption is the rpc server port, libp2p address and libp2p port should be found in the first 100 log entries when starting the node. Then line 105 is just a way to stop the iteration as soon as all values are found which is checked every iteration

Also, this function panics if any of patterns was matched but parsing the value failed, which works in the same way but less elegant ^^

Perhaps we could just read all lines to a String then do search it for the values but should be less efficient but more readable I guess.

pkhry:
> > But a small nit: while we are at it can we make function try_find_substrate_port_from_output in subxt-runner return Result<SubstrateNodeInfo, Error> and fail early in the function if some of the values are not present?(we use the function only once on line 105 and we try to unconditionally extract optinoal values)
> 
> We can't really just return early if some line couldn't be parsed because these are logs from the substrate node which we iterator line by line (i.e. std::io::Read) that contains a bunch of unrelated stuff which we are not interested in. So the assumption is the rpc server port, libp2p address and libp2p port should be found in the first 100 log entries when starting the node. Then line 105 is just a way to stop the iteration as soon as all values are found which is checked every iteration
> 
> Also, this function panics if any of patterns was matched but parsing the value failed, which works in the same way but less elegant ^^
> 
> Perhaps we could just read all lines to a String then do search it for the values but should be less efficient but more readable I guess.

I was thinking more inline of just making fields non optional in the struct SubstrateNodeInfo

niklasad1:
> I was thinking more inline of just making fields non optional in the struct SubstrateNodeInfo

aight, I see yeah that's a good idea :)

lexnv (approved):

PR #1910, opened 2025-01-27
Split RPCs into a separate crate

jsdw:
Create a new crate, `subxt-rpcs`. This crate exposes:

- _methods_: `legacy` and `chainHead` RPCs are exposed (the same set that were available in `subxt`), and `archive` methods will be added here later.
- _client_: a set of clients are exposed: `jsonrpsee`, `reconnecting-rpc-client`, `unstable-light-client` and `mock-rpc-client`. Any of these can be used to call the methods. Subxt makes use of each of these.

Basically, we get all of the low level network communication/serialization/deserialization/typing handled for us, without any of the additional abstraction that Subxt places on top. Useful if you want to interact directly with the RPCs but don't want any other features provided by Subxt.


rpcs/src/methods/legacy.rs
```diff
@@ -509,55 +524,6 @@ pub enum TransactionStatus<Hash> {
     Invalid,
 }
 
-/// The decoded result returned from calling `system_dryRun` on some extrinsic.
```
jsdw, line 512: This stuff requires extra crates/features (eg subct_metadata and all fo the dispatch error stuff) so the logic is left in subxt.

rpcs/src/client/lightclient_impl.rs
```diff
@@ -36,18 +36,27 @@ impl RpcClientT for LightClientRpc {
...
+                .map_err(|e| Error::Client(Box::new(e)))
                 .boxed();
 
             Ok(RawRpcSubscription { id, stream })
         })
     }
 }
 
-fn lc_err_to_rpc_err(err: LightClientRpcError) -> RpcError {
+fn lc_err_to_rpc_err(err: LightClientRpcError) -> Error {
```
niklasad1, line 47: can we make this an into impl instead? Or stopped by the orphan-rule?

rpcs/src/client/jsonrpsee_impl.rs
```diff
@@ -60,9 +60,27 @@ impl RpcClientT for Client {
...
                 .boxed();
             Ok(RawRpcSubscription { stream, id })
         })
     }
 }
+
+/// Convert a JsonrpseeError into the RPC error in this crate.
+/// The main reason for this is to capture user errors so that
+/// they can be represented/handled without casting. 
+fn error_to_rpc_error(error: JsonrpseeError) -> Error {
```
niklasad1, line 73: same here, can we change it to an into/from impl instead?

rpcs/src/client/reconnecting_rpc_client/mod.rs
```diff
@@ -448,7 +407,7 @@ impl RpcClientT for RpcClient {
             let sub = self
                 .subscribe(sub.to_string(), params, unsub.to_string())
                 .await
-                .map_err(|e| SubxtRpcError::ClientError(Box::new(e)))?;
+                .map_err(error_to_rpc_error)?;
```
niklasad1, line 410: good

rpcs/src/methods/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+//! We also have (although their use is not advised):
+//!
+//! - [`LegacyRpcMethods`] (and the types in [`legacy`]): a collection of legacy RPCs.
+//!   These are not well specified and may change in implementations without warning,
+//!   but for those methods we expose, we make a best effort to work against latest Substrate versions.
+
+pub mod chain_head;
+pub mod legacy;
+
+pub use chain_head::ChainHeadRpcMethods;
```
niklasad1, line 19: The question is whether we shall support more rpc methods such as `rpc v2 archieve` and similar.

I think no, but we could open an issue for it

rpcs/src/client/mock_rpc_client.rs
```diff
@@ -0,0 +1,643 @@
...
+pub struct MockRpcClientBuilder {
+    method_handlers_once: HashMap<String, VecDeque<MethodHandlerFnOnce>>,
+    method_handlers: HashMap<String, MethodHandlerFn>,
+    method_fallback: Option<MethodHandlerFn>,
+    subscription_handlers_once: HashMap<String, VecDeque<SubscriptionHandlerFnOnce>>,
+    subscription_handlers: HashMap<String, SubscriptionHandlerFn>,
+    subscription_fallback: Option<SubscriptionHandlerFn>
+}
+
+impl  MockRpcClientBuilder {
```
niklasad1, line 63:
```suggestion
impl MockRpcClientBuilder {
```

rpcs/src/client/mock_rpc_client.rs
```diff
@@ -0,0 +1,643 @@
...
+    fn new() -> Self {
```
niklasad1, line 64: could be replaced by `derive(Default)` to save a few lines of code

subxt/src/backend/mod.rs
```diff
@@ -1125,140 +925,46 @@ mod test {
...
-                        response.map(|x| RawValue::from_string(x).unwrap())
-                    })
+                .method_handler_once("chainHead_v1_storage", move |_params| async move {
+                    // Next call, return a storage item and then a "waiting for continue".
+                    tokio::spawn(async move {
+                        tx.send(storage_items("Id1", &[storage_result("ID1", "Data1")]))
+                            .unwrap();
+                        tx.send(operation_continue("Id1")).unwrap();
+                    });
+                    Ok(Json(response_started("Id1")))
```
niklasad1, line 946: Technically, we should ensure that response_started was sent "before" the messages in the tokio::spawn above.

if the tests passes it's not a big deal but you use a oneshot for that here.

niklasad1 (commented):
Looks great so far, it's definitely a nice cleanup of the error handling.

I got tired now and I'll finish the review tomorrow.

subxt/src/backend/mod.rs
jsdw: Actually, I don't mind if some data comes back _before_ the "response started" because, owing to races, there's a chance that could happen in real life too, so this checks that the client is ready to receive such messages before the response comes back (which I think we do; we subscribe before making the call to buffer any responses)

rpcs/src/client/lightclient_impl.rs
jsdw: I think we could yeah, but it would make the conversion more "public" I guess, and here I was thinking of it as an implementation detail!

subxt/src/backend/mod.rs
```diff
@@ -790,94 +592,72 @@ mod test {
...
+                    Err(subxt_rpcs::Error::DisconnectedWillReconnect(
+                        "..".to_string(),
+                    )),
+                    Ok(Json(runtime_version(2))),
+                    Ok(Json(runtime_version(3))),
+                ],
+                vec![
+                    Ok(Json(runtime_version(4))),
+                    Ok(Json(runtime_version(5))),
+                    Err(subxt_rpcs::Error::Client("..".into())),
```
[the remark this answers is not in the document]
jsdw: Oh I just used ".." because it's just test code so not really relevant what the reason text is :)

IIRC The `RequestRejected` error was only ever used in the case that we got a `LimitReached` response back from the chainHead RPCs, so I renamed it to be `LimitReached` as it's a bit of a special case error (ie the `MethodResponse::LimitReached` we get back isn't an error from the API, but we want to turn it into one in Subxt).

In this particular test I just think that _any_ error was good enough to show that after an error the stream returns None

subxt/src/backend/mod.rs
```diff
@@ -790,94 +592,72 @@ mod test {
...
+                client_runtime_version(4)
+            );
+            assert_eq!(
+                results.next().await.unwrap().unwrap(),
+                client_runtime_version(5)
+            );
+            assert!(matches!(
+                results.next().await.unwrap(),
+                Err(Error::Rpc(RpcError::ClientError(
+                    subxt_rpcs::Error::Client(_)
```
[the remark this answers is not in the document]
jsdw: Yeah; `LimitReached` is an error that signals that the chainHead backend returned the non-error response `LimitReached`, which we treat as an error in Subxt and so turn into one.

I don't remember explicitly changing any errors from Rejected to ClientError (should be form Rejected to LimitReached), but in some of these tests it doesn't matter, so I might have opted for an error that didn't have any "special" retry consequences!

subxt/src/backend/mod.rs
```diff
@@ -658,33 +443,48 @@ mod test {
...
+            // Map from storage key to responses, given out in order, when that key is requested.
+            let mut values: HashMap<&str, VecDeque<_>> = HashMap::from_iter([
                 (
-                    "ID2",
-                    Message::Single(Err(RpcError::DisconnectedWillReconnect(
-                        "Reconnecting".to_string(),
-                    ))),
+                    "ID1",
+                    VecDeque::from_iter([
+                        Err(subxt_rpcs::Error::DisconnectedWillReconnect(
```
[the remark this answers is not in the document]
jsdw: Done!

rpcs/src/client/mock_rpc_client.rs
jsdw: Weird that fmt did this or didn't catch it!

rpcs/src/client/jsonrpsee_impl.rs
jsdw: Done!

rpcs/src/client/lightclient_impl.rs
jsdw: Done! Made the code a bit nicer in the end :)

niklasad1 (approved):

PR #1926, opened 2025-02-20
deps: remove polkadot-sdk umbrella crate

### LINKED ISSUE #1925: Reduce reliance on polkadot-sdk
Hi there! Related to something that was mentioned in #1888 :

> Exactly; it's basically an API client to give a nice interface over interacting with the node RPCs at its core, which means we want it to be fairly independant from substrate code for the reasons Niklas mentioned!

We are in the process of upgrading our subxt version to 0.39 to take advantage of loading metadata from wasm (cool feature btw). We noticed that subxt now imports the whole polkadot-sdk by default. This seems to contradict the statements made in the above issue that subxt tries to be independent. In our case, for example, polkadot-sdk brings with it a huge number of additional indirect and unused dependencies - pallets, parachains, consensus etc. - when all we are interested in is core substrate interop. All the pallet-specific bits should come from the metadata, right?

Are there any plans to mitigate this in future (or ways to do it already)? For example I could imagine there being a polkadot-sdk with optional features so that you don't get eg. *all* of the frame pallets by default.

niklasad1:
Close #1925 

Basically for library users of subxt `cargo` can't figure out which features are enabled and ends up fetching all crates
from the polkadot-sdk umbrella which is annoying.

Since we are using quite few of the dependencies from polkadot-sdk anyway the umbrella crate doesn't help that much, so let's get rid of it again.

I liked the idea with the umbrella crate for maintainability for subxt but not worth it.

### DIFF
--- a/metadata/src/utils/validation.rs
+++ b/metadata/src/utils/validation.rs
@@ -12,7 +12,6 @@ use alloc::vec::Vec;
 use hashbrown::HashMap;
 use outer_enum_hashes::OuterEnumHashes;
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, TypeDefVariant, Variant};
-use sp_crypto_hashing;
 
 pub mod outer_enum_hashes;
 
--- a/metadata/src/utils/validation.rs
+++ b/metadata/src/utils/validation.rs
@@ -11,8 +11,8 @@ use crate::{
 use alloc::vec::Vec;
 use hashbrown::HashMap;
 use outer_enum_hashes::OuterEnumHashes;
-use polkadot_sdk::sp_crypto_hashing;
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, TypeDefVariant, Variant};
+use sp_crypto_hashing;
 
 pub mod outer_enum_hashes;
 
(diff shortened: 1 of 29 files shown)


subxt/src/events/events_client.rs
```diff
@@ -5,7 +5,7 @@
 use crate::backend::{Backend, BackendExt, BlockRef};
 use crate::{client::OnlineClientT, error::Error, events::Events, Config};
 use derive_where::derive_where;
-use polkadot_sdk::sp_crypto_hashing;
+use sp_crypto_hashing;
```
niklasad1, line 8:
```suggestion
```

jsdw (approved):

lexnv (approved):

metadata/src/utils/validation.rs
```diff
@@ -11,8 +11,8 @@ use crate::{
 use alloc::vec::Vec;
 use hashbrown::HashMap;
 use outer_enum_hashes::OuterEnumHashes;
-use polkadot_sdk::sp_crypto_hashing;
 use scale_info::{form::PortableForm, Field, PortableRegistry, TypeDef, TypeDefVariant, Variant};
+use sp_crypto_hashing;
```
niklasad1, line 15:
```suggestion
```

PR #1931, opened 2025-02-24
Support constructing and submitting V5 transactions

jsdw:
This PR adds support for constructing v5 bare and general extrinsics, and updates some of the surrounding terminology to match that of polkadot-sdk More concretely:

- Add support for `VerifySignature` transaction extension (this is how a signature is passed in a V5 general extrinsic)
- Make alterations to allow this to be populated as needed, and for the correct signer payload to be constructed (in V5, we _always_ hash the payload, we ignore any extensions prior to/including `VerifySignature` when constructing the payload, and we inject signature into this payload just prior to submitting).
- Simplify setting mortality; no longer need to provide the "from" block details, just the number of blocks to be valid from.
- Change `RefineParams` into `Params` and change interface to `inject_*` to align with how we now inject signature.
- add versioned methods to create extrinsics; `create_v4_partial`, `create_v5_partial`, `create_v4_partial_offline`, `create_v5_partial_offline`, `create_v4_unsigned`, `create_v5_bare`. Add similar to subxt_core (only having versioned methods at that level).
- Signer methods take `T::AccountId` rather than `T::Address` now because this is what V5 requires, and only AccountId was ever useful in v4 anyway. `SignerT::address` is thus not needed and removed. (See https://github.com/paritytech/polkadot-sdk/blob/cf52a0d960a17b4d023bb7c9351e159c60227c85/docs/sdk/src/reference_docs/extrinsic_encoding.rs#L100)
- Renames:  `create_partial_signed` => `create_partial`, `SignedExtension` => `TransactionExtension`, `extra` => `value`, `additional` => `implicit`, `PartialExtrinsic` => `PartialTransaction`, `SubmittableExtrinsic` => `SubmittableTransaction`
- Add tests to create and then decode extrinsics (via frame_decode) to help ensure things are working.

Todo:

- [x] Use frame-decode to check that we can decode constructed transactions properly
- [x] Address vs AccountId; do we need to allow both or support v4 construction as a fallback?


core/src/config/extrinsic_params.rs
```diff
@@ -28,15 +28,82 @@ pub trait ExtrinsicParams<T: Config>: ExtrinsicParamsEncoder + Sized + Send + 's
...
+    /// parameters to some buffer that has been provided. These parameters are
+    /// _not_ sent along with the transaction, but are taken into account when
     /// signing it, meaning the client and node must agree on their values.
-    fn encode_additional_to(&self, _v: &mut Vec<u8>) {}
+    fn encode_implicit_to(&self, _v: &mut Vec<u8>) {}
+
+    /// Set the signature. This happens after we have constructed the extrinsic params,
+    /// and so is defined here rather than on the params, below. We need to use `&dyn Any`
+    /// to keep this trait object safe, but can downcast in the impls.
+    fn inject_signature(&mut self, _account_id: &dyn Any, _signature: &dyn Any) {}
```
niklasad1, line 57: aye, I see the downcast/upcast annoyance now

core/src/tx/mod.rs
```diff
@@ -89,27 +91,70 @@ pub fn validate<Call: Payload>(call: &Call, metadata: &Metadata) -> Result<(), E
...
+    call: &Call,
+    tx_version: u8,
+    metadata: &Metadata,
 ) -> Result<Transaction<T>, Error> {
     // 1. Validate this call against the current node metadata if the call comes
     // with a hash allowing us to do so.
     validate(call, metadata)?;
 
-    // 2. Encode extrinsic
+    // 3. Encode extrinsic
```
niklasad1, line 153:
```suggestion
    // 2. Encode extrinsic
```

revert, looks like it was correct before

metadata/src/from_into/v15.rs
```diff
@@ -330,9 +332,23 @@ mod into_v15 {
...
+            // pick just one of these if converting back to V14/V15 metadata.
+            //
+            // - Picking the largest may mean that older tooling won't be compatible (it may only
+            //   check/support older version).
+            // - Picking the smallest may mean that newer tooling won't work, or newer methods won't
+            //   work.
+            //
+            // Either could make sense, but we keep the oldest to prioritize backward compat with
+            // older tooling.
+            version: *e
```
niklasad1, line 345: does this mean in practice that we will probably just use transaction v4 if we assume both formats are supported by the node we connected to?

core/src/config/extrinsic_params.rs
jsdw: Yup exactly; the impl has access to `T: Config` and the types come from a client with the same config, so all should work, but yeah let's make the warnings way clearer here!

metadata/src/from_into/v15.rs
jsdw: Yup; we pick the lowest version for now because we know it'll work!

(for instance, right now you can submit v5 extrinsics but the `VerifySignature` extension doesn't appear to be used yet anywhere, which means you couldn't sign them, so they would fail anyways)

core/src/config/default_extrinsic_params.rs
```diff
@@ -2,32 +2,32 @@
...
 >;
 
 /// A builder that outputs the set of [`super::ExtrinsicParams::Params`] required for
 /// [`DefaultExtrinsicParams`]. This may expose methods that aren't applicable to the current
 /// chain; such values will simply be ignored if so.
 pub struct DefaultExtrinsicParamsBuilder<T: Config> {
-    /// `None` means the tx will be immortal.
-    mortality: Option<Mortality<T::Hash>>,
+    /// `None` means the tx will be immortal, else it's mortal for N blocks (if possible).
+    mortality: Option<u64>,
```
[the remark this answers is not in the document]
jsdw: I messed with everything a bunch and then realised I had to undo a load of stuff, but wanted to keep not needing the block hash too :)

subxt/src/tx/tx_client.rs
```diff
@@ -50,59 +51,143 @@ impl<T: Config, C: OfflineClientT<T>> TxClient<T, C> {
...
+        })
+    }
+
+    /// Create a partial transaction. Depending on the metadata, we might end up constructing either a v4 or
+    /// v5 transaction. See [`subxt_core::tx`] if you'd like to manually pick the version to construct
     ///
     /// Note: if not provided, the default account nonce will be set to 0 and the default mortality will be _immortal_.
     /// This is because this method runs offline, and so is unable to fetch the data needed for more appropriate values.
-    pub fn create_partial_signed_offline<Call>(
+    pub fn create_partial_offline<Call>(
```
niklasad1, line 115: create_partial_signed_offline removed?

I guess one needs to call sign expliclity on it now?

niklasad1 (approved):
Nice, looks way cleaner than I had mind.

Perhaps just some extra clarification on the method that exposes the dyn Any stuff.

subxt/src/tx/tx_client.rs
jsdw: It's just calleed `create_partial_offline` now :)

The "signed" bit felt confusing because the thing being created is partial but it's definitely not signed yet!

jsdw:
Tests passed but CI issue cleaning caches so merging anyway.

PR #1936, opened 2025-02-28
Document generating interface from Runtime WASM and change feature to `runtime-wasm-path`

jsdw:
Using a Runtime WASM file to generate the Subxt interface was not documented, and the feature flag for it was named `runtime_metadata_path` despite the actual attribute being `runtime_path` (and a `runtime_metadata_path attribute existing already to use metadata), so I added some docs and changed the feature to `runtime_wasm_path` to better correlate with the actual attribute (while making clear it's about the WASM).


niklasad1 (approved):
nice

PR #1940, opened 2025-03-03
Add archive RPCs to subxt-rpcs

jsdw:
- Add support for archive* V2 RPCs in `subxt-rpcs` crate.
- Add basic tests for these.

TODO

- [ ] Check/add archive_storage unsubscribe function to spec so we know what to use


rpcs/src/methods/chain_head.rs
```diff
@@ -872,6 +885,28 @@ pub enum ArchiveCallResult {
     Error(String),
 }
 
+impl ArchiveCallResult {
+    /// Unwrap a successful value from the call, panicking if it was an error.
+    pub fn unwrap_success(self) -> Bytes {
```
niklasad1, line 890: I would prefer `fn as_success(self) -> Option<Bytes>` and let the user to the unwrapping themselves.

Anyway, either is fine

rpcs/src/methods/chain_head.rs
```diff
@@ -295,6 +291,132 @@ impl<T: RpcConfig> ChainHeadRpcMethods<T> {
...
+        function: &str,
+        call_parameters: &[u8],
+    ) -> Result<ArchiveCallResult, Error> {
+        use serde::de::Error as _;
+
+        // We deserialize to this intermediate shape, since
+        // we can't have a boolean tag to denote variants.
+        #[derive(Deserialize)]
+        struct Response {
+            success: bool,
```
niklasad1, line 321: This success flag is annoying, why do we need it at all?

result/value or error should be sufficient?

niklasad1 (approved):

rpcs/src/methods/chain_head.rs
jsdw: Yeah true; I just added it as an extra check to catch any weird responses eg success:false but a value!

rpcs/src/methods/chain_head.rs
jsdw: Good idea; I'll change the APIs to that!

rpcs/src/methods/chain_head.rs
ryanleecode: isn't success part of the spec? whats the issue here

rpcs/src/methods/chain_head.rs
```diff
@@ -295,6 +291,132 @@ impl<T: RpcConfig> ChainHeadRpcMethods<T> {
...
+                query_type: item.query_type,
+            })
+            .collect();
+
+        let sub = self
+            .client
+            .subscribe(
+                "archive_unstable_storage",
+                rpc_params![block_hash, items, child_key.map(to_hex)],
+                "archive_unstable_storage_unsub", // TODO unsub method undefined in spec: look it up/add to spec
```
ryanleecode, line 414: how can archive method have subscription? isn't that a chainhead only thing? B/c ur looking in the past right, so how can it change? whereas chainhead is constantly moving forward so it makes sense to have a subscription?

lexnv (approved):

rpcs/src/methods/chain_head.rs
lexnv: Yep this is quite annoying in rust, IIRC we only left the success flag because it might be easier for js/ts realm interpret the data easier

rpcs/src/methods/chain_head.rs
ryanleecode: So its basically a subscription that will only ever emit 1 value right? Interesting!

rpcs/src/methods/chain_head.rs
niklasad1: aight, I see

PR #1943, opened 2025-03-05
Allow transaction timeout in ChainheadBackend to be configured

### LINKED ISSUE #1942: Configuration of transaction bailout timeout
I am working on fork-aware transaction pool, and I am using subxt in my [transaction test tool](https://github.com/michalkucharczyk/tx-test-tool/), which allows to execute numerous transactions-sending scenarios.

I recently switched to the `ChainHead` backend and I noticed that there is a hardcoded timeout:
https://github.com/paritytech/subxt/blob/2d5a70862a11122bff2134e9a437efc22e41e557/subxt/src/backend/chain_head/mod.rs#L583-L588
which makes some of my tests to exit early.

In some test-cases it is expected that transaction progress stream will live longer then `240s`, e.g.:
- testing heavy loads of higher priority transactions batches (e.g. 100k), while having lower prios already submitted,
- checking correctness of  triggering `FinalityTimeout` when finality is stalled.

Would it be possible to add means to disable and/or configure the value for this timeout?

jsdw:
Rather than having a hardcoded value, let's make this configurable.

Closes #1942

### DIFF
--- a/subxt/src/backend/chain_head/mod.rs
+++ b/subxt/src/backend/chain_head/mod.rs
@@ -47,6 +47,7 @@ pub use subxt_rpcs::methods::chain_head::ChainHeadRpcMethods;
 /// Configure and build an [`ChainHeadBackend`].
 pub struct ChainHeadBackendBuilder<T> {
     max_block_life: usize,
+    transaction_timeout_secs: usize,
     _marker: std::marker::PhantomData<T>,
 }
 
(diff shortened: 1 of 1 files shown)


lexnv (approved):
LGTM! Nice 🙏

jsdw:
The plan is to release 0.41 probably next week, once https://github.com/paritytech/subxt/pull/1931 merges :)

PR #1947, opened 2025-03-06
Wrap the subxt::events::Events type to avoid exposing subxt_core errors and types unnecessarily

jsdw:
This is consistent with the rest of the Subxt interface which hides the subxt_core logic (in most places necessarily because it extends it with functions that can't exist in subxt_core). By wrapping `Events` / `EventDetails` we avoid exposing `subxt_core::Error` from this part of the API in contrast with `subxt::Error` from others.


niklasad1 (approved):

jsdw:
Reverted and opened #1948 instead; forgot to commit a file!

PR #1961, opened 2025-03-12
Improve error message regarding failure to extract metadata from WASM runtime

### LINKED ISSUE #1960: Wasm loader reporting misleading error + doesn't allow allow_missing_host_functions
Hi

I'm using the subxt macro that is configured to load metadata from wasm runtime. However, when my runtime is not built with `std` feature I get an error `method Metadata_metadata_at_version doesnt exist`, which is not true because my runtime has Metadata_metadata_at_version function (confirmed through `wasm2wat` tool).

I forked subxt locally and changed the [`WasmLoader`](https://github.com/paritytech/subxt/blob/master/macro/src/wasm_loader.rs#L139) line where it reports the error to print the actual error and not ignore it:
```
.map_err(|e| {
                CodegenError::Wasm(
                    format!("{:?}", e).to_owned(),
                )
            })?;
``` 

Then I get the actual error :
```
Other("Failed to create module: Other error happened while constructing the runtime: runtime requires function imports which are not present on the host: 'env:ext_benchmarking_set_whitelist_version_1', 'env:ext_benchmarking_reset_read_write_count_version_1', 'env:ext_benchmarking_get_read_and_written_keys_version_1', 'env:ext_benchmarking_wipe_db_version_1', 'env:ext_benchmarking_current_time_version_1', 'env:ext_benchmarking_read_write_count_version_1', 'env:ext_benchmarking_proof_size_version_1', 'env:ext_benchmarking_commit_db_version_1', 'env:ext_benchmarking_add_to_whitelist_version_1'")
```
I guess the error is related to the `WasmExecutor` in the `wasm_loader` not allowing missing host functions.
Indeed, my issue is fixed if by setting the `allow_missing_host_functions` param of the `uncahed_call` call function in `load_metadata_at_version`. Is it possible to:

1. Change the `load_metadata_at_version` in the `wasm_loader` to allow for missing host functions.
2. Change the `load_metadata_at_version` to properly report the error and not misleading function doesn exist error.

Many thanks

jsdw:
Closes #1960

### DIFF
--- a/macro/src/wasm_loader.rs
+++ b/macro/src/wasm_loader.rs
@@ -117,8 +117,10 @@ impl Executor {
                 "Metadata_metadata",
                 &[],
             )
-            .map_err(|_| {
-                CodegenError::Wasm("method \"Metadata_metadata\" doesnt exist".to_owned())
+            .map_err(|e| {
+                CodegenError::Wasm(format!(
+                    "Failed to call \"Metadata_metadata\" on WASM runtime. Cause: {e}"
+                ))
             })?;
         let encoded_metadata =
             <Vec<u8>>::decode(&mut &encoded_metadata[..]).map_err(CodegenError::Decode)?;
--- a/macro/src/wasm_loader.rs
+++ b/macro/src/wasm_loader.rs
@@ -135,10 +135,10 @@ impl Executor {
                 "Metadata_metadata_at_version",
                 &version.encode(),
             )
-            .map_err(|_| {
-                CodegenError::Wasm(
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

nmammeri:
Sorry to interfere but while here, it might be beneficial to also improve error reporting on [load_legacy_metadata](https://github.com/paritytech/subxt/blob/20bc4b89ca7f506323d0f8ae9e8faf0a91baae44/macro/src/wasm_loader.rs#L110C8-L110C28)

PR #1962, opened 2025-03-12
Allow submitting transactions ignoring follow events

### LINKED ISSUE #1956: Allow submitting transactions via `ChainHeadBackend` to _not_ track follow events
Currently, when a transaction is submitted via `submit_transaction()` on the `ChainHeadBackend`, we connect transaction events with follow events to attempt to have valid `BlockRef`s for any block mentioned in the transaction events. This helps to guarantee that information about the block hashes handed back via transaction events (finalized blocks in particular) is instantly accessible.

In some cases, this might cause an issue:
1. Finalization is lagging on the chain and so we end up waiting so long for a finalized block event that the transaction submission times out, or the `chainHead_follow` subscription emits a `stop` event because we're holding onto too many old blocks while waiting to see a finalized one.
2. We want to return transaction submission events ASAP and want to avoid any delay in linking them with follow event hashes.
3. We simply want to eliminate the possibilty that any `chainHead_follow` "stop" event will hamper our ability to submit and monitor the submission of transactions through to completion.

To do this, we could add a config option to `ChainHeadBackendBuilder` which disables the linking, and documents the perils of doing so. Notably some of the "default" subxt APis may sporadically fail (eg `sign_and_submit_then_watch`).

I'm not completely sure whether we should do this yet, but worth a thought!

jsdw:
Closes #1956

### DIFF
--- a/subxt/src/backend/chain_head/mod.rs
+++ b/subxt/src/backend/chain_head/mod.rs
@@ -48,6 +48,7 @@ pub use subxt_rpcs::methods::chain_head::ChainHeadRpcMethods;
 pub struct ChainHeadBackendBuilder<T> {
     max_block_life: usize,
     transaction_timeout_secs: usize,
+    submit_transactions_ignoring_follow_events: bool,
     _marker: std::marker::PhantomData<T>,
 }
 
@@ -63,6 +64,7 @@ impl<T: Config> ChainHeadBackendBuilder<T> {
         Self {
             max_block_life: usize::MAX,
             transaction_timeout_secs: 240,
+            submit_transactions_ignoring_follow_events: false,
             _marker: std::marker::PhantomData,
         }
     }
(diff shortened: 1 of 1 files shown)


subxt/src/backend/chain_head/mod.rs
```diff
@@ -558,170 +588,230 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for ChainHeadBackend<T> {
...
+            submit_transaction_ignoring_follow_events(extrinsic, &self.methods).await
+        } else {
+            submit_transaction_tracking_follow_events::<T>(
+                extrinsic,
+                self.transaction_timeout_secs as u64,
+                &self.methods,
+                &self.follow_handle,
+            )
+            .await
+        }
```
jsdw, line 814: I just moved the existing logic into `submit_transaction_tracking_follow_events` and added an alternative path which is muc hsimpler for ignoring them

subxt/src/backend/chain_head/mod.rs
```diff
@@ -84,11 +86,35 @@ impl<T: Config> ChainHeadBackendBuilder<T> {
...
+    ///
+    /// Calling this method turns off this synchronization, speeding up the response and removing any reliance
+    /// on the `chainHead_follow` subscription continuing to run without stopping throughout submitting a transaction.
+    ///
+    /// # Warning
+    ///
+    /// This can lead to errors when calling APIs like `wait_for_finalized_success`, which will try to retrieve events
+    /// at the finalized block, because there will be a race and the finalized block may not be available for querying
+    /// yet.
+    pub fn submit_transactions_ignoring_follow_events(mut self) -> Self {
```
niklasad1, line 113: a little long name :)

I don't have a better suggestion in mind though

niklasad1 (approved):

subxt/src/backend/chain_head/mod.rs
jsdw: Yeah; it's quite long but also quite a rare thing to set I would hope, so I don't mind I guess!

michalkucharczyk:
Thank you!

PR #1967, opened 2025-03-19
Support V16 metadata and refactor metadata code

jsdw:
This PR:

- Adds support for some v16 metadata features and converting from v16 metadata (currently unstable); notably pallet view functions and pallet associated types (Future PRs will make use of this new information).
- Removes the ability to convert back from `subxt::Metadata` to `frame_metadata`. Our internal `subxt::Metadata` doesn't care about everything in the metadata, and not mandating that it be convertible back into `frame_metadata` makes it a non-issue to throw away any information that we don't care about.
- Moves metadata stripping off `Metadata` (since you can't encode it any more) and onto the individual `frame_metadata` versions that we've implemented it for. This was necessary given the above, and means that when we strip metadata for a particular version, we won't change anything else about it.
- Nerfs metadata stripping in order to avoid potential issues and reduce code complexity. Metadata stripping would achieve the biggest size reduction by also stripping variants from the "outer enums" (`RuntimeCall`/`RuntimeError`/`RuntimeEvent`), but it's hard to figure out:
    
    1. when we can actually strip these without introducing bugs, and
    2. how to compare metadata which may or may not have had these variants stripped, with a "normal" metadata (ie for metadata validation).
    
    Our current implementation had some potantial issues around (2), and a bunch of complexity to address (1), so we no longer strip variants. The downside is that metadata stripping won't reduce the size of the metadata much now (the outer enums contain a lot of type information which we keep regardless), but it can still help to exclude APIs you don't intend to work with.
- Moves metadata stripping to a new utils crate, reworks it to (hopefully) be much easier to follow and adds some (more) tests to check that it is working as intended.

To illustrate the nerfing of size reduction:

- Using this PR and Substrate node:
    - `subxt metadata`: 681K
    - `subxt metadata --runtime-apis ''`: 651K
    - `subxt metadata --pallets System,Balances,Contracts,Council,Democracy --runtime-apis ''`: 518K
    - `subxt metadata --pallets Balances --runtime-apis ''`: 501K
    - `subxt metadata --pallets '' --runtime-apis ''`: 495K
- Current `master`:
    - `subxt metadata`: 681K
    - `subxt metadata --runtime-apis ''`: 651K
    - `subxt metadata --pallets System,Balances,Contracts,Council,Democracy --runtime-apis ''`: 459K
    - `subxt metadata --pallets Balances --runtime-apis ''`: 334K
    - `subxt metadata --pallets '' --runtime-apis ''`: 324K

YMMV depending on the node configuration. I thought I'd seen much smaller sizes on `master` but perhaps that was on a different chain.


cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+
+use frame_metadata::{ v14, v15, v16 };
+use scale_info::PortableRegistry;
+use subxt_metadata::utilities::OuterEnums;
+
+/// This trait is implemented for metadata versions to enable us to strip pallets and runtime APIs from them.
+/// 
+/// To implement the [`RetainMetadata::retain_metadata`] method for a new metadata version, you'll probably:
+/// - Remove any pallets and runtime APIs from the metadata based on the filter functions.
+/// - Gather the outer enum IDs and then call [`retain_relevant_types_in_registry`] to strip the
```
niklasad1, line 15:
```suggestion
/// - Gather the outer enum type IDs and then call [`retain_relevant_types_in_registry`] to strip the
```

cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+            event_ty: Some(self.outer_enums.event_enum_ty.id),
+            error_ty: Some(self.outer_enums.error_enum_ty.id),
+        };
+        retain_relevant_types_in_registry(self, keep_pallet, outer_enum_ids);
+    }
+}
+
+/// This function:
+/// 
+/// - Iterates deeply over all IDs that matter when considering whether we can strip variants from
```
niklasad1, line 110:
```suggestion
/// - Iterates deeply over all type IDs that matter when considering whether we can strip variants from
```

cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+            self.extrinsic.signature_ty.id
+        ];
+
+        pallet_types.chain(runtime_api_inputs).chain(extrinsic_type_ids)
+    }
+}
+
+/// This is implemented for any metadata version that a [`PortableRegistry`] can be obtained from.
+trait GetPortableRegistry {
+    fn get_portable_registry(&self) -> &PortableRegistry;
```
niklasad1, line 489: I think get and get_mut is sufficient here, the trait is called GetPortableRegistry ^

cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+/// This deeply recurses over any IDs given, handing back any child IDs seen in addition to the original IDs.
+fn recurse_type_ids<'a, I: Iterator<Item = u32> + 'a>(types: &'a PortableRegistry, ids: I) -> impl Iterator<Item = u32> + 'a {
+    ids.flat_map(|id| {
+        use scale_info::TypeDef;
+        let ty = &types.resolve(id).unwrap();
+
+        let params = ty.type_params.iter().filter_map(|p| p.ty.map(|t| t.id));
+        let inner_ids = match &ty.type_def {
+            TypeDef::Composite(c) => {
+                Either6::A(c.fields.iter().map(|field| field.ty.id))
```
niklasad1, line 529: :rofl:

cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+/// strip metadata if they want to decode extrinsics freom _any_ pallet.
+/// 
+/// When implementing this, if you're ever not sure whether a type ID should be considered for
+/// stripping or not, always assume that it is to play it safe. If _everything_ is in 
+/// [`IterateTypeIds::iterate_type_ids_to_consider_for_stripping`], then things will still work
+/// (we'll just end up stripping fewer things from the metadata).
+trait IterateTypeIds {
+    /// This should return an iterator over all type IDs which, if seen, should prevent us
+    /// from stripping variants from RuntimeCall/Error/Event enums ("outer enums").
+    fn iterate_type_ids_to_consider_for_stripping(&self) -> impl Iterator<Item=u32>;
```
niklasad1, line 190: I don't think you need to repeat the type ids for every method here

I think `iter_for_stripping`, `iter_not_stripping` and `iter_all/iter` would be sufficient but your call :)

cli/src/commands/metadata/retain.rs
```diff
@@ -0,0 +1,603 @@
...
+
+/// This trait is implemented for metadata versions to enable us to strip pallets and runtime APIs from them.
+/// 
+/// To implement the [`RetainMetadata::retain_metadata`] method for a new metadata version, you'll probably:
+/// - Remove any pallets and runtime APIs from the metadata based on the filter functions.
+/// - Gather the outer enum IDs and then call [`retain_relevant_types_in_registry`] to strip the
+///   type registry down given the pallets and runtime APIs left in the metadata.
+/// - This will require implementing [`IterateTypeIds`], which is the thing that determines which
+///   IDs in the metadata to pay attention to or ignore when it comes to stripping things out.
+pub trait RetainMetadata {
```
niklasad1, line 19: Looks clean

metadata/src/from_into/v14.rs
niklasad1, line 1: :muscle:

cli/src/commands/metadata/retain.rs
jsdw: Chanegd to `iter_type_ids_mut` (I realised it needed to be mutable for updating the IDs after)

core/src/config/transaction_extensions.rs
```diff
@@ -582,12 +582,13 @@ macro_rules! impl_tuples {
...
                 let metadata = &client.metadata;
                 let types = metadata.types();
+                let tx_extension_version = metadata.extrinsic().transaction_extension_version_to_use_for_encoding();
 
                 // For each transaction extension in the tuple, find the matching index in the metadata, if
                 // there is one, and add it to a map with that index as the key.
                 let mut exts_by_index = HashMap::new();
                 $({
-                    for (idx, e) in metadata.extrinsic().transaction_extensions().iter().enumerate() {
+                    for (idx, e) in metadata.extrinsic().transaction_extensions_by_version(tx_extension_version).unwrap().enumerate() {
```
niklasad1, line 591: Maybe add an expect here or explain that this may fail if an `invalid transaction version`?

Or add some docs on the macro itself...

core/src/config/transaction_extensions.rs
```diff
@@ -604,7 +605,7 @@ macro_rules! impl_tuples {
 
                 // Next, turn these into an ordered vec, erroring if we haven't matched on any exts yet.
                 let mut params = Vec::new();
-                for (idx, e) in metadata.extrinsic().transaction_extensions().iter().enumerate() {
+                for (idx, e) in metadata.extrinsic().transaction_extensions_by_version(tx_extension_version).unwrap().enumerate() {
```
niklasad1, line 608: Same here

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,318 @@
...
+    metadata: &mut v14::RuntimeMetadataV14,
+    path_segments: Vec<String>,
+) -> u32 {
+    let variants: Vec<_> = metadata
+        .pallets
+        .iter()
+        .filter_map(|pallet| {
+            let error = pallet.error.as_ref()?;
+
+            // Note:  using the `alloc::format!` macro like in `let path = format!("{}Error", pallet.name);`
```
niklasad1, line 241: hmm interesting

jsdw: I wonder if that's still true; it's something Tadeo/Pavlo ran into but was very strange. Let me see if simplifying works

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,313 @@
...
+    value: v14::ExtrinsicMetadata<PortableForm>,
+    missing_ids: MissingExtrinsicTypeIds,
+) -> ExtrinsicMetadata {
+    let transaction_extensions: Vec<_> = value
+        .signed_extensions
+        .into_iter()
+        .map(from_signed_extension_metadata)
+        .collect();
+
+    let transaction_extension_indexes = (0..transaction_extensions.len() as u32).collect();
```
niklasad1, line 124: I wonder whether we can replace Vec<u32> with Range<u32> as transaction_extensions_indexes instead to get rid off some alloc?

metadata/src/lib.rs
```diff
@@ -593,74 +605,93 @@ impl ConstantMetadata {
...
+
+        Some(iter)
     }
 
-    /// Which version are these transaction extensions?
-    pub fn transaction_extensions_version(&self) -> u8 {
-        self.transaction_extensions_version
+    /// When constructing a v5 extrinsic, use this transaction extensions version.
+    pub fn transaction_extension_version_to_use_for_encoding(&self) -> u8 {
+        *self.transaction_extensions_by_version.keys().max().unwrap()
```
niklasad1, line 654: Silly comment: maybe add an expect that `transaction_extensions_by_version` must contain at least one version

testing/integration-tests/src/full_client/metadata_validation.rs
```diff
@@ -27,14 +27,14 @@ fn v15_to_metadata(v15: RuntimeMetadataV15) -> Metadata {
...
 
-fn modified_metadata<F>(metadata: Metadata, f: F) -> Metadata
-where
-    F: FnOnce(&mut RuntimeMetadataV15),
-{
-    let mut metadata = RuntimeMetadataV15::from((*metadata).clone());
-    f(&mut metadata);
-    v15_to_metadata(metadata)
-}
+// fn modified_metadata<F>(metadata: Metadata, f: F) -> Metadata
```
niklasad1, line 30: remove?

testing/integration-tests/src/full_client/metadata_validation.rs
```diff
@@ -97,24 +97,8 @@ fn pallets_to_metadata(pallets: Vec<PalletMetadata>) -> Metadata {
...
-    let metadata = RuntimeMetadataV15::from((*api.metadata()).clone());
-    let metadata = v15_to_metadata(metadata);
-
-    assert!(
-        node_runtime::is_codegen_valid_for(&metadata),
-        "Should still be valid after conversion back and forth"
-    );
-}
+/*
+TODO: Add a means to get RuntimeMetadataV16 from api, and then modify that, convert to Metadata and check validity
```
niklasad1, line 101: are you planning to fix this in this PR?

testing/ui-tests/src/utils/metadata_test_runner.rs
```diff
@@ -100,15 +100,10 @@ impl MetadataTestRunnerCaseBuilder {
...
-    where
-        M: TryInto<Metadata>,
-        M::Error: std::fmt::Debug,
-    {
-        let macro_metadata = macro_metadata.try_into().expect("can into Metadata");
+    pub fn build(self, macro_metadata: frame_metadata::RuntimeMetadataPrefixed) -> String {
         let validation_metadata = self
             .validation_metadata
-            .unwrap_or_else(|| macro_metadata.clone());
+            .unwrap_or_else(|| clone_via_encode(&macro_metadata));
```
niklasad1, line 106: was this taken care of the prior `Into/From` or why is it needed?

utils/strip-metadata/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition.workspace = true
+rust-version.workspace = true
+publish = true
+autotests = false
+
+license.workspace = true
+repository.workspace = true
+documentation.workspace = true
+homepage.workspace = true
+description = "subxt utils strip metadata"
```
niklasad1, line 14:
```suggestion
description = "subxt utils to strip metadata"
```

metadata/src/from/v14.rs
jsdw: In this case we could, but for v16 metadata, the indexes vec might no longer be contiguous if we have several versions of transaction extensions defined.

metadata/src/lib.rs
jsdw: Done!

metadata/src/from/v14.rs
niklasad1: I see

utils/strip-metadata/src/lib.rs
```diff
@@ -0,0 +1,946 @@
...
+    let keep_these_ids: BTreeSet<u32> = m
+        .iter_type_ids_mut()
+        .map(|id| *id)
+        .chain(Some(dispatch_err_type_id))
+        .collect();
+
+    let new_ids = m.get_types_mut().retain(|id| keep_these_ids.contains(&id));
+
+    // Map IDs found in the metadata to new ones as needed after the retaining:
+    m.iter_type_ids_mut().for_each(|id| {
```
niklasad1, line 100: could be map_while :)

metadata/src/utils/outer_enums.rs
```diff
@@ -0,0 +1,66 @@
+// Copyright 2019-2025 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or GPL-3.0.
+// see LICENSE for license details.
+
+//! Utility functions for working with v14 metadata.
```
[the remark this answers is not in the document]
jsdw: Actually I can remove this entirely and just have it in from/v14 now :)

niklasad1 (approved):
Nice work.

LGTM, I think this PR is improvement when it comes readability of the code.

Would be awesome with empirical metrics how much the codesize would increase w.r.t trimming.

testing/ui-tests/src/utils/metadata_test_runner.rs
jsdw: This is because we were previously accepting `subxt::Metadata`, which can be cloned, but now we accept `frame_metadata::RuntimeMetadataPrefixed`, which for some reason doesn't implement Clone (despite the individual versions implementing it iirc).

We can prob improve that, but until then, this is a bit of a hack to make a clone of it by encoding and then decoding again 😆

utils/strip-metadata/src/lib.rs
jsdw: Ooh I haven't used that one before!

However, it appears like it stops after it sees the first `None`, whereas I want to iterate the entire set of IDs either way :)

testing/integration-tests/src/full_client/metadata_validation.rs
jsdw: Fixed :)

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,351 @@
...
+            value,
+        } => StorageEntryType::Map {
+            hashers: hashers.into_iter().map(from_storage_hasher).collect(),
+            key_ty: key.id,
+            value_ty: value.id,
+        },
+    }
+}
+
+fn from_storage_entry_modifier(value: v14::StorageEntryModifier) -> StorageEntryModifier {
```
pkhry, line 165: dq: why not `From<_>` instance instead?

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,351 @@
...
+        v14::StorageHasher::Blake2_256 => StorageHasher::Blake2_256,
+        v14::StorageHasher::Blake2_128Concat => StorageHasher::Blake2_128Concat,
+        v14::StorageHasher::Twox128 => StorageHasher::Twox128,
+        v14::StorageHasher::Twox256 => StorageHasher::Twox256,
+        v14::StorageHasher::Twox64Concat => StorageHasher::Twox64Concat,
+        v14::StorageHasher::Identity => StorageHasher::Identity,
+    }
+}
+
+fn from_storage_entry_type(value: v14::StorageEntryType<PortableForm>) -> StorageEntryType {
```
pkhry, line 150: dq: why not `From<_>` instance?

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,351 @@
...
+        address_ty: missing_ids.address,
+        signature_ty: missing_ids.signature,
+        transaction_extensions_by_version: BTreeMap::from_iter([(
+            0,
+            transaction_extension_indexes,
+        )]),
+    }
+}
+
+fn from_storage_hasher(value: v14::StorageHasher) -> StorageHasher {
```
pkhry, line 138: dq: why not From<_> instance?

metadata/src/from/v14.rs
```diff
@@ -0,0 +1,351 @@
...
+            },
+            apis: Default::default(),
+            custom: CustomMetadataInner {
+                map: Default::default(),
+            },
+        })
+    }
+}
+
+fn from_signed_extension_metadata(
```
pkhry, line 104: dq: why not From<_> instance?

metadata/src/from/v15.rs
```diff
@@ -0,0 +1,233 @@
...
+                call_enum_ty: m.outer_enums.call_enum_ty.id,
+                event_enum_ty: m.outer_enums.event_enum_ty.id,
+                error_enum_ty: m.outer_enums.error_enum_ty.id,
+            },
+            custom: m.custom,
+        })
+    }
+}
+
+fn from_signed_extension_metadata(
```
pkhry, line 102: same dq about `From` instances instead for this conversion and those below

metadata/src/from/v16.rs
```diff
@@ -0,0 +1,260 @@
...
+                call_enum_ty: m.outer_enums.call_enum_ty.id,
+                event_enum_ty: m.outer_enums.event_enum_ty.id,
+                error_enum_ty: m.outer_enums.error_enum_ty.id,
+            },
+            custom: v15::CustomMetadata { map: custom_map },
+        })
+    }
+}
+
+fn from_transaction_extension_metadata(
```
pkhry, line 123: same dq about `From` instances

utils/strip-metadata/src/lib.rs
pkhry, line 100: nit: maybe `filter_map().for_each()` ?

utils/strip-metadata/src/lib.rs
```diff
@@ -0,0 +1,946 @@
...
+                match self {
+                    $name::$first_tok(item) => item.next(),
+                    $($name::$tok(item) => item.next(),)*
+                }
+            }
+        }
+    }
+}
+
+either!(Either: A B);
```
pkhry, line 435: maybe more conventional `Left` and `Right` , also why not use `either` that is already a dependency? is it because of finnicky `Iter` trait defs on it?

pkhry (approved):

metadata/src/from/v16.rs
jsdw: I'd write these as `From` impls if I wanted it a part of a public interface, but I suppose I didn't want that sort of interface so it was easier to just write freestanding functions to do what was needed :)

utils/strip-metadata/src/lib.rs
jsdw: Good point! This is a bit of a vestige from when I had a couple of Either defs (which is also why not `Left` and `Right`, because I had an `Either6` with 6 branches, so the `Either` aligned with using letters instead!)

I deleted this and justu sed the `either ` crate!

utils/strip-metadata/src/lib.rs
jsdw: I went the other way and simplified to a normal for loop:

```rust
for id in m.iter_type_ids_mut() {
    if let Some(new_id) = new_ids.get(id) {
        *id = *new_id;
    };
}
```

filter_map.for_each would have felt a bit ugly I think ie something like:

```rust
m.iter_type_ids_mut()
 .filter_map(|id| new_ids.get(id).map(|new_id| (id, new_id))
 .for_each(|(id, new_id)| *id = new_id)
```

PR #1974, opened 2025-03-26
Metadata V16: Be more dynamic over which hasher is used.

jsdw:
This PR takes a small step toward using the associated type information present in V16 metadata (still unstable) to simplify the Subxt config. It:

- Sets the default `Hasher` type to be `DyanmicHasher256`, which will hash things using `BlakeTwo256` or `Keccak256` (falling back to the former if no information available) depending on what the chain is configured with.
- Removes the `Hash` associated type (since it's defined in `Hasher` anyway), just to simplify Config a little.


lexnv (approved):

core/src/config/polkadot.rs
```diff
@@ -17,7 +17,6 @@ pub use primitive_types::{H256, U256};
 pub enum PolkadotConfig {}
 
 impl Config for PolkadotConfig {
-    type Hash = <SubstrateConfig as Config>::Hash;
```
niklasad1, line 20: woo 👍

subxt/src/backend/chain_head/mod.rs
```diff
@@ -543,7 +543,8 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for ChainHeadBackend<T> {
 
     async fn stream_all_block_headers(
         &self,
-    ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
+        _hasher: T::Hasher,
```
niklasad1, line 546: Why do we need pass in the hasher here?

It's already part of T: Config?

subxt/src/backend/chain_head/mod.rs
```diff
@@ -560,7 +561,8 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for ChainHeadBackend<T> {
 
     async fn stream_best_block_headers(
         &self,
-    ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
+        _hasher: T::Hasher,
```
niklasad1, line 564: same q here ☝️

subxt/src/backend/chain_head/mod.rs
```diff
@@ -575,7 +577,8 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for ChainHeadBackend<T> {
 
     async fn stream_finalized_block_headers(
         &self,
-    ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
+        _hasher: T::Hasher,
```
niklasad1, line 580: same q here ☝️

subxt/src/backend/legacy.rs
```diff
@@ -270,16 +273,16 @@ impl<T: Config + Send + Sync + 'static> Backend<T> for LegacyBackend<T> {
 
     async fn stream_all_block_headers(
         &self,
-    ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error> {
+        hasher: T::Hasher,
```
niklasad1, line 276: same q here ☝️

subxt/src/backend/mod.rs
```diff
@@ -123,30 +123,33 @@ pub trait Backend<T: Config>: sealed::Sealed + Send + Sync + 'static {
     /// A stream of all new block headers as they arrive.
     async fn stream_all_block_headers(
         &self,
-    ) -> Result<StreamOfResults<(T::Header, BlockRef<T::Hash>)>, Error>;
+        hasher: T::Hasher,
```
niklasad1, line 126: explain why hasher is needed here

testing/wasm-rpc-tests/tests/wasm.rs
```diff
@@ -31,8 +31,9 @@ async fn wasm_ws_transport_works() {
     let client = subxt::client::OnlineClient::<SubstrateConfig>::from_url("ws://127.0.0.1:9944")
         .await
         .unwrap();
+    let hasher = client.hasher();
 
-    let mut stream = client.backend().stream_best_block_headers().await.unwrap();
+    let mut stream = client.backend().stream_best_block_headers(hasher).await.unwrap();
```
niklasad1, line 36: I see

niklasad1 (approved):
Nice improvement for sure 🚀

subxt/src/backend/chain_head/mod.rs
jsdw: The hasher interface changed so that you can instantiate a hasher with some metadata (allowing it to decide how to hash based on the metadata). Because of this, we now need to construct a hasher to use (rather than use a static method on an empty hasher type), so we pass this hasher around in order to avoid constructing it as much as possible

core/src/config/mod.rs
```diff
@@ -97,15 +98,18 @@ impl<T> BlockHash for T where
...
     /// The type given back from the hash operation
-    type Output;
+    type Output: Hash;
+
+    /// Construct a new hasher.
+    fn new(metadata: &Metadata) -> Self;
 
     /// Hash some bytes to the given output type.
-    fn hash(s: &[u8]) -> Self::Output;
+    fn hash(&self, s: &[u8]) -> Self::Output;
```
jsdw, line 108: @niklasad1 this is the important change; we can now construct a hasher given metadata (which might be expensive), and then `hash` takes `&self` now so we can hash based on this metadata. This is why we pass a hasher around more. We _could_ construct it any place we have access to metadata, but the point is to avoid this because it could be expensive

core/src/config/substrate.rs
```diff
@@ -39,17 +39,73 @@ pub type SubstrateExtrinsicParams<T> = DefaultExtrinsicParams<T>;
...
+        let Some(hash_ty_id) = system_pallet.associated_type_id("Hashing") else {
+            return Self(HashType::Unknown);
+        };
+
+        let ty = metadata
+            .types()
+            .resolve(hash_ty_id)
+            .expect("Type information for 'Hashing' associated type should be in metadata");
+
+        let hash_type = match ty.path.ident().as_deref().unwrap_or("") {
```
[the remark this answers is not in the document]
jsdw: Yeah; right now we don't have V16 metadata, and so the hasher type will be unknown. We default this back to being BlakeTwo256 for now for backward compat. Worst case, things work exactly as before. BEst case, when V16 metadata is used it can use Keccak when needed!

PR #1977, opened 2025-04-01
Support v1 archive RPCs

jsdw:
Now that we've stabilized the archive RPCs, add support for v1 archive RPCs in subxt-rpcs.

We keep the `archive_unstable` methods as copies of the v1 ones which do the same thing, but when the v1 ones have been around a while we can delete them.

### DIFF
--- a/rpcs/src/methods/chain_head.rs
+++ b/rpcs/src/methods/chain_head.rs
@@ -296,10 +296,7 @@ impl<T: RpcConfig> ChainHeadRpcMethods<T> {
     ///
     /// Returns an array of the hexadecimal-encoded scale-encoded extrinsics found in the block,
     /// or `None` if the block wasn't found.
-    pub async fn archive_v1_body(
-        &self,
-        block_hash: T::Hash,
-    ) -> Result<Option<Vec<Bytes>>, Error> {
+    pub async fn archive_v1_body(&self, block_hash: T::Hash) -> Result<Option<Vec<Bytes>>, Error> {
         self.client
             .request("archive_v1_body", rpc_params![block_hash])
             .await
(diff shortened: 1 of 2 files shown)


lexnv (approved):

lexnv:
Nice! Thanks for stabilizing and implementing these! 🙏

rpcs/src/methods/chain_head.rs
```diff
@@ -292,6 +292,128 @@ impl<T: RpcConfig> ChainHeadRpcMethods<T> {
...
+    /// Returns an array of the hexadecimal-encoded scale-encoded extrinsics found in the block,
+    /// or `None` if the block wasn't found.
+    pub async fn archive_v1_body(&self, block_hash: T::Hash) -> Result<Option<Vec<Bytes>>, Error> {
+        self.client
+            .request("archive_v1_body", rpc_params![block_hash])
+            .await
+    }
+
+    /// Call the `archive_v1_call` method and return the response.
+    pub async fn archive_v1_call(
```
niklasad1, line 306: hmm, we could perhaps have a shared implementation for this for both `archive_v1_call` and `archive_unstable_call` but doesn't matter that much since we will remove the unstable stuff eventually :)

rpcs/src/methods/chain_head.rs
```diff
@@ -292,6 +292,128 @@ impl<T: RpcConfig> ChainHeadRpcMethods<T> {
...
+            )
+            .await?;
+
+        Ok(ArchiveStorageSubscription { sub, done: false })
+    }
+
+    // Dev note: we continue to support the latest "unstable" archive methods because
+    // they will be around for a while before the stable ones make it into a release.
+    // The below are just a copy-paste of the v1 methods, above, but calling the
+    // "unstable" RPCs instead. Eventually we'll remove them.
```
niklasad1, line 415: maybe open a tracking issue for that so we don't forget

niklasad1 (approved):

rpcs/src/methods/chain_head.rs
jsdw: Yeah, I contemplated that, but then it'd make it messier to remove them, whereas at the mo we can just delete the unstable ones nice and cleanly :)

PR #1979, opened 2025-04-01
Add DefaultParams to allow more transaction extensions to be used when calling _default() methods

### LINKED ISSUE #1975: `signed_extensions::AnyOf` don't implement default for more than 12 extensions.
In `signed_extensions::AnyOf` using default transaction extension is practical, but if there is more than 12 extensions then the trait is no longer implemented because rust only support up to 12 element in their implementation of `Default` for tuple.

Maybe we shouldn't use `Default` trait and instead a new method trait or method like `SignedExtensionDefault`?

jsdw:
We currently rely on extrinsic parameters implementing `Default` on these two methods of `TxClient`:

```rust
    pub async fn sign_and_submit_then_watch_default<Call, Signer>(
        &mut self,
        call: &Call,
        signer: &Signer,
    ) -> Result<TxProgress<T, C>, Error>
    where
        Call: Payload,
        Signer: SignerT<T>,
        <T::ExtrinsicParams as ExtrinsicParams<T>>::Params: Default, // <-- here
    {
        self.sign_and_submit_then_watch(call, signer, DefaultParams::default_params())
            .await
    }

    pub async fn sign_and_submit_default<Call, Signer>(
        &mut self,
        call: &Call,
        signer: &Signer,
    ) -> Result<T::Hash, Error>
    where
        Call: Payload,
        Signer: SignerT<T>,
        <T::ExtrinsicParams as ExtrinsicParams<T>>::Params: Default, // <-- here
    {
        self.sign_and_submit(call, signer, DefaultParams::default_params())
            .await
    }
```

Extrinsic params are, by default, a tuple containing an entry for each transaction extension configured via `AnyOf`. Our default transaction extensions are this:

```rust
transaction_extensions::AnyOf<
    T,
    (
        transaction_extensions::VerifySignature<T>,
        transaction_extensions::CheckSpecVersion,
        transaction_extensions::CheckTxVersion,
        transaction_extensions::CheckNonce,
        transaction_extensions::CheckGenesis<T>,
        transaction_extensions::CheckMortality<T>,
        transaction_extensions::ChargeAssetTxPayment<T>,
        transaction_extensions::ChargeTransactionPayment,
        transaction_extensions::CheckMetadataHash,
    ),
>;
```

And `AnyOf` is implemented such that the parameters to configure each extension are expected in a tuple with a size matching the extensions above. 

We currently have support for 9 default transaction extensions above, but if this number were to exceed 12 (which is the max tuple size that Rust implements `Default` for), then we could no longer use `AnyOf` in conjunction with our `_default` methods, since the parameters would no longer implement `Default`.

So, we add a `DefaultParams` trait instead, which supports up to 26 entries in the tuple (as long as each of those impls `Default`), matching the number of tuple entries `AnyOf` supports.

The downside is that a user going off the beaten path will need to implement `DefaultParams` instead of `Default` for their own params if they want to use the `_default` methods, but this is likely to be a rare case. 

Closes #1975.

### DIFF
--- a/core/src/config/extrinsic_params.rs
+++ b/core/src/config/extrinsic_params.rs
@@ -113,6 +113,11 @@ const _: () = {
     impl_tuples!(A 0, B 1, C 2, D 3, E 4, F 5, G 6, H 7, I 8, J 9, K 10, L 11, M 12, N 13, O 14, P 15, Q 16);
     impl_tuples!(A 0, B 1, C 2, D 3, E 4, F 5, G 6, H 7, I 8, J 9, K 10, L 11, M 12, N 13, O 14, P 15, Q 16, R 17);
     impl_tuples!(A 0, B 1, C 2, D 3, E 4, F 5, G 6, H 7, I 8, J 9, K 10, L 11, M 12, N 13, O 14, P 15, Q 16, R 17, S 18);
-    impl_tuples!(A 0, B 1, C 2, D 3, E 4, F 5, G 6, H 7, I 8, J 9, K 10, L 11, M 12, N 13, O 14, P 15, Q 16, R 17, S 18, U 19);
-    impl_tuples!(A 0, B 1, C 2, D 3, E 4, F 5, G 6, H 7, I 8, J 9, K 10, L 11, M 12, N 13, O 14, P 15, Q 16, R 17, S 18, U 19, V 20);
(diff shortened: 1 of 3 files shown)


gui1117 (approved):
Looks good to me

lexnv (approved):

niklasad1 (approved):

PR #1981, opened 2025-04-04
Metadata V16: Implement support for Pallet View Functions

jsdw:
Add support for Pallet View Functions in Subxt. Summary of changes:

- We have new `api.view_functions()` APIs on the Subxt client which allows somebody to call a view function. This is intended for Pallet View Functions right now, and eventually will work equally with Runtime View Functions when/if they exist.
- Corresponding functions exist in `subxt_core` for constructing the args needed to execute a view function and decoding the result.
- Codegen has been added so that you can call view functions in a type safe way via the generated interface (`runtime::view_functions().pallet().fn_name(fn_args)`). Validation happens as with anything else to ensure that the generated interface matches the node you're trying to use it against.
- A couple of tests have been added to call view functions (and changes made so that we can test things against unstable metadata, while it exists). Note that no examples of view functions are added because Polkadot doesn't expose any yet (but when the time comes, the tests can be easily turned into examples).


core/src/view_functions/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        .ok_or_else(|| MetadataError::ViewFunctionNotFound(*payload.query_id()))?;
+    if static_hash != view_function.hash() {
+        return Err(MetadataError::IncompatibleCodegen.into());
+    }
+
+    Ok(())
+}
+
+/// Return the name of the Runtime API call which can execute
+pub fn call_name() -> &'static str {
```
lexnv, line 35: dq: We plan to extend this in the future by inspecting the metadata and therefore we picked a plain fn call instead of const VIEW_FN: ..?

lexnv (approved):
LGTM!

core/src/view_functions/mod.rs
jsdw: Made it a const!

niklasad1 (approved):

PR #1989, opened 2025-04-08
chore(deps): tokio ^1.44.2

niklasad1:
This pull request would resolve a Dependabot alert on tokio. 

> This release fixes a soundness issue in the broadcast channel. The channel
accepts values that are Send but !Sync. Previously, the channel called
clone() on these values without synchronizing. This release fixes the channel
by synchronizing calls to .clone() (Thanks Austin Bonander for finding and
reporting the issue).

I don't think subxt are using such channels anyway but you never now if one some dependency is using them, not sure whether jsonrpsee is using them for the clients


jsdw (approved):
Cool cool! Any reason we want to ensure it's bumped or just a routine update?

PR #1999, opened 2025-05-07
Support v16 metadata and use it by default if it's available

jsdw:
This PR updates subxt to support V16 metadata (from `frame-metadata@23.0.0`) and ensures it's downloaded and used by default when available.

It also:
- Updates `frame-decode` so that it aligns on the `frame-metadata` in use ([see this](https://github.com/paritytech/frame-decode/pull/42))
- Updates `scale-typegen` because a new `Duration` prelude type was spotted, breaking tests.
- Updates the Substrate dependencies.
- Fixes other changes in the Substrate node that broke tests.

### DIFF
--- a/metadata/src/from/mod.rs
+++ b/metadata/src/from/mod.rs
@@ -8,6 +8,11 @@ mod v14;
 mod v15;
 mod v16;
 
+/// The metadata versions that we support converting into [`crate::Metadata`].
+/// These are ordest from highest to lowest, so that the metadata we'd want to
+/// pick first is first in the array.
+pub const SUPPORTED_METADATA_VERSIONS: [u32; 3] = [16, 15, 14];
+
 /// An error emitted if something goes wrong converting [`frame_metadata`]
 /// types into [`crate::Metadata`].
 #[derive(Debug, PartialEq, Eq, DeriveError)]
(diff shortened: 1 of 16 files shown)


testing/integration-tests/src/full_client/frame/balances.rs
```diff
@@ -282,30 +282,10 @@ async fn storage_total_issuance() {
...
-
-    signed_extrinsic
-        .submit_and_watch()
-        .await
-        .unwrap()
-        .wait_for_finalized_success()
-        .await?
-        .find_first::<system::events::ExtrinsicSuccess>()?
-        .expect("No ExtrinsicSuccess Event found");
-
```
jsdw, line 308: There is already a hold now so no need to artifically create one.

utils/fetch-metadata/src/url.rs
```diff
@@ -183,9 +171,19 @@ async fn fetch_metadata(client: impl ClientT, version: MetadataVersion) -> Resul
...
+            // The "new" interface failed. if the user is asking for V14 or the "latest"
+            // metadata then try the legacy interface instead. Else, just return the
+            // reason for failure.
+            if matches!(version, MetadataVersion::Version(14) | MetadataVersion::Latest) {
+                fetch_inner_legacy(&client).await
+            } else {
+                Err(e)
+            }
+        }
     }
```
jsdw, line 188: I tidied this up because previously it was returning only the error from `fetch_inner_legacy` (ie "The node can only return version 14 metadata using the legacy API..") which wasn't very helpful!

metadata/src/from/mod.rs
```diff
@@ -8,6 +8,11 @@ mod v14;
 mod v15;
 mod v16;
 
+/// The metadata versions that we support converting into [`crate::Metadata`].
+/// These are ordest from highest to lowest, so that the metadata we'd want to
+/// pick first is first in the array.
+pub const SUPPORTED_METADATA_VERSIONS: [u32; 3] = [16, 15, 14];
```
jsdw, line 14: I moved this here so that we have only one array of supported versions that's used anywhere we need it

jsdw:
The only failing test is failing because of some cache cleanup error at the end that cropped up before for a bit. All of the actual tests pass, so I'll ignore this failure.

PR #2025, opened 2025-06-19
Address some issues around tx mortality

### LINKED ISSUE #2024: Misleading usability with `create_partial_offline` and mortal txs
I am a subxt newbie, and to some extent, a polkadot newbie, and I would assume lots of developers can be in the situation where the API and the docs are still confusing (although to the majority of cases it is very well written and precise), so that when they configure some tx params with mortality, used to create an offline transaction, they notice in the end it is submitted as `immortal`.

Using `create_partial_offline` shouldn't allow passing over params that have configured `mortality`. Should allow only certain params that are compatible with the `offline` transaction creation. Not sure how feasable it is, but the whole point of the issue here is to question whether it is possible for `create_partial_offline` to accept a certain `OfflineTxParams` type/trait (corresponding to txs that are created offline), where we can suggest better through the API what's possible (e.g. `mortal` setter will not exist for such a transaction). At the same time, if the mortality will be the only thing different between `OfflineTxParams` and `OnlineTxParams`, I can understand why this suggestion wouldn't be very valuable. I would go for it nonetheless, curious about opinions.

jsdw:
- Make it a runtime error if you explicitly set an extrinsic to be `.mortal(..)` when providing params to it, but then are constructing it offline which prevents this from working. Previously we'd turn the extrinsic into an immortal one.
- Expose `mortal_from_unchecked` in the `DefaultExtrinsicParamsBuilder` to make it a little more ergonomic to set the relevant mortality params when constructing eg an offline extrinsic. Previously You'd have to do:

   ```rust
   let params = DefaultExtrinsicParamsBuilder::new();
   params.5 = CheckMortalityParams::mortal_from_unchecked(for_n_blocks, from_block_n, from_block_hash);
   ```

- Clarify the docs on these functions to make it clear that you can't use `mortal()` with offline extrinsics.
- Actually construct mortal transactions by default. I think this was the intended behaviour, but it may have been accidentally broken because the default Params actually led to immortal transactions. 
- This led to uncovering an issue decoding Eras, and so now there is a proper implementation of EncodeAsType and DecodeAsType for it.

Closes #2024

### DIFF
--- a/core/src/config/default_extrinsic_params.rs
+++ b/core/src/config/default_extrinsic_params.rs
@@ -57,6 +57,13 @@ impl<T: Config> DefaultExtrinsicParamsBuilder<T> {
         Default::default()
     }
 
+    /// Make the transaction immortal, meaning it will never expire. This means that it could, in
+    /// theory, be pending for a long time and only be included many blocks into the future.
+    pub fn immortal(mut self) -> Self {
+        self.mortality = transaction_extensions::CheckMortalityParams::immortal();
+        self
+    }
+
     /// Make the transaction mortal, given a number of blocks it will be mortal for from
     /// the current block at the time of submission.
     ///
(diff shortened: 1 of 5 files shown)


lexnv (approved):
LGTM!

PR #2034, opened 2025-07-03
Add FFI example

### LINKED ISSUE #1913: Create example(s) showing how to bind to Subxt from other languages
Create an example showing how to bind to Subxt from several languages (eg, Python, Java) and use it to construct and/or submit transactions.

We should decide which language(s) are most important for the community and then create example bindings to those to accomplish a simple task (like submitting a transaction).

A single example should be applicable to several languages: we'd create bindings once and then show how they can be used in target langauges. We don't want to create bindings multiple times as this will be more of a maintenance headache.

wassimans:
Adds a basic example that showcases how foreign languages (Python and Javascript in this example) can interact with any Substrate based network by interacting with a simple façade exposing Subxt functionality via FFI using the C ABI.

Closes #1913


jsdw:
Given that I got it working in the end, I opened #2037 to tweak and see if I can get the example running in CI and will push to merge that once working :)

wassimans:
Thanks @jsdw for your detailed feedback, sorry for the late response (took a long weekend with my kids!).

For the errors, I had none with my setup (M1 Pro macOS):  tested it in a pre-LTS version (v20) and the LTS version. I'll test it with your node version 19 and add more setup instructions in the README if necessary. 

I originally used ffi-napi because that's the only package I knew, and also because we wanted a very simple example with very simple tools to showcase, otherwise we had to go a much deeper approach that requires adding a new layer between the JS and Rust code, a C/C++ code that interacts with our Rust code via the Node API, and that exports the same functionality via a node-gyp configuration so that the JS code calls it. ffi-napi abstracts all of that for us. So I think a more "raw" approach would be to much not only for our example but also for the Subxt users interested in integrating it in FFI setups.

But still, your comment made me search for alternatives to ffi-napi, as it was the only one I knew a bit, and I discovered this project: https://napi.rs. Which seems what we need for our JS code, but I think it doesn't make sense for our example:

- needs JS specific annotations in the Rust code, a code pollution that doesn't help the other languages we interact with (Python in our case).
- we still need another FFI path for the other languages.

This is why I think the best solution is to stay with a core logic in pure Rust with a plain C-ABI surface (extern "C" functions).

Now, to keep ffi-napi or not, I'de suggest to keep it for the sake of simplicity and add more details in the README on alternatives with pros and cons of each.

If you don't mind I'll edit the README and add any conclusions on the node setup requirements along with the ffi-napi alternatives.

jsdw:
> If you don't mind I'll edit the README and add any conclusions on the node setup requirements along with the ffi-napi alternatives.

For now, I got everything working with a couple of tweaks on a branch I forked off this one (see #2037) so my inclination is to push to get that merged now and then think about any tweaks later (because what you've done is already a really great example of how to get started with FFI :))

With that in mind I'll close this and hopefully we'll get your FFI examples merged in #2037 soon! Thank you so much for your work on this!

PR #2035, opened 2025-07-04
Improve docs for `at_latest`

bkchr:
Tell the user that these functions operate with the last finalized block.


jsdw (approved):
Thank you! Good point to clarify this!

PR #2046, opened 2025-07-29
Increase reconnecting client request/response size

jsdw:
The server limit is higher than 10MB; let's make the limit higher so that we are less likely to hit issues on the client side with things being too large and instead will rely on the server to tell us if this is the case.

Just in case somebody wants to send a huge transaction or retrieve a huge storage value :)


bkchr (approved):

lexnv (approved):
LGTM! 🙏

PR #2068, opened 2025-08-25
chore(dev): increase number of dev accounts

bee344:
Increase the number of dev accounts to 20, in order to match the number `hardhat` provides.

Rel: https://github.com/paritytech/hardhat-polkadot/issues/288


jsdw (approved):
Happy with this if it's useful!

Issue #1880, opened 2024-12-09
Panic when constructing a dynamic transaction with invalid arguments


AlexD10S:
I encountered a panic! in the `subxt` library when attempting to construct a dynamic transaction `call_data` with invalid arguments.
The issue occurs in this specific line of codehttps://github.com/paritytech/subxt/blob/master/core/src/tx/payload.rs#L186 

Error output:
```
The fields are valid types from the metadata, qed;: Error { context: Context { path: [Location { inner: Index(0) }] }, kind: WrongShape { actual: Number, expected_id: "13" } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
**Steps to Reproduce**
```rs
let arg = scale_value::stringify::from_str_custom()
				.add_custom_parser(custom_parsers::parse_hex)
				.add_custom_parser(custom_parsers::parse_ss58)
				.parse("123")
				.0?;
subxt::dynamic::tx("System", "remark", parsed_args);
let call_data = tx.encode_call_data(&client.metadata());
```

I should ensure proper validation of the requested arguments before constructing the transaction. However, if the provided arguments are invalid, it would be ideal for the `encode_call_data` function to throw a meaningful error instead of causing a panic.

niklasad1:
> I should ensure proper validation of the requested arguments before constructing the transaction. However, if the provided arguments are invalid, it would be ideal for the encode_call_data function to throw a meaningful error instead of causing a panic.

Yes agree, it should be an easy fix.

PR #2069, opened 2025-08-26
Do not panic when encoding call data with invalid fields

elfedy:
Addresses https://github.com/paritytech/subxt/issues/1880 which is also making `subxt-cli` panic as well (e.g: calling `subxt explore pallet System calls remark '{ "remark": 123 }'`).


jsdw (approved):
Looks great to me; thank you!

jsdw:
Will fix CI and merge in #2070!

PR #2101, opened 2025-10-03
integrating frame decode: claude pass

jsdw:
Let's see how well Claude did at fixing up Subxt per my subxt-core changes


jsdw:
It sucked; closing :)

PR #2123, opened 2025-11-18
chore: fix some typos in comments

weifangc:
fix some typos in comments


jsdw:
That code is generated so it's pointless fixing typos there

PR #2124, opened 2025-11-19
[v0.50.0] Allow visiting extrinsic fields in subxt_historic

jsdw:
There are two main ways to decode some data that exist in Rust-land right now:

1. via traits like `scale_decode::DecodeAsType` and `scale_decode::DecodeAsFields`
2. via`scale_decode::visitor::decode_with_visitor(..)` and `scale_decode::visitor::Visitor`.

The first is the "higher level" approach, where in most APIs you simple give some type you want the bytes to be decoded into and get back that type if decoding works.

The second is "lower level" and is what the first is built upon. This isn't typically exposed, but in some cases can be very useful. One such case is addressing things like https://github.com/paritytech/polkadot-rest-api/pull/36#discussion_r2538499050, where we want to know the names of types for instance.

This PR adds a `visit` method to extrinsic fields in `subxt_historic` which allows them to be visitored with said `Visitor` type, and adds a usage of this in the extrinsics example so that you can see roughly how it's used.

This is somewhat of a workaround to the fact that we have to handle two different type ID / type resolver types in this version of `subxt_historic`. In future versions, either:
1. the `visit()` method will be much simpler to implement internally, or 
2. it will not exist, but you'll be able to easily fetch everything you need to call `scale_decode::visitor::decode_with_visitor(..)` yourself anyway, which is all that is done in the `visit` method anyway.
3. for cases where you just want type information and not to actually decode the type, you'll just be able to fetc hthe `u32` type ID, and look up the type in the `scale_info::PortableRegistry` to find the information you need, without any visitor related stuff. 

See **examples/extrinsics.rs** for an example of what we can do with a visitor.

Note: I also updated the `subxt codegen` CLI command to accept legacy metadatas; this is helpful to test what sort of types are generated (aside from possibly being useful in the future)

\cc @TarikGul


metadata/src/lib.rs
jsdw: Just a couple of tweaks to support the CLI command update; not really relevant for the historic example!

TarikGul (approved):
Super solid! 🚀

PR #2136, opened 2025-12-03
feat(subxt-historic): Add at_hash to OnlineClient

TarikGul:
Adds a new method to create a client at a specific block hash, bypassing the `archive_v1_hash_by_height` RPC call when
the caller already has the hash.

### DIFF
--- a/historic/src/client/online_client.rs
+++ b/historic/src/client/online_client.rs
@@ -148,6 +148,59 @@ impl<T: Config> OnlineClient<T> {
             block_hash,
         }))
     }
+
+    /// Pick the block hash at which to operate. This is useful when you already have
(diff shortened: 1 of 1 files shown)


TarikGul:
> Interesting! When do you find that you have both the block hash and block number available?

So before every request we resolve for the hash of the block - similar to how we did things in sidecar. Most users use a number for {blockId}, this is also helpful for endpoints like `/blocks/head`, and any related endpoints. For those endpoints I believe we wouldn't have the blockNumber though.

 > If you think it's useful right now though we can def merge this, but I wonder if it's worth just making sure we handle it in the new version over merging this?
 
 Yea there is no issue with holding off, the real main concern here is just performance and decreasing the rpc calls by one :) But this is not urgent at all, so we can hold off to ensure we do this the right way.

PR #2139, opened 2025-12-03
 Add missing standard derives to core types

serikssonn:
This PR adds missing standard derives (`Hash`, `Eq`, `Copy`, `Default`) to core data structures, including `AccountId32`,`RuntimeVersion`, and various transaction extensions.

 **Key Benefits:**
-  **Collection Support:** Types like `AccountId` and`RuntimeVersion` can now be used as keys in `HashMap` and `HashSet.
-  **Ergonomics:** Adds `Copy` to small POD types (e.g., `AccountId32`), reducing the need for explicit `.clone()` calls.

### DIFF
--- a/core/src/client.rs
+++ b/core/src/client.rs
@@ -23,7 +23,7 @@ pub struct ClientState<C: Config> {
 }
 
 /// Runtime version information needed to submit transactions.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


jsdw:
Thank you! However, quite a big refactoring of the Subxt codebase is currently underway and so I think it would be best to hold off on this sort of PR until we are into the new year and have something more concrete there to build on :)

PR #2142, opened 2025-12-09
Allow passing $OUT_DIR in the runtime_metadata_path attribute

pgherveou:
the `runtime_metadata_path` attribute requires a string literal. 
This makes it impossible to generate the runtime metadata at build time and place it in `OUT_DIR`

This PR extends the macro to recognize the `$OUT_DIR` placeholder inside `runtime_metadata_path` and replaces it with the actual `OUT_DIR` environment value at macro expansion time. 

This allows users to write:

```rust

runtime_metadata_path = "$OUT_DIR/md.scale"
```

### DIFF
--- a/macro/src/lib.rs
+++ b/macro/src/lib.rs
@@ -201,6 +201,22 @@ fn validate_type_path(path: &syn::Path, metadata: &Metadata) {
     }
 }
 
+/// Resolves a path, handling the $OUT_DIR placeholder if present.
+/// If $OUT_DIR is present in the path, it's replaced with the actual OUT_DIR environment variable.
+/// Otherwise, the path is resolved relative to CARGO_MANIFEST_DIR.
+fn resolve_path(path_str: &str) -> std::path::PathBuf {
+    if path_str.contains("$OUT_DIR") {
+        let out_dir = std::env::var("OUT_DIR").unwrap_or_else(|_| {
+            abort_call_site!("$OUT_DIR is used in path but OUT_DIR environment variable is not set")
+        });
+        std::path::Path::new(&path_str.replace("$OUT_DIR", &out_dir)).into()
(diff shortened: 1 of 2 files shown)


macro/src/lib.rs
```diff
@@ -232,9 +232,18 @@ fn fetch_metadata(args: &RuntimeMetadataArgs) -> Result<subxt_codegen::Metadata,
...
+            let path = if rest_of_path.contains("$OUT_DIR") {
+                let out_dir = std::env::var("OUT_DIR").unwrap_or_else(|_| {
+                    abort_call_site!("$OUT_DIR is used in runtime_metadata_path but OUT_DIR environment variable is not set")
+                });
+                std::path::Path::new(&rest_of_path.replace("$OUT_DIR", &out_dir)).into()
+            } else {
+                let root = std::env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
+                let root_path = std::path::Path::new(&root);
+                root_path.join(rest_of_path)
+            };
```
jsdw: We should probably make a little function for this and apply the same logic above for lines 216-218; people would expect to be able to do the same thing for the runtime-wasm-path field

jsdw (approved):
Thank you!

subxt/src/lib.rs
```diff
@@ -151,6 +151,15 @@ pub mod ext {
 /// mod polkadot {}
 /// ```
 ///
+/// You can use the `$OUT_DIR` placeholder in the path to reference metadata generated at build time:
+///
+/// ```rust,no_run
```
jsdw: This and below will need ignore instead of no_run to pass the doc tests

lexnv (approved):
Nice! 🚀

PR #2248, opened 2026-07-10
chore: Upgrade smoldot to 2.1.0

DenzelPenzel:
## Summary

Impl #2247 

Upgrades the light client stack and hardens the light client integration tests.

- Bump `smoldot` 1.1.0 → 2.1.0 and `smoldot-light` 1.1.0 → 1.3.1.
- Raise the workspace `rust-version` 1.85.0 → 1.88.0 (required by the newer smoldot).
- Harden the light client integration tests: wrap client initialization and each step in explicit timeouts so a stalled light client fails fast and names the stuck step, and initialise a `RUST_LOG`-driven tracing subscriber so smoldot/subxt logs are attached to failures.

PR #2257, opened 2026-07-22
ci: Promote lightclient test jobs from flaky to mandatory

### LINKED ISSUE #2247: ci: Investigate LightClient flaky tests
The subxt CI has been flaky wrt light client testing.

https://github.com/paritytech/subxt/actions/runs/28862946189/job/85607290121

In multiple iterations the CI is marked as flaky, we'd like to increase the confidence of our testing.

- Upgrade smoldot to the latest released version
- Investigate the flakiness

DenzelPenzel:
Closes #2247.

The light client CI steps were previously "flaky": test failures were swallowed and reported as a neutral `Flaky: ...` check that never blocked PRs. After the smoldot 2.1.0 upgrade (#2248) the systematic failures behind that marking are gone — a 60-iteration stress run over the suites passed 59/60, with the single failure being a live-network slow-peer stall (details: https://github.com/paritytech/subxt/issues/2247#issuecomment-5050977340).

This promotes the steps to mandatory:

- Test failures now fail the job. The swallow-and-report-neutral machinery is removed, and the jobs drop the "Flaky" prefix: `Lightclient WASM tests`, `Lightclient integration tests (<feature>)`.
- Every failure is filed to a rolling tracking issue (label `lightclient-ci-failure`) by a new `lightclient_failure_report` job, linking the failed run — so live-network stalls are reported to maintainers rather than swallowed, and a manual re-run doesn't lose the report. (On PRs from forks the token is read-only, so the report degrades to a warning.)
- smoldot diagnostics are captured for the linked runs: the integration job runs with `RUST_LOG=debug` (smoldot logs under bare targets like `network` and `sync-service-polkadot`, which can't be prefix-filtered as a group), and the wasm test now writes logs to the browser console, which wasm-bindgen-test prints for failing tests.
- `fail-fast: false` on the integration matrix so each backend reports independently.

Note for admins: the `Flaky: ...` informational check runs are no longer created, and if the old job names appear in branch protection required checks they need updating to the new names.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -395,8 +395,8 @@ jobs:
           wasm-pack test --headless --chrome
         working-directory: signer/tests/wasm
 
-  flaky_lightclient_wasm_tests:
-    name: Flaky lightclient WASM tests
+  lightclient_wasm_tests:
+    name: Lightclient WASM tests
     runs-on: ubuntu-latest
     needs: [clippy, wasm_clippy, check, wasm_check, docs]
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -414,41 +414,22 @@ jobs:
...
-                  : 'Lightclient WASM tests failed. These tests are known to be flaky and do not block the PR.'
-              }
-            });
-
-  flaky_lightclient_integration_tests:
-    name: "Flaky lightclient integration tests (${{ matrix.feature }})"
+          # They sync the live network, so a single stalled request can fail a run;
+          # retry once before failing the job for real (see #2247).
+          run_tests() { wasm-pack test --headless --firefox && wasm-pack test --headless --chrome; }
+          run_tests || { echo "Tests failed, retrying once..."; run_tests; }
```
lexnv: Hmm could we retry in the test instead? This might actually hide some issues for us down the line. IIUC the flakiness is related to establishing connection between smoldot <-> full nodes which might cause some stalls. Could we instead check that behavior for retriable errors and enable trace logs for smoldot, such that we report back those issues to active maintainers? 🙏

lexnv (approved):

PR #2262, opened 2026-07-28
feat: decode call data into dynamic transaction payload

DenzelPenzel:
## Summary

Impl #2194 

- add `TransactionsClient::from_call_data_bytes` to decode SCALE-encoded call data into a dynamic transaction payload
- validate pallet/call indices, argument decoding, and reject trailing bytes with dedicated errors

### DIFF
--- a/subxt/src/transactions.rs
+++ b/subxt/src/transactions.rs
@@ -129,6 +129,8 @@ impl<T: Config, Client: OfflineClientAtBlockT<T>> TransactionsClient<T, Client> 
             scale_decode::Field::new(arg.id, name)
         });
 
+        // The first two bytes are the pallet and call indices decoded above;
+        // only the call arguments remain to be decoded here
         let cursor = &mut &call_data[2..];
         let args = Composite::<()>::decode_as_fields(cursor, &mut fields, metadata.types())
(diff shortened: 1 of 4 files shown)


subxt/src/transactions.rs
```diff
@@ -102,6 +103,43 @@ impl<T: Config, Client: OfflineClientAtBlockT<T>> TransactionsClient<T, Client>
...
+        let call_info = metadata
+            .extrinsic_call_info_by_index(pallet_index, call_index)
+            .map_err(|e| ExtrinsicError::CannotDecodeCallData(e.into_owned()))?;
+
+        let mut fields = call_info.args.iter().map(|arg| {
+            let name = (!arg.name.is_empty()).then_some(&*arg.name);
+            scale_decode::Field::new(arg.id, name)
+        });
+
+        let cursor = &mut &call_data[2..];
```
lexnv, line 134: nit: Could add a doc stating why we skipped the first 2 bytes.

Is there any possibility that this will cause an out of bounds panic?

DenzelPenzel: done!

lexnv (approved):
Nice!

PR #2266, opened 2026-07-31
fix: don't keep the startup finalized block pinned forever

marian-radu:
## Problem
`FollowStreamDriver`'s replay buffer holds `BlockRef`s and entries only leave it when a later `Finalized` event reports the block.
At subscription start the node unconditionally reports the current best block. When best == finalized, always on instant-seal, a connect-time race otherwise, that block was finalized *before we started watching*, so no `Finalized` event we receive ever reports it and the pin lasts for the life of the subscription, so the node can't prune state past that block.

## Fix
Keep only the latest `BestBlockChanged` in the buffer. This behavior matches the node rather than diverging from it.
Subscribers joining mid-stream now receive only the latest best-block event. Existing subscribers are unaffected.

## Testing
`startup_best_block_event_is_not_kept_alive_forever` asserts the replay buffer no longer references block 0 and that an unpin was issued for it.

### DIFF
--- a/subxt/src/backend/chain_head/follow_stream_driver.rs
+++ b/subxt/src/backend/chain_head/follow_stream_driver.rs
@@ -322,8 +322,9 @@ impl<H: Hash> Shared<H> {
                     .push_back(FollowEvent::NewBlock(new_block_ev));
             }
             FollowStreamMsg::Event(ev @ FollowEvent::BestBlockChanged(_)) => {
-                // The node sends exactly one `BestBlockChanged` to each new `chainHead_follow`
-                // subscription, so replay only the latest one to our new subscribers.
+                // When the node starts a new subscription it reports exactly one
+                // `BestBlockChanged`, so replaying only the latest matches what a fresh
+                // subscription would see.
                 shared
                     .block_events_for_new_subscriptions
                     .retain(|e| !matches!(e, FollowEvent::BestBlockChanged(_)));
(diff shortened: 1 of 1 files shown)


subxt/src/backend/chain_head/follow_stream_driver.rs
```diff
@@ -322,6 +322,11 @@ impl<H: Hash> Shared<H> {
                     .push_back(FollowEvent::NewBlock(new_block_ev));
             }
             FollowStreamMsg::Event(ev @ FollowEvent::BestBlockChanged(_)) => {
+                // The node sends exactly one `BestBlockChanged` to each new `chainHead_follow`
+                // subscription, so replay only the latest one to our new subscribers.
```
marian-radu:
```suggestion
                // When the node starts a new subscription it reports exactly one
                // `BestBlockChanged`, so replaying only the latest matches what a fresh
                // subscription would see.
```

subxt/src/backend/chain_head/follow_stream_driver.rs
```diff
@@ -322,6 +322,12 @@ impl<H: Hash> Shared<H> {
                     .push_back(FollowEvent::NewBlock(new_block_ev));
             }
             FollowStreamMsg::Event(ev @ FollowEvent::BestBlockChanged(_)) => {
+                // When the node starts a new subscription it reports exactly one
+                // `BestBlockChanged`, so replaying only the latest matches what a fresh
+                // subscription would see.
+                shared
+                    .block_events_for_new_subscriptions
+                    .retain(|e| !matches!(e, FollowEvent::BestBlockChanged(_)));
```
lexnv, line 330: Thanks! It makes sense, this effectively cleans up the `block_events_for_new_subscriptions`. If we had:
- new block 1 + bestBlock 1 + new Block 2  + bestBlock 2
- new subscribers would see: new block 1 + new block 2 + best block 2

lexnv (approved):

DenzelPenzel (approved):
LGTM, thanks!

PR #2268, opened 2026-08-04
build(deps): update frame-decode to 0.18.1

pgherveou:
## Summary

- Update the workspace `frame-decode` requirement from 0.17.1 to 0.18.1.
- Consume the corrected Extrinsic V5 signer-payload implementation from frame-decode.
- Use `is_authorization_extension` trait method instead of `out.clear()` side effects for `VerifySignature`.

## Changes

- frame-decode 0.18.1 keeps `(transaction_extension_version, call)` as the immutable base implication and uses `is_authorization_extension` to exclude preceding extension values/implicits from the V5 signer payload.
- Subxt's `VerifySignature` now implements `is_authorization_extension() -> true` and removes the `encode_value_for_signer_payload_to` / `encode_implicit_to` buffer-clearing overrides.

Uses frame-decode 0.18.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.50.3] - 2026-08-06
+
+This release updates `frame-decode` to 0.18.1, which fixes V5 signer payload construction and improves how authorization extensions (like `VerifyMultiSignature`) are handled.
+
+### Fixed
+
+- V5 signer payloads now correctly include the transaction extension version and call data as an immutable base implication, matching FRAME's transaction extension pipeline semantics. ([paritytech/frame-decode#104](https://github.com/paritytech/frame-decode/pull/104))
+- Unknown `Option<T>` transaction extensions are now encoded as `None` (`0u8`) by default, allowing transaction encoding to succeed on chains with optional extensions. (from frame-decode 0.17.2)
+
+### Changed
+
+- `VerifySignature` now uses `is_authorization_extension` to declare itself as an authorization boundary, replacing the previous `out.clear()` side effects in `encode_value_for_signer_payload_to` and `encode_implicit_to`. ([#2268](https://github.com/paritytech/subxt/pull/2268), [#2269](https://github.com/paritytech/subxt/pull/2269))
+- Update `frame-decode` from 0.17.1 to 0.18.1. ([#2268](https://github.com/paritytech/subxt/pull/2268))
+
 ## [0.50.2] - 2026-07-06
 
 This release fixes an issue whereby setting the genesis hash in the `SubstrateConfigBuilder` had no effect, and improves the reliability of tests against public archival RPC endpoints.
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -80,7 +80,7 @@ darling = "0.20.10"
 derive-where = "1.2.7"
 either = { version = "1.13.0", default-features = false }
 finito = { version = "0.1.0", default-features = false }
-frame-decode = { version = "0.17.1", default-features = false }
+frame-decode = { version = "0.18.1", default-features = false }
```
TarikGul, line 83: Shouldn't the lockfile also be updated?

TarikGul (changes requested):
Let's add a changelog entry, and a test. We should also mention the changes and fixes that come from 0.17.2

filvecchiato:
Pushed updates:

- **Lockfile**: regenerated for frame-decode 0.18.1
- **Changelog**: added 0.50.3 entry covering the V5 signer payload fix, `Option<T>` extension encoding (from frame-decode 0.17.2), and `is_authorization_extension` refactor
- **`encode_implicit_to` no-op**: added comment explaining implicit type is `()` and authorization extensions are excluded from the signer payload
- **`VerifySignature` doc comment**: updated to mention extension version, values and implicits (not just "additional bytes")
- **Single authorization extension assumption**: documented on the struct that `inject_signature` broadcasts the same signature to every extension, so multiple authorization extensions would disagree with `rposition`
- **Tests**: added 3 tests — `VerifySignature` returns `true`, `CheckMetadataHash` returns `false`, and tuple-level routing works correctly

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.50.3] - 2026-08-06
```
TarikGul, line 7: We can keep this as Unreleased until the release PR, but not a huge deal.

TarikGul:
The lockfile is kinda big should we just do "cargo update -p frame-decode --precise 0.18.1" ?
We also should get a test for specifically: signer_payload()

subxt/src/config/transaction_extensions.rs
```diff
@@ -53,6 +57,10 @@ impl<T: Config> frame_decode::extrinsics::TransactionExtension<PortableRegistry>
 {
     const NAME: &str = "VerifyMultiSignature";
 
+    fn is_authorization_extension(&self) -> bool {
```
lexnv, line 60: The file looks good! I think we can trim a bit the cargo lock changes 🙏

TarikGul (approved):

lexnv (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
...
+
+This release updates `frame-decode` to 0.18.1, which fixes V5 signer payload construction and improves how authorization extensions (like `VerifyMultiSignature`) are handled.
+
+### Fixed
```
lexnv, line 11: Id move this section under `[0.50.3]`

We usually keep here the PRs formatted as

```
 Upate Artifacts (auto-generated) ([#2245](https://github.com/paritytech/subxt/pull/2245))
```

Issue #1998, opened 2025-05-06
Always use transaction extensions version 0 when working with v4 transactions, and find latest match when using v5


jsdw:
v4 transactions should always be encoded or decoded using the transaciton extensions under version 0 in the metadata. Currently, they would use the latest available version (which is ok at the moment because there only exists one version anyway; 0).

This is more complicated than just setting which version we'll look at, because V4 and V5 transactions will differ in their behaviour.
- V4 transactions should always use version 0
- V5 transactions should tend to using the latest version, but really should look for the version that lines up with the configured transaction extensions for the chain, and use that.
    - In the event that there is a version change, but the transaction extension names remain identical (only the properties of one or more of them change), we would have a dilemma as to which version to select. Thus, perhaps we should allow the user to explicitly configure which version of extensions to try to use, falling back to a sane default (use the latest that matches?) otherwise. Names changing would be an issue though and so I would expect that this wouldn't happen, so we can probably just be clever and pick the latest matching set.

DenzelPenzel:
Closing: the v4 half (always decode against extension version 0) was fixed in #2277, and the v5 half (pick the newest extension version that can be satisfied, falling back otherwise) in #2273. Both ship in v0.51.0 (#2280).

PR #2273, opened 2026-08-13
feat: support custom transaction extension values

### LINKED ISSUE #2265: Cannot sign for a chain with a custom transaction extension without defining a whole new Config
### The problem

If a chain declares a custom transaction extension subxt doesn't implement, and its value type is neither empty nor `Option<_>`, signing fails outright.

**Example**:

`RestrictOrigins` is the only transaction extension on Paseo Asset Hub whose value is neither `()` nor `Option<_>`: it encodes a bare `bool`. 
Generic clients cannot build a signed transaction for the chain without special-casing it. 
For example, the polkadot-staking-miner (subxt 0.50) failed every submission with:

```
Subxt(ExtrinsicError(EncodeError(TransactionExtensions(NotFound("RestrictOrigins")))))
```

The mechanism is in subxt's `frame-decode`, the encoder subxt uses (`methods/extrinsic_encoder.rs:1198`):

- extensions whose value type is **empty** are filtered out before the client is consulted;
- unknown extensions whose value type is **`Option<_>`** fall back to encoding `None`;
- anything else is a hard error.

Of the seven custom extensions Paseo AH publishes, six therefore cost clients nothing (`AuthorizeCall` and `EthSetOrigin` are `()`; `AsPgas`, `AsRingAlias`, `AsDotnsGateway` and `AuthorizeValueTransfer` are `Option<_>`). 
Only `RestrictOrigins` breaks them.

####  Additional notes:

- Paseo AH publishes **exactly one** transaction-extension version `(0)` containing all 18 extensions. Polkadot AH, Kusama AH and Westend AH likewise publish only version 0, so extension versioning is not in use anywhere and offers no fallback set.
- `Preamble::Signed(Address, Signature, ExtensionV0)` in `sp_runtime::generic::unchecked_extrinsic` pins V4 extrinsics to extension version 0, and subxt signs V4 whenever a chain advertises it (see `subxt/src/transactions.rs:236-246`).  All chains advertise `[4, 5]`) so a V4 signer has nothing to negotiate: whatever sits in version 0 is mandatory.
- IIUC the issues is common between V4 and V5 (both `encode_v4_signed` and `encode_v5_general` take the same `Config::TransactionExtensions`)
- **subxt asks the metadata for the highest extension version, not the highest one it can actually encode**. see [here](https://github.com/paritytech/subxt/blob/master/metadata/src/lib.rs#L991)
```rust
pub fn transaction_extension_version_to_use_for_encoding(&self) -> u8 {
    *self.transaction_extensions_by_version.keys().max().expect(…)
}
```

**So on a chain publishing v0 (all-encodable) and v1 (containing an extension subxt doesn't implement), subxt jumps to v1 and fails, despite v0 being right there and workable.** The capability check already exists one crate down: frame-decode's `find_best_extension_version` walks the versions and returns the first whose extensions are each empty-valued, Option-valued, or supplied by the client. subxt just never calls it. 

This might deserve its own separate issue maybe but for a V5 client it might make the difference between "adding a new custom extension is a safe, additive change" vs "it breaks all the clients"  😅 so it feels pretty important. 

### Workaround client-side

The only way out today (AFAIK) is to define your own type implementing `subxt::Config`, since the extension set is an associated type, which is a lot of machinery and boilerplate. See for example https://github.com/paritytech/polkadot-staking-miner/pull/1302.

### What I'd like

Ideally I'd like a way to hand subxt a value for a named extension it doesn't know so **a raw value override.**,  something like  `DefaultExtrinsicParamsBuilder::custom_extension("RestrictOrigins", value)` or similar, taking a SCALE-encodable value or a `scale_value::Value` encoded against the metadata type. subxt supplies no semantics, the caller supplies the value.

`RestrictOrigins` encodes as a boolean, but I would like us to be flexible to support more than a boolean (and also avoid to rely on default value of the boolean... even if it wouldn't be unreasonable to treat default value of boolean as `false` in this case).

### Versions

subxt 0.50.1, frame-decode 0.17.2, against Paseo Asset Hub (spec 2004002).

DenzelPenzel:
## Summary

Allow callers to provide metadata-aware values for custom transaction extensions without implementing a new `Config`:

```rust
let params = PolkadotExtrinsicParamsBuilder::new()
    .custom_extension("RestrictOrigins", true)
    .build();
```

The value is encoded against the extension type from runtime metadata. This fixes signing on chains such as Paseo Asset Hub, where `RestrictOrigins` contains a bare `bool`.

Fixes #2265.

## Implementation

- Preserve known default extension behavior and parameter access.
- Encode custom values using runtime metadata.
- Reject duplicate names, overrides of known extensions, and non-empty implicit data.
- Restore the output buffer on encoding errors.

## Breaking changes

`DefaultTransactionExtensions<T>` and its `Params` type changed from public tuple aliases to structs with private fields, so that custom extension values can be carried alongside the known ones. The tuples live on under new names: `KnownDefaultTransactionExtensions<T>` and `KnownDefaultExtrinsicParams<T>`.

Code which configures transactions via the builder and passes the params through opaquely is source-compatible and produces byte-identical transactions. Code which relied on the tuple shape of the params migrates as follows:

| Before | After |
|---|---|
| constructing the params tuple manually | `DefaultExtrinsicParams::from_known(tuple)` |
| mutating an element, e.g. `params.5 = ...` | `params.known_mut().5 = ...` |
| reading an element, e.g. `&params.7` | `&params.known().7` |

Given the above, this should land in 0.51 rather than a 0.50.x patch.

## Notes

- The V5 extension-version selection part of the issue is deliberately out of scope here; it is tracked in #1998.
- Custom authorization extensions and custom extensions with non-empty implicit data are not supported; the latter returns a clear error naming the extension.
- Includes a drive-by `cargo fmt` fix for two test asserts that landed unformatted on master.

### DIFF
--- a/subxt/src/config/default_transaction_extensions.rs
+++ b/subxt/src/config/default_transaction_extensions.rs
@@ -2,14 +2,20 @@
 // This file is dual-licensed as Apache-2.0 or GPL-3.0.
 // see LICENSE for license details.
 
-use super::{Config, HashFor};
-use super::{TransactionExtensions, transaction_extensions};
+use super::{Config, HashFor, TransactionExtensions, transaction_extensions};
+use crate::config::transaction_extension_traits::Params;
 use crate::config::transaction_extensions::CheckMortalityParams;
+use crate::error::TransactionExtensionError;
+use crate::transactions::DefaultParams;
+use scale_encode::EncodeAsType;
+use scale_info::PortableRegistry;
+use scale_value::Value;
+use std::collections::BTreeMap;
 
-/// A set of transaction extensions which is applicable in most default cases. This can be used
-/// as our [`Config::TransactionExtensions`] type, and then users can configure transactions by
-/// using the corresponding [`DefaultExtrinsicParamsBuilder`].
-pub type DefaultTransactionExtensions<T> = (
+/// The known transaction extensions used by [`DefaultTransactionExtensions`].
+///
+/// This is exposed for users who need to manually compose the default typed extensions.
+pub type KnownDefaultTransactionExtensions<T> = (
     transaction_extensions::VerifySignature<T>,
     transaction_extensions::CheckSpecVersion,
     transaction_extensions::CheckTxVersion,
@@ -20,6 +26,192 @@ pub type DefaultTransactionExtensions<T> = (
     transaction_extensions::ChargeTransactionPayment,
     transaction_extensions::CheckMetadataHash,
 );
+
+/// The parameters used to construct [`KnownDefaultTransactionExtensions`].
+pub type KnownDefaultExtrinsicParams<T> =
+    <KnownDefaultTransactionExtensions<T> as TransactionExtensions<T>>::Params;
+
+/// The default set of transaction extensions, along with any custom extensions supplied for a
+/// specific transaction.
+pub struct DefaultTransactionExtensions<T: Config> {
+    known: KnownDefaultTransactionExtensions<T>,
+    custom: BTreeMap<String, Value>,
+}
+
+/// Parameters used to construct [`DefaultTransactionExtensions`].
+pub struct DefaultExtrinsicParams<T: Config> {
+    known: KnownDefaultExtrinsicParams<T>,
+    custom: Vec<(String, Value)>,
+}
+
+impl<T: Config> DefaultExtrinsicParams<T> {
+    /// Construct parameters from the known default transaction extension parameters.
+    pub fn from_known(known: KnownDefaultExtrinsicParams<T>) -> Self {
+        Self {
+            known,
+            custom: Vec::new(),
+        }
+    }
+
+    /// Return the parameters for the known default transaction extensions.
+    pub fn known(&self) -> &KnownDefaultExtrinsicParams<T> {
+        &self.known
+    }
+
+    /// Return a mutable reference to the parameters for the known default
+    /// transaction extensions.
+    pub fn known_mut(&mut self) -> &mut KnownDefaultExtrinsicParams<T> {
+        &mut self.known
+    }
+
+    /// Return the values provided for custom transaction extensions.
+    pub fn custom(&self) -> &[(String, Value)] {
+        &self.custom
+    }
+}
+
+impl<T: Config> Default for DefaultExtrinsicParams<T> {
+    fn default() -> Self {
+        DefaultExtrinsicParamsBuilder::new().build()
+    }
+}
+
+impl<T: Config> DefaultParams for DefaultExtrinsicParams<T> {
+    fn default_params() -> Self {
+        Self::default()
+    }
+}
+
+impl<T: Config> Params<T> for DefaultExtrinsicParams<T> {
+    fn inject_account_nonce(&mut self, nonce: u64) {
+        self.known.inject_account_nonce(nonce);
+    }
+
+    fn inject_block(&mut self, number: u64, hash: HashFor<T>) {
+        self.known.inject_block(number, hash);
+    }
+}
+
+impl<T: Config> TransactionExtensions<T> for DefaultTransactionExtensions<T> {
+    type Params = DefaultExtrinsicParams<T>;
+
+    fn new(
+        client: &super::ClientState<T>,
+        params: Self::Params,
+    ) -> Result<Self, TransactionExtensionError> {
+        let known = <KnownDefaultTransactionExtensions<T> as TransactionExtensions<T>>::new(
+            client,
+            params.known,
+        )?;
+        let mut custom = BTreeMap::new();
+
+        for (name, value) in params.custom {
+            if frame_decode::extrinsics::TransactionExtensions::contains_extension(&known, &name) {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' conflicts with a known transaction extension"
+                )));
+            }
+            if custom.insert(name.clone(), value).is_some() {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' was provided more than once"
+                )));
+            }
+        }
+
+        Ok(Self { known, custom })
+    }
+
+    fn inject_signature(&mut self, account_id: &T::AccountId, signature: &T::Signature) {
+        self.known.inject_signature(account_id, signature);
+    }
+}
+
+impl<T: Config> DefaultTransactionExtensions<T> {
+    fn encode_custom_value_to(
+        &self,
+        name: &str,
+        type_id: u32,
+        type_resolver: &PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), frame_decode::extrinsics::TransactionExtensionsError> {
+        let value = self.custom.get(name).ok_or_else(|| {
+            frame_decode::extrinsics::TransactionExtensionsError::NotFound(name.to_owned())
+        })?;
+        let original_len = out.len();
+        let result = value.encode_as_type_to(type_id, type_resolver, out);
+        result.map_err(|error| {
+            out.truncate(original_len);
+            frame_decode::extrinsics::TransactionExtensionsError::Other {
+                extension_name: name.to_owned(),
+                error: Box::new(error),
+            }
+        })
+    }
+}
+
+impl<T: Config> frame_decode::extrinsics::TransactionExtensions<PortableRegistry>
+    for DefaultTransactionExtensions<T>
+{
+    fn contains_extension(&self, name: &str) -> bool {
+        frame_decode::extrinsics::TransactionExtensions::contains_extension(&self.known, name)
+            || self.custom.contains_key(name)
+    }
+
+    fn is_authorization_extension(&self, name: &str) -> bool {
+        frame_decode::extrinsics::TransactionExtensions::is_authorization_extension(
+            &self.known,
+            name,
+        )
+    }
+
+    fn encode_extension_value_to(
+        &self,
+        name: &str,
+        type_id: u32,
+        type_resolver: &PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), frame_decode::extrinsics::TransactionExtensionsError> {
+        if frame_decode::extrinsics::TransactionExtensions::contains_extension(&self.known, name) {
+            frame_decode::extrinsics::TransactionExtensions::encode_extension_value_to(
+                &self.known,
+                name,
+                type_id,
+                type_resolver,
+                out,
+            )
+        } else {
+            self.encode_custom_value_to(name, type_id, type_resolver, out)
+        }
+    }
+
+    fn encode_extension_implicit_to(
+        &self,
+        name: &str,
+        type_id: u32,
+        type_resolver: &PortableRegistry,
+        out: &mut Vec<u8>,
+    ) -> Result<(), frame_decode::extrinsics::TransactionExtensionsError> {
+        if frame_decode::extrinsics::TransactionExtensions::contains_extension(&self.known, name) {
+            frame_decode::extrinsics::TransactionExtensions::encode_extension_implicit_to(
+                &self.known,
+                name,
+                type_id,
+                type_resolver,
+                out,
+            )
+        } else if self.custom.contains_key(name) {
+            Err(frame_decode::extrinsics::TransactionExtensionsError::Other {
+                extension_name: name.to_owned(),
+                error: format!(
+                    "Custom transaction extension '{name}' requires non-empty implicit data, which is not supported"
+                )
+                .into(),
+            })
+        } else {
+            Err(frame_decode::extrinsics::TransactionExtensionsError::NotFound(name.to_owned()))
+        }
+    }
+}
 
 /// A builder that outputs the set of parameters required to configure transactions when
 /// [`DefaultTransactionExtensions`] is used. This may expose methods that aren't applicable
@@ -34,6 +226,7 @@ pub struct DefaultExtrinsicParamsBuilder<T: Config> {
     tip_of: u128,
     /// A fallback tip used when no Asset ID is given (or the chain doesn't support it).
     tip: u128,
+    custom: Vec<(String, Value)>,
 }
 
 impl<T: Config> Default for DefaultExtrinsicParamsBuilder<T> {
@@ -44,6 +237,7 @@ impl<T: Config> Default for DefaultExtrinsicParamsBuilder<T> {
             tip_of: 0,
             tip_of_asset_id: None,
             nonce: None,
+            custom: Vec::new(),
         }
     }
 }
@@ -119,8 +313,17 @@ impl<T: Config> DefaultExtrinsicParamsBuilder<T> {
         self
     }
 
+    /// Provide a metadata-aware value for a custom transaction extension.
+    ///
+    /// Values absent from runtime metadata are ignored. Known or duplicate names and extensions
+    /// with non-empty implicit data are rejected. Custom authorization extensions are unsupported.
+    pub fn custom_extension(mut self, name: impl Into<String>, value: impl Into<Value>) -> Self {
+        self.custom.push((name.into(), value.into()));
+        self
+    }
+
     /// Build the extrinsic parameters.
-    pub fn build(self) -> <DefaultTransactionExtensions<T> as TransactionExtensions<T>>::Params {
+    pub fn build(self) -> DefaultExtrinsicParams<T> {
         let check_mortality_params = self.mortality;
 
         let charge_asset_tx_params = if let Some(asset_id) = self.tip_of_asset_id {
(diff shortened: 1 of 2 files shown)


subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,195 @@ pub type DefaultTransactionExtensions<T> = (
...
 
+/// The parameters used to construct [`KnownDefaultTransactionExtensions`].
+pub type KnownDefaultExtrinsicParams<T> =
+    <KnownDefaultTransactionExtensions<T> as TransactionExtensions<T>>::Params;
+
+/// The default set of transaction extensions, along with any custom extensions supplied for a
+/// specific transaction.
+pub struct DefaultTransactionExtensions<T: Config> {
+    known: KnownDefaultTransactionExtensions<T>,
+    custom: BTreeMap<String, CustomTransactionExtensionValue>,
```
lexnv: If we misspell the extension name we can now produce a wrongly singed extrinsic

This takes an arbitrary string, maybe we can easily compare against `metadata.extrinsic().transaction_extensions_to_use_for_encoding()`?

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,195 @@ pub type DefaultTransactionExtensions<T> = (
...
+    for DefaultTransactionExtensions<T>
+{
+    fn contains_extension(&self, name: &str) -> bool {
+        frame_decode::extrinsics::TransactionExtensions::contains_extension(&self.known, name)
+            || self.custom.contains_key(name)
+    }
+
+    fn is_authorization_extension(&self, name: &str) -> bool {
+        frame_decode::extrinsics::TransactionExtensions::is_authorization_extension(
+            &self.known,
```
lexnv, line 176: nit: This forwards only to known entries, so chain specific exntesions return false here? Would be worth having a subxt issue about this one to tackle later

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,195 @@ pub type DefaultTransactionExtensions<T> = (
...
+    custom: BTreeMap<String, CustomTransactionExtensionValue>,
+}
+
+/// Parameters used to construct [`DefaultTransactionExtensions`].
+pub struct DefaultExtrinsicParams<T: Config> {
+    known: KnownDefaultExtrinsicParams<T>,
+    custom: Vec<(String, CustomTransactionExtensionValue)>,
+}
+
+enum CustomTransactionExtensionValue {
```
lexnv: Do we plan to extend this? Maybe its easier to colapse and use directly `Value`?

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,195 @@ pub type DefaultTransactionExtensions<T> = (
...
+        )?;
+        let mut custom = BTreeMap::new();
+
+        for (name, value) in params.custom {
+            if frame_decode::extrinsics::TransactionExtensions::contains_extension(&known, &name) {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' conflicts with a known transaction extension"
+                )));
+            }
+            if custom.insert(name.clone(), value).is_some() {
```
lexnv: nit: we clone the name here on the happy path to ensure we can properly report the error

Instead could we use `contains_key` and avoid name cloning?

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,195 @@ pub type DefaultTransactionExtensions<T> = (
...
+pub struct DefaultExtrinsicParams<T: Config> {
+    known: KnownDefaultExtrinsicParams<T>,
+    custom: Vec<(String, CustomTransactionExtensionValue)>,
+}
+
+enum CustomTransactionExtensionValue {
+    Dynamic(Value),
+}
+
+impl<T: Config> DefaultExtrinsicParams<T> {
```
lexnv, line 47: nit: API cannot access the custom entries

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -119,8 +316,20 @@ impl<T: Config> DefaultExtrinsicParamsBuilder<T> {
         self
     }
 
+    /// Provide a metadata-aware value for a custom transaction extension.
+    ///
+    /// Values absent from runtime metadata are ignored. Known or duplicate names and extensions
+    /// with non-empty implicit data are rejected. Custom authorization extensions are unsupported.
```
lexnv:
```suggestion
    /// Extensions absent from runtime metadata are ignored. Known or duplicate names and extensions
    /// with non-empty implicit data are rejected. Custom authorization extensions are unsupported.
```

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -119,8 +316,20 @@ impl<T: Config> DefaultExtrinsicParamsBuilder<T> {
         self
     }
 
+    /// Provide a metadata-aware value for a custom transaction extension.
```
lexnv, line 330: nit: We should add an ` # Example` section to this one

lexnv (commented):
The shape looks good, just some tiny bits and pieces!

We could probably add some more `V5` tests (ie around `authorization_extension_check_is_forwarded` and `is_authorization_extension` and probably some tests with a custom extension positioned before and after `VerifyMultiSignature`)

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Good catch – a typo silently dropped the extension. `DefaultTransactionExtensions::new` now validates each custom name against `metadata.extrinsic().transaction_extensions_to_use_for_encoding()` and errors with "Custom transaction extension '{name}' is not present in the runtime metadata" if it is absent. Covered by `custom_extension_absent_from_metadata_is_rejected`

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: No plans to extend it, collapsed to `Value` directly, so the maps are now `BTreeMap<String, Value>` / `Vec<(String, Value)>`.

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Done

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Added `custom()` / `custom_mut()` on `DefaultExtrinsicParams`, mirroring the existing `known()` / `known_mut()` pair. They return `&[(String, Value)]` and `&mut Vec<(String, Value)>`

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Added an `# Example` section. It runs as a doctest and also exercises the new `custom()` accessor.

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Filed https://github.com/paritytech/subxt/issues/2276 for this, and left a comment on `is_authorization_extension` pointing at it.

Worth noting it is not purely cosmetic: in V5 the signer payload excludes the last authorization extension and everything before it, so a chain declaring its own authorization extension and supplying it via `custom_extension` would sign over bytes the runtime excludes. Latent for now since `VerifyMultiSignature` is the only authorization extension Subxt knows about.

DenzelPenzel:
Added the V5 coverage you asked for. Since `polkadot_metadata_small.scale` only declares v4 and has no `VerifyMultiSignature`, there is a new `v5_encoding_info()` helper laying out `[CheckWeight, VerifyMultiSignature, WeightReclaim]` – a custom extension either side of the authorization extension:

- `v5_payload_excludes_custom_extension_before_authorization_extension` – flipping `CheckWeight` leaves the signer payload hash unchanged, since it precedes `VerifyMultiSignature`.
- `v5_payload_includes_custom_extension_after_authorization_extension` – flipping `WeightReclaim` does change it, since it follows.
- `v5_general_extrinsic_includes_custom_extensions_either_side_of_authorization` – the extrinsic itself carries every extension value regardless of position, unlike the payload.
- `authorization_extension_check_is_forwarded` also now asserts `false` for a known non-authorization extension (`CheckNonce`) and for a name not held at all.

The `is_authorization_extension` gap for chain-specific extensions is tracked in #2276.

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+            }
+            if custom.contains_key(&name) {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' was provided more than once"
+                )));
+            }
+            let in_metadata = client
+                .metadata
+                .extrinsic()
+                .transaction_extensions_to_use_for_encoding()
```
TarikGul, line 122: This still reads the newest extension version out of the metadata, and that is the part of #2265 I was hoping this would solve. As it stands a chain adding one new non optional extension still breaks every client until each one hardcodes a value for it. frame_decode already has `best_v5_general_transaction_extension_version`, which walks the versions and hands back the first one we can actually encode. Could we call that instead, so the fallback happens on its own and supplying a value stays optional? Worth knowing the ordering is backwards today too: our `extrinsic_extension_version_info` returns BTreeMap keys ascending, but frame_decode documents that trait as highest preference first, so wiring it up as is would quietly pick version 0 every time.

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+{
+    fn contains_extension(&self, name: &str) -> bool {
+        frame_decode::extrinsics::TransactionExtensions::contains_extension(&self.known, name)
+            || self.custom.contains_key(name)
+    }
+
+    // Only the known extensions can authorize a transaction; a chain-specific extension
+    // supplied via `custom_extension` always reports `false` here. See
+    // https://github.com/paritytech/subxt/issues/2276.
+    fn is_authorization_extension(&self, name: &str) -> bool {
```
TarikGul, line 174: If a chain has a custom extension that is an authorization extension, returning false here means we build the signer payload against the wrong boundary and sign something invalid, with nothing to tell the user. I would rather this returns an error once we know a custom extension is an authorization one, so it fails loudly instead of producing a bad signature. Fine by me if the real support lands later under #2276, silent mis signing is the bit I want to avoid.

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
     transaction_extensions::CheckMetadataHash,
 );
 
+/// The parameters used to construct [`KnownDefaultTransactionExtensions`].
+pub type KnownDefaultExtrinsicParams<T> =
+    <KnownDefaultTransactionExtensions<T> as TransactionExtensions<T>>::Params;
+
+/// The default set of transaction extensions, along with any custom extensions supplied for a
+/// specific transaction.
+pub struct DefaultTransactionExtensions<T: Config> {
```
TarikGul, line 36: `DefaultTransactionExtensions` was a type alias for the tuple and is now a struct, and `build` returns a new type as well. That breaks anyone naming either of them, so this wants a minor release rather than a patch. Could you call it out in the description so it does not get picked up into a point release by mistake?

subxt/src/backend.rs
```diff
@@ -34,10 +34,17 @@ pub use legacy::subscribe_to_block_headers_filling_in_gaps;
 #[doc(hidden)]
 mod sealed {
     pub trait Sealed {}
+
+    #[cfg(feature = "unstable-backend")]
```
TarikGul, line 38: The unseal Backend commit looks unrelated to custom extension values. It brings in a new feature flag, ui tests and a workflow change. Could we split it into its own PR? Would make this one a lot easier to review and to revert on its own if we ever need to.

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+                name,
+                type_id,
+                type_resolver,
+                out,
+            )
+        } else if self.custom.contains_key(name) {
+            Err(frame_decode::extrinsics::TransactionExtensionsError::Other {
+                extension_name: name.to_owned(),
+                error: format!(
+                    "Custom transaction extension '{name}' requires non-empty implicit data, which is not supported"
```
TarikGul, line 220: Small one. The other validation all happens up in `new`, but this only fires at encode time. Could we check for non empty implicit data in `new` too, so people find out when they build the params rather than when they go to sign?

TarikGul (commented):
Left a few comments inline.

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+        let mut custom = BTreeMap::new();
+
+        for (name, value) in params.custom {
+            if frame_decode::extrinsics::TransactionExtensions::contains_extension(&known, &name) {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' conflicts with a known transaction extension"
+                )));
+            }
+            if custom.contains_key(&name) {
+                return Err(TransactionExtensionError::custom(format!(
```
lexnv: nit: Only checks the extension exists. If the extension's extra_ty is (), frame-decode filters it out before calling us, so the value is dropped with no error. 

Since we have client.metadata right here with the type registry, could we check encode_as_type_to against extra_ty and check additional_ty is empty?

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+        )?;
+        let mut custom = BTreeMap::new();
+
+        for (name, value) in params.custom {
+            if frame_decode::extrinsics::TransactionExtensions::contains_extension(&known, &name) {
+                return Err(TransactionExtensionError::custom(format!(
+                    "Custom transaction extension '{name}' conflicts with a known transaction extension"
+                )));
+            }
+            if custom.contains_key(&name) {
```
lexnv: contains_key + insert could be one entry match

subxt/Cargo.toml
```diff
@@ -74,6 +74,12 @@ jsonrpsee = [
 # latest features exposed by the metadata.
 unstable-metadata = []
 
+# Enable this to implement the `Backend` trait outside of Subxt.
+# The backend API is unstable and subject to breaking changes without
+# following semantic versioning. Use this only if you need to provide
+# your own backend implementation.
+unstable-backend = []
```
lexnv: unstable-backend was the old name for the chain-head backend feature, maybe custom-backend?

subxt/src/config/default_transaction_extensions.rs
```diff
@@ -21,6 +27,206 @@ pub type DefaultTransactionExtensions<T> = (
...
+
+/// The default set of transaction extensions, along with any custom extensions supplied for a
+/// specific transaction.
+pub struct DefaultTransactionExtensions<T: Config> {
+    known: KnownDefaultTransactionExtensions<T>,
+    custom: BTreeMap<String, Value>,
+}
+
+/// Parameters used to construct [`DefaultTransactionExtensions`].
+pub struct DefaultExtrinsicParams<T: Config> {
```
lexnv, line 44: nit: Could use a Debug here

lexnv (approved):
Nice job here @DenzelPenzel! Feel free to merge after addressing Tarik's comments and the tiny ones from my side 🙏

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Already called out – see the "Breaking changes" section in the description (migration table + the 0.51 note)

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Metadata carries no authorization flag (only `identifier/ty/implicit`), there is nothing to detect this with failing loudly needs #2276 or a metadata flag. The one name we do know, `VerifyMultiSignature` is already rejected by the conflicts-with-known check

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: done

subxt/src/backend.rs
DenzelPenzel: done, split out into #2279

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: good catch, done

subxt/Cargo.toml
DenzelPenzel: Renamed to `custom-backend`; the commit is also split out into #2279

subxt/src/config/default_transaction_extensions.rs
DenzelPenzel: Done – V5 transaction now picks its extension version via `best_v5_general_transaction_extension_version`

TarikGul (approved):
🚀  Thank you

PR #2277, opened 2026-09-10
fix: decode v4 extrinsics with transaction extension version 0

TarikGul:
Addresses the v4 half of #1998.

### Description

A v4 extrinsic encodes no transaction extension version and is defined to use version 0. `transaction_extension_version_to_use_for_decoding` instead returned the highest version in the metadata, which was harmless while chains only exposed version 0.

Polkadot Asset Hub spec 2005000 exposes `[0, 1]`, and version 1 prepends `UnitTransactionExtension` and `VerifyMultiSignature`. V4 extras were read against that set, so the era bytes were interpreted as enum variant indexes and decoding failed with `VariantNotFound`. Reproduced on Polkadot Asset Hub block 20487777, where `extrinsics().iter()` yielded errors for the two v4 signed extrinsics and decoded 2 of 4. With this change it decodes 4 of 4, and block 20494727, whose signed extrinsic is a v5 General declaring version 1, still decodes 3 of 3.

Downstream report: paritytech/polkadot-rest-api#405.

### Changes

`transaction_extension_version_to_use_for_decoding` now returns 0, matching what its doc comment already described. Metadata below V16 only ever populates version 0, so this is a no op for V14, V15 and legacy metadata, and only changes behaviour on V16 runtimes exposing more than one version.

V5 extrinsics are untouched: they carry an explicit version, which is passed through `extrinsic_extension_info` and used as given. `transaction_extension_version_to_use_for_encoding` still prefers the newest version.

### Not in scope

The other half of #1998, picking the matching rather than the newest extension set when encoding v5, needs to reconcile the configured extensions against the metadata and is left alone here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


lexnv (approved):
LGTM! Thanks Tarik! 🙏

pgherveou (approved):

Imod7 (approved):

